PR #572, opened 2014-05-17
Send stderr output to the REPL buffer.

pjstadig:
Reflection warnings are sent to stderr, and are currently displayed in
the minibuffer (though you can't actually see them there because they
get scrolled out of view).  They used to be displayed in the REPL, and
it is more helpful to have them displayed there.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -752,6 +752,7 @@ The handler simply inserts the result value in BUFFER."
                                (lambda (_buffer value)
                                  (cider-repl-emit-interactive-output value))
                                (lambda (buffer err)
+                                 (cider-repl-emit-interactive-output err)
                                  (message "%s" err)
                                  (cider-highlight-compilation-errors
                                   buffer err))


cch1:
I sure wish this would get addressed.  A separate buffer for the REPL process' stdout & stderr would be ideal.  My use case (in addition to  the ones mentioned by @pjstadig) is logging: logback and log4j are frequently configured with a ConsoleAppender and that logging is now unavailable.  Reconfiguring logging to send to a file is one option -but it is clunky to do so only for the repl.

Please provide buffers for the repl process' stdout/stderr.

bbatsov:
What you're asking for is pretty different from the original ticket. Perhaps you should file a new ticket.

Malabarba:
FWIW, I'm fine with this PR. The REPL is the first place (and frequently only) place where a lot of people look for feedback. My only fear is that seeing compilation warnings on the REPL might cause people to pay even less attention to the cider startup warnings (which is already a problem we have). 

As for having a dedicated buffer, that's fine too. It should definitely use compilation-mode so that we can leverage its automated highlighting and other functionality.

bbatsov:
Btw, I'm guessing everyone is aware of https://github.com/clojure-emacs/cider/blob/a439429e3589240e3d3d5c83dab969a1d1d88f1c/cider-interaction.el#L198

bbatsov:
As for output generation in general - it was often discussed to be able to redirect output/stderr using a regular expression. People would just have to define mappings regexp -> buffer-name. Something like this would be pretty simple to setup.

cch1:
A regexp to redirect the repl process' stdout/stderr to a named buffer would easily address my outstanding use case of log4j ConsoleAppender log messages.

Such a feature would be awesome.

bbatsov:
I've created a ticket for this https://github.com/clojure-emacs/cider/issues/1525

PR #818, opened 2014-09-22
Scroll output in `*nrepl-server*` window

vspinu:


cch1:
Not sure why this is closed...
For various reasons (e.g. org.apache.log4j.ConsoleAppender's early caching of System/out), applications send output to *nrepl-server*.  Are you suggesting that we should switch from the ConsoleAppender (or whatever library we are currently using) to something else?  Or that it's not realistic to expect scrolling in *nrepl-server*?

On a related note, is there a way to create a dedicated buffer (i.e. "MyTracingBuffer") and stream output to it from Clojure via the nrepl?

vspinu:
I agree, till the issue is not fixed one way or another it probably should stay open.

arrdem:
Agreed, this is something that has really gotten in my hair of late in hacking on Grimoire which uses a classical logging tool that doesn't write to the REPL.

bbatsov:
The PR was closed because it's not a solution to the real problem. I think there's an open ticket somewhere tracking what to do with the output to the nREPL buffer, but if I'm mistaken we can start a new discussion.

cch1:
What is the real problem?

bbatsov:
@cch1 That the output in question should have never went to the `*nrepl-server*` buffer. Keep in mind this buffer doesn't even exist if you're not using `cider-jack-in`.

cch1:
Hmmm.... I'm wondering then where this "real problem" originates.  As a Clojure developer, I don't chose to send output to *nrepl-server* -it is a by product of log4j's startup behavior (which, I think we would agree, is not something that is reasonable to change) and perhaps leiningen, cider or nrepl.  Based on some research (guided by your references above), it seems as though log4j binds early to System/out when using the ConsoleAppender.  So do you have any idea how System/out gets sent to *nrepl-server* and how I (as a Clojure developer) might change it?

So far, I have heard nothing that indicates it is possible to change this behavior, and thus I am stuck with having log4j output in *nrepl-server*.

I submit that having log output in a buffer separate from interactive REPL output is desireable.  I would go so far as to call it a feature.  But if the only way to get scrolling log output is to send it to the *cider-repl* then I will happily do that if you can guide me.

PR #854, opened 2014-10-14
Fix broken stacktrace navigation to references of the form `foo/eval16959/fn`

vspinu:
I closed #850 prematurely and cannot reopen it now. The problem is not fixed. Sometimes info middleware doesn't return file location. In those cases just pick the file from stacktrace.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -403,9 +403,12 @@ it wraps to 0."
          (method (button-get button 'method))
          (info (or (and var (cider-var-info var))
                    (and class method (cider-member-info class method))
-                   `(dict "file" ,(button-get button 'file))))
+                   (nrepl-dict)))
          ;; stacktrace returns more accurate line numbers
-         (info (nrepl-dict-put info "line" (button-get button 'line))))
+         (info (nrepl-dict-put info "line" (button-get button 'line)))
+         ;; give priority to `info` files as `info` returns full paths.
+         (info (nrepl-dict-put info "file" (or (nrepl-dict-get info "file")
+                                               (button-get button 'file)))))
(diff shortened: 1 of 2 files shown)


vspinu:
All right. This patch is still needed even after fixing the middleware error in clojure-emacs/cider-nrepl#142.. 

I guess you would like a changelog for this as well. Now the problem is  that all my 3 pulls have changelog entry and will conflict. /This is one of the reasons I hate changelog :-)/. 

So let's do like this, once all the PR's are closed I will add a change log for all of the changes. Otherwise I have to rebase all pulls on top of each other or synchronise with you and make it iteratively.

bbatsov:
> I guess you would like a changelog for this as well. Now the problem is that all my 3 pulls have changelog entry and will conflict. /This is one of the reasons I hate changelog :-)/.

Yes, this is the only significant downside, but I'm used to this already. :-)

> So let's do like this, once all the PR's are closed I will add a change log for all of the changes. Otherwise I have to rebase all pulls on top of each other or synchronise with you and make it iteratively.

Fair enough. 

I see a typo in the second commit and to be honest I totally forgot what this PR was supposed to fix.

vspinu:
> I see a typo in the second commit and to be honest I totally forgot what this PR was supposed to fix. 

```
(defn tt []
  (/ 1 0))

(tt)
```

Eval with source tracking command and then go to the line with `/evalNNN` in it. The source will not be found. Comit message describes the technicalities.

vspinu:
> I see a typo in the second commit 

I don't see any.

bbatsov:
> I don't see any.

My bad. I keep forgetting most people use American spelling (`favor` seemed incorrect to me)...

PR #932, opened 2014-12-29
[Fix #930] Explain in README how to use local JVM API doc over the online version on Oracle's site

### LINKED ISSUE #930: specify local path for API javadoc
CIDER 0.9.0snapshot (package: 20141227.1232).

This is more of a feature request than an issue.

The current `javadoc-info` logic (https://github.com/clojure-emacs/cider-nrepl/blob/711b47b20afc51b9a563358b4ad6c24ed1137269/src/cider/nrepl/middleware/info.clj#L172) is to first look at resource path before resorting to the Oracle official source. This allows the use a local copy of API doc, but only by either
* manipulating resource classpath (which is difficult to do in Emacs config) or 
* putting a full version of API HTMLs under Leiningen's "resources" directory (which is specific to a single project).

It would be nice if we can specify a local doc path in Cider's *Emacs* config, to use a single shared copy of API doc, e.g., "/usr/share/doc/java/api" on a Linux machine.

I am not sure whether this can be configured now; I have not found a solution yet.

pw4ever:


README.md
```diff
@@ -382,6 +382,20 @@ helpful for identifying each host.
 (setq cider-known-endpoints '(("host-a" "10.10.10.1" "7888") ("host-b" "7888")))
 ```
 
+* If you are targeting at the JVM and prefer a local copy of the JDK API documentation over Oracle's official copy (e.g., for [JavaSE 8](http://docs.oracle.com/javase/8/docs/api/)), per nREPL's [`javadoc-info` logic (accurate as of 29 Dec 2014)](http://docs.oracle.com/javase/8/docs/api/), you can arrange your project to include the **root** path of the local API doc (i.e., where the `index.html` is located) to be included on your classpath (i.e., where the doc HTML files can be located by `clojure.java.io/resource`). For example, for Leiningen, with the local API path being `/usr/share/doc/java/api/`, put the following line in `project.clj`:
```
bbatsov: targeting the

README.md
```diff
@@ -382,6 +382,20 @@ helpful for identifying each host.
...
+:dev {:resource-paths ["/usr/share/doc/java/api/"]}
+```
+
+**or** the following line in `$HOME/.lein/profiles.clj`:
+
+```clj
+:user {:resource-paths ["/usr/share/doc/java/api/"]}
+```
+
+More detail: [issue 930](https://github.com/clojure-emacs/cider/issues/930)
```
bbatsov: I think this should be `More details can be found here.` (where `here` is a link).

pw4ever:
Updated based on the suggestions.

bbatsov:
:+1: Thanks!

Issue #709, opened 2014-08-07
Rework the REPL to use comint


shosti:
This is a really small thing, but the current behavior for the CIDER repl is to always scroll to show maximum output. I don't really like this behavior, and it would be nice if it were configurable (I'm currently hacking around it by redefining `cider-repl--show-maximum-output`—let me know if there's already another way to turn this behavior off). A `cider-repl-show-maximum-output` variable would probably be enough, although it might be nice to have some more configuration options à la `comint`—what are your thoughts? (Also, are there any plans to eventually base `cider-repl-mode` off of `comint-mode`?) I'd be happy to put together a PR for this.

bbatsov:
This important ticket is still looking for a volunteer to tackle it.

GordonGustafson:
@vspinu did you have any unfinished work on this I should look at if I decide to take a crack at this? Or should I start by looking at ielm source and the comint documentation?

johnbendi:
What about @shosti the OP who promised a PR?

bbatsov:
I think that nobody actually got anywhere with this task, so @GordonGustafson can definitely start from scratch.

vspinu:
@GordonGustafson, I haven't done anything concrete.

shosti:
@GordonGustafson yeah, never got around to looking at this (as you can probably surmise).

johnv02139:
I think basing it off of comint is a great idea, but for the record, for the main complaint, I added a defcustom in January, cider-repl-scroll-on-output, which can be set to nil to avoid the irritating scrolling.  I didn't see this issue at the time.

Is any work ongoing here to use comint?

bbatsov:
To my knowledge no one is working on this, so without a volunteer to push it forward it's not going to happen any time soon.

vspinu:
I am not even convinced that it's worth anymore Comint is old and rigid which makes it hard to build a richer interfaces on top of it. So, IMO, if there is ever an effort in this direction, it should be better spent on building a new comint system from scratch.

johnv02139:
I would agree it doesn't seem worth reimplementing.  I think your (@vspinu) original message made more sense, i.e., "use comint customization variables."  I would think we could use the variables and emulate the behavior without sharing implementation.

bbatsov:
OK, I'll close this ticket then. We've got more important tasks to do anyways.

PR #936, opened 2015-01-02
Give repl prompt its own 'field' text property

GordonGustafson:
advantages:
- Don't have to write our own bol commands since move-beginning-of-line now
respects the repl prompt.
- Commands like backward-word respect the repl prompt as well.
- These commands work properly for old repl prompts as well as the current one.

disadvantages:
- Breaks if the user sets inhibit-field-text-motion to t, but this is very rare.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ non-functioning `cider-test-jump` from test reports.
 * Fix tab completion in `cider-read-from-minibuffer`.
 * [#894](https://github.com/clojure-emacs/cider/issues/894): Make it possible to enter any symbol with `cider-read-symbol-name`.
 * Report Clojure's version including its qualifier (e.g. `alpha4`) if present.
+* Use the `field` text property to make move-beginning-of-line respect the repl prompt instead of writing our own beginning-of-line commands.
 
 ## 0.8.1 / 2014-11-20
 
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -269,11 +269,11 @@ point."
   "Find the next prompt.
 If BACKWARD is non-nil look backward."
   (let ((origin (point))
-        (prop 'cider-repl-prompt))
+        (cider-repl-prompt-property 'field))
     (while (progn
-             (cider-search-property-change prop backward)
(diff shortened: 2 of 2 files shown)


bbatsov:
It's not a bad idea. On the other hand long-term we have to adopt comint for the repl, which will render such changes pointless. Wish I had time to work on the comint transition...

GordonGustafson:
Ah, I wasn't aware of the plan to switch to comint. If you'd rather not make changes that would be obsoleted by that transition you can close this in favor of #709. However, this change does knock off a very small piece of the work involved in that switch, and more importantly it documents the inevitable transition from a custom bol function to a text-property approach (I stole this solution from comint :)

vspinu:
Yes. The comint version will be based on fields. The regexp prompts don't
make much sense for cider.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * Fix tab completion in `cider-read-from-minibuffer`.
 * [#894](https://github.com/clojure-emacs/cider/issues/894): Make it possible to enter any symbol with `cider-read-symbol-name`.
 * Report Clojure's version including its qualifier (e.g. `alpha4`) if present.
+* Use the 'field' text property to make move-beginning-of-line respect the repl prompt instead of writing our own bol commands.
```
bbatsov: Quote `field` with backticks and use `begining of line` instead of `bol` for clarity's sake.

cider-repl.el
```diff
@@ -269,7 +269,7 @@ point."
   "Find the next prompt.
 If BACKWARD is non-nil look backward."
   (let ((origin (point))
-        (prop 'cider-repl-prompt))
+        (prop 'field))
```
bbatsov: Add a small comment here explaining this for posterity's sake.

bbatsov:
@GordonGustafson Ping :-)

GordonGustafson:
Whoops, sorry for flatlining there. Rather than add a comment I changed the variable name from `prop` to `cider-repl-prompt-property`, which should be descriptive enough.

bbatsov:
:+1:

PR #941, opened 2015-01-10
[Fix #921] Non-interactive `cider-jump-to-var`.

### LINKED ISSUE #921: Cannot jump to tests from test report
Invoking `cider-test-jump` from the test report (as default-bound to `t`) does not jump to the test.  Instead it just always yields the message 'Symbol SYMBOL not resolved', where SYMBOL is whatever symbol is under point.

Commit b1304515 appears to have broken non-interactive calls to `cider-jump-to-var`, where `cider-test-jump` seems to contain the only such instance in CIDER proper.  I'm willing to write a PR to fix this, but I'm not sure which is correct approach: (1) fix `cider-jump-to-var` to once more support non-interactive invocation, (2) fix `cider-test-jump` to e.g. use `cider--jump-to-loc-from-info` directly, or (3) both.

Thanks!

llasram:
Commit b130451 broke non-interactive use of `cider-jump-to-var`, which in turn broke `cider-test-jump`.  This commit fixes the former issue and thus also fixes the latter.

I'm not sure I follow what the naming convention for the internal helper function should be.  I'll gladly update if there's a better choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
 * [#934](https://github.com/clojure-emacs/cider/issues/934): Remove
   `cider-turn-on-eldoc-mode` in favor of simply using `eldoc-mode`.
 
+### Bugs fixed
+
+* [#921](https://github.com/clojure-emacs/cider/issues/921): Fixed
+  non-functioning `cider-test-jump` from test reports.
+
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1: Looks good. Thanks!

PR #947, opened 2015-01-14
implement boot support for cider-jack-in, closes #920

### LINKED ISSUE #920: Boot build tool compatibility
The `cider-jack-in` command, hallowed be its name (and shortcut), will unfortunately **not** boot up a REPL and attach a client to it in projects that use [Boot](http://boot-clj.com/) instead of Leiningen. This is unsurprising, given the predominance of lein, but boot support would be nice and, looking into it, the issues at play are simple and mostly confined to nrepl-client.el. Basically, they are:

1. `nrepl-project-directory-for` needs to recognize `build.boot` in addition to `project.clj` files.
2. `nrepl-server-filter` would need to adjust its regex to accept both lein's "nREPL server started on..." and boot's "nREPL server listening..." outputs

This would enable barebones support, with users using customization to specify "boot" as `cider-lein-command` and "repl -s wait" as `cider-lein-parameters`.

However... in the future, perhaps **detection by cider for the appropriate defaults for those values on a per-project basis** would be convenient to end users.

Question: would a pull request to implement the two immediate simple changes above be welcome? Or, is there any orientation in general on this question?

Thanks

pandeiro:


cider.el
```diff
@@ -84,6 +84,20 @@ version from the CIDER package or library.")
   :type 'string
   :group 'cider)
 
+(defcustom cider-boot-command
+  "boot"
+  "The command used to execute Boot."
+  :type 'string
+  :group 'cider
+  :version "0.9.0")
```
bbatsov: See https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L118

cider.el
```diff
@@ -122,6 +136,18 @@ Sub-match 1 must be the project path.")
   (interactive)
   (message "CIDER %s" (cider--version)))
 
+(defun cider-command-present-p (type)
```
bbatsov: Please, rename `type` to `project-type`. This improves the readability a bit.

cider.el
```diff
@@ -122,6 +136,18 @@ Sub-match 1 must be the project path.")
...
+  (cond ((string= "lein" type) 'cider--lein-present-p)
```
bbatsov: I'd suggest using the more compact `pcase` instead of `cond`.

bbatsov:
Overall the changes look good. Apart from the changes I suggested I'd be nice if you updated the commit message to match the format you're using in the project and refer to the issue it fixes `[Fix #...] Implement`. You'll also have to update the changelog and the readme.

pandeiro:
@bbatsov Thanks for the feedback; I've made all the changes you suggested except for one: the README. I'm looking at that now (specifically the Basic Usage section and subsections) and I'm not sure exactly how much information needs to change there, given that the `cider-jack-in` changes should be completely transparent to the user, regardless of the build tool their project is using. Could you chime in on this?

pandeiro:
Took a crack at it -- let me know what you think.

bbatsov:
:+1: The PR looks good to me! Thanks!

Guess we'll get some feedback soon from boot users.

pandeiro:
Great, looking forward to it showing up in melpa!

PR #948, opened 2015-01-14
Add section on nREPL middleware use with Boot

pandeiro:
Boot has a similar convention for automatic loading of libraries via `~/.profile.boot`, which should accompany the Leiningen-specific instructions in the README.


bbatsov:
:+1:

PR #951, opened 2015-01-16
[Fix #949] Define custom cider-default-project-type.

### LINKED ISSUE #949: cider-jack-in: nil
I'm trying to start CIDER via cider-jack-in on a .clj file I'm working on. This is the error I got (I turned on debug-on-error):
```
Debugger entered--Lisp error: (void-function nil)
  nil()
  cider-jack-in(nil)
  funcall-interactively(cider-jack-in nil)
  #<subr call-interactively>(cider-jack-in nil nil)
  ad-Advice-call-interactively(#<subr call-interactively> cider-jack-in nil nil)
  apply(ad-Advice-call-interactively #<subr call-interactively> (cider-jack-in nil nil))
  call-interactively(cider-jack-in nil nil)
  command-execute(cider-jack-in)
```
I'm using CIDER 0.9.0snapshot (package20150114.512)

pandeiro:
Custom var is used with `cider-jack-in` when connecting to project-less REPLs.

(Both Leiningen and Boot support REPLs outside of projects: Cider users can choose which command to default to.)

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -98,6 +98,15 @@ version from the CIDER package or library.")
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-default-repl-command
+  "lein"
+  "Determines the default command and parameters to use when connecting to nREPL.
+This value will only be consulted when no identifying file types, ie project.clj for
+leiningen or build.boot for boot, could be found."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
+
 (defcustom cider-known-endpoints nil
   "Specify a list of custom endpoints where each endpoint is a list.
 For example: '((\"label\" \"host\" \"port\")).
@@ -158,7 +167,7 @@ If PROMPT-PROJECT is t, then prompt for the project for which to
 start the server."
   (interactive "P")
   (setq cider-current-clojure-buffer (current-buffer))
-  (let ((project-type (cider-project-type)))
+  (let ((project-type (or (cider-project-type) cider-default-repl-command)))
     (if (funcall (cider-command-present-p project-type))
         (let* ((nrepl-create-client-buffer-function  #'cider-repl-create)
                (project (when prompt-project
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   `nrepl-client` connection logic from CIDER. New hooks `cider-connected-hook`
   and `cider-disconnected-hook`.
 * [#920](https://github.com/clojure-emacs/cider/issues/920): Support `cider-jack-in` for boot-based projects.
+* [#949](https://github.com/clojure-emacs/cider/issues/949): New custom var: `cider-default-repl-command`.
 
 ### Changes
 


bbatsov:
Of we can simply check which command is available on the `PATH`. I'm not against using a defcustom, but I think the proposed name is misleading as we're talking about a situation in which there's no project.

pandeiro:
@bbatsov $PATH will probably have both lein and boot for many Clojure users, though.

What should the defcustom be called? cider-default-command maybe?

pandeiro:
cider-default-repl-command ?

vspinu:
As there are already `cider-lein-command` and `cider-boot-command`, there should probably be a selector like `cider-prefered-boot-tool`. Even better - `cider-boot-tools'  - a list of the form `'(lein boot)' with tools tried in turn. This list can be easely expanded in the future. If this list is missing a tool, that tool is not even tried.

pandeiro:
@vspinu I think putting the word 'boot' into the var would be a little confusing... But I'm also not sure I see the advantage in making it a list rather than just one custom var that the user can change if necessary -- can you explain?

I've changed the var name per @bbatsov 's request and am resubmitting the PR but if need be I can rework this more.

vspinu:
> @vspinu I think putting the word 'boot' into the var would be a little confusing...

If I understand correctly lein and boot are better described as _boot tools_ and not _project types_ (as `cider-project-type` might suggest). What's the right name for these automations? Build tools? Load-backends?

>  But I'm also not sure I see the advantage in making it a list rather than just one custom var that the user can change if necessary

The idea is that cider should go through the list of available tools and pick the first tool for which the executable is available. This way the user who has only boot on the system will not need to customize anything. The same variable gives the preferred order of `completing-read/ido` when the tool is requested interactively (for whatever reason). If a user removes a backend from the list, that means he/she doesn't want to see it in completions or other operations.

Having a comprehensive list of backends is a good idea in general. Your code becomes aware of all available tools and a lot of code  could be written generically.

pandeiro:
@vspinu Cool, I see your point... I'm not totally sure about the implementation though, but I'll hack on it. @bbatsov you okay with this so far?

pandeiro:
@vspinu Re: terminology, yeah, "projects" are the Leiningen-inspired parlance for Clojure ... repositories. So I stuck with that. Boot might be "build", given `build.boot`. If there were an agnostic term to encompass both, that would be great. I personally don't think "project-type" is a misnomer -- but maybe "project-build-type" or "project-build-tool" would be better?

Regarding line 170, it is a little bit of a shortcut, since our backend-agnostic functions dispatch off of either the strings "lein" or "boot" -- the same set of values that make sense for `cider-repl-command`.

Anyhow thanks for the input and explanations.

vspinu:
> Regarding line 170, it is a little bit of a shortcut, since our backend-agnostic functions dispatch off of either the strings "lein" or "boot" -- the same set of values that make sense for cider-repl-command.

Yes. I saw that, but it contains a meta assumption that `actual commands` == `build tools names`. I think the actual code is fine. The semantics is not right. I was really confused when I read it first time. The appropriate renaming will fix the semantic issue.

The boot documentation calls itself `build tool`. It still seems to me that in the context of cider  `boot tools` is more appropriate, but given the confusion with the  `boot` itself, it might be better to use `build` indeed. So maybe `cider-build-tool` instead of `cider-project-type` and `cider-default-build-tool` custom var or `cider-build-tools` custom list instead of `cider-default-repl-command`.

bbatsov:
> The boot documentation calls itself build tool. It still seems to me that in the context of cider boot tools is more appropriate, but given the confusion with the boot itself, it might be better to use build indeed. So maybe cider-build-tool instead of cider-project-type and cider-default-build-tool custom var or cider-build-tools custom list instead of cider-default-repl-command.

Projectile has long been using a terminology like `lein project`, `maven project`, etc. While I see your point, it seems to me most people like it.

cider.el
```diff
@@ -98,6 +98,15 @@ version from the CIDER package or library.")
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-default-repl-command
```
bbatsov: I still don't quite like this, as it's not clear this is a command used to **start/run** an nREPL server. Maybe the more descriptive `cider-start-nrepl-server-command`. The command doesn't have to be tied to `lein` or `boot` at all. Someone might like to start `nREPL` using `java -jar ...`.

bbatsov:
As I noted in a code comment - when you're outside of a `lein` or `boot` project you should be able to use any command to start an nREPL server - `lein` and `boot` are certainly not the only options.

vspinu:
> As I noted in a code comment - when you're outside of a lein or boot project you should be able to use any command to start an nREPL server - lein and boot are certainly not the only options.

This is probably true inside lein/boot projects as well. One might want to have a custom startup commands without altering `cider-lein-command` or `cider-boot-command` in order to have multiple alternatives at startup.


This is why `cider-boot-backends` list is a good idea. It standardizes the infrastructure. If someone wants to add a new command or tool, it has to add a symbol to that list,  say `joot`, then add `cider-joot-command`, `cider-joot-present-p` and `cider-joot-parameters` and he is done.  On startup from outside of the lein/boot project you will be prompted interactively for the tool as cider will be aware of all available tools.

The current API could be rewritten generically. For example:

```
(defun cider-command-present-p (build-tool)
  (let ((fun (intern (format "cider-%s-present-p" build-tool))))
    (when (fboundp fun)
      (funcall fun))))
```

BTW, `cider-command-present-p` would probably be better as `cider-program-present-p`.

triclops200:
I want to mention that right now cider is broken for files not in projects, and this would fix it. It cannot start a server to connect to, as cider-project-type returns nil.

bbatsov:
I'll merge this to spare users of the MELPA package the pain of dealing with the existing bug, but this is something that we can definitely improve upon.

pandeiro:
@bbatsov I think that's prudent; a more robust, future-proof solution would best be tackled by someone else for now.

PR #952, opened 2015-01-18
improve tramp support a bit

federkasten:
This PR improve tramp support a bit.

The following two points in `cider-connect` behavior are modified:

- Use "sshx" method instead of "ssh" for usability
- Display an initial value in the host/port selection prompts

Thanks.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -207,7 +207,7 @@ Create REPL buffer and start an nREPL client connection."
                          (let* ((change-dir-p (file-remote-p default-directory))
                                 (default-directory (if change-dir-p "~/" default-directory)))
                            (cider-locate-running-nrepl-ports (unless change-dir-p default-directory)))
-                       (let ((vec (vector "ssh" nil host "" nil))
+                       (let ((vec (vector "sshx" nil host "" nil))
                              ;; might connect to a different remote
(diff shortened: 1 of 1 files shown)


bbatsov:
What's the advantage of using `sshx`?

I'd suggest splitting the two changes into two commits. The changes should have changelog entries. I'd also appreciate it if the commit messaged followed the rules outlined in CONTRIBUTING.MD.

federkasten:
Sorry. I will re-request in accordance with your suggestion later.

`sshx` is similar to `ssh` well. `sshx` just opens a connection with `ssh -t -t host -l user /bin/sh`.

This means

- It tries to avoid any non-standard shell configuration on the remote host
- It tries to allocate a pseudo tty

So this is useful for users where normal login shell is configured heavily such as it asks them a number of questions when logging in. This is also useful for users using well-configured `.ssh/config` and `ssh-agent`.

cf. <http://www.gnu.org/software/tramp/>

federkasten:
I re-request with #953 about changing `sshx`.

And I found that displaying an initial value in `completing-read` is officially [deprecated](http://www.gnu.org/software/emacs/manual/html_node/elisp/Initial-Input.html#Initial-Input). 
From now on, I'm going to be more careful. Sorry.

Issue #822, opened 2014-09-23
cider-connect problem


reedho:
Previously, (probably before 57ac60494e2bc37ac6c9fcc6b8f972ece960e800), i can do `M-x cider-connect RET 10.x.y.z RET 9092` without problem and connected to remote nrepl i started on other machine.

Now, the same command fail with message as below: 

```
SSH port forwarding failed
open-network-stream: make client process failed: connection refused, :name, nrepl, :buffer, *cider-repl-ridho:9092*, :host, localhost, :service, 9092, :nowait, nil
```

Its now always pick "localhost" instead of the whatever ip address given to it.

Is it now SSH tunnel become mandatory? In my case, i use other ssh-user and ssh-port so that its failed, i can confirm that if i hardcoded-ly pass the ssh user and password, `cider-connect` is success as usual. 

I'm figuring out how to pass ssh user and ssh-port for `nrepl--ssh-tunnel-command`, is there easy way for this?

PR #953, opened 2015-01-19
Use sshx instead of ssh in cider-select-endpoint

federkasten:
I modified `cider-select-endpoint` function to use `sshx` instead of `ssh`.

`sshx` is similar to `ssh` well. `sshx` just opens a connection with `ssh -t -t host -l user /bin/sh`.

This means

- It tries to avoid any non-standard shell configuration on the remote host
- It tries to allocate a pseudo tty

So this is useful for users where normal login shell is configured heavily such as it asks them a number of questions when logging in.
This is also useful for users using well-configured `.ssh/config` and `ssh-agent`.
(And [the official document](http://www.gnu.org/software/tramp/) says Windows users will get some benefit, though I couldn't check it.)

Probably, this change is related to #822.

cf. <http://www.gnu.org/software/tramp/>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 * [#934](https://github.com/clojure-emacs/cider/issues/934): Remove
   `cider-turn-on-eldoc-mode` in favor of simply using `eldoc-mode`.
+* [#953](https://github.com/clojure-emacs/cider/pull/953) Use `sshx` instead of `ssh` in `cider-select-endpoint`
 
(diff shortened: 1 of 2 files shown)


vspinu:
I think this change is safe. Thanks.

bbatsov:
@federkasten Sounds good. Please, mention this in the changelog as well.

federkasten:
I have modified CHANGELOG.md. Thanks.

bbatsov:
:+1: Thanks!

Issue #955, opened 2015-01-20
buffer handling is 'strange'


arohner:
Sorry this isn't more specific, I don't really understand what's going on. 

After upgrading from Cider 0.6 to 0.8.2, buffer handling is strange. A few behaviors I've noticed:

- If M-x cider-connect, restart the JVM and cider-connect to the same host & port, Cider usually creates a new buffer, rather than reusing the old buffer
- The cider buffer name appears based on the path of the file in the active buffer, which may have no relation to the connection
- cider-connect often fails to connect properly if old cider-connect buffers are around. After killing the old buffers, cider-connect works fine.

arohner:
`; CIDER 0.8.2 (Java 1.8.0, Clojure 1.6.0, nREPL 0.2.6)`

bbatsov:
> If M-x cider-connect, restart the JVM and cider-connect to the same host & port, Cider usually creates a new buffer, rather than reusing the old buffer

Pretty sure this was the behaviour in 0.6 as well. Implementing some smarter "reconnect" has been in the to-do list for a while.

> The cider buffer name appears based on the path of the file in the active buffer, which may have no relation to the connection

Yes, this is a known issue that we'll address at some point (the REPL buffers will serve as the connection). This has also been like this for quite a while. @vspinu did some work in this direction, but there's lot more to be done.

> cider-connect often fails to connect properly if old cider-connect buffers are around. After killing the old buffers, cider-connect works fine.

If you find a reliable way to reproduce this problem we'll certainly try to fix it.

vspinu:
>>   The cider buffer name appears based on the path of the file in the active buffer, which may have no relation to the connection

> Yes, this is a known issue that we'll address at some point (the REPL buffers will serve as the connection)

I don't quite understand the problem. REPL buffer contains the name of the project if a new nREPL server is cranked with `cider-jack-in`. For connections I see `host:12345` in the name. No project name or directory name.

bbatsov:
The problem is that you can change the active connection and this won't be reflected in the buffer's name - meaning that extra info in the has little value in practice when dealing with multiple connections. I think we discussed this in the past and I suggested that changing the active connection should not affect REPL buffers and that the active connection should be displayed in the mode-line for source buffers.

arohner:
>  The cider buffer name appears based on the path of the file in the
> active buffer, which may have no relation to the connection
>
> Yes, this is a known issue that we'll address at some point (the REPL
> buffers will serve as the connection)
>
[...]

PR #958, opened 2015-01-23
[#955] Reuse existing repl buffer when cleint process has died

vspinu:
This addresses only the user case when the port is preserved. When port is not preserved, new repl buffer is generated as before.  This makes me wonder if port should be part of the buffer name in the first place. It makes repl buffer re-usage difficult.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -140,6 +140,10 @@ buffer will be hidden."
 It is called with one argument, a plist containing :host, :port and :proc
 as returned by `nrepl-connect'. ")
 
+(defvar nrepl-use-this-as-repl-buffer 'new
+  "Name of the buffer to use as REPL buffer.
+In case of a special value 'new, a new buffer is created.")
+
 
 ;;; nREPL Buffer Names
 
@@ -156,7 +160,7 @@ as returned by `nrepl-connect'. ")
               (concat nrepl-buffer-name-separator designation)
             "")))
 
-(defun nrepl-make-buffer-name (buffer-name-template &optional project-dir host port)
+(defun nrepl-make-buffer-name (buffer-name-template &optional project-dir host port dup-ok)
   "Generate a buffer name using BUFFER-NAME-TEMPLATE.
 
 If not supplied PROJECT-DIR, PORT and HOST default to the buffer local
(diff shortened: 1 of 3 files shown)


bbatsov:
> When port is not preserved, new repl buffer is generated as before. This makes me wonder if port should be part of the buffer name in the first place. It makes repl buffer re-usage difficult.

I think we can check for dead REPL buffers and prompt the users for input  if it's not clear what do to on `cider-connect`.

bbatsov:
There's a typo in the commit message and there's no changelog entry.

nrepl-client.el
```diff
@@ -156,7 +156,7 @@ as returned by `nrepl-connect'. ")
               (concat nrepl-buffer-name-separator designation)
             "")))
 
-(defun nrepl-make-buffer-name (buffer-name-template &optional project-dir host port)
+(defun nrepl-make-buffer-name (buffer-name-template &optional project-dir host port dup-ok)
```
bbatsov: A few unit tests for this new behaviour would be useful.

bbatsov:
@vspinu Ping :-)

vspinu:
> @vspinu Ping :-)

Sorry, I got involved with other stuff and forgot about this one.

> I think we can check for dead REPL buffers and prompt the users for input if it's not clear what do to on cider-connect.

I think that if a dead-process-buffer is for the same  project directory but the port differs we can safely reuse that buffer without even asking the user. On the other hand, if a dead-process-buffer is for a different project it rarely makes sense to reuse the buffer, so maybe it's not worth asking in this case either.

As to the unit tests. As the process handling system should be rewritten from scratch, such tests look to me like an unnecessary commitment at this stage.

bbatsov:
> I think that if a dead-process-buffer is for the same project directory but the port differs we can safely reuse that buffer without even asking the user. On the other hand, if a dead-process-buffer is for a different project it rarely makes sense to reuse the buffer, so maybe it's not worth asking in this case either.

I'm a bit confused - if we've used `cider-connect` we only have a host and a port; we don't actually have a project dir. So my idea was:

* user runs cider-connect
* we see there's some dead buffer that used to be connected to the same host, but on different port
* we ask the user if he wants to reuse this dead repl buffer

> As to the unit tests. As the process handling system should be rewritten from scratch, such tests look to me like an unnecessary commitment at this stage.

Fair enough.

vspinu:
> we see there's some dead repl buffer that used to be connected to the same host, but on different port

I thought you want to generalize this to `cider-jack-in` when the connection or server process dies. 

How about querying for project directory on `cider-connect`?

bbatsov:
> I thought you want to generalize this to cider-jack-in when the connection or server process dies.

I'm also interested in being able to reuse this for the implementation of sensible restart semantics.

> How about querying for project directory on cider-connect?

Doesn't feel right to me. Remote connections should be defined just in term of their endpoint IMO.

vspinu:
> Doesn't feel right to me. Remote connections should be defined just in term of their endpoint IMO. 

Even if you work on remote you work on a project and having that information as part of the buffer name is useful IMO.

bbatsov:
> Even if you work on remote you work on a project and having that information as part of the buffer name is useful IMO.

Yeah, I can see some value in this. But even if we do it, the name should optional as some people might be connecting to the remote server just to do a bit of REPL debugging/inspection. In such cases they might not have a copy of the project's sources lying around.

vspinu:
> In such cases they might not have a copy of the project's sources lying around. 

 I meant to simply request the project name from the remote and include it into `repl` buffer name. Having or not the source code lying around is irrelevant, unless I am missing something.

bbatsov:
> I meant to simply request the project name from the remote and include it into repl buffer name. Having or not the source code lying around is irrelevant, unless I am missing something.

Ah, I understand now. I assumed you wanted us to prompt the users for a "path to project" on connect and guessed the idea was that people are likely to have the project source around even if the server is remote. I wonder if the users will understand why they are supposed to set a project name, though.

bbatsov:
So, what are we going to do about this?

vspinu:
> So, what are we going to do about this?

I will amend it in light of what we have discussed. Unfortunately I cannot say when. Hopefully within a week.

vspinu:
Ok, I have finally got down to it.  

As discussed, if the zombi repl buffer matches exactly, the new connection is simply reusing the buffer. Otherwise the user is asked whether to reuse the existing zombi buffer and, if there  aremultiple such buffers, which of those to use.  

Once you pull this one, I will make a new PR to changelog all my recent pulls.

PR #962, opened 2015-02-01
[1dba82f] On error don't auto-jump to tooling files

vspinu:
The reversion to `eval` broke auto-jumping functionality.  For example <kbd> C-c C-e </kbd> after 

```
(def tt []
  (/ dfd 0))
```

jumps to `form-init` file.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -608,7 +608,8 @@ create a valid path."
   "Return t if FILE-NAME is not a 'real' source file.
 Currently, only check if the relative file name starts with 'form-init'
 which nREPL uses for temporary evaluation file names."
-  (string-match-p "\\bform-init" (file-name-nondirectory file-name)))
+  (let ((fname (file-name-nondirectory file-name)))
+    (string-match-p "^form-init" fname)))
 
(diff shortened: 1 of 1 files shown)


bbatsov:
The code looks good, but it needs a changelog entry.

vspinu:
> The code looks good, but it needs a changelog entry.

Do you use clearcut rules of when to add a change log and when not? Would be good to know. It seem that now every commit needs a changelog. Quite a lot of redundant typing IMO. 

By any standards, this is a small fix for a small recent regression. Why would an end user be interested to read this in a change log?

bbatsov:
> Do you use clearcut rules of when to add a change log and when not? Would be good to know. It seem that now every commit needs a changelog. Quite a lot of redundant typing IMO.

Yes. If some commit fixes a bug, introduces a change in existing user-visible behaviour or adds a new user-visible feature it should have a changelog entry. 

> By any standards, this is a small fix for a small recent regression. Why would an end user be interested to read this in a change log?

I'm definitely interested in reading about such fixes and I'm pretty sure I'm not the only one.

PR #976, opened 2015-02-16
Fix ansi output coloring for test failures

### LINKED ISSUE #968: midje ansi output broken
After running git bisect it appears that 83d360d474b961f53c08a0db5e5c409bc7478d16 broke midje ansi output. Steps to reproduce:

1. Install latest cider
2. Open a file containing a failing Midje test within a Leiningen project
3. cider-jack-in
4. `C-c C-k`
5. In the REPL, Midje's output will display ANSI codes rather than actually colorizing the text

GordonGustafson:
The function cider-repl--input-line-beginning-position used to return the position at the beginning of the line containing cider-repl-input-start-mark. Commit 83d360d made it impossible to move the cursor to this position IF the marker was on the same line as the prompt. This caused text to be inserted in the wrong position and not have its color codes processed. Since all callers of this function subtract one from its result to find the end of the line that precedes that of the marker, we simply make a function that does this directly and avoids the prompt issue.

Fixes #968

I've tacked on two other minor things; I can remove them if necessary.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -298,12 +298,6 @@ If BACKWARD is non-nil search backward."
   (set-marker cider-repl-output-start (point))
   (set-marker cider-repl-output-end (point)))
 
-(defun cider-repl--mark-output-end ()
-  "Mark the output end."
-  (add-text-properties cider-repl-output-start cider-repl-output-end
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Thanks!

PR #978, opened 2015-02-16
Add support for format-edn middleware & fix #977

### LINKED ISSUE #977: cider-format-region does not respect indentation of region start position
e.g. if I have

```clojure
(let [x 1]
  (let [x 1]
    (let [x 1]
      (let [x 1]
                  (println x)))))
```

and I select the innermost `let` form, then do `cider-format-region`, I get:

```clojure
(let [x 1]
  (let [x 1]
    (let [x 1]
      (let [x 1]
  (println x)))))
```

cichli:
I'm sure there's a better way to do `cider--format-reindent`. Originally I thought of `indent-buffer` after insertion, but that would only work for EDN.

Feedback welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#920](https://github.com/clojure-emacs/cider/issues/920): Support `cider-jack-in` for boot-based projects.
 * [#949](https://github.com/clojure-emacs/cider/issues/949): New custom var: `cider-default-repl-command`.
 * New code formatting commands - `cider-format-buffer`, `cider-format-region` and `cider-format-defun`.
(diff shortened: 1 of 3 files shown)


bbatsov:
> I'm sure there's a better way to do cider--format-reindent. Originally I thought of indent-buffer after insertion, but that would only work for EDN.

I'm definitely confused by it. :-) Will think a bit more on the subject tomorrow.

cichli:
Basically, we know that any leading whitespace before the first char will be stripped, so `start` is definitely the position where the form will begin. So we just add `start-column` spaces to the start of every line after the first (in addition to the spaces already added by the formatting op).

cichli:
This (along with the format-code commands) fixes #535 btw.

bbatsov:
:+1:

PR #980, opened 2015-02-17
Fix inspector popup

cichli:


cichli:
This might actually be a bug in `cider-popup-buffer-display`. If the buffer it's trying to pop up is already selected, maybe it should be a no-op.

bbatsov:
Can you describe the problem this fixes in a bit more detail?

> This might actually be a bug in cider-popup-buffer-display. If the buffer it's trying to pop up is already selected, maybe it should be a no-op.

This sounds sensible. I recall @vspinu worked a few months ago on some popup buffer improvements, though I can't recall if they were related. I'll have to consult the code and the history.

vspinu:
>> This might actually be a bug in cider-popup-buffer-display. If the buffer it's trying to pop up is already selected, maybe it should be a no-op.

>This sounds sensible. 

Yes it does. I think I missed this pattern because it's a "popup-up" buffer.

cichli:
@bbatsov Sorry, I should have been more clear. It's easy to replicate.

* `M-x` `cider-inspect` and eval a form. The inspector buffer will pop up in a new window.
* Perform any inspector operation in the buffer, e.g. refresh it with `g` or inspect a value with `RET`. The inspector buffer will now pop up in the window you previously had selected.
* Repeat the second step and the inspector continues to pop up in another window.

@vspinu thanks for clarifying. I've just pushed again with the fix applied to `cider-popup-buffer-display`.

cider-interaction.el
```diff
@@ -1196,13 +1196,14 @@ If major MODE is non-nil, enable it for the popup buffer."
...
   "Display BUFFER.
 If SELECT is non-nil, select the BUFFER."
-  (-when-let (win (get-buffer-window buffer))
-    (with-current-buffer buffer
-      (set-window-point win (point))))
-  ;; Non nil `inhibit-same-window' ensures that current window is not covered
-  (if select
-      (pop-to-buffer buffer `(nil . ((inhibit-same-window . ,pop-up-windows))))
-    (display-buffer buffer `(nil . ((inhibit-same-window . ,pop-up-windows)))))
+  (when (not (eq buffer (current-buffer)))
```
bbatsov: when + not == unless

cichli: Thanks; fixed

cider-interaction.el
```diff
@@ -1196,13 +1196,14 @@ If major MODE is non-nil, enable it for the popup buffer."
...
   "Display BUFFER.
 If SELECT is non-nil, select the BUFFER."
-  (-when-let (win (get-buffer-window buffer))
-    (with-current-buffer buffer
-      (set-window-point win (point))))
-  ;; Non nil `inhibit-same-window' ensures that current window is not covered
-  (if select
-      (pop-to-buffer buffer `(nil . ((inhibit-same-window . ,pop-up-windows))))
-    (display-buffer buffer `(nil . ((inhibit-same-window . ,pop-up-windows)))))
+  (unless (eq buffer (current-buffer))
```
vspinu: It's safer to use `(eq (get-buffer buffer) (current-buffer))` as buffer might be buffer name. All emacs functions that accept a buffer also accept a string.

cichli: @vspinu I've fixed this too, thanks!

bbatsov:
:+1: Thanks!

PR #982, opened 2015-02-21
Prefer find-buffer-visiting to get-file-buffer

### LINKED ISSUE #981: Reuse buffer when jumping to jar
If you jump twice into the same file in a jar cider will ask something akin to "A buffer visiting jar://path/to/my/file.clj already exists.  Proceed?"  If you hit yes a new buffer is created, if you hit no the jump is cancelled.  Instead of creating new buffers cider should just pop to the existing buffer.

cichli:
Fixes #981.


bbatsov:
:+1:

PR #983, opened 2015-02-21
Add `cider-repl-prompt-function` and `cider-repl-default-prompt`, to support tailoring of the prompt

simon-katz:
I've been hacking a few things in Cider over the last few years, and I have to change things each time there's a new version of Cider. It's time to stop doing that if possible!

Here's the first of likely a few simple pull requests in the coming days and weeks.


simon-katz:
Hmmm, I should do things properly.

I've just looked at the guidelines for contributing. I'll do what it says there, and close this pull request.

PR #984, opened 2015-02-21
Allow the prompt to be tailored

simon-katz:
The prompt string is currently hidden in `cider-repl--insert-prompt`, and so it cannot be tailored.

This pull request adds `cider-repl--prompt-function` and `cider-repl--default-prompt`. This allows the prompt to be tailored, either by setting `cider-repl--prompt-function` or perhaps by advising `cider-repl--default-prompt`.

If it's a good idea, I'm happy to update the README to describe this. I think I would probably only describe setting `cider-repl--prompt-function`, and not the possibility of using advice. (But maybe that would just be cluttering the README with something that most people wouldn't be interested in.)

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -363,6 +363,16 @@ This will not work on non-current prompts."
 
 (put 'cider-save-marker 'lisp-indent-function 1)
 
+(defun cider-repl-default-prompt (namespace)
+  "Return a prompt string that mentions NAMESPACE."
+  (format "%s> " namespace))
+
+(defcustom cider-repl-prompt-function #'cider-repl-default-prompt
+  "A function that returns a prompt string.
+Takes one argument, a namespace name."
+  :type 'function
+  :group 'cider-repl)
+
 (defun cider-repl--insert-prompt (namespace)
   "Insert the prompt (before markers!), taking into account NAMESPACE.
 Set point after the prompt.
@@ -372,7 +382,7 @@ Return the position of the prompt beginning."
     (cider-save-marker cider-repl-output-end
       (unless (bolp) (insert-before-markers "\n"))
       (let ((prompt-start (point))
-            (prompt (format "%s> " namespace)))
+            (prompt (funcall cider-repl-prompt-function namespace)))
         (cider-propertize-region
             '(font-lock-face cider-repl-prompt-face read-only t intangible t
                              field cider-repl-prompt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
   - New command to eval and pprint result: `cider-interactive-pprint-eval`.
   - `cider-format-pprint-eval` has been removed.
 * Warn when used with incompatible nREPL server.
+* Allow the prompt to be tailored by adding `cider-repl-prompt-function` and `cider-repl-default-prompt`.
 
 ### Changes
 


bbatsov:
What would be the use case for having the prompt configurable? Displaying some extra info in it perhaps?

simon-katz:
More info perhaps, but that's not what I would use it for at the moment.

I like to print a newline after my prompt, because, when the namespace name is long, I get annoyed that all my code is way off to the right. And when entering multi-line forms in the REPL, I think having the first line start at the first column is nicer.

I also sometimes like to print a long row of underscores on a line before the namespace name, to clearly separate things when looking back over a series of REPL commands.

arrdem:
I'm a fan. I'd suggest that expecting `cider-repl--prompt-function` to be a function of an alist which may contain the namespace and in future could be extended with stuff like the Clojure host (clj, cljs cljclr) and CIDER version just in case there's some madman out there who wants a ZSH style two line prompt with a bunch of data.

bbatsov:
I'm fine with adding this, but a do wonder if passing an alist won't be an overkill as a lot of the extra info that one might need is encoded in the REPL buffer name. On the other hand this will future proof the code (at least in theory).

cider-repl.el
```diff
@@ -363,6 +363,14 @@ This will not work on non-current prompts."
 
 (put 'cider-save-marker 'lisp-indent-function 1)
 
+(defun cider-repl--default-prompt (namespace)
+  "Return a prompt string that mentions NAMESPACE."
+  (format "%s> " namespace))
+
+(defvar cider-repl--prompt-function #'cider-repl--default-prompt
```
bbatsov: This should be a defcustom.

bbatsov:
We also have to consider if this would still be doable when we rebase the REPL on top of comint-mode (which is one of our top priorities on the roadmap). I haven't done any research for comint migration at all, so guess @vspinu can comment on this.

vspinu:
> We also have to consider if this would still be doable when we rebase the REPL on top of comint-mode (which is one of our top priorities on the roadmap).

The new comint based repl will use fields instead of the older `comint-prompt-regexp`. Thus custom prompt should not interfere with that at all.

bbatsov:
@vspinu Thanks for the update!

simon-katz:
I'll make `cider-repl--prompt-function` a defcustom.

Regarding the args to `cider-repl--prompt-function`: I could make the args `(NAMESPACE &REST OPTIONS)`; that would make the simple case unchanged from now and allow options to be added in future. Would that be a good compromise?

(And looking for some guidance on how to do things... should I make separate commits for each little change, or wrap things up into a single tidy-up commit?)

vspinu:
> I'll make cider-repl--prompt-function a defcustom.

It should not have "--" in the name. Cider uses "--" for "very" internal function that are used once or twice mostly to increase the readability of the code.

> Regarding the args to cider-repl--prompt-function: I could make the args (NAMESPACE &REST OPTIONS);

Either this or no arguments at all. This function could pick all the relevant information (namespace, point, buffer, process state etc) from the context of where it was called.

bbatsov:
> Either this or no arguments at all. This function could pick all the relevant information (namespace, point, buffer, process state etc) from the context of where it was called.

Yep this is an option - guess can have everything sensible as buffer local vars. Or we can have just a single ns argument and adventurous folks can retrieve more data for their fancy prompts from the global environment.  Don't want to over-engineer something that will be probably used by a handful of people in the end.

simon-katz:
I've pushed most of the changes that have been suggested.

I prefer that the prompt function takes arguments, because that makes it more self-contained (and so more understandable and more testable, IMO).

I'm happy of course to make further changes.

bbatsov:
> I prefer that the prompt function takes arguments, because that makes it more self-contained (and so more understandable and more testable, IMO).

The problem with this is that it's not really clear what the structure of the `rest` argument is - it's just a list we have to somehow magically know that the item at index 3 is x, at index 4 is y, etc. Of course a similar problem exists with buffer-local-variables - you have to know they exist...

As for the final version of the PR - it should be a single commit.

simon-katz:
> The problem with this is that it's not really clear what the structure of the rest argument is

Yes.

I was assuming that could be addressed if and when extra arguments are added; perhaps there could be optional arguments, perhaps there could be a single extra argument that is an alist, or perhaps there could be keyword arguments.

But I'm more than happy to go back to having just the single NAMESPACE argument.

> As for the final version of the PR - it should be a single commit.

OK. But I'm not sure how to do that. This PR is associated with a particular branch in my fork, and that already has pushed commits. It it possible to squash already-pushed commits?

Ah, I found this...

* Squash commits locally with git rebase -i and then force push with git push origin +master

...at http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed.

Is that good advice? I'm a bit nervous about messing things up.

bbatsov:
Yep - squash + force-push is the way to go. I do it all the time.

On 22 February 2015 at 16:28, simon-katz <notifications@github.com> wrote:

> The problem with this is that it's not really clear what the structure of
> the rest argument is
>
> Yes.
>
> I was assuming that could be addressed if and when extra arguments are
> added; perhaps there could be optional arguments, perhaps there could be a
> single extra argument that is an alist, or perhaps there could be keyword
> arguments.
>
> But I'm more than happy to go back to having just the single NAMESPACE
> argument.
>
> As for the final version of the PR - it should be a single commit.
>
> OK. But I'm not sure how to do that. This PR is associated with a
> particular branch in my fork, and that already has pushed commits. It it
> possible to squash already-pushed commits?
>
> Ah, I found this...
>
>    - Squash commits locally with git rebase -i and then force push with
>    git push origin +master
>
> ...at
> http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed
> .
>
> Is that good advice? I'm a bit nervous about messing things up.
>
> —
> <https://github.com/clojure-emacs/cider/pull/984#issuecomment-75437817>.
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

simon-katz:
> Yep - squash + force-push is the way to go. I do it all the time.

OK, thanks; done. (And I'll do the same if I make further changes.)

So now I think the only outstanding thing is the signature of the prompt function.

bbatsov:
Let's start with just the namespace; we'll change this when some actual reason to do so appears. Rebase on top of the current master and we're good to go.

simon-katz:
> Let's start with just the namespace; we'll change this when some actual reason to do so appears. Rebase on top of the current master and we're good to go.

OK, I'll do that — probably tomorrow.

simon-katz:
Actually, today, not tomorrow...

It's done.

PR #993, opened 2015-02-27
Completion metadata

cichli:
I'm taking a cue from `company-clang`[1] in using text properties to store the metadata on the candidates.

We could make this configurable with `defcustom`s for:
* the format string (and also tag the candidate with `:ns`, which will then be passed as the second arg to `format`)
* a dict of `:type` to abbreviation

[1] [company-clang.el#L143](https://github.com/company-mode/company-mode/blob/master/company-clang.el#L143)

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -119,7 +119,7 @@ which will use the default REPL connection."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
-(defcustom cider-annotate-completion-candidates nil
+(defcustom cider-annotate-completion-candidates t
   "When true, annotate completion candidates with some extra information."
   :type 'boolean
   :group 'cider
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -797,25 +797,39 @@ form, with symbol at point replaced by __prefix__."
       (setq cider-completion-last-context context)
       context)))
 
+(defun cider-completion--parse-candidate-map (candidate-map)
+  (let ((candidate (nrepl-dict-get candidate-map "candidate"))
+        (type (nrepl-dict-get candidate-map "type")))
+    (put-text-property 0 1 'type type candidate)
+    candidate))
+
 (defun cider-complete (str)
   "Complete STR with context at point."
-  (cider-sync-request:complete str (cider-completion-get-context)))
+  (let* ((context (cider-completion-get-context))
+         (candidates (cider-sync-request:complete str context)))
+    (map 'list #'cider-completion--parse-candidate-map candidates)))
 
 (defun cider-annotate-symbol (symbol)
-  "Append extra information to SYMBOL's name.
+  "Return a string suitable for annotating SYMBOL.
 
-Currently we annotate macros, special-forms and functions,
-as it's not obvious from their names alone which is which."
-  (if cider-annotate-completion-candidates
-      (-when-let (info (cider-var-info symbol))
-        (let ((macro   (nrepl-dict-get info "macro"))
-              (special (nrepl-dict-get info "special-form"))
-              (args    (nrepl-dict-get info "arglists-str")))
-          (cond
-           (macro " <m>")
-           (special " <s>")
-           (args " <f>"))))
-    ""))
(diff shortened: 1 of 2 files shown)


arrdem:
Cool feature! Thanks!

cider-interaction.el
```diff
@@ -797,25 +797,39 @@ form, with symbol at point replaced by __prefix__."
       (setq cider-completion-last-context context)
       context)))
 
+(defun cider-completion--parse-candidate-map (candidate-map)
```
bbatsov: Isn't this processing just a single entry of the candidates map?

cider-interaction.el
```diff
@@ -797,25 +797,39 @@ form, with symbol at point replaced by __prefix__."
...
+  (let ((candidate (nrepl-dict-get candidate-map "candidate"))
+        (type (nrepl-dict-get candidate-map "type")))
+    (put-text-property 0 1 'type type candidate)
```
bbatsov: Nice solution indeed!

bbatsov:
There's one other aspect that'd be cool to have (config controlled) - showing the ns of a var, when it's not qualified.

cider-interaction.el
cichli: Well the `complete` request returns a list of maps, and I'm referring to an element of this list as `candidate-map`... better name definitely welcome. `candidate-dict` or `candidate-data`?

cichli:
@arrdem no worries :+1:

@bbatsov yeah, that's a nice idea! Instead of a format string as a config option we could just use a function that takes both `type` and `ns` optionally. Then either let users check whether the var is qualified (provide a function for this) or provide a config option to control when `ns` is or isn't passed to said function (`always`, `unqualified`, `nil`). Sound good?

cichli:
Oh and the abbreviations are pretty non-obvious (better suggestions welcome, I kept it simple for now). We can document them on the config option and maybe add them to the README too.

cider-interaction.el
bbatsov: Should really stop commenting code late in the evening. Just realised what I've said. :-) It's OK to keep this name as is.

bbatsov:
> Then either let users check whether the var is qualified (provide a function for this) or provide a config option to control when ns is or isn't passed to said function (always, unqualified, nil). Sound good?

Sounds good.

bbatsov:
> Oh and the abbreviations are pretty non-obvious (better suggestions welcome, I kept it simple for now). We can document them on the config option and maybe add them to the README too.

Guess we can do both.

expez:
This looks great!  I suppose this only takes care of the performance issues related to annotations?  IIRC the previous annotation mechanism would ping the middleware (using caching?) to annotate each candidate.

bbatsov:
I'll merge this as is, because I merged the middleware changes last evening without thinking of the consequences (breaking all existing client running 0.9-snapshot). 

@expez No caching is needed at all. We bundle completion metadata together with the candidates right now, which is a much smarter and simpler solution.

cichli:
@expez thanks, I'll fix it to use `mapcar` in the PR to add config options around this. Currently `type` is only returned by the cljs completion library (for deftypes) - I think compliment returns `class` in this case, but that doesn't make sense in cljs. I might actually change `import` to `class` for cljs completions, as imports are really Google Closure classes.

cider-interaction.el
```diff
@@ -119,7 +119,7 @@ which will use the default REPL connection."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
-(defcustom cider-annotate-completion-candidates nil
+(defcustom cider-annotate-completion-candidates t
```
dgutov: Do you figure there's going to be any further use for this variable? It was added when annotations got slow.

bbatsov: No idea. I doubt that someone will want to disable them, but on the other hand I've seen odd customization requests in all of my projects. Guess we can keep it for now.

arrdem:
MRW updating and this just worked

653f449f-2072-4c67-968c-67e8d24617a7

Rock on!

PR #1003, opened 2015-03-03
Send ns with interactive eval requests

### LINKED ISSUE #1002: cider-interactive-eval not using cider-current-ns?
Previously, the behavior of C-x C-e is evaling last sexp using buffer namespace.

In current master, the cider-interactive-eval is not using (cider-current-ns) as last arg to nrepl-request:eval. 

Is this intended change?

cichli:
Fixes #1002.


bbatsov:
:+1:

PR #1009, opened 2015-03-07
Update inspector to use :inspect option of eval op

cichli:
see clojure-emacs/cider-nrepl#164

Since eval can return multiple values, I've added some (very basic) logic for dealing with this - each successive `:value` response is rendered into the inspector buffer, but it is only popped up once the `:status #{:done}` response is received (so the last result of evaluating the expr is inspected). We check if the inspector buffer exists in the done handler in case it's been killed between the `:value` and `:status #{:done}` responses.

I'm passing the current buffer to `nrepl-make-response-handler` rather than the inspector buffer, as we have to close over it for `cider-popup-buffer-display` to work in the `done` handler (since it's an async request, `(current-buffer)` returns the connection buffer inside the handler - see [here](https://github.com/clojure-emacs/cider/blob/master/nrepl-client.el#L891)).

Currently, the [way piggieback works](https://github.com/cemerick/piggieback/blob/master/src/cemerick/piggieback.clj#L123) is that it calls `read-string` on the result returned by the CLJS REPL. If it reads succesfully the value is returned, otherwise the result is simply printed (and `nil` is returned). This means inspecting simple readable data structures works well in CLJS, but it's not possible to inspect something like `(atom {:a [1 2 3]})` as the result can't be read. Perhaps a good enhancement would be to send the `:out` responses to the REPL buffer, or some `*cider-inspect-out*` buffer - what do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
   - `cider-annotate-completion-function` controls how the annotations are formatted.
   - `cider-completion-annotations-alist` controls the abbreviations used in annotations.
   - `cider-completion-annotations-include-ns` controls when to include the candidate namespace in annotations.
+* Inspector middleware now relies on `eval` middleware, adding support for ClojureScript.
 
 ### Changes
 
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -67,41 +67,62 @@
   (interactive
    (list (cider-read-from-minibuffer "Inspect value: "
                                      (cider-sexp-at-point))))
-  (cider-ensure-op-supported "inspect-start")
-  (cider-inspect-sym expression (cider-current-ns)))
+  (cider-inspect-expr expression (cider-current-ns)))
 
 ;; Operations
-(defun cider-render-response (buffer)
-  (nrepl-make-response-handler
-   buffer
-   (lambda (buffer str)
-     (cider-irender buffer str))
-   '()
-   (lambda (buffer _str)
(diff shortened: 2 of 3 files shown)


bbatsov:
> Perhaps a good enhancement would be to send the :out responses to the REPL buffer, or some *cider-inspect-out* buffer - what do you think?

There's already a config that all functionality should respect - the output destination. But default it's send to the REPL buffer, but can be sent to a dedicated buffer as well.

cichli:
A very good point :-). Updated to use the default out/err handlers for interactive evals, and to add a changelog entry.

cichli:
Pushed again to ensure `cider--prep-interactive-eval` is called before making the request. This ensures that the namespace form of the current buffer has been evaluated - an error is thrown otherwise. Not sure if the inspector shouldn't just use the current REPL namespace.

bbatsov:
:+1:

bbatsov:
> Not sure if the inspector shouldn't just use the current REPL namespace.

Or we can attach a ns to the inspector buffer and use it when drilling down.

cichli:
Well the ns is only relevant for the initial inspect call - pushing/popping values don't require any further evals.

bbatsov:
> Well the ns is only relevant for the initial inspect call - pushing/popping values don't require any further evals.

In this case using the REPL ns is not a good idea - we should always get the ns from buffer in which the inspector was invoked.

PR #1015, opened 2015-03-13
Restore inhibit-same-window behaviour

cichli:


cichli:
Sorry, to clarify what was happening here - in async request handlers, the current buffer is the connection buffer, so we should actually be checking if the selected window is equal to the window of the buffer we're popping up (rather than checking if the current buffer is the buffer we're popping up).

@vspinu after using this for a day I now see that using `inhibit-same-window` is clearly better than the default here; sorry for reverting it.

vspinu:
> in async request handlers, the current buffer is the connection buffer, so we should actually be checking if the selected window is equal to the window 

That makes sense, indeed. The change looks ok to me.  Would you please add this clarification inline? So that we don't forget why that window juggling is there.

cichli:
@vpsinu thanks! I've added a note above the `selected-window` check.

bbatsov:
:+1: Looks legit!

Issue #1004, opened 2015-03-04
*cider-error* buffer seem to have buffer-undo activated


Arcanum-XIII:
I have this error quite often :

Warning (undo): Buffer *cider-error*' undo info was 14111968 bytes long. The undo info was discarded because it exceededundo-outer-limit'.

(Origin of the error seem to be an issue between piggieback/weasel | boot-cljs-repl | cider, filling the cider-error buffer)

I don't think this buffer need the undo to be activated ?

cichli:
I don't think there should ever be that much data in the error buffer. Can you share the versions of cider/cider-nrepl/clojurescript/piggieback/weasel, or (even better) a project that will let me replicate?

Arcanum-XIII:
latest cider from melpa | cider nrepl 9.0-snapshot | piggieback should be 0.16-snapshot and weasel-6.0 (if my understanding of the boot-cljs-repl is correct - meaning I override piggieback/weasel in my local build.boot)
Boot is updated daily with 'boot -u'

I agree that something fishy is going on - especially since most of the time the *cider-error* buffer is "looking" empty - the joy of clojurescript ?

noprompt:
I'm seeing the same warning as well with `[piggieback "0.1.6-SNAPSHOT"]`, `[weasel "0.6.0"]`, and cider 0.8.2. It seems to happen at random intervals. Sometimes my session will lock up in minutes; other times hours. This is the warning I get

```
Warning (undo): Buffer `*cider-error*' undo info was 15964992 bytes long.
The undo info was discarded because it exceeded `undo-outer-limit'.
```

noprompt:
I think this happens because it is getting hung up on dumping the exception info which contains the entire ClojureScript compiler analyzer data (this can be enormous).

cichli:
[...]

PR #1016, opened 2015-03-14
Add cider-stacktrace-print-length

cichli:


bbatsov:
As it stands I'm not sure many people will understand what this does. I know I'm having a hard time understanding it myself and had to consult the middleware change.

vspinu:
How does the *print-length* affect stacktrace more concretely? Will this truncate error lines? A small example would be really helpful. Thanks.

cichli:
Agh, I submitted this earlier than I intended to and apparently only updated the title. What I _meant_ to say was:

This is a more general fix for one potential cause of #1004, in much the same vein of #738 - while we limit printing of very deep data structures, we don't limit printing of long (potentially infinitely so) data structures. For example, evaluating this will cause the error buffer to display but never render anything, and the java process to use 100% of a CPU until killed:

```clojure
(throw (ex-info "" {:a (range)}))
```

Doing the same for a very large range causes the buffer to only render after several seconds. With this fix, anything sequential in the `ex-data` map for each cause will only have its first 50 (by default) elements printed, by binding `clojure.core/*print-length*` when printing the cause data in the middleware [here](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/stacktrace.clj#L129-L131). It's still possible to inspect the cause data in its entirety either at the REPL or in the inspector using `*e`.

It might also be useful to:

* make the printed data foldable separately from the call stack itself (currently you can only fold both at once)
* when initially rendering the error buffer, move the point to the line of first cause itself, rather than the first line in its call stack. If the cause data takes up so many lines that the cause type is off-screen, you currently get no context as to what kind of exception has been thrown (when you're not at the REPL)

@bbatsov I agree that the docstrings aren't perfect. I'll update both this one and `cider-stacktrace-print-level` to be more descriptive and add something to README.md too.

@vspinu this only affects printing of the `ex-data` map above the call stack for each cause that is an instance of `IExceptionInfo`. No actual call stacks, exception messages or exception types are truncated.

bbatsov:
> make the printed data foldable separately from the call stack itself (currently you can only fold both at once)

I agree.

> when initially rendering the error buffer, move the point to the line of first cause itself, rather than the first line in its call stack. If the cause data takes up so many lines that the cause type is off-screen, you currently get no context as to what kind of exception has been thrown (when you're not at the REPL)

Perhaps. Although something like this should definitely be configurable as I imagine not everyone will like it.

> @bbatsov I agree that the docstrings aren't perfect. I'll update both this one and cider-stacktrace-print-level to be more descriptive and add something to README.md too.

Sounds good.

cichli:
I've updated the docstrings for both `cider-stacktrace-print-length` and `cider-stacktrace-print-level`.

bbatsov:
Just squash the related commits together and we're good to go.

cichli:
Done.

bbatsov:
:+1: Thanks!

PR #1019, opened 2015-03-15
Debugging in Cider

Malabarba:
This PR goes along with clojure-emacs/cider-nrepl#170, please see there for instructions.

cider-debug

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -1,0 +1,174 @@
+;;; cider-debug.el --- CIDER interaction with clj-debugger  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2015  Artur Malabarba
+
+;; Author: Artur Malabarba <bruce.connor.am@gmail.com>
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Instrument code with `cider-debug-defun-at-point', and when the code is
+;; executed cider-debug will kick in.  See this function's doc for more
+;; information.
+
+;;; Code:
+
+(require 'nrepl-client)
+(require 'cider-interaction)
+
+(defvar cider--current-debug-value nil
+  "Last value received from the debugger.
+Is printed by `cider--debug-read-command' while stepping through
+code.")
+
+(defconst cider--instrument-format
+  (concat "(cider.nrepl.middleware.debug/instrument-and-eval"
+          ;; filename and point are passed in a map. Eventually, this should be
+          ;; part of the message (which the nrepl sees as a map anyway).
+          " {:filename %S :point %S} '%s)")
+  "Format to instrument an expression given a file and a coordinate.")
+
+
+;;; Implementation
+(defun cider--debug-init-connection ()
+  "Initialize a connection with clj-debugger."
+  (nrepl-send-request
+   '("op" "init-debugger")
+   (let ((connection-buffer (nrepl-current-connection-buffer)))
+     (lambda (response)
+       (nrepl-dbind-response response (value coor filename point status id)
+         (if (not (member "done" status))
+             (cider--handle-debug value coor filename point connection-buffer)
+           (puthash id (gethash id nrepl-pending-requests)
+                    nrepl-completed-requests)
+           (remhash id nrepl-pending-requests)))))))
+
+(defun cider--forward-sexp (n)
+  "Move forward N logical sexps.
+This will skip over sexps that don't represent objects, such as ^{}."
+  (while (> n 0)
+    ;; Non-logical sexps.
+    (while (progn (forward-sexp 1)
+                  (forward-sexp -1)
+                  (looking-at-p "\\^"))
+      (forward-sexp 1))
+    ;; The actual sexp
+    (forward-sexp 1)
+    (setq n (1- n))))
+
+(defun cider--handle-debug (value coordinates file point connection-buffer)
+  "Handle debugging notification.
+VALUE is saved in `cider--current-debug-value' to be printed
+while waiting for user input.
+COORDINATES, FILE and POINT are used to place point at the instrumented sexp.
+CONNECTION-BUFFER is the nrepl buffer."
+  ;; Be ready to prompt the user when debugger.core/break is
+  ;; triggers a need-input request.
+  (nrepl-push-input-handler #'cider--need-debug-input connection-buffer)
+
+  ;; Navigate to the instrumented sexp, wherever we might be.
+  (find-file file)
+  ;; Position of the sexp.
+  (goto-char point)
+  (condition-case nil
+      ;; Make sure it is a list.
+      (let ((coordinates (append coordinates nil)))
+        ;; Navigate through sexps inside the sexp.
+        (while coordinates
+          (down-list)
+          (cider--forward-sexp (pop coordinates)))
+        ;; Place point at the end of instrumented sexp.
+        (cider--forward-sexp 1))
+    ;; Avoid throwing actual errors, since this happens on every breakpoint.
+    (error (message "Can't find instrumented sexp, did you edit the source?")))
+  ;; Prepare to notify the user.
+  (setq cider--current-debug-value value))
+
+(defun cider--debug-read-command ()
+  "Receive input from the user representing a command to do."
+  (let ((cider-interactive-eval-result-prefix
+         "(n)ext (c)ontinue (i)nject => "))
+    (cider--display-interactive-eval-result
+     cider--current-debug-value))
+  (let ((input
+         (cl-case (read-char)
+           ;; These keys were chosen to match edebug rather than clj-debugger.
+           (?n "(c)")
+           (?c "(q)")
+           ;; Inject
+           (?i (condition-case nil
+                   (concat (read-from-minibuffer "Expression to inject (non-nil): ")
+                           "\n(c)")
+                 (quit nil))))))
+    (if (and input (not (string= "" input)))
+        (progn (setq cider--current-debug-value nil)
+               input)
+      (cider--debug-read-command))))
+
+(defun cider--need-debug-input (buffer)
+  "Handle an need-input request from BUFFER."
+  (with-current-buffer buffer
+    (nrepl-request:stdin
+     ;; For now we immediately try to read-char. Ideally, this will
+     ;; be done in a minor-mode (like edebug does) so that the user
+     ;; isn't blocked from doing anything else.
+     (concat (cider--debug-read-command) "\n")
+     (cider-stdin-handler buffer))))
+
+
+;;; User commands
+;;;###autoload
+(defun cider-debug-defun-at-point ()
+  "Instrument the top-level expression at point.
+If it is a defn, dispatch the instrumented definition. Otherwise,
+immediately evaluate the instrumented expression.
+
+While debugged code is being evaluated, the user is taken through the
+source code and displayed the value of various expressions. At each step,
+the following keys are available:
+    n: Next step
+    c: Continue without stopping
+    i: Inject a value at this point"
+  (interactive)
+  (cider--debug-init-connection)
+  (let* ((expression (cider-defun-at-point))
+         (eval-buffer (current-buffer))
+         (position (cider-defun-at-point-start-pos))
+         (prefix
+          (if (string-match "\\`(defn-? " expression)
+              "Instrumented => " "=> "))
+         (instrumented (format cider--instrument-format
+                         (buffer-file-name)
+                         position
+                         expression)))
+    ;; Once the code has been instrumented, it can be sent as a
+    ;; regular evaluation. Any debug messages will be received by the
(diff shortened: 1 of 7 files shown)


bbatsov:
Adding a menu for this would be nice. You might also post an animated gif showcasing the prototype as some users might have a hard time updating the client and the middleware by themselves.

Bruce-Connor:
Added gif on both posts.

Bruce-Connor:
When you say menu, do you mean an item under the Cider menu-bar?

bbatsov:
Yes. And maybe a separate debug menu when the debugger is enabled (as in
edebug).

On 15 March 2015 at 18:16, Artur Malabarba <notifications@github.com> wrote:

> When you say menu, do you mean an item under the Cider menu-bar?
>
> —
> <https://github.com/clojure-emacs/cider/pull/1019#issuecomment-81133482>.
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

nrepl-client.el
```diff
@@ -786,6 +786,26 @@ for functionality like pretty-printing won't clobber the values of *1, *2, etc."
 (defvar nrepl-err-handler 'cider-default-err-handler
   "Evaluation error handler.")
 
+(defvar-local nrepl--input-handler-queue nil
```
vspinu: BTW, `nrepl-client.el` already loads [queue.el](http://www.dr-qubit.org/predictive/queue-0.1.1.el). You might consider using that data structure.

nrepl-client.el
```diff
@@ -830,7 +850,13 @@ server responses."
              (when (member "namespace-not-found" status)
                (message "Namespace not found."))
              (when (member "need-input" status)
-               (cider-need-input buffer))
+               (let ((handler
```
vspinu: I am a bit confused, cannot debug need-input events be handled directly in `cider--handle-debug`? This input handler queue doesn't look like a very robust thing to me.

vspinu:
@Bruce-Connor  thanks. Truly great job!

Let's bind this to `C-u C-M-x` to mimic edebug. Printing could be shifted to `C-u C-u`. Not a big deal.

nrepl-client.el
Bruce-Connor: Thanks for the pointer. Will do.

cider-debug.el
```diff
@@ -0,0 +1,159 @@
...
+               input)
+      (cider--debug-read-command))))
+
+(defun cider--need-debug-input (buffer)
+  "Handle an need-input request from BUFFER."
+  (with-current-buffer buffer
+    (nrepl-request:stdin
+     ;; For now we immediately try to read-char. Ideally, this will
+     ;; be done in a minor-mode (like edebug does) so that the user
+     ;; isn't blocked from doing anything else.
```
vspinu: Yes, and also bound to some rarely used but easelly accessible combos (like `C-Q` and `C-C`) to avoidinterference with the editing. The only thing I dislike about edebug is the read-only lock.

nrepl-client.el
Bruce-Connor: There a two things at play here.

1. The input handler queue is not absolutely necessary. We can just use a boolean variable and hardcode `cider--debug-need-input` in there. I just need a way to decide which function gets to handle a `"need-input"` message depending on whether we are debugging or not. 

2. The other factor at play here, which is also why the above is necessary, is that our breakpoints actually consist of two messages. The first is a message defined by us (well, by `cider-nrepl`) that does two things. It tells us where to find the breakpoint in code, and tells us that there's a `debugger.core/break` incoming. The second message is a `"need-input"` sent by the call to `debugger.core/break`. 

    This would be more robust if it were done in one message, but that would involve us redefining a close immitation of `debugger.core/break`, because this macro doesn't provide us all we need at the moment.


`need-input` messages can be received for different reasons. We need to send them to  `cider--handle-debug` only when a previous message has notified us that a `debugger.core/break` message is incoming.

I didn't have to use a queue for that, I could have just used a variable like `waiting-for-debug` and if that's `t` we send the message to `

Bruce-Connor:
> Let's bind this to C-u C-M-x to mimic edebug. Printing could be shifted to C-u C-u. Not a big deal.

I agree. Printing can also always be done with `C-u C-x C-e` anyway.

nrepl-client.el
vspinu: To elaborate a bit more on this. I have just started looking at the code so this is a bit uneducated. 

Wouldn't it be nicer if all the mechanics of the `cider--handle-debug`, `cider--need-debug-input` and the call handler in `cider-debug-defun-at-point` be comprised into one handler to say `debug-eval` op?

Currently you seem to keep one global `debugger-message` which seems to imply that only one function can be debugged at a time. Correct me if I am wrong but, the scenarios like, "in the middle of the debugging of A you decided to debug B" will not work. 

So wouldn't it be more reliable to have a `debug` middleware with an `instrument` op accepting `instrument-id`. Then have `quit`, `step` and `continue` ops that would also keep track of the original `instrument-id`. That is, embed each instrumentation and all the action associated with it in its own "bubble". On the emacs side there will be one handler for everything.

cider-debug.el
Bruce-Connor: Making it read-only is a precaution we might need as well. The way it's currently designed, if the user edits the source file inside or before the instrumented sexp, then the debugger will start to misbehave. It will probably place point in the wrong places and might end up throwing an error while trying to navigate the code (due to calling forward-sexp in an invalid location).

nrepl-client.el
vspinu: > The other factor at play here, which is also why the above is necessary, is that our breakpoints actually consist of two messages. The first is a message defined by us (well, by cider-nrepl) that does two things. It tells us where to find the breakpoint in code, and tells us that there's a debugger.core/break incoming. The second message is a "need-input" sent by the call to debugger.core/break. 

Sorry I haven't seen this when writing my previous post. 

Now I see.  So you are quite restrained by the `debugger.core/break` then.

cider-debug.el
vspinu: > if the user edits the source file inside or before the instrumented sexp, then the debugger will start to misbehave.

The debugging can be aborted if that happens, similarly to what edebug does when the source has been modified. 

Alternatively we can check the region for modifications, and if there have been such since the last instrumentation, re-instrument the definition in the background. Not sure if that's possible at `debugger.core` level.

nrepl-client.el
Bruce-Connor: > Wouldn't it be nicer if all the mechanics of the `cider--handle-debug`, `cider--need-debug-input` and the call handler in `cider-debug-defun-at-point` be comprised into one handler to say debug op?

- `cider-debug-defun-at-point` uses the "eval" op simply because that was easier for me when I was first writting this. It can be made to use a "debug" op which would take care of both instrumenting and `eval`'ing. As a nice side-effect, both the `cider--init-debugger` function and the `debugger-message` variable over at `cider-nrepl` would become unnecessary. 

- Regarding `cider--need-debug-input`, yes, it would be nice for its mechanics to be integrated into this same op. Again, however, that lies on us rewriting `debugger.core/break` to do some arbitrary stuff instead of just spawning a repl which triggers a regular "need-input" message. I'm not saying I can do that, though, because by looking at their code I'm not sure their `break` would work at all without spawning a repl. 

> Currently you seem to keep one global debugger-message which seems to imply that only one function can be debugged at a time. Correct me if I am wrong but the scenarios like "in the middle of the debugging of A you decided to debug B" will not work.

Depends on what you mean.

- You can instrument two different functions (which presumably call each other). Then, when you call one of them the debugger will work seamlessly as they call each other and you'll be able to debug both.

- While you're actually in the middle of debugging (because our input handler immediately calls `read-char`) you can't really do anything other than continue debugging. So no, you can't suddenly decide to instrument another function. This doesn't have anything to do with the fact we use a single message. It's a limitation on how we ask for input at the emacs level. Later on, there should many ways to get around this. I think the easiest here is to drop into a `recursive-edit` and activate a relevant minor-mode instead of calling `read-char`.

> So wouldn't it be more reliable to have a debug middleware with an instrument op accepting instrument-id. Then have quit, step and continue ops that would also keep track of the original instrument-id. 

- The `instrument-id` part can (and probably will) be done as a consequence of item `1.` above. But as it is right now, you can already have multiple instrumented functions simultaneously. The fact that it's all hapenning in a single message doesn't cause any problems.

- Having `quit`, `step` and `continue` ops would require us do item `2.` above and them some. `debugger.core/break` doesn't even have a real "quit execution" command. The `(quit)` function it provides is equivalent to `edebug`'s  "continue" (which is why I've bound it to `c`).

Bruce-Connor: > Now I see. So you are quite restrained by the debugger.core/break then.

Yes. :(

vspinu: >   cider-debug-defun-at-point uses the "eval" op simply because that was easier for me when I was first writting this. 

Actually cider was using this trick quite a lot till fairly recently. Now `eval` op shortcut is somewhat thrown upon. It's a very good start nevertheless.

>   Regarding cider--need-debug-input, yes, it would be nice for its mechanics to be integrated into this same op. Again, however, that lies on us rewriting debugger.core/break 

Yeh. I think I have now a fairly complete picture of what's going on. It's not a big deal anyways.  The only thing is that this elisp input-queue thingy is a bit of an over-engineering IMO. I think a simple check if repl is in a debug-mode should be indeed sufficient.

bbatsov:
@Malabarba Love the new handle. :-)

Malabarba:
Thanks. Just hoping I'll stop being called Bruce now.
Even changed the email on my profile to the blog's domain, even though it only does forwarding.  =P

bbatsov:
Well, to be honest I thought your name was Bruce Connor for a while myself. :-)

Malabarba:
Exactly. :-) It's my own fault, of course.
Just an old nick from my gaming era which stuck around a little too long.

Malabarba:
Squashed. I can remove the keybind if it's not ok.

bbatsov:
Please, drop the `.` in the second commit's title (titles are not sentences) and mention the debugger in both the changelog and the readme.

Malabarba:
> Please, drop the . in the second commit's title (titles are not sentences) and mention the debugger in 
both the changelog and the readme.

Done

bbatsov:
Hmm, now the branch can't be merged. Guess it was not in sync with the current `master`, so you'll have to rebase.

Malabarba:
Rebased

bbatsov:
:tada:

Malabarba:
🍺

arichiardi:
Good job!

arrdem:
:beers:

DesmondHayes:
Thanks for the great work. Happy to see debugging capabilities in Cider.
Here, I've found some issues for you to solve:) (Should I open a new issue?):
1.
```clojure
(defn factorial
  [n]
  (loop [cnt n acc 1]
    (if (zero? cnt)
      acc
      (recur (dec cnt) (* acc cnt)))))
```
C-u C-c C-c to debug factorial gives an error in cider buffer:
```
Failed to instrument  (recur # #)
CompilerException java.lang.UnsupportedOperationException: Can only recur from tail position, compiling:(blah-blah-blah)
```
2.
```clojure
(defn foo
  [v]
  (do v))
```
C-u C-c C-c to debug foo gives an error in cider buffer:
```
Failed to instrument  (do v)
```
and also sometimes hangs cider repl so that only C-c C-c can stop it (at which point it gives a huge error with a stacktrace).

bbatsov:
Open a new issue.

On Tuesday, March 31, 2015, Desmond Hayes <notifications@github.com> wrote:

> Thanks for the great work. Happy to see debugging capabilities in Cider.
> Here, I've found some issues for you to solve:) (Should I open a new
> issue?):
> 1.
>
> (defn factorial
>   [n]
>   (loop [cnt n acc 1]
>     (if (zero? cnt)
>       acc
>       (recur (dec cnt) (* acc cnt)))))
>
> C-u C-c C-c to debug factorial gives an error in cider buffer:
>
> Failed to instrument  (recur # #)
> CompilerException java.lang.UnsupportedOperationException: Can only recur from tail position, compiling:(blah-blah-blah)
>
> 2.
>
> (defn foo
>   [v]
>   (do v))
>
> C-u C-c C-c to debug foo gives an error in cider buffer:
>
> Failed to instrument  (do v)
>
> and also sometimes hangs cider repl so that only C-c C-c can stop it (at
> which point it gives a huge error with a stacktrace).
>
> —
> <https://github.com/clojure-emacs/cider/pull/1019#issuecomment-87991315>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

Malabarba:
@DesmondHayes thanks for catching these, could you open a new issue for them? 

Item 2 is probably an easy fix, we should be able to instrument that for sure.
Item 1 is more of a conceptual challenge. We'll have to discuss a solution.

Malabarba:
@DesmondHayes Nevermind. These should both be fixed on PR https://github.com/clojure-emacs/cider-nrepl/pull/177

DesmondHayes:
@Malabarba I've actually opened issue #1048. Also, have you seen #1049?

johnbendi:
Please how do I turn it off?

Malabarba:
@johnbendi What do you mean by that? The debugger is off until you activate it.

johnbendi:
@malabarba how do I deactivate?                                                                                                                                                                                                                                                                                                                                         Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: Artur MalabarbaSent: Saturday, June 13, 2015 7:48 AMTo: clojure-emacs/ciderReply To: clojure-emacs/ciderCc: John ChijiokeSubject: Re: [cider] Debugging in Cider (#1019)@johnbendi What do you mean by that? The debugger is off until you activate it.

—Reply to this email directly or view it on GitHub.

Malabarba:
@johnbendi 
- If you mean *"how do I quit in the middle of debugging something"*, then hit the `q` key. 
- If you mean *"after I debug a function how do I stop it from activating the debugger every time it is called again?"*, then you just need to evaluate the function's `defn` normally (with `C-M-x`).

Issue #937, opened 2015-01-03
Error buffer is focused when it shouldn't be auto-selected


philc:
Steps to reproduce:

1. `(setq cider-auto-select-error-buffer nil)`
2. Eval `(/ 1 0)` in a clojure file buffer (not the cider repl). See the error buffer appear; it does not have keyboard focus.
3. Eval `(/ 1 0)` in the source buffer again. The error buffer is dismissed for a split second and then shown again. Now it has keyboard focus; it shouldn't.

This is in Cider 0.8.2 from MELPA stable. I upgraded Cider recently and this functoinality broke. It worked as expected in a previous version; I believe 0.8.0.

cichli:
@philc this should be fixed on 0.9.0-SNAPSHOT now, if you're willing to give it a test.

philc:
Thanks guys.

PR #1020, opened 2015-03-15
Ensure point is moved correctly when rendering stacktrace

cichli:
I was investigating #937, and couldn't replicate that exact issue, but did notice that when `cider-auto-select-error-buffer` is nil, the point is not moved to the start of the call stack in the error buffer, and instead moves in the currently selected window.

We have to use `with-selected-window` because the buffer and window point do not move together if the buffer is in a window that is not selected (i.e. `cider-auto-select-error-buffer` is nil) - see the third bullet point [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Window-Point.html). This then necessitates using `with-current-buffer` for the case when the buffer is not popped up (i.e. `cider-show-error-buffer` is not `t`).

Is there a cleaner way of doing this?


bbatsov:
> Is there a cleaner way of doing this?

This solution is clean enough IMO. :+1:

PR #1021, opened 2015-03-16
Add `nrepl-server-ready-function'

abo-abo:
* nrepl-client.el (nrepl-server-ready-function): New defvar.
(nrepl-server-filter): Call `nrepl-server-ready-function' if it's set, and set
it to nil.

This allows things like auto-starting CIDER on custom "C-x C-e" *and* finishing
the eval operation. Also restoring the window configuration etc.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1025,6 +1025,9 @@ Return a newly created process."
              (propertize cmd 'face 'font-lock-keyword-face))
     serv-proc))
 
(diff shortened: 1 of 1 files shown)


bbatsov:
What's wrong with using the connected hook?

abo-abo:
> What's wrong with using the connected hook?

Nothing at all, I just didn't step into `cider-jack-in` far enough to find it:)
Sorry for the noise.

Issue #1022, opened 2015-03-16
cider-browse-ns,  Error -- wrong type argument stringp 40


EricGebhart:
; CIDER 0.9.0snapshot (package: 20150315.528) (Java 1.8.0_25, Clojure 1.6.0, nREPL 0.2.7)
user> 

Something is really not right here.  I traced it down to nrepl-send-sync-request  where the response comes back as a list encoded as a string.  So when it get's back to cider-browse-ns  the 
(mapcar (lambda (format ...   gets a series of characters the first of which is \(  which is 40.

I've never looked at any of this code until today, so I figured it is best to let someone who knows the
design intentions take a look.

I wouldn't be surprised if I had something wrong in my setup.  I just got cider to stop complaining at me about upgrading or installing 0.9.0snapshot  today.   That said, I've never seen browse-ns actually work,
although I can clearly see the namespaces and ns-vars if I debug the elisp.  It's just that they are in a list encoded as a string and not a list.

EricGebhart:
I'm using clojure.tools.namespace  0.2.10.

bbatsov:
Sounds like a duplicate of #1017 (although the solution there is flawed). I'll have a look at this soon.

cichli:
I think this is happening because the middleware linearization algorithm changed in tools.nrepl 0.2.7, and pr-values is now causing the `:value` slot of the response to be printed. We just need to use a different name for the key.

bbatsov:
@cichli Nice catch!

P.S. Thought we had stopped using `:value` everywhere, but I guess this was not the case. :-)

PR #1023, opened 2015-03-16
Rename `value` slot in `ns-list` and `ns-vars` responses

cichli:
Fixes #1017 and #1022.

Middleware ordering changed in nREPL 0.2.7+, so the `:value` slot for these responses is now getting printed by the `pr-values` middleware.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -220,14 +220,18 @@ loaded. If CALLBACK is nil, use `cider-load-file-handler'."
 
 (defun cider-sync-request:ns-list ()
   "Get a list of the available namespaces."
-  (cider--sync-request-value (list "op" "ns-list"
-                                   "session" (nrepl-current-session))))
+  (-> (list "op" "ns-list"
+            "session" (nrepl-current-session))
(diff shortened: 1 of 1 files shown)


cichli:
We could instead add `pr-values` to `expects` for the middleware, but there are other pretty printers like [whidbey](https://github.com/greglook/whidbey) that do the same thing as `pr-values`. It seems simplest to just not use the `:value` slot at all.

EricGebhart:
Just wondering, why not refactor cider--sync-request-value with the changes instead of going around it?
It doesn't seem to be used anywhere else.  These changes would make it an unused function.

bbatsov:
> Just wondering, why not refactor cider--sync-request-value with the changes instead of going around it?
> It doesn't seem to be used anywhere else. These changes would make it an unused function.

It was always meant to be a temporary solution until all the middleware stopped using the `value` slot. Guess we reached this moment and I just removed it.

EricGebhart:
Good to know. Just getting to know the code.

----- Bozhidar Batsov's Original Message -----

PR #1025, opened 2015-03-17
FIX #1024 do not move up at the beginning of the buffer

### LINKED ISSUE #1024: Print to *out* outside REPL fails if REPL buffer is clear
When I evaluate `(prn "Hello world")` from Clojure buffer using `C-x C-e` AND `*cider-repl*` buffer is clear, print fails with:

```
line-move-visual: Beginning of buffer
```

The issue is caused in `cider-repl--end-of-line-before-input-start`. The code moves up unconditionally one line up. This does not work if you are at the beginning of the buffer. `line-move` would be better than `previous-line` (patch follows).

danskarda:


cider-repl.el
```diff
@@ -657,7 +657,7 @@ text property `cider-old-input'."
   "Return the position of the end of the line preceding the beginning of input."
   (save-excursion
     (goto-char cider-repl-input-start-mark)
-    (previous-line)
```
bbatsov: Add a comment explaining this. I don't want someone changing it accidentally.

bbatsov:
Please, update the commit message to `[Fix #1024] Do not move up at the beginning of the buffer`.

cider-repl.el
danskarda: @bbatsov you mean in the commit or in the code?

bbatsov: In the code.

danskarda:
Commit updated.

bbatsov:
:+1:

Issue #814, opened 2014-09-18
`nrepl-ns-buffer` is never changing from user.


vspinu:
I don't have time right now to investigate this but, <kbd> C-c C-k</kbd> nor <kbd>C-c C-n</kbd> and interactive eval commands never change the "user" value of `nrepl-ns-buffer`.

I think this has to do with the fact that `load-file` op (on which everything is based now) always returns "user" ns and I think it's always reset in the default handler.

PR #1027, opened 2015-03-17
Send the complete ns form in interactive eval requests

### LINKED ISSUE #1026: ns-form sent within cider-eval-defun-at-point seems incomplete
ClojureScript project is https://github.com/friemen/zackzack

After starting a REPL using `cider-jack-in` I start the backend and the weasel+piggieback based browser-connected REPL.

Then I connect with the browser using http://localhost:8080/testindex.html

I load for example src/cljs/zackzack/demo/playground.cljs
I eval code using `cider-eval-defun-at-point` for example function `playground-view`.

I receive warnings as shown below and changed code is not loaded/compiled:

```
; CIDER 0.9.0snapshot (package: 20150316.2241) (Java 1.8.0_40, Clojure 1.6.0, nREPL 0.2.8)
zackzack.backend> (start!)
:started
zackzack.backend> (cljs-repl)
<< Started Weasel server on ws://0.0.0.0:9001 >>
Type `:cljs/quit` to stop the ClojureScript REPL
nil
WARNING: Use of undeclared Var zackzack.demo.playground/convert-text at line 23 /home/riemensc/Workspaces/Clojure/workspace/zackzack/src/cljs/zackzack/demo/playground.cljs
WARNING: Use of undeclared Var zackzack.demo.playground/view at line 19 /home/riemensc/Workspaces/Clojure/workspace/zackzack/src/cljs/zackzack/demo/playground.cljs
WARNING: Use of undeclared Var zackzack.demo.playground/textfield at line 20 /home/riemensc/Workspaces/Clojure/workspace/zackzack/src/cljs/zackzack/demo/playground.cljs
WARNING: Use of undeclared Var zackzack.demo.playground/textfield at line 21 /home/riemensc/Workspaces/Clojure/workspace/zackzack/src/cljs/zackzack/demo/playground.cljs
WARNING: Use of undeclared Var zackzack.demo.playground/button at line 22 /home/riemensc/Workspaces/Clojure/workspace/zackzack/src/cljs/zackzack/demo/playground.cljs
```
I added `(message ns-form)` in cider-interaction.el, function `cider--dummy-file-contents`.
In *Messages* buffer the line `(ns zackzack.demo.playground)` appears, but the complete ns-form in cljs file is:

```
(ns zackzack.demo.playground
  (:require [examine.constraints :as c]
            [examine.core :as e]
            [clojure.string :as string]
            [zackzack.specs :refer [action-link button checkbox column
                                    datepicker panel
                                    selectbox table textfield view]])
  (:require-macros [examine.macros :refer [defvalidator]]
                   [cljs.core.async.macros :refer [go]]))
```

I suspect the ns-form sent to compiler is incomplete, which causes the warnings.
Around end of November 2014 this worked flawlessly, so the problem appeared very recently.

Oh, and not to forget: I really appreciate all the excellent work that you have done for the Emacs/Clojure community!

cichli:
Fixes #1026.

See [CLJS-1139](http://dev.clojure.org/jira/browse/CLJS-1139) - `ns` forms are not currently additive in ClojureScript REPLs, so until that's fixed, I think we should send the full form.

This was previously changed to send the full form in #807 and then changed back in #814 - I can confirm that the latter has not regressed from this change. 

AFAIK the only possible side-effect from this would be the`:reload` / `:reload-all` options, which we already remove.

cc @vspinu - does this sound reasonable?

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1439,25 +1439,36 @@ the printed result, and defaults to `fill-column'."
    nil
    (or right-margin fill-column)))
 
+(defun cider--dummy-file-ns-form (form)
+  "Construct a `(ns ...)` form for `cider--dummy-file-contents'.
+
+If there is no current ns form, or if FORM is a ns form itself, returns an
+empty string. If the current ns form has not yet been evaluated, returns
+the current ns form. Otherwise, returns the current ns form with
+all :reload and :reload-all directives removed."
+  (let ((current-ns-form (cider-ns-form))
+        (ns-form (cond
+                  ((null current-ns-form) "")
+                  ((cider-ns-form-p form) "")
+                  ((null cider--cached-ns-form) current-ns-form)
+                  ;; previously, we were sending a truncated version of
+                  ;; the current ns form if it was equal to the cached ns
+                  ;; form, but this causes problems with ClojureScript,
+                  ;; where evaluating ns forms is not additive - see #1026
+                  (:default (replace-regexp-in-string ":reload\\(-all\\)?\\>"
+                                                      ""
+                                                      current-ns-form)))))
+    (setq cider--cached-ns-form current-ns-form)
+    ns-form))
+
 (defun cider--dummy-file-contents (form start-pos)
   "Wrap FORM to make it suitable for `cider-request:load-file'.
 START-POS is a starting position of the form in the original context."
-  (let* ((cur-ns-form (cider-ns-form))
-         (ns-form (cond
-                   ((or (null cur-ns-form)
-                        (cider-ns-form-p form))
-                    "")
-                   ((string= cur-ns-form cider--cached-ns-form)
-                    (format "(ns %s)" (cider-current-ns)))
-                   ((null cider--cached-ns-form)
-                    cur-ns-form)
-                   (t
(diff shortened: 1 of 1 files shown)


bbatsov:
> See CLJS-1139 - ns forms are not currently additive in ClojureScript REPLs, so until that's fixed, I think we should send the full form.

Yes.

> This was previously changed to send the full form in #807 and then changed back in #814 - I can confirm that the latter has not regressed from this change.

Yeah, we were incorrectly believing the `ns` forms weren't additive; turned out we had another problem.

While we're on this, you might also comment the logic in the function a bit. If it's confusing to you I can imagine few people would normally understand what are we doing here. :-)

vspinu:
Oh. We have been dancing around this for at least half a year now.  #830 and #956 are behind the most recent code change. 

@cchili Could you please just comment those two lines and reference all the relevant issues in the comment? Thanks.

cichli:
Yeah, I'll add some notes to the code now. FWIW I see little-to-no performance difference between `cider-eval-last-sexp` and `cider-eval-defun-at-point` (the former using plain `eval` and the latter using `load-file` with the full `ns` declaration), so it seems performance is no longer an issue. Certainly neither is taking 1-2 seconds even for a big ns declaration.

cichli:
Hang on, don't merge this yet...

PR #1028, opened 2015-03-18
Add password listener to `nrepl--ssh-tunnel-filter`

vspinu:
`cider-connect` locks on password authentification without this.


bbatsov:
:+1:

PR #1030, opened 2015-03-19
Send the complete ns form in interactive eval requests

cichli:
Continuing on from #1027 (apparently you can't reopen a pull request after force pushing).

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1377,6 +1377,40 @@ otherwise fall back to \"user\"."
 (defvar-local cider--cached-ns-form nil
   "Cached ns form in the current buffer.")
 
+(defun cider--ns-form-for-dummy-file (form)
+  "Construct a `(ns ...)` form for `cider--dummy-file-contents'.
+
+If FORM is a ns form itself, or there is no current ns form, returns an
+empty string. If the current ns form has not yet been evaluated, returns
+the current ns form. Otherwise, returns the current ns form with
+all :reload and :reload-all directives removed."
+  (let ((current-ns-form (cider-ns-form)))
+    (cond
+     ((cider-ns-form-p form) "")
+     ((null current-ns-form) "")
+     ((null cider--cached-ns-form) current-ns-form)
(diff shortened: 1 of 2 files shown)


cider-interaction.el
```diff
@@ -1439,25 +1439,36 @@ the printed result, and defaults to `fill-column'."
    nil
    (or right-margin fill-column)))
 
+(defun cider--dummy-file-ns-form (form)
```
vspinu: I think `cider--ns-form-for-dummy-file` might be more suggestive.

cider-interaction.el
```diff
@@ -1439,25 +1439,36 @@ the printed result, and defaults to `fill-column'."
...
+                   ((cider-ns-form-p form) "")
+                   ((null cider--cached-ns-form) current-ns-form)
+                   ;; previously, we were sending a truncated version of
+                   ;; the current ns form if it was equal to the cached ns
+                   ;; form, but this causes problems with ClojureScript,
+                   ;; where evaluating ns forms is not additive - see #1026
+                   (:default (replace-regexp-in-string ":reload\\(-all\\)?\\>"
+                                                       ""
+                                                       current-ns-form)))))
+    (setq cider--cached-ns-form current-ns-form)
```
vspinu: This side effect doesn't belong here. It should be right close to the evaluation call. This will screw things up if this function is used in a different (non evaluation context).

cichli: Yeah, I meant to ask about this - I just moved it from `cider--dummy-file-contents` to preserve behaviour. I'll remove it and make sure that it is being set correctly elsewhere.

cichli:
@vspinu updated based on your feedback, thanks!

cichli:
I've moved `cider--dummy-file-contents` so it's directly above `cider-interactive-source-tracking-eval`, where it's used.

bbatsov:
> I've moved cider--dummy-file-contents so it's directly above cider-interactive-source-tracking-eval, where it's used.

Hmm, totally missed this. Don't know where I was looking.

:+1:

Issue #699, opened 2014-08-04
Cider documentation and navigation commands are non-standard and inefficient  


vspinu:
CIDER's documentation and navigation commands should not immediately act on symbol at point. The current behavior suffers from multiple flaws on which I elaborate with an example.

Assume you have typed `(reduce fun my-coll|)` where `|` denotes the cursor. Now you want the documentation for `reduce`. You automatically press `C-c C-d C-d` and you get an error because the doc for `my-coll` doesn't exist. You have a choice now, navigate to `reduce` and invoke the help again, or `C-u C-c C-d C-d`. Both of these choices enforce bad habits and increase cognitive burden. For some reason I tend to choose the first option, which is the worst.

So here is my personal list of why the current situation is bad:

- It makes you think hard before you invoke the command "Is the current symbol available? Is the current symbol that I need? Shall I press C-u or not?" most often you end up calling help twice.
  
- It enforces a bad habit of navigating to the symbol (I hate this one the most). 

- Most of the time you don't care about the symbol at point and you just want to quickly navigate to the doc/sourse/location independently of the current symbol, buffer or other contextual elements.
  
- It is inconsistent with standard emacs and documentation systems in other IDEs which ask the user before the jump.

- As it pops the choice menu only sometimes, depending on the complex context,  it feels aleatory, and thus goes against the principle of least surprise. 

- It creates conceptual problems. Are the doc commands acting on symbol, or on user input? What is the meaning of doc commands in other context than the code buffers? Does it make sense to have doc commands in docview buffers? Should the they act on a buffer in docview or on a symbol at point? Note, that all of these problems are automatically sorted out if you adopt an "act on user input" paradigm.

All of the above is simply because CIDER wants to be efficient and to avoid one <kbd>RET</kbd> press. But  instead it makes you feel that it "stays in your way".

The fix is to investigate the symbols at point and _always_ propose them as defaults. In the example above, both `reduce` and `my-coll` must be at the top of the completion list (in that order), so that people who are using IDO can see both of them and quickly select one of those or anything else.

(EDIT: added two more points)

EricGebhart:
Here's a pull request to fix this.
https://github.com/clojure-emacs/cider/pull/1031

danskarda:
Hi,

I would appreciate if you make this change optional or provide two kinds of functions (immediate and with prompt), which user can bind according his preferences.

[...]

Issue #1014, opened 2015-03-13
cider-jump with prefix to open buffer in other window?


expez:
Wanting to read the source of one of the functions I'm calling is the most common use-case for me with `cider-jump`.  To view the call-site and the function definition side-by-side I now have to do `M-.` `C-x 3` `C-x o` `C-x b <RET>`. It would be nice if the jump would happen in the other window when `cider-jump-to-var` was given a prefix.

This might be cool for `cider-jump-to-resource` as well

vspinu:
This would be easy to implement. Both `cider-jump-to-var` and `cider-jump-to-resource` are using `cider-jump-to` which has an optional `other-window` argument that does exactly what you want. 

The problem is with the usability. `C-u` prefix is already taken by `cider-jump-to-var` for manual insertion of the candidate. My personal opinion (explained in detail in  #699) is that current behavior is non-standard and inefficient.  Cider should always display a prompt with a default candidate like other emacs packages do. If #699 is fixed then current `C-u` is freed and could be used for something more useful like the `other-window` behavior that you propose.

bbatsov:
I said it before - PRs welcome. I agree that a lot of things need polish, but sadly have little time to deal with this ATM. 

On a related note - an alternative to the prefix would be alternative keybindings (like `C-x 4 f`).

EricGebhart:
@expez I have this working for both jump-to-var and jump-to-resource, but it's dependent upon #1031, so we'll have to wait to see how that goes.  Not cool to pile up requests. In the mean time, it's in my repo if you want it.

expez:
@EricGebhart awesome!  Just opened another one, which makes sense for the code you're touching now, if you want more to do :)

malcolmsparks:
As far as I can tell, this means I now have to M-. (cider-jump) then hit return, into order to jump. This is hardly an improvement from just typing M-. - is there an alternative function I can bind to M-. ?

vspinu:
> This is hardly an improvement from just typing M-. -

It's a big improvement if you get proper completions (see #699). The list of default completions should include not just the symbol at point but also the surrounding function call, class or method. 

The actual completion list should ideally contain all symbols defined within a session.

malcolmsparks:
I'd still like a quick way of navigating to the symbol under the point,
without having to get a mini-buffer prompt - is there a way that can be
achieved? nrepl-jump is the thing I used 10 times every minute, so I'm not
at all happy with this change.

[...]

PR #1031, opened 2015-03-19
Doc prompt

EricGebhart:
This is changes the behavior of the doc functions to always prompt with a default of symbol at point.  The prompts were given contextual labels instead of 'Symbol:'  where needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@
 * [#953](https://github.com/clojure-emacs/cider/pull/953) Use `sshx` instead of `ssh` in `cider-select-endpoint`
 * [#956](https://github.com/clojure-emacs/cider/pull/956) Eval full ns form only when needed.
 * Enable annotated completion candidates by default.
+* [#1031] (https://github.com/clojure-emacs/cider/pull/1031) Interactive functions prompt with
+  symbol at point as a default value.
 
 ### Bugs fixed
 
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -52,10 +52,10 @@
     (message "Symbol %s not resolved" symbol)))
 
 ;;;###autoload
-(defun cider-grimoire-web (query)
+(defun cider-grimoire-web ()
   "Open the grimoire documentation for QUERY in the default web browser."
-  (interactive "P")
-  (cider-read-symbol-name "Symbol: " 'cider-grimoire-web-lookup query))
+  (interactive)
+  (cider-read-symbol-name "Grimoire doc for: " 'cider-grimoire-web-lookup))
 
 (defun cider-create-grimoire-buffer (content)
   "Create a new grimoire buffer with CONTENT."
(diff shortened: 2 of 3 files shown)


bbatsov:
Please, take a look at CONTRIBUTING.md (I'm referring to the commit message format in particular).

bbatsov:
Another thing - I can imagine some people actually like the current behaviour, so I'd suggest a config option to toggle between the two.

vspinu:
> I can imagine some people actually like the current behaviour

That might be tricky in light of the proposed `other-window` prefix usage for the jump-to-var(#1014).

EricGebhart:
OK. sorry about that.

I thought about an option to turn it on and off, but that does prevent more useful prefix behaviors like jump to var other window.  This does cause the behavior to match that of many other emacs packages, where the current behavior is mostly frustrating.   Hitting return once is a lot easier than typing C-u prefix after figuring out that is what you need to do.

bbatsov:
> I thought about an option to turn it on and off, but that does prevent more useful prefix behaviors like jump to var other window. This does cause the behavior to match that of many other emacs packages, where the current behavior is mostly frustrating. Hitting return once is a lot easier than typing C-u prefix after figuring out that is what you need to do.

Fair enough. Let's see the feedback we'll get. For now just update the commits.

EricGebhart:
What is your preference for connecting the changelog, the issue and the pull request together?

expez:
@EricGebhart I opened an issue about `jump-to-var-other-window` the other day, [here](https://github.com/clojure-emacs/cider/issues/1014).  It'd be awesome if you had the time to fix this!  I'd do it myself but I'm busy working on classloader isolation for [clj-refactor](https://github.com/clojure-emacs/clj-refactor.el) atm.

EricGebhart:
@expez That one is on my radar. I'm in the middle of all that code anyway.  Learning my way around...

bbatsov:
`[Fix #699] Changed interactive doc functions to give prompt.` -> `[Fix #699] Prompt for symbol in interactive doc functions`

bbatsov:
@EricGebhart Commit message titles are titles (meaning they should not end with `.`). :-)

I'm sorry if I sound picky, but I care about every details.

EricGebhart:
It's ok.  All the periods are gone. I'll be sure to leave them off in the future.

bbatsov:
:+1:

malcolmsparks:
Please revert this commit, for the reasons I give in the thread in #1014

grammati:
@bbatsov, since you said above "Let's see the feedback we'll get" - I second @malcolmsparks . Sorry, but I really hate the new behavior. Having to type RET after every M-. is very clunky.

j0ni:
I also am finding this rather annoying. I understand there are arguments for this behaviour, so I would settle for the ability to configure it.

cichli:
You can now set `cider-prompt-for-symbol` to `nil` to disable the prompting behaviour. Passing a prefix argument to any of the source/doc lookup commands will make them do the opposite of that config option, if you wish to override it for just one command.

PR #1034, opened 2015-03-20
Use `face` text property for error buffer buttons

### LINKED ISSUE #1033: Erroneous underlining in error buffer
* The filter buttons are always underlined, even when toggled off
* All frames are underlined

e.g. with `font-lock-mode` enabled/disabled:

screenshot 2015-03-20 12 28 01
screenshot 2015-03-20 12 29 03

It seems the default `button` face is always being used as `button` only cares about the `face` property and not the `font-lock-face` property. If I change `cider-stacktrace.el` to use `face` when rendering the buttons and frames, all is better with both `font-lock-mode` enabled and disabled:

screenshot 2015-03-20 12 36 54
screenshot 2015-03-20 12 37 06

Would you accept a PR to this effect? The change to `font-lock-face` was in 401e819, to fix #749 - but I think in this particular case, things like `parenface` and `rainbow-delimiters` don't really apply. Thoughts?

cichli:
Fixes #1033.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ when in buffer that's not visiting a file (e.g. a REPL buffer).
 * [#1004](https://github.com/clojure-emacs/cider/issues/1004): `:repl-env` key is now filtered from exception causes, as it contains unprintably large strings of compiled javascript when using ClojureScript.
 * Tunneled ssh connection now deals correctly with the ssh password request.
 * [#1026](https://github.com/clojure-emacs/cider/issues/1026): The full `(ns ...)` form for the current buffer is now sent with all source-tracking eval requests, to fix ClojureScript compatibility.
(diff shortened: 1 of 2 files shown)


vspinu:
Confirmed. That fixes it. But it's strange as `font-lock-face` is designed to be used in font-lock modes instead of `face`.

vspinu:
Are there other usages of `font-lock-face` also broken or it only concerns buttons?

cichli:
Yeah, I haven't dug in too deeply but this seems to be a `button.el` specific thing. As far as I can tell this is the only place where we're not using the default button face on buttons (we also use buttons in apropos, doc, and test modes, but I'm not seeing this specific issue in any of them).

bbatsov:
:+1: Guess I overdid it when trying to fix the REPL font-locking. :-)

Issue #1032, opened 2015-03-20
Combine cider-jump-to-var and cider-jump-to-resource


expez:
We should be able to spot the difference between a var and resource easily and just do what the user wants.  This has the additional benefit of freeing up a keybinding.

PR #1036, opened 2015-03-20
[#1032] Combine jump-to-var and jump-to-resource into one function.

EricGebhart:
This adds an entirely new function jump-to-resource-or-var.  
A prompt is given with thing at point as the default.  Thing-at-point uses 'filename as
that is more leniant than 'symbol.
First an attempt is made to find the resource. Upon failure jump to var is attempted.
C-u prefix will cause the result to go to other-window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@
 * New defcustom, `cider-prompt-for-symbol`. Controls whether to prompt for
   symbol when interactive commands require one. Defaults to t, which always
   prompts. Currently applies to all documentation and source lookup commands.
+* [#1032](https://github.com/clojure-emacs/cider/issues/1032) New functions, `cider-find-dwim` and 
+  `cider-find-dwim-other-window`. These functions combine the functionality of `cider-jump-to-var` and 
+  `cider-jump-to-resource`. Which are now renamed to `cider-find-var` and `cider-find-resource` respectively.
+
 
 ### Changes
 
@@ -51,7 +55,7 @@
 * [#953](https://github.com/clojure-emacs/cider/pull/953) Use `sshx` instead of `ssh` in `cider-select-endpoint`
 * [#956](https://github.com/clojure-emacs/cider/pull/956) Eval full ns form only when needed.
 * Enable annotated completion candidates by default.
-* [#1031] (https://github.com/clojure-emacs/cider/pull/1031) Interactive functions prompt with
+* [#1031](https://github.com/clojure-emacs/cider/pull/1031) Interactive functions prompt with
   symbol at point as a default value.
 
 ### Bugs fixed
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -775,7 +775,44 @@ window."
       (when pos
         (goto-char pos)))))
 
-(defun cider-jump-to-resource (path)
+(defun cider-find-dwim-other-window (symbol-file)
+  "Jump to SYMBOL-FILE at point, place results in other window."
+  (interactive (cider--find-dwim-interactive "Jump to: "))
+  (cider--find-dwim symbol-file 'cider-find-dwim-other-window t))
+
+(defun cider-find-dwim (symbol-file)
+  "Try to jump to the SYMBOL-FILE at point. If thing at point is empty dired
+on project. If var is not found, try to jump to resource of the same name.
+When called interactively, A prompt is given according to the variable
+cider-prompt-for-symbol. A prefix inverts the meaning. A default value of
+thing at point is given when prompted."
+  (interactive (cider--find-dwim-interactive "Jump to: "))
+  (cider--find-dwim symbol-file 'cider-find-dwim))
+
+(defun cider--find-dwim (symbol-file callback &optional other-window)
+  "Try to jump to the SYMBOL-FILE at point, show results in OTHER-WINDOW as indicated.
+CALLBACK upon failure to invoke prompt if not prompted previously.
+If thing at point is empty dired on project."
+  (-if-let (info (cider-var-info symbol-file))
+      (cider--jump-to-loc-from-info info other-window)
+    (progn
+      (cider-ensure-op-supported "resource")
+      (-if-let* ((resource (cider-sync-request:resource symbol-file))
+                 (buffer (cider-find-file resource)))
+          (cider-jump-to buffer 0 other-window)
+        (if (cider--should-prompt-for-symbol current-prefix-arg)
+            (error "Resource or var %s not resolved" symbol-file)
+          (let ((current-prefix-arg (if current-prefix-arg nil '(4))))
+            (call-interactively callback)))))))
+
+(defun cider--find-dwim-interactive (prompt)
+  "Get interactive arguments for jump-to functions using PROMPT as needed."
+  (if (cider--should-prompt-for-symbol current-prefix-arg)
+      (list
+       (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
+    (list (or (thing-at-point 'filename) ""))))  ; No prompt.
+
+(defun cider-find-resource (path)
   "Jump to the resource at the resource-relative PATH.
 When called interactively, this operates on point."
   (interactive (list (thing-at-point 'filename)))
@@ -798,7 +835,7 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
         (cider-jump-to buffer (cons line nil) other-window)
       (error "No source location"))))
 
-(defun cider--jump-to-var (var &optional line)
+(defun cider--find-var (var &optional line)
   "Jump to the definition of VAR, optionally at a specific LINE."
   (-if-let (info (cider-var-info var))
       (progn
@@ -806,9 +843,8 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
         (cider--jump-to-loc-from-info info))
     (error "Symbol %s not resolved" var)))
 
-(defun cider-jump-to-var (&optional arg var line)
+(defun cider-find-var (&optional arg var line)
   "Jump to the definition of VAR, optionally at a specific LINE.
-
 Prompts for the symbol to use, or uses the symbol at point, depending on
 the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
 opposite of what that option dictates."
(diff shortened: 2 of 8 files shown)


cider-interaction.el
```diff
@@ -762,6 +762,32 @@ window."
       (when pos
         (goto-char pos)))))
 
+(defun cider-jump-to-resource-or-var (symbol-file &optional other-window)
```
bbatsov: Seems like a bad idea to allow an optional argument in this interactive command, given that this is actually used only by another interactive command.

bbatsov:
> This adds an entirely new function jump-to-resource-or-var.

Without changing the existing bindings this command will unlikely be used by someone. Maybe the existing commands should be make ordinary functions.

bbatsov:
A further optimization might be to check if the point is within a string - then it would make sense to use jump to resource, otherwise we would do var jump.

EricGebhart:
The optional parameter is a leftover from trying to continue using cider-read-symbol-name. 
I'll remove it and add your other suggestions.

EricGebhart:
ok. jump-to-var and jump-to-resource are regular functions.  I changed the key bindings to point at jump-to-resource-or-var.   Which means there are duplicate key bindings for each.

I didn't want to break anything.  Your call.

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@
 * [#958](https://github.com/clojure-emacs/cider/pull/958) Reuse existing repl
   buffers with dead processes. Users are now informed about existing zombie repl
   buffers and are offered the choice to reuse those for new connections.
+* [#1032](https://github.com/clojure-emacs/cider/issues/1032) New function jump-to-resource-or-var
```
bbatsov: Use full function names here and surround them with backticks.

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
       (when pos
         (goto-char pos)))))
 
+(defun cider-jump-to-resource-or-var (symbol-file)
```
bbatsov: I think `cider-jump-dwim` might be a better name for this.

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
...
+  "Try to jump to the variable at point.
```
bbatsov: The first sentence of the docstring should be a summary of the command. You have to mention `SYMBOL-FILE` in the docstring.

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
...
+   If variable is not found, try to jump to resource of the same name.
+When called interactively, A prompt is given with a default value of
```
bbatsov: , A -> , a

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
...
+thing at point. If a prefix is given then the result is opened in 'other-window'. "
+  (interactive
+   (list (read-string "Jump to: " (thing-at-point 'filename)
+                      nil (thing-at-point 'filename))))
+  (-if-let (info (if (not (string-match "\\(\\.\\)" symbol-file))
```
bbatsov: Not sure what you're trying to do here - are you assuming that a filename has `.` in it>

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
...
+                     (cider-var-info symbol-file)
+                   nil))
+      (cider--jump-to-loc-from-info info current-prefix-arg)
+    (progn
+      (cider-ensure-op-supported "resource")
+      (-if-let* ((resource (cider-sync-request:resource symbol-file))
+                 (buffer (cider-find-file resource)))
+          (cider-jump-to buffer 0 current-prefix-arg)
+        (message "Resource or variable %s not resolved" symbol-file)))))
```
bbatsov: variable and var are different things

cider-interaction.el
```diff
@@ -762,10 +762,34 @@ window."
...
+
+(defun cider-jump-to-resource-or-var-other-window ()
+  "jump to var or resource at point, place results in other window."
```
bbatsov: jump -> Jump

cider-repl.el
```diff
@@ -1002,9 +1002,9 @@ constructs."
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map clojure-mode-map)
     (define-key map (kbd "C-c C-d") 'cider-doc-map)
-    (define-key map (kbd "M-.") 'cider-jump-to-var)
+    (define-key map (kbd "M-.") 'cider-jump-to-resource-or-var)
     (define-key map (kbd "M-,") 'cider-jump-back)
-    (define-key map (kbd "C-c M-.") 'cider-jump-to-resource)
+    (define-key map (kbd "C-c M-.") 'cider-jump-to-resource-or-var)
     (define-key map (kbd "RET") 'cider-repl-return)
```
bbatsov: Don't see the other-window command here.

bbatsov:
I'm also starting to think that the "jump" terminology was a bad idea. Right now I prefer "lookup". 

@cichli @vspinu Any feedback?

expez:
How about `find-` now that these commands are more like `find-variable`, `find-function` etc in emacs?  I think lookup is better suited for actual lookups, like the grimoire command

bbatsov:
> How about find- now that these commands are more like find-variable, find-function etc in emacs?

Good point! Let's use `find-var`, `find-resource` and `find-dwim`.

bbatsov:
Btw, just realized the downside of having a combined command - what are we going to use as the completion candidates when not operating at the point? `find-var` will obviously show known vars, `find-resource` will probably show nothing. For a mix of them - it's hard to determine this.

expez:
Can't we just merge all known vars and all resources on classpath?

vspinu:
> Can't we just merge all known vars and all resources on classpath?

Yes. I think that's pretty much the only way to go. It's a bit tricky with non-ido completion. Should the resources be full file names or base names only, or maybe both.

expez:
>  Should the resources be full file names or base names only, or maybe both.

Anything below the project root is just noise, I think.  If you see `config/logback.xml` that should be more than enough.

EricGebhart:
This version is completely different from the last commit.  The new cider-prompt-for-symbol functionality changed things around quite a bit. 

@bbatsov I probably misunderstood your earlier suggestion about the '.'  Checking for a '.' in the name would otherwise be something I would not have done. It seems a bit risky to me.

Currently the new functions are find-dwim and find-dwim-other-window.   The former jump-to-var and jump-to-resource are now non-interactive functions find-var and find-resource.  The string match on '.' is still there.

bbatsov:
The completion problem remains - `cider-find-var` used to have a smart completion and the new command has zilch. Let's do this in two steps. Rework the current PR into something that simply renames the existing commands to `cider-find-var` and `cider-find-resource`. 

Afterwards we'll figure out how to auto-complete resources and how to combine the completion for resources with the completion for vars. As it stands I don't like the proposed command and won't accept it in its current state.

EricGebhart:
I didn't think this iteration would go, but it seemed silly to continue having discussions about code that had not been rebased or integrated.

The completion and history for vars is easy,  resources never had completion or history.   I already had that working by the time I read this, so I'll push that and see what you say.  I haven't had much time to look at this today.

EricGebhart:
This does of course still leave adding completions for resources to be done.  But it does provide the same functionality as before with the added benefit of history for resources albeit mixed with history for vars.

bbatsov:
> I didn't think this iteration would go, but it seemed silly to continue having discussions about code that had not been rebased or integrated.

Therefore my suggestion to do several small changes instead of focusing on merging find-var and find-resource directly.

> The completion and history for vars is easy, resources never had completion or history.

Is it really easy? I've changed it several times over the years and while the current version is pretty nice (it simply reuses the REPL completion), it's not exactly ido friendly (and probably never will be). We cannot just dump all the candidates, as we can do for Emacs Lisp. But we can do this for the resources...

EricGebhart:
Well, it is easy to reuse what you've already been using.   Adding completion for resources will take more.

EricGebhart:
What is missing in the current functionality of find-dwim verses find-var and find-resource?  I don't see anything lacking. There is completion, history, the new prompting behavior with prefix and the addition of the other-window version of the function.  It seems to me that anything beyond that belongs in another PR.

bbatsov:
> What is missing in the current functionality of find-dwim verses find-var and find-resource? I don't see anything lacking. There is completion, history, the new prompting behavior with prefix and the addition of the other-window version of the function. It seems to me that anything beyond that belongs in another PR.

Right now you'll get var completion in the new command, which is totally unacceptable to me. The new command makes a lot of sense when operating at the point, but its usability is pretty questionable otherwise as mixing two completion sources together is generally a bad idea (not to mention one is not implemented yet). @vspinu guess you know what I mean. :-)

I understand your desire to push forward with this, but until there's a foolproof solution for the problems that bother me I'd rather have this on hold. @expez You suggested this initially, but I hope you agree with my concerns. I'm starting to think this was a bad idea and there's no sensible way to implement it. Dealing with all the added complexity just to free one keybinding is not worth it.

EricGebhart:
Ok. That makes sense to me.  @expez, the function will be there, so you can bind it how you want. I guess we'll see what we can do with it from here.  I can probably make an other window version of find-var and find-resource with that other PR.  But is that something you still want if you have this one?

expez:
I agree that the dwim stuff only makes sense if we're acting on point and not prompting.  I also get that mixing completion sources can be undesirable.  I'm thinking we should just drop this dwim stuff now.

I really want `find-var` and `find-resources` in other window, though.

doc/cider-refcard.tex
```diff
@@ -51,8 +51,8 @@
 \subgroup{Navigation}
 
 \key{M-,}{cider-jump-back}
-\key{M-.}{cider-jump-to-var}
-\key{C-c M-.}{cider-jump-to-resource}
+\key{M-.}{cider-jump-to-resource-or-var}
```
bbatsov: You forgot to update this.

cider-repl.el
```diff
@@ -1002,9 +1002,9 @@ constructs."
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map clojure-mode-map)
     (define-key map (kbd "C-c C-d") 'cider-doc-map)
-    (define-key map (kbd "M-.") 'cider-jump-to-var)
+    (define-key map (kbd "M-.") 'cider-jump-to-resource-or-var)
```
bbatsov: This should be `cider-find-var`.

cider-repl.el
```diff
@@ -1002,9 +1002,9 @@ constructs."
...
     (define-key map (kbd "M-,") 'cider-jump-back)
-    (define-key map (kbd "C-c M-.") 'cider-jump-to-resource)
+    (define-key map (kbd "C-c M-.") 'cider-jump-to-resource-or-var)
```
bbatsov: This should be `cider-find-resource`.

bbatsov:
Overall things look good, apart from my small remarks. As discussed we're now not going to change the default keybindings, so the new commands should have keybindings of their own (or none).

EricGebhart:
I must not have double checked... :-/

bbatsov:
:+1:

PR #1039, opened 2015-03-20
Allow buffers with no file to be formatted

cichli:
Just found this when I had a scratch buffer with some EDN in that I wanted to format. It's easy to work around using the command for formatting the region, but I don't see any reason for this restriction (seeing as the formatting commands for both code and EDN work on strings alone and not files).

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1879,9 +1879,7 @@ If no buffer is provided the command acts on the current buffer."
 
 Uses FORMATTER, a function of one argument, to convert the string contents
 of the buffer into a formatted string."
-  (unless buffer-file-name
-    (error "Buffer %s is not associated with a file" (buffer-name)))
-  (let* ((original (cider-file-string buffer-file-name))
+  (let* ((original (substring-no-properties (buffer-string)))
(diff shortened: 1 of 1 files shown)


bbatsov:
You'll have to rebase this.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 * [#953](https://github.com/clojure-emacs/cider/pull/953) Use `sshx` instead of `ssh` in `cider-select-endpoint`
 * [#956](https://github.com/clojure-emacs/cider/pull/956) Eval full ns form only when needed.
 * Enable annotated completion candidates by default.
+* `cider-format-buffer` and `cider-format-edn-buffer` no longer require the buffer being formatted to be visiting a file.
```
bbatsov: This is an unreleased feature, so this changelog entry is redundant.

cichli:
Cool, rebased and updated. Wasn't sure if the changelog entry was needed as some people do use snapshot - though I guess it would have been reported already if it affected anyone :-).

arrdem:
:+1:

bbatsov:
:+1:

PR #1040, opened 2015-03-21
[Fix #1014] Jump-to other window with prefix

### LINKED ISSUE #1014: cider-jump with prefix to open buffer in other window?

EricGebhart:
Pass the prefix-arg along to sub-functions to indicate 'other-window'.  This adds an optional 'other-window' argument to jump-to-var, cider--jump-to-var and jump-to-resource.  jump-to-var no longer uses cider-read-symbol-name so that the other window option can be passed along.  Jump-to-var only uses the other-window option in the case of non-interactive use, when var is not nil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
 * [#1032](https://github.com/clojure-emacs/cider/issues/1032) New functions, `cider-find-dwim` and
   `cider-find-dwim-other-window`. These functions combine the functionality of `cider-jump-to-var` and
   `cider-jump-to-resource`. Which are now renamed to `cider-find-var` and `cider-find-resource` respectively.
+* [#1014](https://github.com/clojure-emacs/cider/issues/1014) A prefix of <kbd>-</kbd> causes `cider-find-var` and 
+  `cider-find-resource` to show results in other window. Additionally, a double prefix argument <kbd>C-u C-u</kbd>
+  inverts the meaning of `cider-prompt-for-symbol` and shows the results in other window.
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -694,8 +694,8 @@ which nREPL uses for temporary evaluation file names."
 (defun cider-find-file (url)
   "Return a buffer visiting the file URL if it exists, or nil otherwise.
 If URL has a scheme prefix, it must represent a fully-qualified file path
-or an entry within a zip/jar archive. If URL doesn't contain a scheme
-prefix and is an absolute path, it is treated as such. Finally, if URL is
+or an entry within a zip/jar archive.  If URL doesn't contain a scheme
+prefix and is an absolute path, it is treated as such.  Finally, if URL is
 relative, it is expanded within each of the open Clojure buffers till an
 existing file ending with URL has been found."
   (cond ((string-match "^file:\\(.+\\)" url)
@@ -781,18 +781,24 @@ window."
   (cider--find-dwim symbol-file 'cider-find-dwim-other-window t))
 
 (defun cider-find-dwim (symbol-file)
-  "Try to jump to the SYMBOL-FILE at point. If thing at point is empty dired
-on project. If var is not found, try to jump to resource of the same name.
-When called interactively, A prompt is given according to the variable
-cider-prompt-for-symbol. A prefix inverts the meaning. A default value of
-thing at point is given when prompted."
+  "Find and display the SYMBOL-FILE at point.
+
+SYMBOL-FILE could be a var or a resource.  If thing at point is empty
+then show dired on project.  If var is not found, try to jump to resource
+of the same name.  When called interactively, a prompt is given according
+to the variable `cider-prompt-for-symbol'.  A single or double prefix argument
+inverts the meaning.  A prefix of `-` or a double prefix argument causes the
+results to be displayed in a different window.
+A default value of thing at point is given when prompted."
   (interactive (cider--find-dwim-interactive "Jump to: "))
-  (cider--find-dwim symbol-file 'cider-find-dwim))
+  (cider--find-dwim symbol-file `cider-find-dwim
+                    (cider--open-other-window-p current-prefix-arg)))
 
 (defun cider--find-dwim (symbol-file callback &optional other-window)
-  "Try to jump to the SYMBOL-FILE at point, show results in OTHER-WINDOW as indicated.
+  "Find the SYMBOL-FILE at point.
+
 CALLBACK upon failure to invoke prompt if not prompted previously.
-If thing at point is empty dired on project."
+Show results in a different window if OTHER-WINDOW is true."
   (-if-let (info (cider-var-info symbol-file))
       (cider--jump-to-loc-from-info info other-window)
     (progn
@@ -800,27 +806,75 @@ If thing at point is empty dired on project."
       (-if-let* ((resource (cider-sync-request:resource symbol-file))
                  (buffer (cider-find-file resource)))
           (cider-jump-to buffer 0 other-window)
-        (if (cider--should-prompt-for-symbol current-prefix-arg)
+        (if (cider--prompt-for-symbol-p current-prefix-arg)
             (error "Resource or var %s not resolved" symbol-file)
           (let ((current-prefix-arg (if current-prefix-arg nil '(4))))
             (call-interactively callback)))))))
 
 (defun cider--find-dwim-interactive (prompt)
   "Get interactive arguments for jump-to functions using PROMPT as needed."
-  (if (cider--should-prompt-for-symbol current-prefix-arg)
+  (if (cider--prompt-for-symbol-p current-prefix-arg)
       (list
        (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
     (list (or (thing-at-point 'filename) ""))))  ; No prompt.
 
 (defun cider-find-resource (path)
-  "Jump to the resource at the resource-relative PATH.
-When called interactively, this operates on point."
-  (interactive (list (thing-at-point 'filename)))
+  "Find the resource at PATH.
+
+Prompt for input as indicated by the variable `cider-prompt-for-symbol`.
+A single or double prefix argument inverts the meaning of
+`cider-prompt-for-symbol`.  A prefix argument of `-` or a double prefix
+argument causes the results to be displayed in other window.  The default
+value is thing at point."
+  (interactive
+   (list
+    (if (cider--prompt-for-symbol-p current-prefix-arg)
+        (cider-read-from-minibuffer "Resource: " (thing-at-point 'filename))
+      (or (thing-at-point 'filename) ""))))
+
   (cider-ensure-op-supported "resource")
   (-if-let* ((resource (cider-sync-request:resource path))
              (buffer (cider-find-file resource)))
-      (cider-jump-to buffer)
-    (error "Cannot find resource %s" path)))
+      (cider-jump-to buffer nil (cider--open-other-window-p current-prefix-arg))
+    (if (cider--prompt-for-symbol-p current-prefix-arg)
+        (error "Cannot find resource %s" path)
+      (let ((current-prefix-arg (cider--invert-prefix-arg current-prefix-arg)))
+        (call-interactively `cider-find-resource)))))
+
+(defun cider--open-other-window-p (arg)
+  "Test prefix value ARG to see if it indicates displaying results in other window."
+  (let ((narg (prefix-numeric-value arg)))
+    (pcase narg
+      (-1 t) ; -
+      (16 t) ; empty empty
+      (narg nil))))
+
+(defun cider--invert-prefix-arg (arg)
+  "Invert the effect of prefix value ARG on `cider-prompt-for-symbol'.
+
+This function preserves the `other-window' meaning of ARG."
+  (let ((narg (prefix-numeric-value arg)))
+    (pcase narg
+      (16 -1)   ; empty empty -> -
+      (-1 16)   ; - -> empty empty
+      (4 nil)   ; empty -> no-prefix
+      (nil 4)))) ; no-prefix -> empty
+
+(defun cider--prefix-invert-prompt-p (arg)
+  "Test prefix value ARG for its effect on `cider-prompt-for-symbol`."
+  (let ((narg (prefix-numeric-value arg)))
+    (pcase narg
+      (16 t) ; empty empty
+      (4 t)  ; empty
+      (narg nil))))
+
+(defun cider--prompt-for-symbol-p (&optional prefix)
+  "Check if cider should prompt for symbol.
+
+Tests againsts PREFIX and the value of `cider-prompt-for-symbol'.
+Invert meaning of `cider-prompt-for-symbol' if PREFIX indicates it should be."
+  (if (cider--prefix-invert-prompt-p prefix)
+      (not cider-prompt-for-symbol) cider-prompt-for-symbol))
 
 (defun cider--jump-to-loc-from-info (info &optional other-window)
   "Jump to location give by INFO.
@@ -835,8 +889,18 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
         (cider-jump-to buffer (cons line nil) other-window)
       (error "No source location"))))
 
+(defun cider--find-var-other-window (var &optional line)
+  "Find the definition of VAR, optionally at a specific LINE.
+
+Display the results in a different window."
+  (-if-let (info (cider-var-info var))
+      (progn
+        (if line (setq info (nrepl-dict-put info "line" line)))
+        (cider--jump-to-loc-from-info info t))
+    (error "Symbol %s not resolved" var)))
+
 (defun cider--find-var (var &optional line)
-  "Jump to the definition of VAR, optionally at a specific LINE."
+  "Find the definition of VAR, optionally at a specific LINE."
   (-if-let (info (cider-var-info var))
       (progn
         (if line (setq info (nrepl-dict-put info "line" line)))
@@ -844,17 +908,22 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
     (error "Symbol %s not resolved" var)))
 
 (defun cider-find-var (&optional arg var line)
-  "Jump to the definition of VAR, optionally at a specific LINE.
-Prompts for the symbol to use, or uses the symbol at point, depending on
-the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
-opposite of what that option dictates."
+  "Find definition for VAR at LINE. Prompt according to prefix ARG
+and `cider-prompt-for-symbol'.
+
+A single or double prefix argument inverts the meaning of
+`cider-prompt-for-symbol. A prefix of `-` or a double prefix argument causes
+the results to be displayed in a different window.  The default value is
+thing at point."
   (interactive "P")
   (cider-ensure-op-supported "info")
   (if var
       (cider--find-var var line)
     (funcall (cider-prompt-for-symbol-function arg)
              "Symbol: "
-             #'cider--find-var)))
+             (if (cider--open-other-window-p arg)
+                 #'cider--find-var-other-window
+               #'cider--find-var))))
 
 (define-obsolete-function-alias 'cider-jump-to-resource 'cider-find-resource "0.9.0")
 (define-obsolete-function-alias 'cider-jump-to-var 'cider-find-var "0.9.0")


cider-interaction.el
```diff
@@ -785,22 +789,25 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
...
   "Jump to the definition of VAR, optionally at a specific LINE."
   (-if-let (info (cider-var-info var))
       (progn
         (if line (setq info (nrepl-dict-put info "line" line)))
-        (cider--jump-to-loc-from-info info))
+        (cider--jump-to-loc-from-info info other-window))
     (message "Symbol %s not resolved" var)))
 
-(defun cider-jump-to-var (&optional var line)
+(defun cider-jump-to-var (&optional var line &optional other-window)
```
expez: the lambdalist contains two `&optional`.

bbatsov:
I'm not sure using a prefix argument is a good idea here. As I said in the original ticket - it's common to introduce extra commands with `C-x 4` keybindings for such purposes.

vspinu:
> s I said in the original ticket - it's common to introduce extra commands with C-x 4 keybindings for such purposes.

There is `find-tag-other-window` on `C-x 4 .` which should be probably replaced given that `M-.` is rewritten in cider.  On the other hand, that's the only way to still access tags from cider.

EricGebhart:
I am not attached to this. I don't like prefixes that much.  I was in the code, and thought it would be good to resolve the issue.   At the same time, the combined function in #1032 could also be viewed as a resolution to this issue.

EricGebhart:
Ok, @expez,  The basic C-u prefix is going to be used to reverse prompting behavior, and prefixes have already been discouraged for this fix.   C-c M-.  is going to be free with this change, so I suggest we put it there.  This preserves the current keybindings as they are so there would be no surprises for anyone other than  C-c M-. will show it's results in the other window, and that both keybindings can now jump-to-var or resource.

expez:
@EricGebhart sounds good to me!

bbatsov:
Sounds reasonable.

EricGebhart:
I think this PR is no longer pertinent due to the functionality of #1032 which does this and more.

EricGebhart:
Before I go adding this in I'd like a clear consensus on how this should behave.  
@expez, @vspinu ?

It seems a prefix is needed and it should probably work with C-u so prompting can be controlled.  C-x 4  is usually the 'other-window' prefix,  So do we create C-x 4 M-.  and C-x 4 C-c M-. ?    Or do we add C-u modifiers?   C-u +  for other window?  Which would also imply M-+.   What about the combination of inverted prompting and other-window?   Should we just have numbers to indicate the combo?    C-u = 4 = invert prompt.  C-u 1 = M-1 = 1=other-window, C-u 2 = M-2 = invert prompt and other-window?  I'm not saying those should be the choices, but C-u or M- with a number or character is less typing than the C-x 4 combinations.

C-x also implies creating a separate other-window function for each function, where the prefix argument model does not.

I don't know.   What makes sense?   

Another question is should this carry over to the doc commands. With the exception of find-resource all of them share the same codebase,  so it would be easy to modify them to use the prefix arg conistently.

expez:
The current pr looks perfect to me.  I want `find-var` and `find-resource` to open in another window when given a prefix.  Whatever lands in `master` that's what I'm hacking together to put in my `.emacs`.  If you can keep that use-case in mind when you rewrite this, to do whatever everyone else wants, I'd be very happy.

EricGebhart:
@expez, That sounds good to me.  Lets see what everyone else says.

vspinu:
> t seems a prefix is needed and it should probably work with C-u so prompting can be controlled. C-x 4 is usually the 'other-window' prefix, So do we create C-x 4 M-. and C-x 4 C-c M-. ? Or do we add C-u modifiers? C-u + for other window? Which would also imply M-+. What about the combination of inverted prompting and other-window? Should we just have numbers to indicate the combo? C-u = 4 = invert prompt. C-u 1 = M-1 = 1=other-window, C-u 2 = M-2 = invert prompt and other-window? I'm not saying those should be the choices, but C-u or M- with a number or character is less typing than the C-x 4 combinations.

I am not a very big fan of <kbd>C-x 4</kbd> for the simple reason that there will certainly be other functions in the future for which users would like to have "other window" behavior. That would mean creating an one more function in <kbd>C-x 4</kbd> map for each such usage, very likely implying unavoidable key clashes. <kbd>C-x 4 M-.</kbd> is also pretty awful to type.

Conversely, if we decide on a prefix, we can use the it consistently across all current and future commands that would require "other window" behavior.

My personal taste is to have <kbd>M-0</kbd> or <kbd>M--</kbd> prefix for other window. That would basically mean keeping left thumb on <kbd>M</kbd> and move right ring finger from <kbd>0</kbd> to <kbd>.</kbd> making it very efficient to type.

EricGebhart:
Those sound good to me.  I use Dvorak on a Kinesis keyboard so I can't say what is hard to type or not. 
Although I can clearly see that ```C-x 4 - whatever``` is painful no matter what keyboard is used.

EricGebhart:
Ok.  ```M-0``` and ```M--``` cause other-window behavior.  ```M--``` also inverts the meaning of the ```cider-prompt-for-symbol```.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
       (-if-let* ((resource (cider-sync-request:resource symbol-file))
                  (buffer (cider-find-file resource)))
           (cider-jump-to buffer 0 other-window)
-        (if (cider--should-prompt-for-symbol current-prefix-arg)
+        (if (cider--prompt-for-symbol? current-prefix-arg)
```
bbatsov: That's not idiomatic in Emacs Lisp - predicate method names should end with `-p`.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
-  "Jump to the resource at the resource-relative PATH.
-When called interactively, this operates on point."
-  (interactive (list (thing-at-point 'filename)))
+  "Find resource at PATH, when called interactively,
+a prompt is given according to the variable `cider-prompt-for-symbol`.
+A `C-u` or `M--` prefix inverts the meaning. A prefix of `M--` or `M-0` causes
+the results to be displayed in other window. The default value is thing at point. "
+  (interactive
+   (if (cider--prompt-for-symbol? current-prefix-arg)
+       (list
```
bbatsov: Wrap the `if` in `list` - saves some repetition.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
              (buffer (cider-find-file resource)))
-      (cider-jump-to buffer)
-    (error "Cannot find resource %s" path)))
+      (cider-jump-to buffer nil (cider--open-other-window? current-prefix-arg))
+    (if (cider--prompt-for-symbol? current-prefix-arg)
+        (error "Cannot find resource %s" path)
+      (let ((current-prefix-arg (cider--invert-prefix-arg current-prefix-arg)))
+        (call-interactively `cider-find-resource)))))
+
+(defun cider--open-other-window? (arg)
```
bbatsov: -p

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
   (cider-ensure-op-supported "resource")
   (-if-let* ((resource (cider-sync-request:resource path))
              (buffer (cider-find-file resource)))
-      (cider-jump-to buffer)
-    (error "Cannot find resource %s" path)))
+      (cider-jump-to buffer nil (cider--open-other-window? current-prefix-arg))
+    (if (cider--prompt-for-symbol? current-prefix-arg)
+        (error "Cannot find resource %s" path)
+      (let ((current-prefix-arg (cider--invert-prefix-arg current-prefix-arg)))
+        (call-interactively `cider-find-resource)))))
```
bbatsov: Why use `?

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
+(defun cider--open-other-window? (arg)
+  "Test ARG to see if it indicates displaying results in other window."
+  (let ((narg (prefix-numeric-value arg)))
+    (cond
+     ((= narg -1) t) ;; M--
+     ((= narg 0) t)  ;; M-0
+     (t nil))))
+
+(defun cider--invert-prefix-arg (arg)
+  "Invert the meaning of ARG as far as it's effect on the behavior set by
```
bbatsov: `checkdoc` won't be happy here.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
+  `cider-prompt-for-symbol` while preserving the other-window meaning of ARG."
+  (let ((narg (prefix-numeric-value arg)))
+    (cond
```
bbatsov: There's also `pcase`.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
+     ((= narg -1) 0)   ;; M-- -> M-0
+     ((= narg 0) -1)   ;; M-0 -> M--
+     ((= narg 4) nil)  ;; C-u -> no-prefix
+     ((= narg nil) 4)))) ;; no-prefix -> C-u
+
+(defun cider--prefix-invert-prompt? (arg)
```
bbatsov: -p

EricGebhart:
I had not completely switched my brain over from clojure. I fixed as many checkdoc errors as I felt comfortable with. Cider-interaction.el is down to 46 errors, some of which are ```two spaces after a period``` which I don't agree with anyway. Good writing should only have one space after a period.

vspinu:
> Good writing should only have one space after a period.

Autofill doesn't respect doublespaces on some occasions. Each time I fill a doc string I have to manually add double spaces because autofill removes them.

EricGebhart:
Double spaces after a period has been a hot topic in recent years.  It used to be that a single emspace was used. Around 1960, the two space rule was invented. It has gone out of style in the last few years. There are lots of articles about it if you google it.  I'm really not that attached either way. But it seems silly to enforce it.  A '.' space and Capital should be enough.

bbatsov:
> Double spaces after a period has been a hot topic in recent years. It used to be that a single emspace was used. Around 1960, the two space rule was invented. It has gone out of style in the last few years. There are lots of articles about it if you google it. I'm really not that attached either way. But it seems silly to enforce it. A '.' space and Capital should be enough.

I agree. Unfortunately all of the Emacs code was written using two spaces and this is unlikely to be ever changed. We have to respect the practices adopted by the Emacs community.

CHANGELOG.md
```diff
@@ -40,7 +40,9 @@
 * [#1032](https://github.com/clojure-emacs/cider/issues/1032) New functions, `cider-find-dwim` and 
   `cider-find-dwim-other-window`. These functions combine the functionality of `cider-jump-to-var` and 
   `cider-jump-to-resource`. Which are now renamed to `cider-find-var` and `cider-find-resource` respectively.
-
+* [#1014](https://github.com/clojure-emacs/cider/issues/1014) A prefix of <kbd>M-0</kbd> or 
+  <kbd>M--</kbd> causes `cider-find-var` and `cider-find-resource` to show results in other window.
```
bbatsov: Why have to prefices that do the same thing? I'm referring to `C-u` added by @cichli recently.

cider-interaction.el
```diff
@@ -79,8 +79,8 @@ navigate to this buffer."
   'cider-show-error-buffer "0.7.0")
 
 (defcustom cider-auto-jump-to-error t
-  "When non-nil automatically jump to error location during interactive compilation.
-When set to 'errors-only, don't jump to warnings."
+  "When non-nil automatically jump to error location during interactive
+compilation. When set to 'errors-only', don't jump to warnings."
```
bbatsov: errors-only is a symbol, not a string...

bbatsov:
Most of the doc changes you did are somewhat strange. Please, limit the scope of this PR only to the changes related to other-window.

cider-interaction.el
```diff
@@ -283,7 +283,7 @@ Signal an error if it is not supported."
        (format "WARNING: Can't determine nREPL's version. Please, update nREPL to %s." cider-required-nrepl-version)))))
 
 (defun cider--check-middleware-compatibility-callback (buffer)
-  "A callback to check if the middleware used is compatible with CIDER."
+  "A callback to check if the middleware used for BUFFER is compatible with CIDER."
```
bbatsov: The doc change here is kind of confusing.

cider-interaction.el
```diff
@@ -389,7 +389,7 @@ to jump back to the last Clojure source buffer."
     (setq cider-last-clojure-buffer buffer)))
 
 (defun cider-switch-to-repl-buffer (&optional arg)
-  "Invoke `cider-switch-to-repl-command'."
+  "Invoke `cider-switch-to-repl-command' using ARG if given."
```
bbatsov: Here as well.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
       (list
        (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
     (list (or (thing-at-point 'filename) ""))))  ; No prompt.
 
 (defun cider-find-resource (path)
-  "Jump to the resource at the resource-relative PATH.
-When called interactively, this operates on point."
-  (interactive (list (thing-at-point 'filename)))
+  "Find resource at PATH prompting as indicated by `cider-prompt-for-symbol`.
+A `C-u` or `M--` prefix inverts the meaning of `cider-prompt-for-symbol`.
```
bbatsov: This is so confusing for an Emacs user - `C-u` is not a prefer argument; it's a way to pass a prefix argument. See the documentation of some similar command.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
-    (error "Cannot find resource %s" path)))
+      (cider-jump-to buffer nil (cider--open-other-window-p current-prefix-arg))
+    (if (cider--prompt-for-symbol-p current-prefix-arg)
+        (error "Cannot find resource %s" path)
+      (let ((current-prefix-arg (cider--invert-prefix-arg current-prefix-arg)))
+        (call-interactively `cider-find-resource)))))
+
+(defun cider--open-other-window-p (arg)
+  "Test ARG to see if it indicates displaying results in other window."
+  (let ((narg (prefix-numeric-value arg)))
```
bbatsov: Why don't you simply pass a numeric prefix arg to the commands directly? This looks odd.

cider-interaction.el
```diff
@@ -800,27 +800,69 @@ If thing at point is empty dired on project."
...
-  "Jump to the resource at the resource-relative PATH.
-When called interactively, this operates on point."
-  (interactive (list (thing-at-point 'filename)))
+  "Find resource at PATH prompting as indicated by `cider-prompt-for-symbol`.
+A `C-u` or `M--` prefix inverts the meaning of `cider-prompt-for-symbol`.
+A prefix of `M--` or `M-0` causes the results to be displayed in other window.
+The default value is thing at point."
+  (interactive
+   (list
+    (if (cider--prompt-for-symbol-p current-prefix-arg)
```
bbatsov: Why don't you use (`interactive "p")`. What you're doing here looks super strange.

cider-interaction.el
```diff
@@ -844,17 +894,21 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
...
 
 (defun cider-find-var (&optional arg var line)
-  "Jump to the definition of VAR, optionally at a specific LINE.
+  "Prompt as indicated by the prefix ARG and `cider-prompt-for-symbol,
+find the definition of VAR, optionally at a specific LINE.
 Prompts for the symbol to use, or uses the symbol at point, depending on
-the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
-opposite of what that option dictates."
+the value of `cider-prompt-for-symbol'. A `C-u` or `M--` prefix inverts
+the meaning. A prefix of `M--` or `M-0` causes the results to be displayed
```
bbatsov: Another confusing docstring.

cider-interaction.el
```diff
@@ -79,8 +79,8 @@ navigate to this buffer."
   'cider-show-error-buffer "0.7.0")
 
 (defcustom cider-auto-jump-to-error t
-  "When non-nil automatically jump to error location during interactive compilation.
-When set to 'errors-only, don't jump to warnings."
+  "When non-nil automatically jump to error location during interactive
```
bbatsov: The first line of a docstring should be a complete sentence.

bbatsov:
See `cider-doc` for a function with an idiomatic documentation for a prefix argument.

EricGebhart:
M-- is not a duplicate behavior, although my current thought is that  ```C-u C-u```  should invert the prompt and send results to other window,   while M-- just sends results to other window.  C-u would continue to invert prompt as it does now. 

According to the emacs doc, reading from the mini-buffer should be done from interactive, not the body, 

```In most cases, you should not call minibuffer input functions in the middle of a Lisp function. Instead, do all minibuffer input as part of reading the arguments for a command, in the interactive specification. See Defining Commands.```

Using "P" and then adding lots of code to the body to read from the mini-buffer seems strange to me. etags has a nice example that is similar to what I am doing.

Ok. I'll revert the checkdoc errors.  It seemed like a good idea to fix at least some of them as I was wading through them to my changes.  If there were no errors it would be a lot easier for everyone to ensure compliance.

**Comment by bbatsov** [comment]
Fair enough. You'll also have to rebase this branch on `master` when ready with the changes.

**Comment by EricGebhart** [comment]
Rebase is part of my flow. It is absolutely necessary in my other projects.

**Comment by bbatsov** [comment]
> BTW, cider-doc does not pass checkdoc. Maybe we need an issue or two or three to fix doc strings so checkdoc can be more easily enforced. There are fairly recent commits that have introduced errors.

Unfortunately I don't have neither the time nor the energy to review all commits in great details. Therefore issue https://github.com/clojure-emacs/cider/issues/1072 - the doc check should simply be automated.

File: CHANGELOG.md [hunk 18]
```diff
@@ -40,6 +40,9 @@
 * [#1032](https://github.com/clojure-emacs/cider/issues/1032) New functions, `cider-find-dwim` and
   `cider-find-dwim-other-window`. These functions combine the functionality of `cider-jump-to-var` and
   `cider-jump-to-resource`. Which are now renamed to `cider-find-var` and `cider-find-resource` respectively.
+* [#1014](https://github.com/clojure-emacs/cider/issues/1014) A prefix of <kbd>-</kbd> causes `cider-find-var` and 
+`cider-find-resource` to show results in other window. Additionally, two empty prefixes <kbd>C-u C-u</kbd>
```
bbatsov: Empty?

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
...
 (defun cider-find-dwim (symbol-file)
-  "Try to jump to the SYMBOL-FILE at point. If thing at point is empty dired
-on project. If var is not found, try to jump to resource of the same name.
-When called interactively, A prompt is given according to the variable
-cider-prompt-for-symbol. A prefix inverts the meaning. A default value of
+  "Find and display the SYMBOL-FILE at point.
+
+SYMBOL-FILE could be a var or a resource.  If thing at point is empty dired on
+project.  If var is not found,try to jump to resource of the same name.  When
+called interactively, A prompt is given according to the variable
```
bbatsov: A -> a

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
   (cider--find-dwim symbol-file 'cider-find-dwim-other-window t))
 
 (defun cider-find-dwim (symbol-file)
-  "Try to jump to the SYMBOL-FILE at point. If thing at point is empty dired
-on project. If var is not found, try to jump to resource of the same name.
-When called interactively, A prompt is given according to the variable
-cider-prompt-for-symbol. A prefix inverts the meaning. A default value of
+  "Find and display the SYMBOL-FILE at point.
+
+SYMBOL-FILE could be a var or a resource.  If thing at point is empty dired on
```
bbatsov: could -> can
dired on project?

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
...
+project.  If var is not found,try to jump to resource of the same name.  When
```
bbatsov: missing comma after found

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
...
+called interactively, A prompt is given according to the variable
+`cider-prompt-for-symbol'.  A prefix inverts the meaning.  A default value of
```
bbatsov: A prefix argument

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
...
   (interactive (cider--find-dwim-interactive "Jump to: "))
   (cider--find-dwim symbol-file 'cider-find-dwim))
 
 (defun cider--find-dwim (symbol-file callback &optional other-window)
-  "Try to jump to the SYMBOL-FILE at point, show results in OTHER-WINDOW as indicated.
+  "Find the SYMBOL-FILE at point.
+
 CALLBACK upon failure to invoke prompt if not prompted previously.
-If thing at point is empty dired on project."
+Show the results in OTHER-WINDOW as indicated.
```
bbatsov: Show results in a different window if OTHER-WINDOW is true.

cider-interaction.el
```diff
@@ -781,18 +781,22 @@ window."
...
+If thing at point is empty display dired on project."
```
bbatsov: again with the `dired`

cider-interaction.el
```diff
@@ -800,27 +804,73 @@ If thing at point is empty dired on project."
...
+  (if (cider--prompt-for-symbol-p current-prefix-arg)
       (list
        (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
     (list (or (thing-at-point 'filename) ""))))  ; No prompt.
 
 (defun cider-find-resource (path)
-  "Jump to the resource at the resource-relative PATH.
-When called interactively, this operates on point."
-  (interactive (list (thing-at-point 'filename)))
+  "Find resource at PATH prompting as indicated by `cider-prompt-for-symbol`.
```
bbatsov: This sentence is pretty confusing. Indicated by?

cider-interaction.el
```diff
@@ -835,6 +885,16 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
         (cider-jump-to buffer (cons line nil) other-window)
       (error "No source location"))))
 
+(defun cider--find-var-other-window (var &optional line)
+  "Jump to the definition of VAR, optionally at a specific LINE.
```
bbatsov: jump -> find

cider-interaction.el
```diff
@@ -835,6 +885,16 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
...
+
+Display the results in `other-window`."
```
bbatsov: `other-window` -> another/different window

EricGebhart:
It's going to take some effort to get the doc strings in line. I couldn't fix them all without doing a fair amount of digging around. Contributors with a broader knowledge of the codebase would have an easier time of it.  Once the code starts passing it should be much easier to keep up with.

vspinu:
The checkdock requirement of keeping the same order of arguments is pretty stupid. English has a rigid sentence structure and abiding to that rule often leads to pretty tonguebreaking results. Emacs core doesn't respect that rule either. Can we selectively disable this one?

EricGebhart:
@vspinu I find that rule quite difficult as well. Especially with functions that take prefix arg as the first argument.  It can be very hard to create a nice sentence out of that.

bbatsov:
> The checkdock requirement of keeping the same order of arguments is pretty stupid. English has a rigid sentence structure and abiding to that rule often leads to pretty tonguebreaking results. Emacs core doesn't respect that rule either. Can we selectively disable this one?

AFAIK it can't be disabled. But we can always sent patches to Emacs. :-)

bbatsov:
Forget what I said, there's a config option for this - `checkdoc-arguments-in-order-flag`.

vspinu:
> here's a config option for this - checkdoc-arguments-in-order-flag.

Nice!

cider-interaction.el
```diff
@@ -844,17 +908,22 @@ OTHER-WINDOW is passed to `cider-jamp-to'."
     (error "Symbol %s not resolved" var)))
 
 (defun cider-find-var (&optional arg var line)
-  "Jump to the definition of VAR, optionally at a specific LINE.
-Prompts for the symbol to use, or uses the symbol at point, depending on
-the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
-opposite of what that option dictates."
+  "Prompt for var according to ARG and `cider-prompt-for-symbol'.
```
bbatsov: Now that `checkdoc-arguments-in-order-flag` is set to nil in the project you can rework such awkward wording of docstrings.

cider-interaction.el
```diff
@@ -800,27 +806,75 @@ If thing at point is empty dired on project."
...
+(defun cider--prefix-invert-prompt-p (arg)
+  "Test prefix value ARG for its effect on `cider-prompt-for-symbol`."
+  (let ((narg (prefix-numeric-value arg)))
+    (pcase narg
+      (16 t) ; empty empty
+      (4 t)  ; empty
+      (narg nil))))
+
+(defun cider--prompt-for-symbol-p (&optional prefix)
+  "Check if cider should prompt.
```
bbatsov: Prompt for what? :-)

bbatsov:
@EricGebhart Ping :-)

PR #1041, opened 2015-03-22
Add support for refresh middleware

cichli:
See clojure-emacs/cider-nrepl#173.

Currently this just works like an interactive eval - the printed results from calling `refresh` are sent to the REPL buffer and the result (either `:ok` or the exception) are printed in the echo area.

`refresh` actually returns the last exception if any are thrown. We can add a custom transport to the middleware if we want to rethrow it and display the error buffer/compilation error highlighting. Will do that as a separate enhancement if you like.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -214,7 +214,7 @@ if the candidate is not namespace-qualified."
 (defvar cider-required-nrepl-ops
   '("apropos" "classpath" "complete" "eldoc" "format-code" "format-edn" "info"
     "inspect-pop" "inspect-push" "inspect-refresh"
-    "macroexpand" "ns-list" "ns-vars"
+    "macroexpand" "ns-list" "ns-vars" "refresh"
     "resource" "stacktrace" "toggle-trace-var" "toggle-trace-ns" "undef")
   "A list of nREPL ops required by CIDER to function properly.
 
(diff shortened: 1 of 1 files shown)


cichli:
btw, this leaves `cider-load-fn-into-repl-buffer` as the only feature still relying on the tooling session. I'll log a separate issue regarding what I think should be done about it :-).

bbatsov:
> btw, this leaves cider-load-fn-into-repl-buffer as the only feature still relying on the tooling session. I'll log a separate issue regarding what I think should be done about it :-).

We deleted it a few days ago. This was functionality with little value that relied on a lot of legacy code (which we deleted as well).

bbatsov:
Btw, this is in a mergeable state, right? You mentioned a fancy UI in the matching cider-nrepl PR, but I guess it's a work in progress.

cichli:
Yeah, so I've got the error conveyance working now (a stacktrace will display in the error buffer if reloading fails), and both this PR + the middleware PR are synced and _could_ be merged now.

But I'm still thinking about how to handle the error recovery (i.e. which namespaces, if any, we should attempt to restore aliases in). Maybe for now we should do no recovery (ala tools.namespace < 0.2.8) and come back to it later.

Regarding the fancy UI, I have the necessary middleware op implemented in a stash locally, but haven't looked into how the elisp-side will work at all yet. I guess [ewoc](http://www.gnu.org/software/emacs/manual/html_node/elisp/Abstract-Display.html) would be a good candidate to use?

If this is blocking the 0.9 release - feel free to merge as is now, and we can add the fancy UI later, or hold off and I will get it all done with the new UI for 0.10. Sorry for the delay around these PRs - it's been a busy few weeks :-).

bbatsov:
> But I'm still thinking about how to handle the error recovery (i.e. which namespaces, if any, we should attempt to restore aliases in). Maybe for now we should do no recovery (ala tools.namespace < 0.2.8) and come back to it later.

Fine by me.

> Regarding the fancy UI, I have the necessary middleware op implemented in a stash locally, but haven't looked into how the elisp-side will work at all yet. I guess ewoc would be a good candidate to use?

Guess so. We use it for the connection management and it works fine.

> If this is blocking the 0.9 release - feel free to merge as is now, and we can add the fancy UI later, or hold off and I will get it all done with the new UI for 0.10. Sorry for the delay around these PRs - it's been a busy few weeks :-).

No worries. I've been super busy myself lately. I do feel that the code is a good shape for a release now, so I'll probably do it pretty soon. 6 months without a new release are way too many. :-)

cichli:
Cool - the descriptor doc needs to be tweaked slightly (doesn't have the `error` slot in the response or the `print-length`/`print-level` slots in the request documented yet), which I'll do next time I'm at my laptop, but feel free to merge in the meantime if you want and I can fix it up afterwards.

bbatsov:
:+1: There a few other things blocking the release - e.g. wrapping the completion changes. Fixing #1059 and #1061 is also on my radar for the release, although we can definitely do it without them.

PR #1043, opened 2015-03-22
Example bind to localhost instead of every interface

hura:
No reason to listen to every network interface so that the websocket would accept connection form anybody that can reach your IP.


cichli:
Good catch. :+1:

bbatsov:
:+1: Thanks!

PR #1046, opened 2015-03-29
Delay creation of error buffer until the error can be rendered

cichli:
Right now when we make a `stacktrace` request, we create (and possibly display) the error buffer when we create the response handler. This means if we get a `no-error` response, or the op fails, or a response is never received, the user has to kill the empty error buffer manually.

This patch changes the error buffer creation to happen only when the contents are ready to render. When using the middleware, we now wait until all causes have been collected. For the non-middleware handler, we now make the `eval` request to print the stacktrace synchronously and render everything when it completes.

I think also making the `stacktrace` op just return a single message with a `:causes` slot would simplify this a bit. We could still make the request asynchronously, but would just render the causes when the response is received (rather than having to collate all the cause messages before rendering like we are now).

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1124,21 +1124,68 @@ They exist for compatibility with `next-error'."
       (goto-next-note-boundary))
     (goto-next-note-boundary)))
 
-(defun cider-default-err-eval-handler (buffer session)
-  "Display in BUFFER the last SESSION exception, without middleware support."
-  (nrepl-request:eval
-   "(clojure.stacktrace/print-cause-trace *e)"
-   (lambda (response)
-     (nrepl-dbind-response response (out)
-       (when out
-         (with-current-buffer buffer
-           (cider-emit-into-color-buffer buffer out)
-           (compilation-minor-mode +1)))))
-   nil
-   session))
+(defun cider--show-error-buffer-p ()
+  "Return non-nil if the error buffer must be shown on error.
 
-(defun cider-default-err-op-handler (buffer session)
-  "Display in BUFFER the last SESSION exception, with middleware support."
+Takes into account both the value of `cider-show-error-buffer' and the
+currently selected buffer."
+  (let* ((selected-buffer (window-buffer (selected-window)))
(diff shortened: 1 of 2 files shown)


bbatsov:
> I think also making the stacktrace op just return a single message with a :causes slot would simplify this a bit. We could still make the request asynchronously, but would just render the causes when the response is received (rather than having to collate all the cause messages before rendering like we are now).

Yeah, that makes sense.

cichli:
> Unless we render them async-ly, streaming them individually doesn't make a lot of sense.

Exactly my thinking. Seeing as we do render them synchronously maybe some way to limit the number of causes returned could also be useful (for pathological cases like where the chain of causes is cyclic).

I've pushed again to kick Travis off btw.

bbatsov:
Btw, seems to me it won't be hard to render the causes asynchronously if we didn't get them in reversed order from the middleware. At any rate, the suggested change will do for now.

cider-interaction.el
```diff
@@ -1124,21 +1124,68 @@ They exist for compatibility with `next-error'."
...
-           (cider-emit-into-color-buffer buffer out)
-           (compilation-minor-mode +1)))))
-   nil
-   session))
+(defun cider--show-error-buffer-p ()
+  "Return non-nil if the error buffer must be shown on error.
 
-(defun cider-default-err-op-handler (buffer session)
-  "Display in BUFFER the last SESSION exception, with middleware support."
+Takes into account both the value of `cider-show-error-buffer' and the
```
bbatsov: This is produce a `checkdoc` warning.

bbatsov:
Btw, that's a bugfix worthy of a changelog entry.

cider-interaction.el
cichli: I'm not seeing a warning here - it's not the first line of the docstring, even if the diff makes it look like that :-). There was a single space between sentences in a different docstring though, which I have fixed.

cichli:
Changelog entry added.

cider-interaction.el
bbatsov: Don't know what I was actually reading. :-)

PR #1053, opened 2015-04-01
Use :debug-value instead of :value

### LINKED ISSUE #1049: Debugger improvement: don't print result of a step to string
Consider an example:
```clojure
(defn foo [x y]
  (symbol (str x "/" y)))
```
I want to debug the following call to 'foo:
```clojure
(foo "bar" "baz")
```
Debugger step results as I press "n":

| N | Now (bad)     | Should be (good) |
|---|---------------|------------------|
| 1 | "\"bar\""     | "bar"            |
| 2 | "\"baz\""     | "baz"            |
| 3 | "\"bar/baz\"" | "bar/baz"        |
| 4 | "bar/baz"     | bar/baz          |

Malabarba:
Should fix inconsistencies between nrepl versions.
Fixes #1049

Note this goes along with https://github.com/clojure-emacs/cider-nrepl/pull/177


bbatsov:
:+1:

Issue #1052, opened 2015-04-01
The debugger doesn't work with nrepl-sync-request:eval


abo-abo:
It appears that the debugger only works with `nrepl-request:eval` currently (that's what `cider-interactive-eval` uses).

Could we get support for `nrepl-sync-request:eval` as well? Currently, it just times out.

Malabarba:
What does `sync-request:eval` do? From the name, I suppose it forces Emacs to wait for a response.
Does it have to wait until `"done"` is received? Or will any response do?

Synchronous debugging should be doable on regular top-level forms, but I can't imagine how it would be done with functions. And I should warn you we were thinking of implementing an operation specifically for the debugger, so it would no longer use `"eval"` at all, and that might make this feature harder to achieve.

bbatsov:
> What does sync-request:eval do? From the name, I suppose it forces Emacs to wait for a response.
> Does it have to wait until "done" is received? Or will any response do?

It waits for `done` or `status` if memory serves. Or timeouts if nothing happens for a configurable sync timeout duration. 

> Could we get support for nrepl-sync-request:eval as well? Currently, it just times out.

What's your use case?

abo-abo:
> It waits for done or status if memory serves

With the new debugger, there's also "need-input", where it bugs out. To the point where I have to restart the REPL.

> What's your use case?

The use case is [lispy](https://github.com/abo-abo/lispy/blob/master/le-clojure.el#L82) and `org-babel-execute:clojure`.
Both do stuff with the value of the sync eval.
The async eval looks to be very inconvenient, since it needs 5 callback functions for each use case.
In `lispy` there are many use cases:

- plain eval: the standard output is concatenated with the value, so async doesn't work, since it has
separate callbacks for stdout and value
- eval and insert
- eval and commented insert
- eval and replace
- eval and flatten
- eval and show function doc
- eval and show function args

I'm fine with the easy solution of immediately returning a bad status for sync eval when
"need-input" is encountered.
But it would be pretty cool if it actually went through and the function that called sync eval
would be kept waiting until the debugger exited.

> And I should warn you we were thinking of implementing an operation specifically for the debugger,
> so it would no longer use "eval" at all, and that might make this feature harder to achieve.

I think it's worth to put in more work to get a better thing, and I'm glad that you did.

Malabarba:
[...]

PR #1055, opened 2015-04-02

abo-abo:
* nrepl-client.el (nrepl-send-sync-request): On encountering "need-input" try to
  call the appropriate handler or `cider-need-input'.

This is the "easy" suggestion fix for #1052. It breaks the sync slightly when the debugger
finally exits, but at least the REPL doesn't hang up any more.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -948,15 +948,27 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
 Hold till final \"done\" message has arrived and join all response messages
 of the same \"op\" that came along."
   (let* ((time0 (current-time))
-         (response (cons 'dict nil)))
+         (response (cons 'dict nil))
+         status)
     (nrepl-send-request request (lambda (resp) (nrepl--merge response resp)))
-    (while (not (member "done" (nrepl-dict-get response "status")))
-      (accept-process-output nil 0.01)
-      ;; break out in case we don't receive a response for a while
-      (when (and nrepl-sync-request-timeout
-                 (> (cadr (time-subtract (current-time) time0))
-                    nrepl-sync-request-timeout))
-        (error "Sync nREPL request timed out %s" request)))
+    (catch 'done
+      (while t
+        (setq status (nrepl-dict-get response "status"))
+        (if (member "done" status)
+            (throw 'done t)
+          (if (member "need-input" status)
+              (let ((handler
+                     (with-current-buffer (nrepl-current-connection-buffer)
+                       (or (queue-dequeue nrepl--input-handler-queue)
+                           #'cider-need-input))))
+                (funcall handler (current-buffer))
+                (throw 'done nil))
+            (accept-process-output nil 0.01)
+            ;; break out in case we don't receive a response for a while
+            (when (and nrepl-sync-request-timeout
+                       (> (cadr (time-subtract (current-time) time0))
+                          nrepl-sync-request-timeout))
+              (error "Sync nREPL request timed out %s" request))))))
     (-when-let* ((ex (nrepl-dict-get response "ex"))
                  (err (nrepl-dict-get response "err")))
       (cider-repl-emit-interactive-err-output err)


bbatsov:
> This is the "easy" suggestion fix for #1052. It breaks the sync slightly when the debugger
> finally exits, but at least the REPL doesn't hang up any more.

That doesn't sound too good. :-) Seems to me that we made a reasonable complex bit of code even more complex, which is also troublesome. Let's hear what @Malabarba thinks about this as well.

nrepl-client.el
```diff
@@ -948,15 +948,27 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
...
-      (when (and nrepl-sync-request-timeout
-                 (> (cadr (time-subtract (current-time) time0))
-                    nrepl-sync-request-timeout))
-        (error "Sync nREPL request timed out %s" request)))
+    (catch 'done
+      (while t
+        (setq status (nrepl-dict-get response "status"))
+        (if (member "done" status)
+            (throw 'done t)
+          (if (member "need-input" status)
```
bbatsov: I can tell you for a fact that most casual readers of the code will have no idea why this special handling is needed here. 
This should either be heavily commented or simplified somehow.

nrepl-client.el
```diff
@@ -948,15 +948,27 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
...
+              (let ((handler
+                     (with-current-buffer (nrepl-current-connection-buffer)
+                       (or (queue-dequeue nrepl--input-handler-queue)
+                           #'cider-need-input))))
+                (funcall handler (current-buffer))
+                (throw 'done nil))
```
Malabarba: Do we really want to throw 'done here? If the user has given the desired input, shouldn't we let the sync process continue until it returns "done"?

nrepl-client.el
```diff
@@ -948,15 +948,27 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
...
+      (while t
+        (setq status (nrepl-dict-get response "status"))
+        (if (member "done" status)
+            (throw 'done t)
+          (if (member "need-input" status)
+              (let ((handler
+                     (with-current-buffer (nrepl-current-connection-buffer)
+                       (or (queue-dequeue nrepl--input-handler-queue)
+                           #'cider-need-input))))
+                (funcall handler (current-buffer))
```
Malabarba: Could you refactor everything from the `let` to the `funcall` into a separate function? 
Then you can use this function also in `nrepl-make-response-handler` and avoid redundant code.
It would also simplify the logic here quite a bit.

Malabarba:
Firstly, thanks for taking the initiative, Oleh.

1. I might be wrong, but I think we don't need the `catch`-`throw` logic. We can keep the loop logic as it was (which will partially reduce the complication here) and simply stay in the loop after the `need-input` message is handled. If the user *wants* to quit, (s)he can simply hit `C-g` at the `need-input` prompt.
   If we do this, just remember to reset `time0` after `need-input` is handled, otherwise we could timeout while the user is thinking about the input.

2. Refactoring the `(let ((handler ...)) (funcall handler ...)`  logic should also simplify this a bit, not to mention reduce redundancy.
 
3. I agree with adding a comment there. It should reference the original GH issue as well (do we prefer full links or just issue number?).

bbatsov:
> I agree with adding a comment there. It should reference the original GH issue as well (do we prefer full links or just issue number?)

I prefer full links. We'll probably have to re-evaluate sync requests in general, btw. They exist mostly to simplify presenting certain info in Emacs (e.g. doc, completions, source location). They were never indented to be used for anything other than this.

nrepl-client.el
abo-abo: > Could you refactor everything from the let to the funcall into a separate function?

I would prefer that you do it, since there's a double `with-current-buffer` wrapper in your variant
of this code.

nrepl-client.el
abo-abo: > Do we really want to throw 'done here? If the user has given the desired input, shouldn't we let the sync process continue until it returns "done"?

It doesn't work without it. It just tries to finish, while the repl is asking for input again.

I've been testing with this code

```clojure
(defn sqr [x]
  (* x x))

(->> [1 2 3 4 5]
     (map sqr))
```

abo-abo:
Any updates? The patch is simple and doesn't cause regressions.
Instead it fixes a full REPL hangup, which is a pretty big issue.
I don't see why this isn't merged yet.

nrepl-client.el
Malabarba: > It doesn't work without it. It just tries to finish, while the repl is asking for input again.

It probably just asks for input again because you hit the next breakpoint.
What did you reply when it asked for input the first time?

abo-abo: I never get a `read-char` prompt, it never gets to that point.

Malabarba:
Sorry. I don't know about @bbatsov, but I'd just like to understand better why that throw is needed. It seems like a misfeature to me.
For instance, if you're running some code that asks the user 2 questions (like size and number of repetitions) why would we quit the sync eval after the first question?

Alternatively, can you provide me with a short emacs command that uses sync-eval for something?
Then I can test this myself and decide whether to accept or rewrite the fix.

abo-abo:
> why would we quit the sync eval after the first question?

This is a work-around, not a perfect solution. We quit because sync
eval simply doesn't handle debug statements.

It's better to throw an error immediately than to fail with a 20
second timeout and have the user restart the REPL.

Here's an example org-mode file:

    * edebug this
    #+begin_src clojure
    (defn sqr [x]
      (* x x))
    #+end_src
    * run this with C-c C-c
    #+begin_src clojure
    (->> [1 2 3 4 5]
         (map sqr))
    #+end_src

bbatsov:
Quick update from me - I'm super busy and I haven't had time to think about this. At any rate, I want us to devise a solid solution for the problem at hand instead of introducing a workaround.

Malabarba:
Alright, I've taken some time to play with this now and I agree this needs a proper solution.
I changed the above code to continue looping (function below), instead of throwing, and that is enough to fix the hang.

## The Good ##

This means synchronous code can successfully work in combination with things that read input.
For instance, typing `M-x cider-eval-last-sexp-and-replace` after `(read-line)` no longer hangs the repl.

## The Bad ##

*However*, there's still something wrong with the code. When doing the above, the user will be asked for input twice on a single `(read-line)`.  The first input seems to be ignored completely, and the second is the one that actually gets used.

I thought that the same response message handled by `nrepl-send-sync-request` was getting caught by some other handler. But I added some debugging messages and it turns out *both* `Stdin:` prompts are coming from `nrepl-send-sync-request`. So I'm a bit at a loss here. Here is the function that I'm using:

    (defun nrepl-send-sync-request (request)
      "Send REQUEST to the nREPL server synchronously.
    Hold till final \"done\" message has arrived and join all response messages
    of the same \"op\" that came along."
      (let* ((time0 (current-time))
             (response (cons 'dict nil))
             status)
        (nrepl-send-request request (lambda (resp) (nrepl--merge response resp)))
        (while (not (member "done" status))
          (setq status (nrepl-dict-get response "status"))
          ;; If we get a need-input message then the repl probably isn't going
          ;; anywhere, and we'll just timeout. So we forward it to the user.
          (if (member "need-input" status)
              (progn (message "HI!")
                     (nrepl--handle-input (current-buffer))
                     (message "HI2!")
                     ;; If the used took a few seconds to respond, we might
                     ;; unnecessarily timeout, so let's reset the timer.
                     (setq time0 (current-time)))
            ;; break out in case we don't receive a response for a while
            (when (and nrepl-sync-request-timeout
                       (> (cadr (time-subtract (current-time) time0))
                          nrepl-sync-request-timeout))
              (error "Sync nREPL request timed out %s" request)))
          (accept-process-output nil 0.01))
        (-when-let* ((ex (nrepl-dict-get response "ex"))
                     (err (nrepl-dict-get response "err")))
          (cider-repl-emit-interactive-err-output err)
          (message err))
        (-when-let (id (nrepl-dict-get response "id"))
          ;; FIXME: This should go away eventually when we get rid of
          ;; pending-request hash table
          (with-current-buffer (nrepl-current-connection-buffer)
            (remhash id nrepl-pending-requests)))
        response))

Malabarba:
Ah wait. I think I see the issue.

Malabarba:
Yes. It's asking twice because `cider-eval-last-sexp-and-replace` evaluates the code twice. ¬¬

Malabarba:
Ok. I think I have a proper fix for this, shall I open a separate PR?

bbatsov:
Sure.

On Saturday, April 11, 2015, Artur Malabarba <notifications@github.com>
wrote:

> Ok. I think I have a proper fix for this, shall I open a separate PR?
>
> —
> <https://github.com/clojure-emacs/cider/pull/1055#issuecomment-91688952>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

Malabarba:
#1074

abo-abo:
@Malabarba, thanks for the effort, it actually works correctly now.

PR #1070, opened 2015-04-05

Malabarba:


Malabarba:
Related to #1055 and should fix #1052. But the purpose of #1055 might still be worth implementing. There may well be other cases where synchronous code asks for input.

If desired, I can merge Oleh's PR into this one, and add the proposed changes to it.

bbatsov:
Thinking just a bit more about this - I think @cichli is right and we should think of a way to handle input properly in sync requests. Simply skipping them doesn't sound quite right.

PR #1074, opened 2015-04-10

Malabarba:
* nrepl-client.el (nrepl-send-sync-request): On encountering
"need-input" try to call the appropriate handler or
`cider-need-input'.

* cider-debug.el (cider--debug-read-command): Be more careful with
what we try to print. If we encounter a similar problem in the future,
the user won't be faced with an error.


Malabarba:
With this code, the situation in #1052 should be resolved. The debugger will work correctly with sync:eval, and the user can just hit `c` if he entered the debugger unintentionally and doesn't want to debug.

bbatsov:
:+1: Looks good!

PR #1075, opened 2015-04-12
Fix evil's string objects not working in the repl

expez:
`ya"` etc didn't work in the repl due to a bug in `cider-end-of-proprange-p`.

@lyro found this when I filed a bug on the evil issue tracker.


bbatsov:
Looks legit. Thanks!

PR #1081, opened 2015-04-22
New debug command

Malabarba:
Add “o” to move out of current sexp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -870,6 +870,7 @@ Keyboard shortcut               | Description
 <kbd>n</kbd> | Next step
 <kbd>c</kbd> | Continue without stopping
 <kbd>i</kbd> | Inject a value into running code
+<kbd>o</kbd> | Move out of the current sexp (like `up-list`)
 
 ### Managing multiple sessions
 
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -40,6 +40,12 @@ can act on it when input is requested."
 Is used to print the value with `cider--debug-read-command' and to step
 through code using the coordinate.")
 
+(defvar cider--debug-point-to-pass nil
+  "Breakpoints are silently skipped until we're past this point.
+Used but debug commands which allow the user to navigate the code, such as
+\"o\". Debug messages will be automatically replied to for as long as the
+point reached by `cider--debug-read-command' is less than this.")
+
 (defconst cider--instrument-format
   (concat "(cider.nrepl.middleware.debug/instrument-and-eval"
           ;; filename and point are passed in a map. Eventually, this should be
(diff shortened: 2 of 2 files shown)


cider-debug.el
```diff
@@ -25,13 +25,26 @@
...
-  "Last value received from the debugger.
-Is printed by `cider--debug-read-command' while stepping through
-code.")
+(cl-defstruct cider--dm
+  "Cider debug-message object.
+Holds all information about the latest debug message received, so that we
+can act on it when input is requested."
+  value coordinates filename point)
+
+(defvar cider--current-debug-message nil
```
bbatsov: Btw, what happens if you're running two debugger sessions? Won't they interfere with each other because this is not tied to a particular session?

cider-debug.el
```diff
@@ -25,13 +25,26 @@
...
 
+(require 'cl-lib)
 (require 'nrepl-client)
 (require 'cider-interaction)
 
-(defvar cider--current-debug-value nil
-  "Last value received from the debugger.
-Is printed by `cider--debug-read-command' while stepping through
-code.")
+(cl-defstruct cider--dm
```
bbatsov: Maybe we should use a longer name here - `dm` could mean a ton of things.

cider-debug.el
```diff
@@ -25,13 +25,26 @@
 
 ;;; Code:
 
+(require 'cl-lib)
```
bbatsov: Do we really need a `cl-struct`? Seems we're not using much of it's functionality. I'm fine with it, but wanted to mention this.

cider-debug.el
```diff
@@ -68,54 +86,71 @@ This will skip over sexps that don't represent objects, such as ^{}."
...
+        "(c)"
+      ;; Get user input.
+      (setq cider--debug-point-to-pass nil)
+      (let ((cider-interactive-eval-result-prefix
+             "(n)ext (c)ontinue (i)nject (o)ut\n => "))
+        (cider--display-interactive-eval-result
+         (or (cider--dm-value msg)
+             "#unknown#")))
+      (let* ((input
+              (cl-case (read-char)
```
bbatsov: We usually use `pcase`.

README.md
```diff
@@ -870,6 +870,7 @@ Keyboard shortcut               | Description
 <kbd>n</kbd> | Next step
 <kbd>c</kbd> | Continue without stopping
 <kbd>i</kbd> | Inject a value into running code
+<kbd>o</kbd> | Move out of the current sexp (like `up-list`)
```
expez: I take it this is in response to my failure to quit the debugger yesterday?

If I'm looking to `quit` the binding `o` with that help text is only getting tried when I've considered the other options.

Is there some other use-case than quitting for calling `o`?  If not how about changing the binding to `q` and just have it call `o` until it's moving out of a toplevel form?

cider-debug.el
Malabarba: I could use an alist or a bunch of variables if you prefer. I'd rather not use just a vector of anonymous entries, as I feel that makes the code hard to read.

cider-debug.el
Malabarba: This variable is only used for a very brief moment of time. So it is possible, but unlikely.
I think I can make it buffer local to the nrepl buffer. That should be enough to avoid conflicts unless the user is running multiple evals simultaneously in the same nrepl session.

cider-debug.el
bbatsov: No that important - we can keep it. I used so far mostly alists and plists. Maybe this was suboptimal. I really wish Elisp had some literal map syntax.

Malabarba:
@expez Did you delete a comment here? I got an email notification but I can't find it here.

expez:
@Malabarba yes, I deleted it because I found the answer after reviewing the code in more detail.

Malabarba:
@expez :+1: I'll have a proper fix for that later this week.

bbatsov:
Looks good. I'd suggest squashing the 4 commits into two (the original structure of the PR).

Malabarba:
I'm having a problem. I'm trying to make these variables buffer-local so that multiple sessions won't conflict, but `(nrepl-current-connection-buffer)` always returns the most recent connection if called in a `.clj` buffer.
Is that intended or a bug?

bbatsov:
> I'm having a problem. I'm trying to make these variables buffer-local so that multiple sessions won't conflict, but (nrepl-current-connection-buffer) always returns the most recent connection if called in a .clj buffer.

At any moment there's only one active connection, regardless of how many projects your working with. This is something I've been meaning to change for a quite a while, but never had the time to actually do it.

Malabarba:
Actually, I think I'll just close this PR for now.

I just realised that implementing our own breakpoint might be very manageable. Which means we won't need clj-debugger, which means all this ugly global variable tracking will become unnecessary.

I'm gonna try to finish this on the plane on Friday.
If it works, most of the cider.el side of the debugger should become obsolete which makes this PR kinda pointless.

bbatsov:
> I just realised that implementing our own breakpoint might be very manageable. Which means we won't need clj-debugger, which means all this ugly global variable tracking will become unnecessary.

Sounds awesome!

> I'm gonna try to finish this on the plane on Friday.
> If it works, most of the cider.el side of the debugger should become obsolete which makes this PR kinda pointless.

Makes sense.

bbatsov:
@Malabarba We can close this, right?

Malabarba:
Thought I had already. :-)

PR #1082, opened 2015-04-22
[Fix #1078] Remove cider-load-fn-into-repl-buffer

### LINKED ISSUE #1078: Function selection (C-c M-f) doesn't seem to work correctly
steps to reproduce:

- try to change namespace (eg. in repl via C-c M-n)
- select function via C-c M-f (all functions from selected namespace are listed correctly)
- when confirmed an exception is thrown: 'Unable to resolve symbol: xxxxx in this context' and namespace gets swtiched to the initial one (boot.user in my case)

CIDER 0.9.0snapshot (package: 20150419.1202)

expez:
This function isn't terrible useful, as it only provides a completing read of
all the vars available in the namespace to insert into the repl.

Furthermore, company-mode already provides the completing read if you just start
typing.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1848,71 +1848,6 @@ For example \"foo.bar.tar\" -> \"foo.bar\"."
 
 ;;; Completion
 
-(defun cider-completing-read-var-select (prompt callback ns selected targets)
-  "Peform completing read using SELECTED and TARGETS.
-If SELECTED is \"..\" then another selection is made for vars in the parent namespace of
-NS using PROMPT.
-If SELECTED is a namespace then another selection is made against that namespace
-using PROMPT.
-Once a selecton is made CALLBACK is called with SELECTED."
-  ;; TODO: immediate RET gives "" as selected for some reason
-  ;; this is an OK workaround though
-  (cond ((equal "" selected)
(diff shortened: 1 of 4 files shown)


cider-interaction.el
```diff
@@ -1889,30 +1889,6 @@ The result of the completing read will be passed to COMPLETING-READ-CALLBACK."
...
-                           (ns-interns '%s)
-                           (ns-refers '%s))))))" ns ns))
-
-(defun cider-load-fn-into-repl-buffer ()
-  "Browse functions available in current repl buffer.
-Once selected, the name of the fn will appear in the repl buffer in parens
-ready to call."
-  (interactive)
-  (let ((ns (cider-current-ns)))
-    (cider-completing-read-sym-form (format "Fn: %s/" ns)
```
bbatsov: I think we can also kill this function and it's friends.

expez: You're right, these were unused outside the namespace according to `ack`. I kept them because they use `cider-` prefix instead of `cider--`.  Hopefully no one else depend on them.

 Force pushed update.

bbatsov: Unfortunately older code didn't make consistent use of the `cider--` convention. There's always room from improvement. Deleting so much code makes me happy!

bbatsov:
:+1:

PR #1083, opened 2015-04-23
[Fix #798] Add hooks for clj{,c,s,x} files

### LINKED ISSUE #798: Clojurescript hook
Hi,

I'm writing an Emacs minor mode that makes sense only for Clojurescript. 
Is there a hook I can use for this case?

expez:
For library authors writing minor modes for `cljs` or similar dialects of
clojure, a hook based on file extension is needed.

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -143,6 +143,42 @@ entirely."
         ["Version info" cider-version]))
     map))
 
+(defcustom cider-cljs-hook nil
+  "List of functions to call when `cider-mode' is activated in a cljs
+file."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
+
+(defcustom cider-clj-hook nil
+  "List of functions to call when `cider-mode' is activated in a clj
+file."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
+
+(defcustom cider-cljx-hook nil
+  "List of functions to call when `cider-mode' is activated in a cljx
+file."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
+
+(defcustom cider-cljc-hook nil
+  "List of functions to call when `cider-mode' is activated in a cljc
+file."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
+
+(defun cider--after-hook ()
+  "Called after `cider-mode' is turned on."
+  (pcase (file-name-extension (buffer-file-name (current-buffer)))
+    ("cljs" (run-hooks cider-cljs-hook))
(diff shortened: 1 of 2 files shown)


bbatsov:
Not sure about the value of this. I'm assuming this hook will result in evaluating some code which also means you have to able to select the proper connection, etc. 

I'd like to see some reasonable use-cases for those hooks, which I find pretty confusing.

bbatsov:
To elaborate on this a bit - what happens if you visit a cljs file without an existing ClojureScript connection? If something is supposed to happen that it should probably be in `clojure-mode`...

expez:
@danielsz you still want?

danielsz:
@expez This looks useful to me. Thank you.
@bbatsov For most usage scenarios, yes, you're looking to do things when clojure-mode kicks in, but for Emacs integration across the board, you probably want to provide those hooks. [This](https://github.com/danielsz/om-mode) minor mode, for example, only makes sense for cljs files.

bbatsov:
So, as I said you seem to agree that a clojurescript hook makes more sense in `clojure-mode`? There could also be a hook for when a clojurescript connection is established.

cider-mode.el
```diff
@@ -151,6 +187,7 @@ entirely."
   nil
   cider-mode-line
   cider-mode-map
+  :after-hook cider--after-hook
```
bbatsov: Technically speaking, this is not a hook.

expez: What makes you say that?  It's an extension point the user can hook into and emacs calls it a hook.

bbatsov: Well, it's a function running some hooks, but it's not a hook per se. Anyways, that's not terribly important.

danielsz:
@bbatsov Yup, it took some time for your comment to sink in my brain.
@expez These hooks will run only after you make a cider connection. Ideally, they should run as soon as you open the clj(s,x) files.

expez:
So what's wanted are`clj{,c,s,x}-hook``clojure-mode` and a `cider-clj{,c,s,x}-connected-hook` in `CIDER`?

There's nothing in your `om-mode` which requires its activation to be delayed until a connection is established (as far as I can see). I propose we just move this PR to `clojure-mod` and add any `connected-hooks` to cider if and when that's ever required.

bbatsov:
Btw, it might make more sense to have different major modes deriving from clojure-mode, as it's feasible to assume there will font-lock differences. Having a clj-hook in something called clojure-mode sounds pretty odd to me. Even now there's some code that's cljx specific, but it's added to the main mode as this was simpler to do.

expez:
Agreed, individual `major-modes` is probably the most sensible thing to do.

danielsz:
Sounds great. Thanks!

Issue #1066, opened 2015-04-04
More keybindings in cider-browse-ns


expez:
I'd forgotten about `cider-browse-ns` (cider is getting large!) and when I tried it again today my expectation was that `RET` would take me to the source but it displays the docs instead.

In any event the keymap could use some more stuff like `f` for `find` and `d` for doc.

bbatsov:
Sure, that makes a lot of sense. For some crazy reason I even thought I had actually done this already. :-)

PR #1084, opened 2015-04-23
Tweaks to cider browse ns

expez:
I'll squash the two commits changing the bindings into one, if you agree, and reference https://github.com/clojure-emacs/cider/issues/1066


bbatsov:
While on the topic of `browse-ns` changes - I was about to rename it to `cider-ns-browser`. You might do this as well.

I'm also not fond of commit titles prefixed with file names, so please update those.

bbatsov:
@expez ping

expez:
I'll leave the renaming for later

bbatsov:
The first commit message is starting with a lowercase letter.

bbatsov:
:+1: Thanks!

PR #1085, opened 2015-04-25
Rewrite cider-debug and document new debugging keys

Malabarba:
Most of the logic has now moved to the clojure side, to the point that
functionality can be added without touching elisp code (though the
readme keybinds will still need to be updated).

Goes with clojure-emacs/cider-nrepl#197.

### DIFF
--- a/README.md
+++ b/README.md
@@ -867,7 +867,10 @@ Keyboard shortcut               | Description
 --------------------------------|-------------------------------
 <kbd>n</kbd> | Next step
 <kbd>c</kbd> | Continue without stopping
+<kbd>o</kbd> | Move out of the current sexp (like `up-list`)
 <kbd>i</kbd> | Inject a value into running code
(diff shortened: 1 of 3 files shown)


Malabarba:
I have a couple of questions about this one.

- Do I remove that input-handler thingy from nrepl.el now that it's no longer needed?
- This debugger will fail miserably if the user is still using a previous cider-nrepl version. Is there something I should do about that? Or do I just trust lein to keep the user up to date?

bbatsov:
> Do I remove that input-handler thingy from nrepl.el now that it's no longer needed?

Yep.

> This debugger will fail miserably if the user is still using a previous cider-nrepl version. Is there something I should do about that? Or do I just trust lein to keep the user up to date?

Trust in `lein`. No further steps are necessary.

Malabarba:
Alright. If the tests pass I'm done.

bbatsov:
Epic work! Thanks!

PR #1092, opened 2015-05-01
enable clojure-mode on the grimoir buffer

AndreaCrotti:
Gives nice syntax highlighting for the grimoir buffer.
There might be smarter ways to do that, but I think that should be good enough..

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -70,6 +70,7 @@ opposite of what that option dictates."
     (read-only-mode -1)
     (insert content)
     (read-only-mode +1)
+    (clojure-mode)
     (goto-char (point-min))
     (current-buffer)))
 


expez:
```markdown
# [org.clojure/clojure "1.7.0-alpha4"] clojure.core/sequential?

## Usages
  [coll]

## Documentation
  Returns true if coll implements Sequential

## User Documentation


## Examples

user=> (sequential? '(1 2 3))
true

user=> (sequential? [1 2 3])
true

user => (sequential? (range 1 5))
true

user=> (sequential? 1)
false

user => (sequential? {:a 2 :b 1})
false
```

I'm not sure highlighting this as clojure code is the right thing to do...

AndreaCrotti:
So well I think the only part that looks a bit weird is the Documentation section to me, the rest looks pretty good and better than no syntax highlighting whatsoever for me..
 
grimoir_clojure_mode

bbatsov:
I think we discussed this with @arrdem in the past and reached the conclusion, that ideally there should be some way to determine what's code and what's markdown. I'm against enabling clojure-mode directly as such buffers are definitely not clojure code.

AndreaCrotti:
Ok yes I understand your point @bbatsov, but the buffer still looks much better in clojure-mode then in no mode..
Also in Clojure-mode it would allow to evaluate stuff straight away?
Anyway alternatively we could use:
https://github.com/purcell/mmm-mode
maybe?

Or could we just make this mode change optional and disable it by default?

bbatsov:
The evaluations come from cider-mode, not clojure-mode and they don't make much sense imo - most snippets have their results shown. My idea for grimoire's buffer was simply to apply `cider-font-lock-as` to the clojure bits in it (but we can't find those now).

Using `mmm-mode` is an overkill for something like this. 

> Or could we just make this mode change optional and disable it by default?

I'd rather not do this. Feel free to keep something like this in your local config.

arrdem:
Agreed that `mmm-mode` is overkill... it also hasn't been terribly stable when I have used it in the past. If there's a prefix or something else that helps for applying `cider-font-lock-as`, then I'm happy to hack it in.

bbatsov:
We just need some delimiters inside which we know we have Clojure code. We scan the content and apply the font-locking.

arrdem:
I'm tepted to reach for say Liquid's `{% highlight Clojure %} .... {% endhighlight %}` or since the rest of the output is Markdown formatted anyway the Githib Markdown tripple backtick is  also an easy option. I'll probably go for the latter. If you actually write code using a tripple backtick, something is very wrong.

bbatsov:
> If you actually write code using a tripple backtick, something is very wrong.

Indeed.

Btw, why are some many of the code snippets including a REPL prompt? Seems to me they'd be better off without it.

arrdem:
http://conj.io/contributing see the "Writing an Example" heading.

The existing dataset has a _lot_ of examples most of which include the REPL prompt and I just haven't had time to rewrite them all yet.

AndreaCrotti:
Another idea about this, since grimoir can also serve HTML instead, could we not just use that and render it easily in Emacs?

It would be better to use plain text but maybe there is even less work to do if we use the HTML output..

bbatsov:
> Another idea about this, since grimoir can also serve HTML instead, could we not just use that and render it easily in Emacs?

Only on Emacs 24.4+.

bbatsov:
Closing this PR. Something which makes use of the markdown formatting added by @arrdem recently would be accepted, but I don't have time to work on it now.

PR #1093, opened 2015-05-02
[Fix #1061] Add cider-find-ns

### LINKED ISSUE #1061: Add cider-find-ns
Here are some very common situations for me:

1. I know some function exists, but I don't recall the exact name.
2. I think maybe some function exists in `lib` and I want to check before I write it myself

I'm solving these now by using `cider-classpath`, `C-s` down to the `lib`, open the jar, look at the list of files and then picking something that looks promising.

`cider-find-ns` would give a completing read of all the available namespaces available on classpath, not just in my own project, and jumps to the selected ns.

expez:
Provide a completing read of available namespaces and jump to the file
containing its definition.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -214,7 +214,7 @@ if the candidate is not namespace-qualified."
 (defvar cider-required-nrepl-ops
   '("apropos" "classpath" "complete" "eldoc" "format-code" "format-edn" "info"
     "inspect-pop" "inspect-push" "inspect-refresh"
-    "macroexpand" "ns-list" "ns-vars" "refresh"
+    "macroexpand" "ns-list" "ns-vars" "ns-path" "refresh"
     "resource" "stacktrace" "toggle-trace-var" "toggle-trace-ns" "undef")
   "A list of nREPL ops required by CIDER to function properly.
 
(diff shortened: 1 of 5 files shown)


cider-interaction.el
```diff
@@ -924,6 +924,31 @@ thing at point."
                  #'cider--find-var-other-window
                #'cider--find-var))))
 
+(defun cider-sync-request:ns-path (ns)
+  "Toggle namespace tracing for NS."
```
bbatsov: Really? :-)

cider-interaction.el
```diff
@@ -924,6 +924,31 @@ thing at point."
...
+  "Find the file containing NS.
+
+A prefix of `-` or a double prefix argument causes
+the results to be displayed in a different window."
+  (interactive "P")
+  (cider-ensure-op-supported "ns-path")
+  (if ns
+      (cider--find-ns ns)
+    (let* ((namespaces (cider-sync-request:ns-list))
+           (ns (completing-read "Namespace: " namespaces)))
```
bbatsov: I think the prompt should say `Find namespace: `.

cider-interaction.el
```diff
@@ -924,6 +924,31 @@ thing at point."
...
+            "ns" ns)
+      nrepl-send-sync-request
+      (nrepl-dict-get "path")))
+
+(defun cider--find-ns (ns &optional other-window)
+  (-if-let (path (cider-sync-request:ns-path ns))
+      (cider-jump-to (cider-find-file path) nil other-window)
+    (error "Can't find %s" ns)))
+
+(defun cider-find-ns (&optional arg ns)
```
bbatsov: This command could benefit from some keybinding.

expez: Yes, but I couldn't think of anything obvious.  Any ideas?

expez:
I've addressed your concerns

cider-interaction.el
bbatsov: Something free. :-) Ideally it'd resemble `M-.` as it serves a similar purpose. Have to think a bit more about this.

expez: `C-c C-.` is unbound for me.

bbatsov: Sounds good. Bind it in both modes and mention it in the README.

expez: Done, also updated the changelog with the new keybinding.

PR #1101, opened 2015-05-13
nrepl-client.el (nrepl-send-sync-request): New optional argument

Malabarba:
If provided, abort the loop immediately and return nil. This avoids
hanging the interface on background requests.


Malabarba:
Posted it now to let the tests run, I haven't actually tested if ti fixes the lag yet. Can only try it at home.

bbatsov:
Looking forward to the results of your experiment!

Malabarba:
Took me a while because of stupid -when-let. But now I can confirm it works for me. Also added some minor changes to another function.

bbatsov:
:+1:

Issue #1105, opened 2015-05-27
cider-doc: There is no ElDoc support in this buffer


tsdh:
Whenever I do `C-c C-d C-d` in a clojure buffer where cider is jacked in, I get the message `There is no ElDoc support in this buffer`.  A second later, I can see the prompt which is correct, i.e., `Doc for: <symbol-at-point>`.  When I hit return, the docs are successfully shown.

Also, eldoc is enabled in the buffer and I do get eldoc signature messages when point is on some function.

My clojure-mode/cider versions are:

```
  cider              20150519.451  installed             Clojure Interactive Development Environment that Rocks
  clojure-mode       20150517.6    installed             Major mode for Clojure code
```

In a new repl, I get the following initial message:

```
; CIDER 0.9.0snapshot (package: 20150519.451) (Java 1.8.0_45, Clojure 1.7.0-RC1, nREPL 0.2.10)
```
I use a current Emacs 25 from git.

I only have very few cider configs which should be harmless:

```elisp
(require 'cider)
(setq nrepl-hide-special-buffers t)
(setq cider-macroexpansion-print-metadata t)

(define-key cider-mode-map (kbd "<f11>") 'cider-switch-to-repl-buffer)
(define-key cider-repl-mode-map (kbd "<f11>") 'delete-window)

(defun th/cider-repl-mode-init ()
  (paredit-mode)
  (eldoc-mode))

(add-hook 'cider-repl-mode-hook 'th/cider-repl-mode-init)

(defun th/cider-mode-init ()
  (eldoc-mode))

(add-hook 'cider-mode-hook 'th/cider-mode-init)
```
The problem is there for a while (say, some months) so it's not that it has been introduced with some recent commit or ELPA version upgrade.

bbatsov:
Is `eldoc-mode` really active in those buffers? Can't think of a reason why something like this would be happening, unless `eldoc-mode` is inactive.

tsdh:
Yes, it is.  But now I found the culprit.

`cider-read-from-minibuffer` runs the hooks in `eval-expression-minibuffer-setup-hook`.  Apparently, that hook contains `eldoc-mode`, and since no `eldoc-documentation-function` is defined in the minibuffer, the message is displayed by `eldoc-mode` which won't be activated in this case.

Having `eldoc-mode` active in the minibuffer is a pretty new and cool Emacs feature.  In this case, the docs are shown in the mode-line.

[...]

PR #1108, opened 2015-05-27
Set eldoc function when reading from minibuffer

tsdh:
Emacs supports eldoc also in the minibuffer.  For that, one can add eldoc-mode
to eval-expression-minibuffer-setup-hook.  cider-read-from-minibuffer runs the
functions in that hook but doesn't set eldoc-documentation-function for the
minibuffer.  Therefore, one gets an eldoc message telling that eldoc isn't
supported when cider-read-from-minibuffer is called and eldoc-mode is in
eval-expression-minibuffer-setup-hook.

This patch sets eldoc-documentation-function locally in the minibuffer before
cider-read-from-minibuffer runs the functions in
eval-expression-minibuffer-setup-hook.  This fixes issue #1105.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -519,6 +519,7 @@ reading input."
         (set-syntax-table clojure-mode-syntax-table)
         (add-hook 'completion-at-point-functions
                   #'cider-complete-at-point nil t)
+        (setq-local eldoc-documentation-function #'cider-eldoc)
         (run-hooks 'eval-expression-minibuffer-setup-hook))
     (read-from-minibuffer prompt initial-value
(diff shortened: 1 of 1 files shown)


expez:
Thanks!  Appreciate the descriptive commit message!

Could you update the commit message so the first line summary begins with "[Fix #1105]" to match the projects [guidelines](https://github.com/clojure-emacs/cider/blob/master/CONTRIBUTING.md)?

tsdh:
Yes, I'll do so later when I come back home.

tsdh:
@expez There you go.

expez:
Thanks!

PR #1109, opened 2015-05-29
Add an option to disable automatic cider-mode

abo-abo:
* cider.el (cider-auto-mode): New defcustom.
(cider--connected-handler): Don't auto-enable `cider-mode' for all
`clojure-mode' buffers when `cider-auto-mode' is nil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### New features
 
+* [#1109](https://github.com/clojure-emacs/cider/issues/1061) New defcustom `cider-auto-mode`.
+On by default, when nil don't automatically enable `cider-mode` in all Clojure buffers.
+
 * [#1061](https://github.com/clojure-emacs/cider/issues/1061) New command `cider-find-ns`, bound to <kbd>C-c C-.</kbd>, which prompts for an ns and jumps to the corresponding source file.
 * [#1019](https://github.com/clojure-emacs/cider/pull/1019): New file, cider-debug.el.
   Provides a new command, `cider-debug-defun-at-point`, bound to <kbd>C-u C-M-x</kbd>.
--- a/cider.el
+++ b/cider.el
@@ -130,6 +130,11 @@ This variable is used by `cider-connect'."
   :group 'cider
   :version "0.9.0")
 
+(defcustom cider-auto-mode t
+  "When non-nil, automatically enable `cider-mode' for all Clojure buffers."
+  :type 'boolean
+  :version "0.9.0")
+
 (defvar cider-ps-running-nrepls-command "ps u | grep leiningen"
   "Process snapshot command used in `cider-locate-running-nrepl-ports'.")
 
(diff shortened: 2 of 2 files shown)


bbatsov:
What's the reasoning here?

abo-abo:
I don't want the extra key bindings and the mode line change. It is a minor mode after all: preferably, it should be opt-in. I ask to at least make it opt-out.

bbatsov:
While this is technically true, CIDER is pretty much unusable without it. The auto-enabling of the mode was borrowed by SLIME. I do have my reservations about it, but I'm also uncertain whether there's a better alternative to the current setup.

abo-abo:
> I do have my reservations about it, but I'm also uncertain whether there's a better alternative to the current setup.

I don't know of a better way either, but CIDER is very usable to me without a minor mode. That's why I propose an opt-out parameter.

Malabarba:
With regards to cider being opt-in, technically, you opt-in when you invoke the `jack-in` or the `connect` command.
Still, I see no reason why we can't have a variable to disable the auto activation of minor modes. Though perhaps a documented defvar might be preferable to a defcustom (so as not to clutter the customization menu).

abo-abo:
> With regards to cider being opt-in, technically, you opt-in when you invoke the jack-in or the connect command.

Not really. What if I just want a REPL? How is `clojure-mode` related to that?  It's a little related, and we have the example of SLIME doing the same thing, but I don't think it's completely the right thing.

I know that this is a recent change, because I haven't seen this behavior before. I bound <kbd>C-l</kbd> to clear the REPL buffer in `cider-mode`.  And I was surprised to find this binding in a `clojure-mode` buffer. I could adjust the key bindings, but the change in mode line is annoying as well.

Malabarba:
> > With regards to cider being opt-in, technically, you opt-in when you invoke the jack-in or the connect command.

> Not really. What if I just want a REPL? How is clojure-mode related to that? It's a little related, and we have the example of SLIME doing the same thing, but I don't think it's completely the right thing.

I'm not saying we shouldn't offer a repl-only option (quite the contrary, think we should). I was just replying to your other comment, that cider isn't opt-in.  
Not being opt-in is when a minor-mode turns itself on as soon as the packge is installed. Cider *is* opt-in, it just doesn't offer (yet) the fine-grained configuration you desire.

bbatsov:
> Not really. What if I just want a REPL? How is clojure-mode related to that? It's a little related, and we have the example of SLIME doing the same thing, but I don't think it's completely the right thing.

Fair point. I was just explaining why it is done as it is now.

> I know that this is a recent change, because I haven't seen this behavior before. I bound C-l to clear the REPL buffer in cider-mode. And I was surprised to find this binding in a clojure-mode buffer. I could adjust the key bindings, but the change in mode line is annoying as well.

This behaviour hasn't been changed since the earliest days of the project. `cider-mode` and `cider-repl-mode` are totally different modes. Not sure what's your problem with the modeline - not knowing the active connection is pretty problematic for everyone using more than one connection. 

I'll think a bit more on how to best tackle this, but you're the first person who has issues with this behaviour for about 3 years now.

bbatsov:
@abo-abo ^^

Btw, you should also mention this change in the changelog.

Issue #1104, opened 2015-05-20
cider-current-ns should be less naive


abo-abo:
I'm using this in the eval routine of `lispy`:

```elisp
(nrepl-sync-request:eval str (cider-current-ns) (nrepl-current-session))
```

But I'm getting a problem if I eval this in a file that I'm just browsing, and hasn't been loaded yet.  It should be possible to do this, for instance I could eval just to get arguments for the built-in function.

Turns out the reason is that `cider-current-ns` just falls back to the regex-based `clojure-find-ns`, that knows nothing of the REPL state. Could we rely on the assumption that it's OK to eval in `cider-current-ns`?

Also before, the eval status at least returned `(member "namespace-not-found" status)`. After some update, it's no longer possible, it just returns status nil, with no further indication.

bbatsov:
> Also before, the eval status at least returned (member "namespace-not-found" status). After some update, it's no longer possible, it just returns status nil, with no further indication.

The idea is that if you get something like this when evaling code in a file, you probably didn't load the file, so we just eval the ns automatically. You can see this interplay in the `*nrepl-messages*`.

> Turns out the reason is that cider-current-ns just falls back to the regex-based clojure-find-ns, that knows nothing of the REPL state. Could we rely on the assumption that it's OK to eval in cider-current-ns?

There's no other way to be certain what the current ns of a file buffer is. Might look lame, but it's actually the best option we have.

abo-abo:
> The idea is that if you get something like this when evaling code in a file, you probably didn't load the file, so we just eval the ns automatically. You can see this interplay in the *nrepl-messages*.

This is being too smart, in my opinion. Imagine Emacs evaluating the current Elisp buffer just because you happen to have it open when you call `eval-expression`. This is the exact same situation: I eval some code just to get the argument list or the docstring of e.g. `ns`. It doesn't mean that I want to load the file.

Also I have no `*nrepl-messages*` buffer on Emacs 25 and latest CIDER.

> There's no other way to be certain what the current ns of a file buffer is. Might look lame, but it's actually the best option we have.

There should be a way to track this, with a buffer-local var perhaps.

In any case, previously I could fall back to "user" for the eval, on encountering "namespace-not-found".  Now, I can't do that any more.

How do you suggest I can eval this, whether the current file is loaded or not:

```
"(if (symbol? 'System/getProperty)
  (if (special-symbol? 'System/getProperty)
    'special
    (or (resolve 'System/getProperty)
[...]

PR #1118, opened 2015-06-08
nrepl-client.el (nrepl-send-sync-request): Fixup

abo-abo:
* nrepl-client.el (nrepl-send-sync-request): Fix "namespace-not-found" status
  being suppressed.

Re #1104

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -955,7 +955,7 @@ sign of user input, so as not to hang the interface."
                       nrepl-sync-request-timeout))
           (error "Sync nREPL request timed out %s" request)))
       ;; Clean up the response, otherwise we might repeatedly ask for input.
-      (nrepl-dict-put response "status" nil)
+      (nrepl-dict-put response "status" (remove "need-input" status))
       (accept-process-output nil 0.01))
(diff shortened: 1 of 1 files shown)


abo-abo:
@Malabarba, could you look at this? It partially reverts your ea449959e06d27d7520c10882db15abf17832da4.
It may be possible to make it more general, I just made the change for a single special case that I care about.

Malabarba:
Oops, guess it was my fault. :-)
I think the correct form here would be just

    (nrepl-dict-put response "status" (remove "need-input" status))

expez:
Could you update the commit message to have the prefix "[Fix #1104]" to match the project's conventions?

bbatsov:
Yep, @Malabarba is right about the proper way to fix this.

abo-abo:
> Yep, @Malabarba is right about the proper way to fix this.

Updated.

> Could you update the commit message to have the prefix "[Fix #1104]" to match the project's conventions?

It doesn't fix #1104, it just works around it.

bbatsov:
> It doesn't fix #1104, it just works around it.

I had totally forgotten what the issue was originally about. :-)

PR #1121, opened 2015-06-11
[Fix #1106] Display local variables in the debugger

### LINKED ISSUE #1106: Show locals table at each step in debugger
Rather than simply showing the last form return value, use the contents of &env to display a table of all the local bindings at a given point, as discussed with @Malabarba in gitter.

```clojure
(defmacro locals [] (zipmap (map #(list 'quote %) (keys &env)) (keys &env)))
```

This will likely include some autogenerated locals (for destructuring, macros, etc.). These aren't likely to be too distracting, but a simple heuristic of whether a name consists of an underscore `#[_]` should filter out all of these (maybe have a toggle?)

```clojure
(defn remove-autogenerated-locals
  "Remove any locals generated by macros (heuristic - named something
  with underscores in it). When we're doing something other than
  println with the data, these could be left in but it's a bit noisy
  at the moment"
  [coll]
  (into {} (remove (fn [[k v]] (re-find  #"[_]" (name k))) coll)))
```

Malabarba:
cider-debug-locals

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -87,7 +87,7 @@ sexp."
 RESPONSE is a message received form the nrepl describing the input
 needed. It is expected to contain at least \"key\", \"input-type\", and
 \"prompt\", and possibly other entries depending on the input-type."
-  (nrepl-dbind-response response (debug-value key coor filename point input-type prompt)
+  (nrepl-dbind-response response (debug-value key coor filename point input-type prompt locals)
     (let ((input))
       (unwind-protect
           (setq input
@@ -97,7 +97,7 @@ needed. It is expected to contain at least \"key\", \"input-type\", and
                   ((pred sequencep)
                    (when (and filename point)
                      (cider--debug-move-point filename point coor))
-                   (cider--debug-read-command input-type debug-value prompt))))
+                   (cider--debug-read-command input-type debug-value prompt locals))))
         ;; No matter what, we want to send this request or the session will stay
         ;; hanged.
         (nrepl-send-request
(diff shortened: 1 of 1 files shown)


Malabarba:
It would be nice to display functions in a prettier way, but I guess that's more of a low-level cider thing, not really a role for the debugger.

cider-debug.el
```diff
@@ -107,16 +107,42 @@ needed. It is expected to contain at least \"key\", \"input-type\", and
                "input" (or input ":quit"))
          #'ignore)))))
 
-(defun cider--debug-read-command (command-list value prompt)
+(defcustom cider-debug-display-locals nil
```
expez: I think when the user sets this to `t` in a `defcustom` they expect the locals to always be shown and that any use of `l` to toggle locals is scoped to the current debugging session.

Malabarba: Possibly, but that was a lot more work (we just don't have a notion of debugging session yet =/), so I compromised a bit and left the toggle as persistent.

I could make it so that `l` toggles locals only for this debugging *step*. That would be very simple, but I don't know if that's preferable to the current state.

expez: How about just renaming it to `cider--debug-display-locals` and making it a `defvariable` instead of a `defcustom`?

Malabarba: Sure. Will do.

bbatsov: I'm not sure making this a `defvar` would be a step in the right direction.

expez: @bbatsov what is the right way of doing this?  Setting a buffer-local variable?

bbatsov: I think there should just be some command to dump the locals on demand. No need for variables at all.

expez: That seems cumbersome, if I understand how this is going to work correctly.  If you're trying to watch the flow of data in your function, wouldn't you have to hit `l` for each step?

bbatsov: We can just drop this and use `l`. Ideally this would affect only the expression you're currently debugging.

bbatsov: > That seems cumbersome, if I understand how this is going to work correctly. If you're trying to watch the flow of data in your function, wouldn't you have to hit l for each step?

I generally don't care about all the locals, but only about the value of the expressions as I step through them. I can understand the benefit of seeing all the locals all the time, but I don't displaying them like this is the way to go (too much vertical space is wasted, you can't really interact with the values, etc).

Malabarba: >  but I don't displaying them like this is the way to go (too much vertical space is wasted, you can't really interact with the values, etc).

Did you have any alternatives in mind?
What do you mean by "interact with the values", copy-pasting them?

bbatsov: I had in mind inspecting them.

Anyways, getting everything right the first time around is not super important. We just have to have a sane version of this feature to merge. 

Generally, even the current state is acceptable to me as I don't share @expez's expectations about how the features would work - I'd expect it to work exactly as it does (although it'd be nice to have something affecting the scope of the session that doesn't exist).

expez: > [...] even the current state is acceptable to me

I really think we should get rid of the `defcustom`, it doesn't represent anything a user should toggle in its current state.

bbatsov: Well, you'll either have locals all the time or you won't. I'm fine with not having them by default and being able to globally toggle them, I'm also fine with being able to toggle them just for the current expression. Having this feature is an improvement, we can always refine it down the road (Lean for the win).

Malabarba: I have no strong opinions on the variable. As long as the user doesn't have to hit `l` at every step this is fine by me. I'll remove it when I sit down for this again.

As for the notion of debugging session. I've thought of a way to implement this at the clojure side that wouldn't be too complicated, I just never had too much of a reason to (this is the first feature that misses it). But I'd rather keep things simple for 0.9, I already short on time as it is.

bbatsov: OK, let's just remove the variable and merge this.

Malabarba:
Pushed.

bbatsov:
:+1: Looks good!

PR #1125, opened 2015-06-12
Fix cider-insert-in-repl indenting too much

expez:
The intention was to indent the code that was just inserted.  The bug was that
the code grabbed the value of point from the source buffer instead of the repl
buffer to mark the start position when indenting.  The result, in most cases,
was that the entire repl sessions was re-indented.  As the repl can grow quite
long this often brought emacs to a grinding halt.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1715,9 +1715,10 @@ If EVAL is non-nil the form will also be evaluated."
   (while (string-match "\\`[ \t\n\r]+\\|[ \t\n\r]+\\'" form)
     (setq form (replace-match "" t t form)))
   (with-current-buffer (cider-current-repl-buffer)
-    (let ((start-pos (point)))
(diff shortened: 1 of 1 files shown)


expez:
@bbatsov this function is still somewhat flawed in my eyes because it just naively inserts the last form into the repl.  I think the user's intention is always to insert the form at the end of the buffer, where it can be (modified and) evaluated, but if the user's last action in the buffer was to copy some result further up the code will just be dumped up there.

Can we just use `(goto-char (point-max))` to get us into the right buffer position, or are there other concerns?

bbatsov:
> Can we just use (goto-char (point-max)) to get us into the right buffer position, or are there other concerns?

I honestly can't recall why I wrote this the way I did. Your suggestion makes sense.

bbatsov:
:+1: Thanks!

PR #1131, opened 2015-06-13
Update the recommended steps when reporting issues

expez:
Recently quite a few of the bugs reported have been either missing reproduction
steps or been caused by interactions with other packages.

I'm hoping this will increase the percentage of issues which are immediately
actionable without one of us asking some variant of "How can I reproduce this?".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,3 +47,6 @@ elpaclean : clean
 
 %.elc : %.el
 	$(CASK) build
+
+run-cider: elpa
+	cask exec emacs -Q -L . --eval "(require 'cider)"
--- a/README.md
+++ b/README.md
@@ -1078,6 +1078,9 @@ Report issues and suggest features and improvements on the
 questions on the issue tracker - the mailing list and the IRC channel are the
 places for questions.
 
+If you want to file a bug, please clone this repo and provide sufficient details to reproduce the issue. Start by running `make run-cider`.
+This will bring up Emacs with only the latest version of CIDER loaded. By starting fresh, with the latest code, we can ensure that the problem at hand isn't already fixed or caused by interactions with other packages.
+
 ### Patches
 
 Patches under any form are always welcome! GitHub pull requests are even better! :-)


README.md
```diff
@@ -1078,6 +1078,8 @@ Report issues and suggest features and improvements on the
 questions on the issue tracker - the mailing list and the IRC channel are the
 places for questions.
 
+If you want to file a bug please clone this repo and provide sufficient details to reproduce the issue starting with `make emacs`.  It's likely that your problem is already fixed on `master` or that it's caused by something or other in your `.emacs`.
```
bbatsov: Sounds good, but the name of the task is a bit confusing IMO. It'd be nice if we explained that people are supposed to test cider in isolation from other packages.

expez:
> [...]  the name of the task is a bit confusing IMO.

Yeah, I just went with the convention `evil-mode` uses.  I get of understand the cop-out, finding a good name here is somewhat hard.  Any suggestions? How about `make start-cider`?

bbatsov:
`start-cider` is fine. `run-cider` is also fine.

expez:
Thanks, I went with `run-cider` and updated the readme slightly with the intent.

Makefile
```diff
@@ -47,3 +47,6 @@ elpaclean : clean
 
 %.elc : %.el
 	$(CASK) build
+
+start-cider: elpa
```
bbatsov: It says `start-cider` here.

README.md
```diff
@@ -1078,6 +1078,9 @@ Report issues and suggest features and improvements on the
 questions on the issue tracker - the mailing list and the IRC channel are the
 places for questions.
 
+If you want to file a bug please clone this repo and provide sufficient details to reproduce the issue starting with `make run-cider`.
+This will start emacs with only the latest version of cider loaded, ensuring that the problem at hand isn't already fixed or caused by interactions with other packages.
```
bbatsov: emacs -> Emacs; cider -> CIDER (for consistency)

README.md
```diff
@@ -1078,6 +1078,9 @@ Report issues and suggest features and improvements on the
 questions on the issue tracker - the mailing list and the IRC channel are the
 places for questions.
 
+If you want to file a bug please clone this repo and provide sufficient details to reproduce the issue starting with `make run-cider`.
```
bbatsov: There should be a `,` after bug. "starting with ..." might be a separate sentence as it reads a bit strange right now. Maybe "Start by running ...".

expez:
updated

PR #1132, opened 2015-06-13
Fix completing read port

patapizza:
Hitting `C-c M-c` yields "Symbol's function definition is void: second".

Stack trace for 'project':
```
Debugger entered--Lisp error: (void-function second)
  second(("project" "51151"))
  cider--completing-read-port("localhost" (("project" "51151")))
  cider-select-endpoint()
  #<subr call-interactively>(cider-connect nil nil)
  ad-Advice-call-interactively(#<subr call-interactively> cider-connect nil nil)
  apply(ad-Advice-call-interactively #<subr call-interactively> (cider-connect nil nil))
  call-interactively(cider-connect nil nil)
  command-execute(cider-connect)
```

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -264,7 +264,7 @@ Return a list of the form (HOST PORT), where PORT can be nil."
          (sel-port (completing-read (format "Port for %s: " host) ports
                                     nil nil nil nil (caar ports)))
          (port (or (cdr (assoc sel-port ports)) sel-port))
-         (port (if (listp port) (second port) port)))
+         (port (if (listp port) (cadr port) port)))
     (if (stringp port) (string-to-number port) port)))
 
 (defun cider-locate-running-nrepl-ports (&optional dir)


expez:
Thanks for spotting this.  I'd prefer if the fix was to require `cl-lib`, which would bring in the symbol `second`.  Contributors coming from clojure have to think twice when they see `cadr` and friends so I'd rather stick with `first` `second` and `nth` (though I IIRC common lisp has `first` to `tenth`).

Could you also write your commit messages in the the imperative, e.g. "Fix completing-read-port".

bbatsov:
Nope. We're moving away from cl lib. It's use in Cider is discouraged.

bbatsov:
@patapizza Fix the title of the commit message itself (use have to use `git commit -amend` and to do a force push afterwards.

bbatsov:
I've noticed there were other similar problems in the code and fixed all of them in one commit https://github.com/clojure-emacs/cider/commit/3d67c24f7e12eab33d433d9f9c6ac0119778cdd6

patapizza:
Erf ok, glad you fixed them all.

bbatsov:
@patapizza Thanks for spotting this!

PR #1135, opened 2015-06-16
Fix corner case with locals display

Malabarba:
Turns out you can't call `max` with no arguments. :P

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -114,14 +114,16 @@ Can be toggled while debugging with `l'.")
 (defun cider--debug-format-locals-list (locals)
   "Return a string description of list LOCALS.
 Each element of LOCALS should be a list of at least two elements."
-  (let ((left-col-width
-         ;; To right-indent the variable names.
-         (apply #'max (mapcar (lambda (l) (string-width (car l))) locals))))
-    ;; A format string to build a format string. :-P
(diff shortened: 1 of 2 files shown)


cider-debug.el
```diff
@@ -114,14 +114,15 @@ Can be toggled while debugging with `l'.")
...
 Each element of LOCALS should be a list of at least two elements."
-  (let ((left-col-width
-         ;; To right-indent the variable names.
-         (apply #'max (mapcar (lambda (l) (string-width (car l))) locals))))
-    ;; A format string to build a format string. :-P
-    (mapconcat (lambda (l) (format (format "%%%ds: %%s\n" left-col-width)
-                        (propertize (car l) 'face 'font-lock-variable-name-face)
-                        (cider-font-lock-as-clojure (cadr l))))
-               locals "")))
+  (if (not locals) ""
```
bbatsov: Why not `if locals`? Sounds more natural to me.

Malabarba: When one of the if branches is something trivial like this I usually prefer to have it first, but I can reverse that if prefered.

bbatsov: Yeah, please reverse it. I'm averse to negative conditions.

Malabarba:
Done

bbatsov:
:+1: Thanks!

PR #1137, opened 2015-06-17
Try and use `markdown-mode' for Grimoire buffers

vedang:
Currently, the **\*cider grimoire*** buffer is loaded up with \`fundamental-mode' as the
major mode. However, since Grimoire notes are written in Markdown, we can try
and use \`markdown-mode' to render them when possible.

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -65,8 +65,14 @@ opposite of what that option dictates."
            #'cider-grimoire-web-lookup))
 
 (defun cider-create-grimoire-buffer (content)
-  "Create a new grimoire buffer with CONTENT."
-  (with-current-buffer (cider-popup-buffer "*cider grimoire*" t)
+  "Create a new grimoire buffer with CONTENT.
+
+If `markdown-mode' has been installed and loaded, use it as the major mode
+for the Grimoire buffer."
+  (with-current-buffer
+      (if (fboundp 'markdown-mode)
+          (cider-popup-buffer "*cider grimoire*" t 'markdown-mode)
+        (cider-popup-buffer "*cider grimoire*" t))
     (read-only-mode -1)
     (insert content)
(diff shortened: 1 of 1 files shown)


bbatsov:
@arrdem Are you using anything else besides fenced code blocks? As they look pretty ugly in `markdown-mode` it will probably be better to just detect those on the fly and apply `clojure-mode`'s font-locking there.

arrdem:
@bbatsov at present headings are Markdown "formatted" (# prefixed) and code blocks are fenced in the github flavored markdown style as mentioned. That's all I remember right now.

arrdem:
No matter what some extra leg work to Clojure font lock the fenced code blocks is in order. If we could actually break that out as a markdown-mode plugin to clojure-mode and then happen to have cider-grimoire activate it that'd be amazing because I regularly hurt for Clojure font locking while writing Markdown nodes or my blog.

bbatsov:
Yeah, I was thinking about this. Making `markdown-mode` applying mode-specific font-locking when the fenced blocks is like "```Clojure". The overall problem with markdown mode is that it's about writing markdown documents and ideally we'd use some mode which renders them in a prettier form.

Malabarba:
Emacs in general would really benefit from a package that higlightings a chunck of text as markdown. 
SX.el had to implement a lot of this stuff too, and it's not a trivial matter.

`markdown-mode` is a start but it doesn't do pretty code blocks (not to mention it's a heavy dependency and has so many contributors there's little hope of it ever getting into Gelpa).

bbatsov:
Yep, a new package is ideally the way to go. Btw, `markdown-mode` is also absurdly slow in bigger documents. I think something is really wrong with the font-locking logic.

bbatsov:
Guess we won't be doing this, so I'm closing the PR.

PR #1150, opened 2015-06-20
[Fix #1122] Run client initialization in new client buffer

### LINKED ISSUE #1122: Can't connect to multiple nrepls
Running cider git checkout of HEAD, 3487955dd600d361ddae54671e3686c38f4aaebd

I'm trying to create multiple nrepl connections, to multiple processes, and can't. A typical workflow for me is to have a CLJ nrepl running, and a second to figwheel, for CLJS. 
 
Steps:

M-x cider-connect "localhost" 6005
M-x cider-connect "localhost" 7888 (figwheel nrepl)

The first connection works just fine. The second connection is "defective". M-x cider-connect reports "Direct connection established", but emacs doesn't automatically open the second buffer. If I manually switch to "*cider-repl localhost*<2>", there is no repl prompt, and typing in the buffer has no effect. 

If I kill both buffers, do the steps in opposite order, and get same result:

M-x cider-connect "localhost" 7888
M-x cider-connect "localhost" 6005

In this case, the 7888 buffer works correctly, and the 6005 buffer is defective. 

The bug doesn't appear to have anything to do with figwheel, I can also reproduce it using two normal CLJ nrepls:

M-x cider-connect "localhost" 6005
M-x cider-connect "localhost" 6006

Note that connecting to the same process twice works fine:

M-x cider-connect "localhost" 6005
M-x cider-connect "localhost" 6005

In this case, both buffers work as expected.

vspinu:


bbatsov:
:+1:

PR #1155, opened 2015-06-21
Add overlays to the debugger.

Malabarba:
Too much? :)  
### end-of-line
cider-debug-overlay-2

### inline
cider-debug-overlays

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -28,6 +28,31 @@
 (require 'nrepl-client)
 (require 'cider-interaction)
 
+(defface cider-result-overlay-face
+  '((((class color) (background light)) :foreground "firebrick")
+    (((class color) (background dark))  :foreground "orange red"))
+  "Face used to display result of debug step at point."
+  :group 'cider
+  :package-version "0.9.1")
+
+(defface cider-debug-code-overlay-face
+  '((((class color) (background light)) :background "grey80")
+    (((class color) (background dark))  :background "grey20"))
+  "Face used to mark code being debugged."
+  :group 'cider
+  :package-version "0.9.1")
+
+(defcustom cider-debug-use-overlays 'end-of-line
+  "Whether to higlight debugging information with overlays.
+Only applies to \"*cider-debug ...*\" buffers, which are used in debugging
+sessions.
+Possible values are inline, end-of-line, or nil."
+  :type '(choice (const :tag "End of line" end-of-line)
+                 (const :tag "Inline" inline)
+                 (const :tag "No overlays" nil))
+  :group 'cider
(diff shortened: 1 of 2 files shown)


Malabarba:
I figured this is safe enough to go in `0.9.1`, so it's based of `master`.

bbatsov:
I like it, although it should probably be configurable, as people used to `edebug` might prefer the use of the minibuffer. Btw, it'd be great if you expand the use of overlays for interactive evaluation as well. This will make it possible for us to do something like `instarepl` / `litable` long term.

Malabarba:
>  I like it, although it should probably be configurable, as people used to edebug might prefer the use of the minibuffer. 

Oh, the screenshot doesn't show, but it still uses the minibuffer. And there's a custom option to turn the overlays off.

> Btw, it'd be great if you expand the use of overlays for interactive evaluation as well. This will make it possible for us to do something like instarepl / litable long term.

The only problem is that these overlays are a little annoying if they stick around. In order to not change the contents of the buffer, they have 0 width, which means the user is completely unable to erase them. There's probably a better way to do this though, as I'm sure I've seen somewhere overlays that vanish when you write over them.

bbatsov:
> There's probably a better way to do this though, as I'm sure I've seen somewhere overlays that vanish when you write over them.

Yep, that'd be best. Another thing to consider is their exact position - some people might not like the overlay shifting things to the right, so an alternative showing the overlay outside the text on the line might be nice to have as well.

cider-debug.el
```diff
@@ -82,9 +82,60 @@ sexp."
...
+                 'face 'cider-debug-code-overlay-face)
+    ;; Higher priority than `show-paren'.
+    (overlay-put cider--debug-code-overlay 'priority 2000)))
+
+(defcustom cider-debug-use-overlays t
+  "Whether to higlight debugging information with overlays.
+Only applies to \"*cider-debug ...*\" buffers, which are used in debugging
+sessions."
+  :type 'boolean
+  :group 'cider)
```
bbatsov: Don't forget the `:package-version` properties.

bbatsov:
On an unrelated note - add a changelog entry for this and drop the `.` from the commit message's title.

Malabarba:
All done, and I've factored out the overlay handling as well so it can be reused. Should these functions go in another file or do I leave them there?

Malabarba:
Also, see new pic at the top.

bbatsov:
:+1: Fantastic work!

arrdem:
:+1:

mavbozo:
:thumbsup: Awesome!!

PR #1156, opened 2015-06-21
[Fix #1142] Don't sniff for ports when port is present in cider-known-endpoint

### LINKED ISSUE #1142: How can I tell cider to establish a direct connection?
OSX Yosemite 10.10.3
cider 20150617.753

Previously I could type C-c M-c
Cider would prompt me for Host:
I could type: `dockerized.foo.com`
the it would prompt me for `port` and I could connect.
Now it tries to establish an xssh connection as soon as I enter the host.
`Password for /sshx:local-trek.outpace.com: `

Here is my messages output:
```Tramp: Opening connection for local-trek.outpace.com using sshx...
Tramp: Sending command `exec ssh   -o ControlMaster=auto -o ControlPath='tramp.%r@%h:%p' -o ControlPersist=no -e none -t -t local-trek.outpace.com /bin/sh'
Tramp: Waiting for prompts from remote shell...
Tramp: Sending password```

vspinu:
We can think of making the port detection more robust, but I think it should stay. It's robust for most common case of nREPL running locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#1142](https://github.com/clojure-emacs/cider/issues/1142): Don't retrive nrepl ports when `cider-known-endpoints` entry already contains the port.
 * [#1153](https://github.com/clojure-emacs/cider/pull/1153): Fix behavior of `cider-switch-to-current-repl-buffer`.
 * [#1139](https://github.com/clojure-emacs/cider/issues/1139): Fix evaluation of ns forms and of forms with unevaluated namespaces.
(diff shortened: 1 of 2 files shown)


bbatsov:
Did you get any comments from me. It seems they disappeared. Btw, don't forget to update the changelog.

vspinu:
> Did you get any comments from me. It seems they disappeared.

No I didn't. I fixed a couple of typos by forced push. That might have screwed the comments.

>  Btw, don't forget to update the changelog.

Added.

vspinu:
> No I didn't.

Actually I did. Fixing now.

vspinu:
Done.

bbatsov:
:+1:

PR #1160, opened 2015-06-23
Use fringe arrows on the debuger

Malabarba:
cider-debug-overlay-2
cider-debug-overlay-3


bbatsov:
:+1: Looks great!

P.S. I never noticed `edebug` had those as well. :-)

arrdem:
:+1:

PR #1164, opened 2015-06-25
cider-browse-ns.el (cider-browse-ns--doc-at-point): Fixup

abo-abo:
* cider-browse-ns.el (cider-browse-ns--doc-at-point): It was a malformed form
  previously.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#1164](https://github.com/clojure-emacs/cider/pull/1164): Fix an error in `cider-browse-ns--doc-at-point`.
+
 ## 0.9.1 / 2015-06-24
 
 ### New features
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -126,7 +126,7 @@
   "Expand browser according to thing at current point."
   (interactive)
   (-when-let (var (cider-browse-ns--var-at-point))
-    ((cider-doc-lookup var))))
+    (cider-doc-lookup var)))
 
 (defun cider-browse-ns--find-at-point ()
   (interactive)
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -130,7 +130,7 @@
 
 (defun cider-browse-ns--find-at-point ()
   (interactive)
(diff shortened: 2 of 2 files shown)


abo-abo:
Also, I see one instance of `when-let` from subr-x has sneaked in place of `-when-let` from dash.  If it's OK to use it, probably all of them need to be updated.

expez:
>  I see one instance of when-let from subr-x has sneaked in place of -when-let from dash.

That would be a mistake. IIRC `when-let` was added in 24.4 or 24.5 and we're still supporting at least 24.3.

Could you update the commit message so it doesn't mention the file?  Something like "Fix a malformed call-site" would be adequate.

bbatsov:
Btw, as this fixes a user-facing bug, it probably warrants a changelog entry as well. You can fix the `when-let` problem in a separate commit.

abo-abo:
OK, updated. Btw, what's up with not mentioning the file in the commit message? I think it's a good practice.

expez:
> OK, updated. Btw, what's up with not mentioning the file in the commit message? I think it's a good practice.

I'm sure @bbatsov has an opinion on this too but the biggest problem I had with the previous commit message was that the first line (the summay which appears in `git log` and all other tools) was essentially useless because it only contained file and function information along with the word "fixup".

abo-abo:
> essentially useless because it only contained file and function information along with the word "fixup"

That's kind of the point. The commit message for such a minor fixup must look benign so that I can safely skip it when browsing the log. On the other hand, "Fix a malformed call-site" is ambiguous and sounds important enough to be tripping through it each time I'm looking for some change.

Of course, that's just my opinion. I hope that you see that there's at least some logic to it:)

bbatsov:
> That's kind of the point. The commit message for such a minor fixup must look benign so that I can safely skip it when browsing the log. On the other hand, "Fix a malformed call-site" is ambiguous and sounds important enough to be tripping through it each time I'm looking for some change.

Well, that's always subjective. :-)

I'm a big believer in meaningful commit messages and clear commit titles. I don't like file mentions in commit messages because file names change and they don't add a ton of value - I'm generally interested in the problem and solution; the location is just details (and is often implied by the nature of the problem).

PR #1165, opened 2015-06-25
Extract function cider-set-relevant-connection

ryfow:
This makes it easier to configure C-c C-k to switch to the appropriate
connection before trying to load the buffer.

I'm an elisp n00b, so feedback is welcome.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -446,8 +446,13 @@ directory."
   (interactive "P")
   (cider-ensure-connected)
   (let* ((project-directory
-          (or (when do-prompt (read-directory-name "Project: "))
-              (nrepl-project-directory-for (nrepl-current-dir))))
+          (nrepl-project-directory-for
+           (or (when do-prompt
+                 (read-directory-name "Project: "
+                                      (nrepl-project-directory-for (buffer-file-name))
+                                      nil
+                                      'confirm))
+               (nrepl-current-dir))))
          (connection-buffer
           (or
            (and (= 1 (length nrepl-connection-list)) (car nrepl-connection-list))
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -392,19 +392,19 @@ to jump back to the last Clojure source buffer."
   (interactive "p")
   (funcall cider-switch-to-repl-command arg))
 
-(defun cider-switch-to-current-repl-buffer (&optional arg)
+(defun cider-switch-to-current-repl-buffer (&optional set-namespace)
   "Select the REPL buffer, when possible in an existing window.
 
 Hint: You can use `display-buffer-reuse-frames' and
 `special-display-buffer-names' to customize the frame in which
 the buffer should appear.
 
-With a prefix ARG sets the namespace in the REPL buffer to that
-of the namespace in the Clojure source buffer."
-  (interactive "p")
+With a prefix argument SET-NAMESPACE, sets the namespace in the REPL buffer to
+that of the namespace in the Clojure source buffer."
+  (interactive "P")
   (cider-ensure-connected)
   (let ((buffer (current-buffer)))
-    (when (eq 4 arg)
+    (when set-namespace
       (cider-repl-set-ns (cider-current-ns)))
     (if cider-repl-display-in-current-window
 	(pop-to-buffer-same-window (cider-get-repl-buffer))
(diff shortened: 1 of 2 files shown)


cider-interaction.el
```diff
@@ -438,6 +438,25 @@ is ambiguity, therefore nil is returned."
       (when (= 1 (length matching-connections))
         (car matching-connections)))))
 
+(defun cider-set-relevant-connection (&optional arg)
+  "Try to set the current REPL buffer based on the current file open in the current buffer.
```
bbatsov: REPL & connection buffer is pretty much the same thing as far as CIDER is concerned. The `current` before file is kind of redundant, although I guess you can specify it's a Clojure source file.

ryfow: I tried to clean this up.

ryfow:
After using this for a while, it seems silly to (message) the current REPL when nothing has changed. Would you like me to only message if the REPL buffer changes?

bbatsov:
> After using this for a while, it seems silly to (message) the current REPL when nothing has changed. Would you like me to only message if the REPL buffer changes?

Yep, this sounds reasonable. You'll also have to rebase on top of the current `master`.

cider-interaction.el
```diff
@@ -438,6 +438,25 @@ is ambiguity, therefore nil is returned."
       (when (= 1 (length matching-connections))
         (car matching-connections)))))
 
+(defun cider-set-relevant-connection (&optional arg)
+  "Try to set the current REPL buffer based on the the current Clojure source buffer.
+If succesful, return the new connection buffer."
+  (interactive "p")
+  (cider-ensure-connected)
+  (let* ((project-directory
+          (or (when (eq 16 arg) (read-directory-name "Project: "))
```
Malabarba: `(eq 16 arg)` means a double prefix argument `C-u C-u`. Are you planning on doing special stuff with `(eq 4 arg)`?
If not, it's best to just change the interactive to `"P"` and replace this `(eq 16 arg)` with `arg`.

Malabarba: Oh nevermind. I see it's used on the other function.

bbatsov: Unfortunately the logic in the old command got pretty convoluted because we wanted to do 3 different things with the same command...

Malabarba: Here's what I would do to organize this.

Since this is a command of its own now, it's prefix should be simplified.

    (defun cider-set-relevant-connection (&optional do-prompt)
      "Try to set the current REPL buffer based on the the current Clojure source buffer.
    If succesful, return the new connection buffer.
    With a prefix argument DO-PROMPT, the chosen REPL buffer is based on a
    supplied project directory."
      (interactive "P")
      (cider-ensure-connected)
      (let* ((project-directory
              (or (when do-prompt (read-directory-name "Project: "))
                  (nrepl-project-directory-for (nrepl-current-dir))))
              ...)))


And then in the other function, we can just write the call like this, which will still have the old effect:

    (defun cider-switch-to-relevant-repl-buffer (&optional arg)
      "..."
      (interactive "p")
      (cider-set-relevant-connection)
      (cider-switch-to-current-repl-buffer (eq 16 arg))
      ...)

Malabarba: And I would also do the same thing on `cider-set-relevant-connection` (using `(eq 4 arg)` instead).

bbatsov: :+1:

cider-interaction.el
```diff
@@ -456,19 +475,9 @@ of the namespace in the Clojure source buffer.
...
-         (connection-buffer
-          (or
-           (and (= 1 (length nrepl-connection-list)) (car nrepl-connection-list))
-           (and project-directory
-                (cider-find-connection-buffer-for-project-directory project-directory)))))
-    (when connection-buffer
-      (setq nrepl-connection-list
-            (cons connection-buffer (delq connection-buffer nrepl-connection-list))))
-    (cider-switch-to-current-repl-buffer arg)
+  (cider-set-relevant-connection)
```
Malabarba: It might be nice to have a comment explaining why we need to call this 2 lines before calling it again.

cider-interaction.el
```diff
@@ -438,6 +438,30 @@ is ambiguity, therefore nil is returned."
       (when (= 1 (length matching-connections))
         (car matching-connections)))))
 
+(defun cider-set-relevant-connection (&optional do-prompt)
+  "Try to set the current REPL buffer based on the the current Clojure source buffer.
+If succesful, return the new connection buffer.
+With a prefix argument DO-PROMPT, the chosen REPL buffer is base on a supplied project
```
bbatsov: base -> based

ryfow:
Thanks for the feedback & sorry about the typo. I think I've followed all of your advice to this point.

While testing, the one annoyance (which isn't new) is that if you're prompted for a directory, the default ends up being a non nrepl project directory.

This seems better to me, but I thought I'd ask before just sticking the change in there:
```elisp
  (let* ((project-directory
          (nrepl-project-directory-for
           (or (when do-prompt (read-directory-name "Project: "))
               (nrepl-current-dir))))
```

Malabarba:
I'd rather not set a value different to what the user inputs. It's better to just offer a smarted default I think. Try the following: `(read-directory-name "Project: " (nrepl-project-directory-for "./") nil 'confirm)`

ryfow:
Should I make that change in this pull request or submit another one?

bbatsov:
New commit here is fine.

On Monday, June 29, 2015, Ryan Fowler <notifications@github.com> wrote:

> Should I make that change in this pull request or submit another one?
>
> —
> <https://github.com/clojure-emacs/cider/pull/1165#issuecomment-116825345>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

ryfow:
`(nrepl-project-directory-for "./")` returned nil for me, but `(nrepl-project-directory-for (buffer-file-name))` seems to work.

Malabarba:
Looks good too.

ryfow:
Added changelog entry for the current project directory issue. Thanks for your patience.

bbatsov:
:+1: Thanks for your contributions! :-)

PR #1172, opened 2015-06-29
Debugger minor mode and some improvements

Malabarba:
I batched these commits together because they don't require anything on the nrepl side.
After this there's another batch of commits that do.


Malabarba:
Oh. Commits on my fork show up here too?  I could swear they didn't.

Malabarba:
Pushed and tested. :+1:

bbatsov:
:+1:

PR #1174, opened 2015-06-29
Define command to run -main

### LINKED ISSUE #1011: Command to run (-main)
I couldn't find anything about this in the readme or the issues list. 
Is there a command to run the project's `-main` function?

Currently what I do is write `;; (-main)` somewhere in the file, and then manually move point to it and hit `C-x C-e` whenever I want to run `-main`. Is there a built-in key for that? 
If not, consider this a feature request.

Malabarba:
Fix #1011.

Something like this?
Not sure where to bind it.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -2164,6 +2164,35 @@ restart the server."
           (cider-jack-in prompt-project))
       (error "Can't restart CIDER for unknown project"))))
 
+(defvar cider--namespace-history nil
+  "History of user input for namespace prompts.")
+
+(defun cider--var-namespace (var)
+  "Return the namespace of VAR.
+VAR is a fully qualified Clojure variable name as a string."
+  (replace-regexp-in-string "\\(?:#'\\)?\\(.*\\)/.*" "\\1" var))
+
+(defun cider-run (&optional function)
+  "Run -main or FUNCTION, prompting for its namespace if necessary.
+With a prefix argument, prompt for function to run instead of -main."
+  (interactive (list (when current-prefix-arg (read-string "Function name: "))))
+  (let ((name (or function "-main")))
+    (-when-let (response (nrepl-send-sync-request
+                          (list "op" "ns-list-vars-by-name" "name" name)))
+      (-if-let (vars (split-string (substring (nrepl-dict-get response "var-list") 1 -1)))
+          (cider-interactive-eval
+           (if (= (length vars) 1)
+               (concat "(" (car vars) ")")
+             (let* ((completions (mapcar #'cider--var-namespace vars))
+                    (def (or (car cider--namespace-history)
+                             (car completions))))
+               (format "(#'%s/%s)"
+                       (completing-read (format "Namespace (%s): " def)
+                                        completions nil t nil
+                                        'cider--namespace-history def)
+                       name))))
+        (user-error "No %s var defined in any namespace" name)))))
+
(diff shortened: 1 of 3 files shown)


cider-interaction.el
```diff
@@ -2146,6 +2146,32 @@ restart the server."
...
+  "History of user input for namespace prompts.")
+
+(defun cider--var-namespace (var)
+  (replace-regexp-in-string "\\(?:#'\\)?\\(.*\\)/.*" "\\1" var))
+
+(defun cider-run (&optional function)
+  "Run -main or FUNCTION, prompting for its namespace if necessary.
+With a prefix argument, prompt for function to run instead of -main."
+  (interactive (list (when current-prefix-arg (read-string "Function name: "))))
+  (-when-let (response (nrepl-sync-request:eval
```
bbatsov: It's be better to move this inlined code to a simple middleware op.

cider-interaction.el
```diff
@@ -2146,6 +2146,32 @@ restart the server."
           (cider-jack-in prompt-project))
       (error "Can't restart CIDER for unknown project"))))
 
+(defvar cider--namespace-history nil
+  "History of user input for namespace prompts.")
+
+(defun cider--var-namespace (var)
```
bbatsov: A docstring a few ert tests for this function would be nice.

cider-interaction.el
Malabarba: Which middleware would you put it in? `info`? `ns`?

Malabarba: But Idk... Isn't this a little narrow for a middleware?

bbatsov: I'd put it in ns. Middleware shouldn't be complex - it should just get some work done. The problem with evaluating strings is twofold:
1) The meaning of the code is obscured by implementation details
2) Even simple code can be problematic (in your case you're not using FQ names, which means that in some namespaces this code will misbehave)

Malabarba:
All added. Along with https://github.com/clojure-emacs/cider-nrepl/pull/221

cider-interaction.el
```diff
@@ -2164,6 +2164,33 @@ restart the server."
...
+  (-when-let (response (nrepl-send-sync-request
+                        (list "op" "ns-list-vars-by-name" "name" (or function "-main"))))
+    (-if-let (vars (split-string (substring (nrepl-dict-get response "var-list") 1 -1)))
+        (cider-interactive-eval
+         (if (= (length vars) 1)
+             (concat "(" (car vars) ")")
+           (let* ((completions (mapcar #'cider--var-namespace vars))
+                  (def (or (car cider--namespace-history)
+                           (car completions))))
+             (format "(#'%s/-main)"
```
bbatsov: Btw, shouldn't this cover the option of `function` being used?

cider-interaction.el
```diff
@@ -2164,6 +2164,33 @@ restart the server."
...
+                     (completing-read (format "Namespace (%s): " def)
+                                      completions nil t nil
+                                      'cider--namespace-history def)))))
+      (user-error "No -main var defined in any namespace"))))
```
bbatsov: Same here.

cider-interaction.el
Malabarba: Oops. :-)

Malabarba:
Tests added and name fixed.

bbatsov:
:+1:

PR #1176, opened 2015-06-30
More debugger

Malabarba:
See https://github.com/clojure-emacs/cider-nrepl/pull/222


Malabarba:
Reworded.

README.md
```diff
@@ -917,10 +917,13 @@ Keyboard shortcut               | Description
 <kbd>c</kbd> | Continue without stopping
 <kbd>o</kbd> | Move out of the current sexp (like `up-list`)
 <kbd>i</kbd> | Inject a value into running code
-<kbd>e</kbd> | Eval code in current context
```
bbatsov: Btw, I believe this is still handy as a shorter alias for `C-c M-:`.

bbatsov: Just verified that in edebug both `e` and `M-:` behave the same way.

Malabarba: Yes, I probably shouldn't have removed it from the readme.
The key still exists, it's just not advertised on the prompt because I wanted to reduce it a little. 

Anyway, I'll add this back, but I can't do it for at least a couple of hours.

bbatsov: No problem. Regarding the prompt itself - `edebug` doesn't even show it by default. Maybe we should add an option to hide it - after you've seen the keys a few times you generally don't need them as hints (and a menu would help as well, as previously discussed).

Malabarba: All true. 
After the next batch of commits, I'll focus entirely on the interface. Then I'll add a help key, a menu, and maybe some toolbar buttons while debugging.

bbatsov: Sounds like a plan!

Malabarba:
Readme Fixed

bbatsov:
Thanks!

Issue #1175, opened 2015-06-29
Error highlighting does not work


whysoserious:
#### Steps to reproduce

* Create any lein project eg. `lein new quil my-sketch`
* Open any `.clj` file eg. `src/my_sketch/core.clj`
* `M-x cider-jack-in`
* Make any syntax error which should be highlighted
* `M-x cider-eval-defun-at-point` - __yields no highlighting__
* `M-x cider-eval-region` - __yields no highlighting__
* `M-x cider-eval-buffer` - works as expected - I can see red underscores

##### My emacs config: 

https://github.com/whysoserious/my-emacs-config/blob/master/init.el#L326-L374

##### `M-x version`

```
GNU Emacs 24.4.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2014-10-21 on builder10-9.porkrind.org
```

##### repl

```
; CIDER 0.9.1 (package: 20150625.946) (Java 1.8.0_25, Clojure 1.6.0, nREPL 0.2.10)
WARNING: The following nREPL ops are not supported:
extract-definition rename-file-or-dir stubs-for-interface warm-ast-cache
Please, install (or update) refactor-nrepl and restart REPL.
You can mute this warning by changing cljr-suppress-middleware-warnings.
WARNING: clj-refactor and refactor-nrepl are out of sync.
Their versions are 1.1.0-SNAPSHOT and 1.0.5, respectively.
You can mute this warning by changing cljr-suppress-middleware-warnings.
user>
```

Malabarba:
There are two things going on here. Let's tackle them 1 at a time:

### The `eval` op doesn't check for syntax errors until execution

If you write something like the following, and evaluate it with `cider-eval-defun`, the nrepl will simply not throw any errors.

    (defn should-error []
      (reduce))

There's not much cider can do, the `eval` op simply doesn't check for errors (I think it doesn't compile the function?). It's not just about highlighting, you don't even get a stacktrace buffer. If you issue `load-buffer` instead, then you *will* get a stacktrace and highlighting, because the `load-file` op *does* compile.

In order to trigger an error, you have to actually use the function, for instance by evaluating `(should-error)`. Then you will get an error and a stacktrace, but you still don't get highlighting. That's the second problem.

### When execution runs into a syntax error, the eval op produces an uninformative error.

[...]

PR #1180, opened 2015-07-01
Small things all around

Malabarba:
- `cider-docview-file` will tell the user that it can't find the file, instead of throwing a cryptic error.
- `cider-jump-to` supports jumping to vars by name, besides position. This goes with a change in cider-nrepl which will provide us with the file even for variables whose origin we don't fully know (still, this commit is safe to merge on its own).
- Small fix and a defgroup on cider-debug.

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -29,24 +29,31 @@
 (require 'cider-interaction)
 (require 'dash)
 
+
+;;; Customization
+(defgroup cider-debug nil
+  "Presentation and behaviour of the cider debugger."
+  :prefix "cider-debug-"
+  :package-version '(cider-debug . "0.10.0"))
+
 (defface cider-result-overlay-face
   '((((class color) (background light)) :foreground "firebrick")
     (((class color) (background dark))  :foreground "orange red"))
   "Face used to display result of debug step at point."
-  :group 'cider
+  :group 'cider-debug
   :package-version "0.9.1")
 
 (defface cider-debug-code-overlay-face
   '((((class color) (background light)) :background "grey80")
     (((class color) (background dark))  :background "grey30"))
   "Face used to mark code being debugged."
-  :group 'cider
+  :group 'cider-debug
   :package-version "0.9.1")
 
 (defface cider-debug-prompt-face
   '((t :underline t :inherit font-lock-builtin-face))
   "Face used to mark code being debugged."
-  :group 'cider
+  :group 'cider-debug
   :package-version "0.10.0")
 
 (defcustom cider-debug-use-overlays 'end-of-line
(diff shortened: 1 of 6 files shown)


cider-debug.el
```diff
@@ -29,24 +29,31 @@
 (require 'cider-interaction)
 (require 'dash)
 
+
```
Malabarba: This is not an empty line, it's a ^L.

cider-doc.el
```diff
@@ -175,10 +175,15 @@
   "Open the source for the current symbol, if available."
   (interactive)
   (if cider-docview-file
-      (let ((buffer (and cider-docview-file
-                         (not (cider--tooling-file-p cider-docview-file))
-                         (cider-find-file cider-docview-file))))
-        (cider-jump-to buffer (cons cider-docview-line nil) nil))
+      (-if-let (buffer (and (not (cider--tooling-file-p cider-docview-file))
+                            (cider-find-file cider-docview-file)))
+          (cider-jump-to buffer (if (> cider-docview-line 1)
```
Malabarba: I wasn't super happy with this part, but the repl reports a :line of 1 on defs evaluated with "eval", so the only way to check if we had line information was to check if it's `> 1`. Any better ideas?

Malabarba:
This builds on #1176. So merge that first.

cider-doc.el
bbatsov: Check whether the filename is a tooling file?

bbatsov:
> cider-jump-to supports jumping to vars by name, besides position. This goes with a change in cider-nrepl which will provide us with the file even for variables whose origin we don't fully know (still, this commit is safe to merge on its own).

You should check that this works from stacktraces as well. Probably no extra changes are needed there, but still... better safe than sorry.

cider-doc.el
Malabarba: Thanks to the upcoming changes to the `info` op, we won't get tooling files as filenames anymore (well, maybe sometimes, but much less often) so we can't use that to determine whether the line is correct or not.
However, you just made me realise, the `info` op itself could remove the line information whenever it manually replaces a tooling filename with a real filename. Do you forsee issues with that? (I'll push the actual `info` change soon)

Malabarba:
I'll check stacktraces, I'll probably have to add a bit of code.

Malabarba:
Stacktraces now work as well. And I've added a commit.
Also https://github.com/clojure-emacs/cider-nrepl/pull/223

bbatsov:
> Stacktraces now work as well.

Nice. I guess we can mark #1175 as fixed and update the changelog.

Malabarba:
Actually, I only tested that following buttons on Stack traces works, I forgot to test the highlighting. So don't close it just yet.

Malabarba:
I just tested, and no, this is not enough to make highlighting work. I'm adding a comment to that discussion, but it's mosly unrelated to this commit, I think.

Malabarba:
Reworded.

bbatsov:
:+1: Thanks!

PR #1182, opened 2015-07-02
Fix navigation bug introduced with map handling

Malabarba:


bbatsov:
hadling -> handling

Malabarba:
(add-hook 'git-commit-mode-hook #'flyspell-mode)

PR #1185, opened 2015-07-03
Mention manual completion with company in readme

expez:
This adds a new section to the readme for users who prefers to trigger
completions manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,7 +507,7 @@ You might also want to install [`ido-flex`](https://github.com/lewang/flx).
 ### Auto-completion
 
 `CIDER` users are advised to use [`company-mode`](http://company-mode.github.io/) to enable auto-completion
-inside of source code and REPL buffers.  To install company-mode do:
+inside of source code and REPL buffers.  To install `company-mode` do:
 
 `M-x package-install <RET> company <RET>`
 
@@ -526,6 +526,21 @@ After installation, company can be turned on  globally, like so --
 
 When `company-mode` is thus enabled, it will receive completion information
 from `cider-complete-at-point`, and requires no additional setup or plugins.
+
+If you'd prefer to trigger completions manually you can add this to you config:
+
+```el
+(setq company-idle-delay nil) ; never start completions automatically
+(global-set-key (kbd "M-TAB") #'company-complete) ; use meta+tab, aka C-M-i, as manual trigger
+```
+
+To make `tab` complete, without losing the ability to manually indent, you can add this to your config:
+
+```el
+(global-set-key (kbd "TAB") #'company-indent-or-complete-common)
+```
+
+`company-indent-or-complete-common` requires a version of `company-mode` newer than `0.8.12`.
 
 #### Migrating from `auto-complete-mode`
 


expez:
@dgutov would it be better if we recommended binding `company-manual-begin` here?  It doesn't have a docstring so I'm not entirely clear on the difference between that and `company-complete`.

dgutov:
No, `company-complete` and `company-complete-common` are both better choices.

README.md
```diff
@@ -527,6 +527,13 @@ After installation, company can be turned on  globally, like so --
 When `company-mode` is thus enabled, it will receive completion information
 from `cider-complete-at-point`, and requires no additional setup or plugins.
 
+If you'd prefer to trigger completions manually you can do:
+
+```el
+(setq company-idle-delay nil) ; never automatically start completions
+(global-set-key "\t" #'company-complete) ; use tab as manual trigger
```
bbatsov: Maybe we should recommend something else (otherwise people won't be able to indent with TAB). I think `M-TAB` is the standard keybinding for completions. There's also a handy tip here http://www.emacswiki.org/emacs/CompanyMode about "indent-or-complete" (which can be bound directly to TAB).

README.md
```diff
@@ -527,6 +527,13 @@ After installation, company can be turned on  globally, like so --
 When `company-mode` is thus enabled, it will receive completion information
 from `cider-complete-at-point`, and requires no additional setup or plugins.
 
+If you'd prefer to trigger completions manually you can do:
```
bbatsov: can do -> can add this to your config.

One can never be too careful. :-)

README.md
```diff
@@ -527,6 +527,13 @@ After installation, company can be turned on  globally, like so --
 When `company-mode` is thus enabled, it will receive completion information
 from `cider-complete-at-point`, and requires no additional setup or plugins.
 
+If you'd prefer to trigger completions manually you can add this to you config:
+
+```el
+(setq company-idle-delay nil) ; never start completions automatically
+(global-set-key (kbd "M-TAB") #'company-complete) ; use meta+tab as manual trigger
```
expez: Now we'll get questions about where in the hell the meta key is :p

dgutov: I think pressing `C-M-i` is a lot easier than `M-TAB`, and these combinations are pretty much equivalent.

bbatsov: Yeah, that's a valid remark.

@dgutov I really think you should add `company-indent-or-complete` and `company-complete-or-indent` in the base distribution. Having them around would simplify things like this.

dgutov: @bbatsov `company-indent-or-complete` sounds reasonable, but how would it work exactly? Would it count invocations, or try indenting and see if the column hasn't changed? The EmacsWiki implementation makes the least sense to me.

Malabarba: > I think pressing C-M-i is a lot easier than M-TAB, and these combinations are pretty much equivalent.

Besides, `Alt-TAB` will sounds like a bad to joke to newcomers since that key is never going to reach Emacs on most computers. It's probably best to advertise `C-M-i`.

cobalamin: :+1: @Malabarba

bbatsov: `M-Tab` works pretty well on OS X. :-) But overall I agree with the general consensus.

bbatsov: @dgutov Thought you created the EmacsWiki implementation. :-) I think it should work exactly like `(setq tab-always-indent 'complete)`. Meaning - if the point is not properly indented do indentation, otherwise do completion.

dgutov: @bbatsov Oh, it was there long before me. Probably written by the author of Ca2+, mentioned there as well.

bbatsov: @dgutov Got it. Let us know if you ever add something like this to `company`. It will be much appreciated.

dgutov: @bbatsov No problem, done.

bbatsov:
:+1:

PR #1186, opened 2015-07-03
Change cider-version to a defconst, so it reloads properly

Malabarba:
Small thing.
Without this, upgrading and reloading cider will not update the version until Emacs restarts. Which means the user will get "CIDER's version ... does not match" warnings even though they've already upgraded.

Only really matters starting with Emacs 25, since package.el doesn't reload after upgrades on emacs 24.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -70,7 +70,7 @@
 (require 'cider-debug)
 (require 'tramp-sh)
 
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: No idea why I made this a `defvar` at all. Silly me...

PR #1187, opened 2015-07-07
Debugger menu and configurable input prompt

Malabarba:
- The menu is pretty straightforward.
- The configurable input prompt (the list of debug keys) can be shown on the minibuffer, on an overlay above the function, on both, or not at all. I think I'll make `overlay` the default, so the echo-area will be less crowded. When the top of the function is not visible, the top of the window is used instead.

What do you say?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1187](https://github.com/clojure-emacs/cider/pull/1187): The list of keys displayed by the debugger can be configured with `cider-debug-prompt`.
 * [#1187](https://github.com/clojure-emacs/cider/pull/1187): While debugging, there is a menu on the menu-bar listing available commands.
 * [#1184](https://github.com/clojure-emacs/cider/pull/1184): When the user kills the repl buffer, CIDER will offer to kill the nrepl buffer and process too. Also, when the client (repl) process dies, the server (nrepl) process is killed too.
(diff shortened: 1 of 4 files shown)


bbatsov:
I say "Aye!".

bbatsov:
Hmm, something's wrong with Travis.

Malabarba:
At least it got to the compiling stage. I've got a package where half the time I can't get travis to clone the repo. 

Did you try restarting the build?

bbatsov:
Yep, I restarted it.

Malabarba:
Ah ok. I was reading it on my phone yesterday, and it looked like the build was stopping at the compile stage. But I see now it's just a normal tests thing. I'll get to it.

Malabarba:
Ok. it's passing now.

Malabarba:
Hold on, don't merge yet. After playing around on my work PC I realise I don't like the way this looks. I'll have to fine tune it a bit more.

Malabarba:
Ok, it works nicer now I think.
The previous version would remove the overlays immediately after the user sent a command, and recreate them when a new input-request was received. This caused a noticeable flicker on all the overlays, because Emacs actually has time to do a redisplay in between those two events.

The current version doesn't immediately remove the overlays, instead it removes them when the next input request is received or after a fraction of a second (whichever comes first). In my system, this is enough to avoid the flicker.

bbatsov:
:+1:

PR #1188, opened 2015-07-08
New debugging tool-bar

Malabarba:
toolbar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


Malabarba:
Too much? :)

bbatsov:
Hehe, not at all. Some people have been complaining CIDER is not user-friendly enough and I'm sure they'll appreciate nice touches like this. It might even make sense to add more toolbars here and there. I never think of them just because my toolbar is always disabled. :-)

Malabarba:
Yes, me too. I only recently learned the toolbar was customizable, because I just never have it on.

PR #1191, opened 2015-07-10
Configurable print-level and length for debugger values

Malabarba:
Goes with https://github.com/clojure-emacs/cider-nrepl/pull/231


bbatsov:
:+1:

Issue #1192, opened 2015-07-10
Wrong type argument: stringp, cider-debug-print-level


manuel-uberti:
Hi,

I am on Emacs 25.0.50.1 (build ```53cf3cf```). 
This is my CIDER version: ```; CIDER 0.10.0snapshot (package: 20150710.544) (Java 1.8.0_60-internal, Clojure 1.7.0, nREPL 0.2.10)```

When I issue ```M-x cider-jack-in``` in a .clj file this is what I get in the ```*Messages*``` buffer:

```
nREPL server started on 51494
nREPL: Establishing direct connection to localhost:51494 ...
nREPL: Direct connection established
error in process filter: nrepl-bencode: Wrong type argument: stringp, cider-debug-print-level
error in process filter: Wrong type argument: stringp, cider-debug-print-level
```

Kungi:
I have exactly the same problem here on Emacs 24.4.50.1
Cider Version 20150710.544
I am using `cider-connect`.

bbatsov:
Ops, this is caused by https://github.com/clojure-emacs/cider/commit/3286f0041ace37ce2336615d539132dbb61aa4c4 //cc @Malabarba 

Once again we're bitten by trying to encode nil. If Artur doesn't fix this before me I'll fix this tomorrow.

Malabarba:
Actually, this wans't a nil problem. This was because the list was quoted, so we were passing the symbols. :$

manuel-uberti:
Thank you, it is fixed now.

PR #1193, opened 2015-07-10
Quick fix for the debugger print-level

Malabarba:
The diff speaks for itself

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -150,9 +150,9 @@ This variable must be set before starting the repl connection."
 (defun cider--debug-init-connection ()
   "Initialize a connection with clj-debugger."
   (nrepl-send-request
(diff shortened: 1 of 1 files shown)


Malabarba:
Fixes #1192

bbatsov:
:+1:

bbatsov:
Still, there's an encoding issue as well - `nil` is encoded as `[]` (because it's both `nil` and an empty list in Emacs Lisp and we can't differentiate between the two usages). So, the defaults are definitely not working as expected. I totally missed this earlier today.

Malabarba:
Oooh. I see what you mean. What should I fix this on the nrepl side? Or is there a proper way to pass nil?

Malabarba:
How about #1194?

PR #1194, opened 2015-07-10
Only pass print-level and print-length if non-nil

Malabarba:


bbatsov:
Yep, you figured out the best solution. :-) Btw, on a related note - should the defaults be `nil` or something along the lines of the values you had hardcoded?

Malabarba:
No idea. The values I had originally might have been too conservative.
But then, it's a safe bet that the user will never want to see a list of length 2000 or something. So it makes sense to have some number as default.

PR #1195, opened 2015-07-10
Double (Clojure + ClojureScript) client REPLs

Malabarba:
First, one needs to add these to `projec.clj` (as described on the Readme):

    :dependencies [[com.cemerick/piggieback "0.2.1"]
                   [org.clojure/tools.nrepl "0.2.10"]
                   [org.clojure/clojure "1.7.0"]]
    :repl-options {:nrepl-middleware [cemerick.piggieback/wrap-cljs-repl]}

and then invoke `M-x cider-jack-in-clojurescript`. This should create two repl buffers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1195](https://github.com/clojure-emacs/cider/pull/1195): CIDER can [create cljs REPLs](https://github.com/clojure-emacs/cider#clojurescript-usage). 
 * [#1191](https://github.com/clojure-emacs/cider/pull/1191): New custom variables `cider-debug-print-level` and `cider-debug-print-length`.
 * [#1188](https://github.com/clojure-emacs/cider/pull/1188): New debugging tool-bar.
 * [#1187](https://github.com/clojure-emacs/cider/pull/1187): The list of keys displayed by the debugger can be configured with `cider-debug-prompt`.
--- a/README.md
+++ b/README.md
@@ -717,45 +717,48 @@ section of your Leiningen project's configuration.
 
 ClojureScript support relies on the
 [piggieback](https://github.com/cemerick/piggieback) nREPL middleware being
-present in your REPL session. Version 0.2.0 or higher is recommended, and the
-below examples assume this, but version 0.1.5 is currently also supported.
+present in your REPL session.
 
-* Example usage of a non-browser connected Node.js REPL:
+1. Add the following dependencies to your `project.clj`
 
-  - At the Clojure REPL:
+   ```clojure
+   [com.cemerick/piggieback "0.2.1"]
+   [org.clojure/clojure "1.7.0"]
+   ```
 
-    ```clojure
-    (require '[cemerick.piggieback :as piggieback])
-    (require '[cljs.repl.node :as node])
-    (piggieback/cljs-repl (node/repl-env))
-    ```
+   as well as the following option:
 
-* Example usage of browser-connected Weasel REPL (requires
-e.g. `[weasel "0.6.0"]` in your project's `:dependencies`):
+   ```clojure
+   :repl-options {:nrepl-middleware [cemerick.piggieback/wrap-cljs-repl]}
+   ```
 
-  - At the Clojure REPL:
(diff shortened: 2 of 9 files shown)


bbatsov:
Overall things are looking good. 

> The cider-cljs-repl-code variable, in particular, should probably be made into middleware.

Probably. It's possible values could simply enumerate the types of REPLs supported by piggieback.

The presentation of connection info should also be augmented to show whether something is `clj/cljs`. Probably the same goes for the REPL buffer names.

Malabarba:
> Probably. It's possible values could simply enumerate the types of REPLs supported by piggieback.

I like that, and I've settled on that for now. I've also added that to the Readme.

> The presentation of connection info should also be augmented to show whether something is clj/cljs. Probably the same goes for the REPL buffer names.

Done.

I think this looking nice now. I'm gonna test it out a bit more, but this might be good to merge.

Malabarba:
I played around a bit and things seem to be working. 
The only issue is that the rhino repl is  the only one that works for me. The Node repl which we recommend in readme didn't work for me (it threw some exceptions and resulted in a cljs REPL that just kept throwing exceptions).

bbatsov:
> The Node repl which we recommend in readme didn't work for me (it threw some exceptions and resulted in a cljs REPL that just kept throwing exceptions).

You do have `nodejs` installed, right? :-) I think only `rhino` works out of the box, because `rhino` is part of the JVM.

cichli:
@Malabarba thanks for looking into this - I've thought about adding support for this in the past but never got around to actually doing it :-).

* Some projects (for example, those that use [figwheel](https://github.com/bhauman/lein-figwheel)) might not invoke `cemerick.piggieback/cljs-repl` directly to start a REPL - we should just include the call to it in the`cider-cljs-repl` custom
* Additionally, some projects are single language (either CLJ or CLJS), but would still benefit from two sessions/REPLs for files in different parts of the class path (e.g. `src/server/` and `src/client/`).

Could we have a list of `'(session-name filespec init)` triples to open this up to configuration? e.g. 
```emacs-lisp
'((clj "*.clj" nil)
  (cljs "*.cljs" "(cemerick.piggieback/cljs-repl ..."))
```

We would then just prompt for the session to start when invoking `cider-jack-in` (rather than hardcoding a specific command for CLJS), and have a similar command for starting a new session on top of an existing connection. Thoughts?

bbatsov:
> Additionally, some projects are single language (either CLJ or CLJS), but would still benefit from two sessions/REPLs for files in different parts of the class path (e.g. src/server/ and src/client/).

I don't quite understand what you mean. 

> We would then just prompt for the session to start when invoking cider-jack-in (rather than hardcoding a specific command for CLJS), and have a similar command for starting a new session on top of an existing connection. Thoughts?

What exactly are we going to ask the user for? Is this config going to be something in the projects themselves?

Malabarba:
@bbatsov 
> You do have nodejs installed, right? :-)

Oh, that might be it. 😅

@cichli
> Some projects (for example, those that use figwheel) might not invoke cemerick.piggieback/cljs-repl directly to start a REPL we should just include the call to it in the cider-cljs-repl custom

Good to know. I'll do that then. 

> Could we have a list of '(session-name filespec init) triples to open this up to configuration? e.g. 

I like the idea of the variable. I don't think we need to prompt the user. They can configure it as a dir local variable for the project, and if jack-in detects this variable is configured it automatically creates the specified REPLs (once the nrepl server has been started, creating REPL sessions is fast). 
That said, the current PR is large enough as it is. Can we file that as a separate issue?

bbatsov:
Overall I think the current state is good enough to merge when the small outstanding issues are addressed. We'll continue to refine this functionality down the road. I can envision other small checks - e.g. verify Clojure is 1.7+, if you select a node repl check you have `node`, etc.

Malabarba:
> Some projects (for example, those that use figwheel) might not invoke cemerick.piggieback/cljs-repl directly to start a REPL - we should just include the call to it in thecider-cljs-repl custom

Done now.

> We'll continue to refine this functionality down the road

Agreed. With should open an issue with the points discussed here.

bbatsov:
There's a failing test.

Malabarba:
Yeah I see it. Pushed againnow...

bbatsov:
OK, the first step was made. Now it's refinement time.

PR #1201, opened 2015-07-13
Integrate overlays with interactive evaluation

### LINKED ISSUE #1196: Expose an option to display interactive eval results results using an overlay
Basically it'd be awesome if everyone could use @Malabarba's debug overlay display for regular interactive evaluation. This should be an option, alternative to the current display of results in the minibuffer.

Malabarba:
Closes #1196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1201](https://github.com/clojure-emacs/cider/pull/1201): Integrate overlays with interactive evaluation. `cider-use-overlays` can be used to turn this on or off.
 * [#1195](https://github.com/clojure-emacs/cider/pull/1195): CIDER can [create cljs REPLs](https://github.com/clojure-emacs/cider#clojurescript-usage). 
 * [#1191](https://github.com/clojure-emacs/cider/pull/1191): New custom variables `cider-debug-print-level` and `cider-debug-print-length`.
 * [#1188](https://github.com/clojure-emacs/cider/pull/1188): New debugging tool-bar.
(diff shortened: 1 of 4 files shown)


bbatsov:
Just one comment from me - won't it better to have the option select either `minibuffer`, `overlays` or `both` as the result's display scheme. Generally I'd never opt to use them both at the same time, as dislike information redundancy. I guess others might share my feelings on the subject.

Malabarba:
> Just one comment from me - won't it better to have the option select either minibuffer, overlays or both as the result's display scheme

Done. I also set the default duration of overlays to be "until next command" which mimics the echo-area.
I also ended up rewriting some stuff and moving some stuff out of cider-interaction, so see if you agree with it.

bbatsov:
:+1: Looks good to me. Not sure about the default value for `cider-user-overlays`, but let's merge it as is and see what the users say about it.

arrdem:
:+1: yey more pretty!

Malabarba:
:+1: 

And I'm not sure about the default value either.

Malabarba:
Here's a data point for the default-value discussion. When you get a result of several lines, having it print in both places will completely take up the window.

bbatsov:
@Malabarba Maybe we should just leave the old version as the default and add a note and a screenshot to the README, regarding the overlays. I'm totally fine with making them the default as well, I'm just worried about potential bugs, as this hasn't been tested that extensively yet.

Malabarba:
@bbatsov Yes, I've been using it and uncovering a few corner cases. There'll be a PR related to that soon.
It might be good to turn this off before release (but maybe keep it on until then?).

bbatsov:
You're a mind reader - this is exactly what I've been thinking about. It makes a lot of sense to solicit feedback about the feature right now, so let's leave it on until the release.

PR #1202, opened 2015-07-14
[Fix #1052] Use recursive-edit in debug-mode

### LINKED ISSUE #1052: The debugger doesn't work with nrepl-sync-request:eval

Malabarba:


bbatsov:
Nice solution!

PR #1205, opened 2015-07-15
Fix bug when completing with no prefix near point.

lewang:
Repro:

1. Insert in clojure buffer:

        (let [_] (partition-by (apply <)))

2. Put point at "_".
3. `C-d` to delete the underscore.
3. `C-c C-d C-d s p <tab>`

Error is thrown.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1026,7 +1026,7 @@ form, with symbol at point replaced by __prefix__."
              (context (cider-defun-at-point))
              (_ (beginning-of-defun))
              (expr-start (point)))
-        (concat (substring context 0 (- pref-start expr-start))
+        (concat (when pref-start (substring context 0 (- pref-start expr-start)))
                 "__prefix__"
                 (substring context (- pref-end expr-start)))))))
 


bbatsov:
I can't reproduce this. I'm getting completion candidates as expected.

Malabarba:
Me neither.

lewang:
Try deleting the "_".

Malabarba:
Still fine for me. Maybe it's a difference with emacs versions.

Malabarba:
I instrumented `cider-completion-get-context-at-point` and it doesn't get called for me when I follow your recipe (assuming I understood it properly).

lewang:
that should be `cider-completion-symbol-start-pos`?

lewang:
I clarified the repro steps.

bbatsov:
It still works for me just fine.

Malabarba:
Anyway, the patch looks reasonable. From its definition I see that `(cider-completion-symbol-start-pos)` can indeed return nil, so something should be done in case `pref-start`  is nil.

bbatsov:
Fair enough.

@lewang Remove the `.` from commit message's title.

P.S. I find it really disturbing that `concat` permits `nil`s.

Malabarba:
> P.S. I find it really disturbing that concat permits nils.

Really? I find it convenient. :-)

lewang:
Fixed up commit message.

bbatsov:
@Malabarba And makes it hard to spot certain bugs right away, because it won't blow up. :-)

@lewang :+1:

PR #1206, opened 2015-07-15
Organize the debug message handler a bit

Malabarba:
This goes with https://github.com/clojure-emacs/cider-nrepl/pull/236

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -1,4 +1,4 @@
-;;; cider-debug.el --- CIDER interaction with clj-debugger  -*- lexical-binding: t; -*-
+;;; cider-debug.el --- CIDER interaction with the cider.debug nREPL middleware  -*- lexical-binding: t; -*-
 
 ;; Copyright © 2015 Artur Malabarba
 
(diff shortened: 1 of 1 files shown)


Malabarba:
Not sure what's goingon, but the test failed with `Invalid function: pushnew`.

bbatsov:
Seems like an issue with cask 0.7.2. //cc @lunaryorn @rejeep 
I'll rollback to 0.7.0.

bbatsov:
Interestingly the same version of Cask is working just fine for Projectile.

lunaryorn:
@bbatsov I don't think that it's Cask's fault. We don't use `pushnew` anywhere.  I rather think that you are lacking a `(require 'cl)` somewhere.

bbatsov:
Yep, you're right. Guess cask 0.7.0 used to require `cl`, but stopped doing so by `0.7.2`. I've just fixed this in cider.

rejeep:
@bbatsov Don't think so, see the diff: https://github.com/cask/cask/compare/v0.7.0...v0.7.2

PR #1207, opened 2015-07-16
Small internal improvements to cider prompts

Malabarba:
This allows `C-c C-d C-d` for instance to work even if you've typed a space after the symbol.
On the other hand, it makes it harder to get an empty prompt, which some might find annoying.

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -61,7 +61,7 @@ the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
 opposite of what that option dictates."
   (interactive "P")
   (funcall (cider-prompt-for-symbol-function arg)
-           "Grimoire doc for: "
+           "Grimoire doc for"
            #'cider-grimoire-web-lookup))
 
 (defun cider-create-grimoire-buffer (content)
@@ -100,7 +100,7 @@ the value of `cider-prompt-for-symbol'. With prefix arg ARG, does the
 opposite of what that option dictates."
   (interactive "P")
   (funcall (cider-prompt-for-symbol-function arg)
-           "Grimoire doc for: "
+           "Grimoire doc for"
            #'cider-grimoire-lookup))
 
 (provide 'cider-grimoire)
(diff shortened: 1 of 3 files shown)


bbatsov:
> On the other hand, it makes it harder to get an empty prompt, which some might find annoying.

Like me. :-) Seems that after this change getting an empty prompt will be pretty hard.  Something like this can be configurable, though. 

I love the fix about ignoring the REPL prompt, btw.

Malabarba:
> Like me. :-) Seems that after this change getting an empty prompt will be pretty hard. Something like this can be configurable, though. 

Another alternative is to use the symbol as the `default` parameter instead of the `initial-input`, which makes it easily available without taking up the prompt.

I *think* fewer people would be bothered by that, given that the only disadvantage is that it's harder to edit the provided symbol and I think that's not very common. (you can still do it with `M-n`, but most people won't know)

Malabarba:
Pushed an example. Turns out the `default-value` arg on read-from-minibuffer doesn't work. ¬¬

expez:
Getting an empty prompt is like one chord away, right?  Just hit `backward-kill-word` or `backward-kill-line`?  I've bound `C-u` to `backward-kill-line` in the minibuffer myself and not being able to edit would be a huge bummer.

Malabarba:
When the prompt has a `default-value` instead of an `initial-input` both cases are 0 keys away. You have an empty input to start typing, or you can just use the symbol at point with `RET`. That's why the initial-value versionis considered obsolete.

bbatsov:
Btw, this commit actually contains 3 changes. Maybe we should merge the 2 that are clear cut and discuss the behaviour of `symbol-at-point` separately.

On a related note - functionality like this should be probably be in `clojure-mode`. Originally I stuffed it here, as I had no control over `clojure-mode` and it was hard for me to get changes inside it. Now it's way easier. :-)

Malabarba:
Sure. I can see the repl-prompt thing is clear cut, but which is the other one?

bbatsov:
I saw you extracted away the need for `: `. :-)

Malabarba:
Both done. :+1: 
The part that changes the actual content of the prompt is stashed away.

bbatsov:
:+1:

PR #1208, opened 2015-07-16
Address some corner cases

Malabarba:
- Add some safeguards to stop the debugger when the buffer is killed or reverted, so you don't get stuck in an unrecoverable debugging session.
- Change display-eval-result to be smarter when it receives a marker, so it does the right thing when the original buffer was killed before evaluation finished. 
- If the result overlay would be out of view (or in a hidden buffer) it will also display the result as a message regardless of user option.
- Hiding the message is better handled now. Instead of messaging the result and then messaging `nil` to hide it, we simply message the result with an `invisible` property. This means it's not displayed in the echo area, but shows up on the Messages buffer.
- `interactive-eval` takes a `bounds` third argument (instead of a point). This allows it to properly provide a good marker to `display-eval-result` (isntead of just giving it the `point-marker`), so the result of the evaluation is always displayed at the end of the sexp, instead of at line where point happens to be located.

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -298,6 +298,8 @@ In order to work properly, this mode must be activated by
       (if cider--debug-mode-response
           (nrepl-dbind-response cider--debug-mode-response (input-type)
             (setq-local tool-bar-map cider--debug-mode-tool-bar-map)
+            (add-hook 'kill-buffer-hook #'cider--debug-quit nil 'local)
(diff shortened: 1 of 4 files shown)


bbatsov:
Apart from the 2 small typos the changes look good. You might consider adding those nice explanations of your changes in the PR to the commit messages themselves.

Malabarba:
All done.

bbatsov:
:+1:

PR #1209, opened 2015-07-16
Better positioning for the result overlay

Malabarba:
cider--make-result-overlay ensures the the overlay is actually at the
end of the sexp.


bbatsov:
:+1:

PR #1211, opened 2015-07-17
Convert version warning to error

Malabarba:
I've had this cause issues to 4 different people already.
Apparently, people don't pay enough attention to warnings. Maybe an error will impose more respect...

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -287,7 +287,9 @@ Signal an error if it is not supported."
    (lambda (_buffer result)
      (let ((middleware-version (read result)))
        (unless (and middleware-version (equal cider-version middleware-version))
-         (cider-repl-emit-interactive-err-output (format "WARNING: CIDER's version (%s) does not match cider-nrepl's version (%s)" cider-version middleware-version)))))
+         (cider-repl-emit-interactive-err-output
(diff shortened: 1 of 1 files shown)


bbatsov:
Maybe this should be a real error, I don't know. Just changing the label is unlikely to improve anything.

Malabarba:
Well, one person did say "I thought that was just a warning." ¬¬
I'm afraid throwing an error might interrupt other code from running and thus cause even more breakage. So maybe a message?

bbatsov:
> Well, one person did say "I thought that was just a warning." ¬¬

Well, technically speaking it's exactly a warning. In some situations there might be no real differences (e.g. shortly after a release) and some people really want to use the basic functionality when connecting to a remote process where the middleware can't be installed.

Perhaps this should be a config setting - an error for version mismatches by default; power users would be able to override this. Not having installed the middleware will result in a (potentially) suppressible warning. Btw, probably it'd be best if we had different messages for missing `cider-nrepl` and a version mismatch.

bbatsov:
Well, it won't hurt to change this for now, but long term I still believe we can improve those startup warnings/errors/whatever.

Malabarba:
Sorry, I completely forgot about this. Been a busy couple of weeks.

PR #1212, opened 2015-07-17
Add pagination of long collections to inspector

alexander-yakushev:
Frontend part of the pagination feature.

Keys are arbitrary, suggestions are welcome.

- `.` for next page
- `,` for previous page
- `z` to set a new page size inside the inspector

Missing features:

- [ ] specifying a default page size. Right now it is hardcoded to 32.

Sister pull request is clojure-emacs/cider-nrepl#238.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -46,6 +46,9 @@
     (define-key map [mouse-1] #'cider-inspector-operate-on-click)
     (define-key map "l" #'cider-inspector-pop)
     (define-key map "g" #'cider-inspector-refresh)
+    (define-key map "SPC" #'cider-inspector-next-page)
+    (define-key map "M-SPC" #'cider-inspector-prev-page)
+    (define-key map "s" #'cider-inspector-set-page-size)
     (define-key map [tab] #'cider-inspector-next-inspectable-object)
     (define-key map "\C-i" #'cider-inspector-next-inspectable-object)
     (define-key map [(shift tab)] #'cider-inspector-previous-inspectable-object) ; Emacs translates S-TAB
@@ -122,6 +125,34 @@ Used for all inspector nREPL ops."
   (interactive)
   (nrepl-send-request (list "op" "inspect-refresh"
                             "session" (nrepl-current-session))
+                      (cider-inspector-response-handler (current-buffer))))
+
+(defun cider-inspector-next-page ()
+  "Jump to the next page when inspecting a paginated sequence/map.
+
+Does nothing if already on the last page."
+  (interactive)
+  (nrepl-send-request (list "op" "inspect-next-page"
+                            "session" (nrepl-current-session))
+                      (cider-inspector-response-handler (current-buffer))))
+
+(defun cider-inspector-prev-page ()
+  "Jump to the previous page when expecting a paginated sequence/map.
+
+Does nothing if already on the first page."
+  (interactive)
+  (nrepl-send-request (list "op" "inspect-prev-page"
+                            "session" (nrepl-current-session))
+                      (cider-inspector-response-handler (current-buffer))))
+
+(defun cider-inspector-set-page-size (page-size)
+  "Set the page size in pagination mode to the specified value.
+
+Current page will be reset to zero."
+  (interactive "nPage size:")
+  (nrepl-send-request (list "op" "inspect-set-page-size"
+                            "session" (nrepl-current-session)
+                            "page-size" page-size)
                       (cider-inspector-response-handler (current-buffer))))
 
 ;; Render Inspector from Structured Values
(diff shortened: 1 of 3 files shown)


cider-inspector.el
```diff
@@ -124,6 +127,26 @@ Used for all inspector nREPL ops."
...
+                      (cider-inspector-response-handler (current-buffer))))
+
+(defun cider-inspector-prev-page ()
+  (interactive)
+  (nrepl-send-request (list "op" "inspect-prev-page"
+                            "session" (nrepl-current-session))
+                      (cider-inspector-response-handler (current-buffer))))
+
+(defun cider-inspector-set-page-size (page-size)
+  (interactive
```
Malabarba: `(interactive "nPage size: ")` is probably enough (and might be slightly more appropriate).

cider-inspector.el
```diff
@@ -46,6 +46,9 @@
     (define-key map [mouse-1] #'cider-inspector-operate-on-click)
     (define-key map "l" #'cider-inspector-pop)
     (define-key map "g" #'cider-inspector-refresh)
+    (define-key map "." #'cider-inspector-next-page)
+    (define-key map "," #'cider-inspector-prev-page)
```
expez: `n` and `p` would be more idiomatic, I think.

Malabarba:
Looks good to me. (The clj side too).

For the keys, I think `jk` or `hl` will make sense to a lot of (vim) people. If we want to imitate Emacs help buffers, `C-c C-b` and `C-c C-f` are the keys used.

That said, having actual `previous/next page` buttons would probably be the most help to most people.

cider-inspector.el
alexander-yakushev: I personally bind `n` and `p` to move between lines in the inspector. If anything, horizontal movement (`f` and `b`) is more appropriate.

Malabarba:
I just noticed this, but docstrings would be nice too. Specially since they're user level commands.

alexander-yakushev:
Done.

cider-inspector.el
bbatsov: I'm pretty sure the standard keybinding for this are `SPC` and `S-SPC`.

bbatsov:
Btw, how did you derive the 32 default?

alexander-yakushev:
> Btw, how did you derive the 32 default?

By raising 2 to the power of 5 ;).

bbatsov:
> specifying a default page size. Right now it is hardcoded to 32.

So the only question now is are we doing this as part of the current PR. I can imagine it'd be nice for this to be some `defcustom`, but there should also be some command to change the value on the fly.

alexander-yakushev:
There is a way to change it on the fly with `z`. If we want to hold the default in a variable, we have to pass it at the beginning of the inspection, I suppose.

bbatsov:
> There is a way to change it on the fly with z. If we want to hold the default in a variable, we have to pass it at the beginning of the inspection, I suppose.

Yeah. The debugger does this for its object print length settings. That should be easy to do for the inspector as well.

> Please also suggest a replacement for z. I don't like it personally, but it was the first thing that came into my mind thinking of siZe.

No brilliant ideas here. I'd just suggest something that's easier to press (doesn't require the use of the little finger).

alexander-yakushev:
`s` to keep it somewhat mnemonic, and the binding is uncommon enough.

Done with the default page size.

bbatsov:
I'm fine with `s`.

cider-inspector.el
```diff
@@ -103,7 +120,8 @@ Used for all inspector nREPL ops."
 (defun cider-inspect-expr (expr ns)
   (cider--prep-interactive-eval expr)
   (nrepl-send-request (append (nrepl--eval-request expr ns)
-                              (list "inspect" "true"))
+                              (list "inspect" "true"
+                                    "page-size" cider-inspector-page-size))
```
bbatsov: Sending `nil` to nREPL will byte you, as it gets encoded as an empty vector. :-) It's best to simply not send it at all, when `nil`. Here's an example.

alexander-yakushev: So do I check for nil first, and if it is `nil`, I don't send that value? Sounds messy. Maybe just `(or` it with a hardcoded default?

bbatsov: That's what we do everywhere. It's not ideal, but you can't simply encode EL `nil` as a Clojure `nil`, as it's also an empty list in Emacs Lisp, and you can never know what exactly was it used to represent.

alexander-yakushev:
Added a call to `or`. The case is already very marginal, someone will have to deliberately change the default page size and set it to `nil`. I'd say let them burn then, but OK.

bbatsov:
I'd say the Emacs part looks solid. Just get the Clojure tests passing and I'll have this merged.

alexander-yakushev:
They are passing. Eastwood and cljfmt cannot into modern threading macros.

bbatsov:
Victory!

alexander-yakushev:
:tada:

PR #1214, opened 2015-07-18
[Fix #1213] Buffer checking in nrepl-make-response-handler

### LINKED ISSUE #1213: cider-repl-set-ns appears to be broken in latest cider snapshot
When using cider with the following versions:
```sh
#CIDER 0.10.0snapshot (Java 1.8.0_11, Clojure 1.7.0, nREPL 0.2.10)
```
After running `cider-jack-in`,
When I run `cider-repl-set-ns` (either from a clojure file or the repl and manually choosing a namespace)
the namespace is not changed in the REPL.

In the `*nrepl-messages*` buffer, I see the following exchange happening:
```
(--->
  op  "eval"
  session  "8114fab7-590b-4747-be75-6b33e1453dcf"
  code  "(in-ns 'awesome-o.parser)"
  id  "43"
)
(<-
  id  "7"
  instrumented-defs  nil  ns  "awesome-o.parser"
  session  "bcf2b5f6-c960-40b2-bdb7-51ccddc5a2aa"
  status  ("instrumented-defs")
)
(<-
  id  "7"
  instrumented-defs  nil  ns  nil  session  "bcf2b5f6-c960-40b2-bdb7-51ccddc5a2aa"
  status  ("instrumented-defs")
)
(<-
  id  "43"
  ns  "awesome-o.parser"
  session  "8114fab7-590b-4747-be75-6b33e1453dcf"
  value  "#object[clojure.lang.Namespace 0xa1a8e1b \"awesome-o.parser\"]"
)
(<-
  id  "43"
  session  "8114fab7-590b-4747-be75-6b33e1453dcf"
  status  ("done")
)
```

If I manually run `(in-ns 'awesome-o.parser)` in the REPL, the namespace is properly updated.

Malabarba:


Malabarba:
Why does cider pass buffer names around instead of buffers?

bbatsov:
> Why does cider pass buffer names around instead of buffers?

Likely just an oversight. We should definitely correct this, as naming `buffer` something that's not is super confusing.

PR #1215, opened 2015-07-19
Case handling in nrepl-project-directory-for

aJchemist:
Some case example:

- `/a/project.clj`
- `/a/b/build.boot`

if your default-directory is `/a/b`, then in existing code `(nrepl-project-directory-for-test (nrepl-current-dir))` always returns `/a/`. So this isn't what we want.

Replaced code will returns `/a/b/`.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -468,12 +468,6 @@ is ambiguity, therefore nil is returned."
       (file-name-directory buffer-file-name)
     default-directory))
 
-(defun cider-project-directory-for (dir-name)
-  "Return the project directory for the specified DIR-NAME."
-  (when dir-name
-    (or (locate-dominating-file dir-name "project.clj")
-        (locate-dominating-file dir-name "build.boot"))))
-
 (defun cider-set-relevant-connection (&optional do-prompt)
   "Try to set the current REPL buffer based on the the current Clojure source buffer.
 If succesful, return the new connection buffer.
@@ -482,10 +476,10 @@ directory."
   (interactive "P")
   (cider-ensure-connected)
   (let* ((project-directory
-          (cider-project-directory-for
+          (clojure-project-dir
            (or (when do-prompt
                  (read-directory-name "Project: "
-                                      (cider-project-directory-for (buffer-file-name))
+                                      (clojure-project-dir (buffer-file-name))
                                       nil
                                       'confirm))
                (cider-current-dir))))
--- a/cider.el
+++ b/cider.el
@@ -229,7 +229,7 @@ own buffer."
     (if (funcall (cider-command-present-p project-type))
         (let* ((project (when prompt-project
                           (read-directory-name "Project: ")))
-               (project-dir (cider-project-directory-for
+               (project-dir (clojure-project-dir
                              (or project (cider-current-dir))))
                (params (if prompt-project
                            (read-string (format "nREPL server command: %s "
@@ -337,7 +337,7 @@ of list of the form (project-dir port)."
          (proj-ports (mapcar (lambda (d)
                                (-when-let (port (and d (nrepl-extract-port (cider--file-path d))))
                                  (list (file-name-nondirectory (directory-file-name d)) port)))
-                             (cons (cider-project-directory-for dir) paths))))
+                             (cons (clojure-project-dir dir) paths))))
     (-distinct (delq nil proj-ports))))
 
 (defun cider--running-nrepl-paths ()
@@ -355,7 +355,7 @@ Use `cider-ps-running-nrepls-command' and `cider-ps-running-nrepl-path-regexp-li
 (defun cider-project-type ()
   "Determine the type, either leiningen or boot, of the current project.
 If both project file types are present, prompt the user to choose."
-  (let* ((default-directory (cider-project-directory-for (cider-current-dir)))
+  (let* ((default-directory (clojure-project-dir (cider-current-dir)))
          (lein-project-exists (file-exists-p "project.clj"))
          (boot-project-exists (file-exists-p "build.boot")))
     (cond ((and lein-project-exists boot-project-exists)
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -385,7 +385,7 @@
   (should (equal (cider--find-rest-args-position [fmt arg]) nil)))
 
 (ert-deftest test-cider-switch-to-relevant-repl-buffer ()
-  (noflet ((cider-project-directory-for (dontcare)
+  (noflet ((clojure-project-dir (dontcare)
                                         nrepl-project-dir))
     (let* ((b1 (generate-new-buffer "temp"))
            (b2 (generate-new-buffer "temp"))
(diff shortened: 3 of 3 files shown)


bbatsov:
So, you have a separate boot project inside a lein project? Looks like a highly unusual setup. I definitely don't want us to have to consider such odd setups in CIDER. Perhaps we should just make the project discovery function configurable instead.

aJchemist:
Hmm, I often construct clojure project directory with some submodules. IMHO, length-based project discovery is quite cheap.

In addition, 
- `/a/project.clj`
- `/a/b/build.boot`
- `a/b/c/blahblah`

In existing code, if your current file-buffer is `a/b/c/blahblah`, It always speak project directory is `/a/`. So I think it makes sense project directory should be `/a/b/`

bbatsov:
I'll think more about this. Nested projects are always a pain to deal with, as everyone has a different notion of how they should be handled. Perhaps in case of ambiguity we should simply ask users to manually select the project root, or delegate to Projectile if it's around.

Malabarba:
I agree that handling nested projects is a can of worms. But our current implementation (checking for project.clj first, then checking for boot.clj) also leaves something to be desired. For instance, if the user happens to have a `project.clj` file lying around in his home directory (I backup things to my home dir all the time) then cider will never see any of their boot projects (because it will find that project.clj file first).

nrepl-client.el
```diff
@@ -277,8 +277,13 @@ Bind the value of the provided KEYS and execute BODY."
 (defun nrepl-project-directory-for (dir-name)
   "Return the project directory for the specified DIR-NAME."
   (when dir-name
-    (or (locate-dominating-file dir-name "project.clj")
-        (locate-dominating-file dir-name "build.boot"))))
+    (let* ((dir-name (file-name-as-directory (abbreviate-file-name dir-name)))
```
Malabarba: Was this line necessary to you? I think it'll work fine without `file-name-as-directory` or `abbreviate-file-name`.

bbatsov:
True that. The `jack-in` code is smarter, but the project code has room for improvement. And, of course this should be pushed out of the nREPL client code, as encoding project knowledge in the nREPL client simply doesn't make sense.

nrepl-client.el
```diff
@@ -277,8 +277,13 @@ Bind the value of the provided KEYS and execute BODY."
...
+           (lein-project-dir (locate-dominating-file dir-name "project.clj"))
+           (boot-project-dir (locate-dominating-file dir-name "build.boot")))
+      (if (string-equal dir-name lein-project-dir) lein-project-dir
```
Malabarba: This line, and the 4 below, can probably be replaced with:

    (if (file-in-directory-p lein-project-dir boot-project-dir)
        lein-project-dir boot-project-dir)

Malabarba:
> this should be pushed out of the nREPL client code

We really ought to make a list of all this stuff.

bbatsov:
https://github.com/clojure-emacs/cider/issues/1216

bbatsov:
This PR has to be rebased to account for some recent changes.

bbatsov:
Also - reformat the code as suggested by @expez and add a changelog entry.

aJchemist:
@bbatsov As your guide, I've rebased PR(patch-1) code to cider origin/master. Then, some conflicts followed for nrepl-client.el file. So I've skipped original 2 commit on nrepl-client.el, and write a new commit on cider-interaction.el file.

Now, I have no idea what I've done (?!) :O

cider-interaction.el
```diff
@@ -468,11 +468,17 @@ is ambiguity, therefore nil is returned."
       (file-name-directory buffer-file-name)
     default-directory))
 
-(defun cider-project-directory-for (dir-name)
+(defun nrepl-project-directory-for (dir-name)
```
bbatsov: You shouldn't rename this. :-)

cider-interaction.el
```diff
@@ -468,11 +468,17 @@ is ambiguity, therefore nil is returned."
...
   "Return the project directory for the specified DIR-NAME."
   (when dir-name
-    (or (locate-dominating-file dir-name "project.clj")
-        (locate-dominating-file dir-name "build.boot"))))
+    (let ((lein-project-dir (locate-dominating-file dir-name "project.clj"))
+          (boot-project-dir (locate-dominating-file dir-name "build.boot")))
+      (cond
```
bbatsov: I'm pretty sure there's also an option when something is neither a `lein`, nor a `boot` project. :-)

aJchemist: Hmmmm, you're right.

aJchemist: @bbatsov As you already mentioned, It is right way to prompt for user to choose project dir over `cider-project-directory-for`.

bbatsov: I just find the cond clauses quite confusing. Basically, you need to check if both vars are true, only one of them is true or none of them is true. What you've written does the same thing, but it's not as readable IMO.

Btw, this functionality should definitely be covered by a few unit tests.

bbatsov:
@aJchemist ping :-)

aJchemist:
@bbatsov pong.

I think maybe Cider can use `clojure-project-dir` (clojure-mode.el) with a param.
```
defun clojure-project-dir () -> defun clojure-project-dir (dirname)

(defun cider-project-directory-for () ... ) ->
(defalias cider-project-directory-for clojure-project-dir)
```

bbatsov:
Yeah, when this was added to `clojure-mode` recently I thought the same thing. No need to duplicate functionality.

cider-interaction.el
```diff
@@ -471,8 +471,14 @@ is ambiguity, therefore nil is returned."
 (defun cider-project-directory-for (dir-name)
   "Return the project directory for the specified DIR-NAME."
   (when dir-name
-    (or (locate-dominating-file dir-name "project.clj")
-        (locate-dominating-file dir-name "build.boot"))))
+    (let ((lein-project-dir (locate-dominating-file dir-name "project.clj"))
+          (boot-project-dir (locate-dominating-file dir-name "build.boot")))
+      (-reduce #'(lambda (x y)
```
aJchemist: I see clojure-mode doesn't have dash.el deps. So when this code move to clojure-mode.el, `-reduce` will be bulit-in `reduce`(`cl-reduce` alias). Is that OK?

aJchemist: I see clojure-mode doesn't have dash.el deps. So when this code move to clojure-mode.el, `-reduce` will be bulit-in `reduce`(`cl-reduce` alias). Is that OK?

bbatsov: `clojure-mode` depends on Emacs 24.3, which has `cl-lib`. Using stuff from `cl` is forbidden.

bbatsov: Btw, isn't using `reduce` here an overkill?

cider-interaction.el
```diff
@@ -468,11 +468,7 @@ is ambiguity, therefore nil is returned."
       (file-name-directory buffer-file-name)
     default-directory))
 
-(defun cider-project-directory-for (dir-name)
-  "Return the project directory for the specified DIR-NAME."
-  (when dir-name
-    (or (locate-dominating-file dir-name "project.clj")
-        (locate-dominating-file dir-name "build.boot"))))
+(defalias cider-project-directory-for clojure-project-dir)
```
bbatsov: It'd be better to simply use `clojure-project-dir` directly in CIDER. This alias is useless.

bbatsov:
And squash the commits together. :-)

aJchemist:
@bbatsov YOU'RE MY BADASS GIT TRAINER! :+1:

bbatsov:
:+1: Yep, that's me! Thanks!

PR #1219, opened 2015-07-19
Refresh log + support for `before` and `after` refresh middleware opts

cichli:
Depends on clojure-emacs/cider-nrepl#233. Current UI looks like this:

screenshot 2015-07-19 18 40 50

Feedback welcome!

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -2021,53 +2021,39 @@ opposite of what that option dictates."
             "symbol" sym)
       (cider-interactive-eval-handler (current-buffer))))))
 
-(defun cider-refresh--handle-response (response buffer)
+(defun cider-refresh--handle-response (response log-buffer)
   (nrepl-dbind-response response (out err reloading status error error-ns after before)
-    (cond (out
-           (cider-emit-into-popup-buffer buffer out))
+    (cl-flet ((log (message &optional face)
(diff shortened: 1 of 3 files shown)


bbatsov:
The code looks great, but this definitely deserves a changelog entry and some documentation in the README. We wouldn't want people to be totally unaware about such great functionality, right?

cider-interaction.el
```diff
@@ -183,6 +183,27 @@ if the candidate is not namespace-qualified."
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defconst cider-refresh-log-buffer "*cider-refresh-log*")
+
+(defcustom cider-refresh-before-fn nil
+  "Function for `cider-refresh' to call before reloading.
```
Malabarba: "Clojure function ..." I had to read this twice to understand it wasn't an emacs hook. :)

bbatsov:
Forget my remark about the changelog. I'm way too tired today...

cider-interaction.el
cichli: :+1: good spot, updated.

cichli:
> The code looks great, but this definitely deserves a changelog entry and some documentation in the README.

Looks like `cider-refresh` isn't mentioned anywhere in the README at all right now - not even the keybindings! Putting something together now.

bbatsov:
:+1:

PR #1228, opened 2015-07-28
Uppercase the print cider version

jonathanchu:
#### What's this PR do?
This PR simply uppercases the `cider-version` in the middleware check callback.  

```
ERROR: CIDER's version (0.10.0-SNAPSHOT) does not match cider-nrepl's version (0.9.0). Things will break!
```

This is also keeps things consistent with the other areas where `upcase` is applied to `cider-version`.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -310,7 +310,7 @@ Signal an error if it is not supported."
        (unless (and middleware-version (equal cider-version middleware-version))
          (cider-repl-emit-interactive-err-output
           (format "ERROR: CIDER's version (%s) does not match cider-nrepl's version (%s). Things will break!"
-                  cider-version middleware-version)))))
+                  (upcase cider-version) middleware-version)))))
    '()
    '()
    '()))


bbatsov:
While I understand your motivation, Emacs doesn't upcase the word `snapshot`, therefore the current state of affairs.

jonathanchu:
@bbatsov point taken, thanks for the explanation!  So just for clarity, are these `upcase` usages on `cider-version` a moot case here?  

https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L258
https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L267

This is what started me on this line of thinking.  Thanks for your time on this!

jonathanchu:
Happy to close this PR by the way! :beers:

Malabarba:
:beers:

bbatsov:
> @bbatsov point taken, thanks for the explanation! So just for clarity, are these upcase usages on cider-version a moot case here?
> 
> https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L258
> https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L267
> 
> This is what started me on this line of thinking. Thanks for your time on this!

Here we generate the middleware version, so the `upcase` usage there is needed.

jonathanchu:
Ah, thanks for clarifying @bbatsov! :smile:

PR #1229, opened 2015-07-28
Include a link on how to fix wrong nrepl version

expez:
People keep asking about this, which is why we added this section to
the readme, but they're still not finding it...

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -297,9 +297,13 @@ Signal an error if it is not supported."
     (if nrepl-version
         (when (version< nrepl-version cider-required-nrepl-version)
           (cider-repl-emit-interactive-err-output
-           (format "WARNING: CIDER requires nREPL %s (or newer) to work properly" cider-required-nrepl-version)))
+           (cider--insert-readme-button
+            (format "WARNING: CIDER requires nREPL %s (or newer) to work properly"
+                    cider-required-nrepl-version)
+            "warning-saying-you-have-to-use-nrepl-027")))
       (cider-repl-emit-interactive-err-output
-       (format "WARNING: Can't determine nREPL's version. Please, update nREPL to %s." cider-required-nrepl-version)))))
+       (format "WARNING: Can't determine nREPL's version. Please, update nREPL to %s."
+               cider-required-nrepl-version)))))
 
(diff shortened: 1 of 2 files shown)


cider-interaction.el
```diff
@@ -297,7 +297,9 @@ Signal an error if it is not supported."
     (if nrepl-version
         (when (version< nrepl-version cider-required-nrepl-version)
           (cider-repl-emit-interactive-err-output
-           (format "WARNING: CIDER requires nREPL %s (or newer) to work properly" cider-required-nrepl-version)))
+           (format "WARNING: CIDER requires nREPL %s (or newer) to work properly.\nSee %s for more information."
```
Malabarba: Tiny tiny thing, but I think "For more information, see %s " would be easier to read, given it's a longish url.

expez:
I agree, fixed.

bbatsov:
This link is super long and looks ugly in the REPL (on smaller displays you can even fit it on a single line), so we can't just dump it like this. We should either use a button or something like this http://www.gnu.org/software/emacs/manual/html_node/elisp/Clickable-Text.html

Malabarba:
(defun insert-readme-button (label section-id)
      "Insert a button that links to the online readme.
    LABEL is the displayed string, and SECTION-ID is where it points
    to."
      (insert-button
       label
       'follow-link t
       'action (lambda (&rest _) (interactive)
                 (browse-url (concat "https://github.com/clojure-emacs/cider#"
                                     section-id)))))
    
    
    (insert-readme-button "nREPL 0.2.7 Warning" "warning-saying-you-have-to-use-nrepl-027")

bbatsov:
:+1: Yep, something like this would do.

expez:
bump

expez:
moved `cider--insert-readme-button` to util

expez:
The text is now as it was, but it is a link to the relevant section in the readme.  This prevents the output from becoming too long, but some users might not understand that the link is clickable, thinking instead that the bold underlining is some angry warning face.

bbatsov:
Let's go with this version and see how it goes. Perhaps a better idea would be to for the link to follow the messages in parens - something like `.... (details)`, but I not sure about this either.

PR #1231, opened 2015-07-31
Fix inspector keybinds (and add a couple)

Malabarba:
SPC is kbd notation, in a pure keybind it actually means `S P C`. To
bind something to the space key we either use an actual space or use
`(kbd "SPC")`.

Also added S-SPC and page-down/up keys, since it most view-like modes
these are also bound to next and previous pages.


Malabarba:
I can remove the extra keybinds, if desired.

bbatsov:
:+1:

Issue #1217, opened 2015-07-19
Make it possible to associate a project folder with any connection


bbatsov:
Right now our primitive algorithm that matches local project directories to nREPL connections is pretty limited. It works only for connections that were started using `cider-jack-in`, because it maps a project dir to a connection at boot time. It also cannot handle very well situations when several projects are mapped to the same connection. All this makes it hard to work with several connections, as often you have to switch manually between them.

One way to improve this is to able to associate a project directory with any connection, using an interactive command, you can invoke at any time. That is simple solution that would solve the bigger part of this issue. In rare cases when some project is mapped to two connections we can opt to dispatch to the default connection (meaning business as usual). //cc @Malabarba

Malabarba:
> One way to improve this is to able to associate a project directory with any connection, using an interactive command, you can invoke at any time. That is simple solution that would solve the bigger part of this issue.

Yes, this seems most sane. Also, when you first `cider-connect`, associate whatever project you happen to be in at the time with that connection.

A more sophisticated option would be to let users name the connections somehow. This is the tricky part.
Once that is done, users could permanently specify which connection a project is associated to by setting a `cider-connection-name` variable in the `.dir-locals` file, and all we'd have to do is check the value of that variable when evaluating stuff.

bbatsov:
> A more sophisticated option would be to let users name the connections somehow. This is the tricky part.

There's already functionality to rename connections, btw. I think that setting this on boot won't be that hard either.

bbatsov:
So, the bulk of the work regarding is done. The only outstanding tasks are:

* support for nested projects
* improving the handling of multiple matching projects - although I'm not sure what's the best course of action here

bbatsov:
I think this works more or less well in it's present form, so I'll close this issue.

PR #1232, opened 2015-07-31
Add cider-load-buffer-and-switch-to-repl-buffer

jeremyheiler:
@SirSkidmore and I wrote a command `C-c M-z` to abbreviate:

```
C-c C-k
C-u C-c C-z
```

Specifically, the function will load (eval) the current buffer, change the namespace of the REPL buffer to match the currently visited source file, and then switch to the relevant REPL buffer.

We thought about parameterizing the the function to selecting the REPL buffer, but that would have to be done before `cider-load-buffer` is invoked, and weren't sure how to handle that. While attempting to do it, the message from `cider-load-buffer` would be in the way unless you know to type something to get rid of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -930,6 +930,7 @@ Keyboard shortcut                    | Description
 <kbd>C-c M-n</kbd>                   | Switch the namespace of the REPL buffer to the namespace of the current buffer.
 <kbd>C-c C-z</kbd>                   | Switch to the relevant REPL buffer. Use a prefix argument to change the namespace of the REPL buffer to match the currently visited source file.
 <kbd>C-u C-u C-c C-z</kbd>           | Switch to the REPL buffer based on a user prompt for a directory.
+<kbd>C-c M-z</kbd>                   | Load (eval) the current buffer and switch to the relevant REPL buffer. Use a prefix argument to change the namespace of the REPL buffer to match the currently visited source file.
 <kbd>C-c M-d</kbd>                   | Display default REPL connection details, including project directory name, buffer namespace, host and port.
 <kbd>C-c M-r</kbd>                   | Rotate and display the default nREPL connection.
 <kbd>C-c M-o</kbd>                   | Clear the entire REPL buffer, leaving only a prompt. Useful if you're running the REPL buffer in a side by side buffer.
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -581,6 +581,12 @@ of the namespace in the Clojure source buffer."
   (interactive "P")
   (cider--switch-to-repl-buffer (cider-current-repl-buffer) set-namespace))
 
+(defun cider-load-buffer-and-switch-to-repl-buffer (&optional set-namespace)
+  "Load the current buffer into the relevant REPL buffer and switch to it."
+  (interactive "P")
+  (cider-load-buffer)
+  (cider-switch-to-relevant-repl-buffer set-namespace))
+
 (defun cider-switch-to-last-clojure-buffer ()
   "Switch to the last Clojure buffer.
 The default keybinding for this command is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### New features
+* [#1232](https://github.com/clojure-emacs/cider/pull/1232): Add cider-load-buffer-and-switch-to-repl-buffer
 * [#1325](https://github.com/clojure-emacs/cider/issues/1325): Jump to error location when clicking on the error message in the stack-trace pop-up.
 * [#1301](https://github.com/clojure-emacs/cider/issues/1301): CIDER can do dynamic font-locking of defined variables, functions, and macros. This is controlled by the `cider-font-lock-dynamically` custom option.
 * [#1271](https://github.com/clojure-emacs/cider/issues/1271): New possible value (`always-save`) for `cider-prompt-save-file-on-load`.
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -105,6 +105,7 @@ entirely."
     (define-key map (kbd "C-c M-t v") #'cider-toggle-trace-var)
     (define-key map (kbd "C-c M-t n") #'cider-toggle-trace-ns)
     (define-key map (kbd "C-c C-z") #'cider-switch-to-repl-buffer)
+    (define-key map (kbd "C-c M-z") #'cider-load-buffer-and-switch-to-repl-buffer)
     (define-key map (kbd "C-c M-o") #'cider-find-and-clear-repl-buffer)
     (define-key map (kbd "C-c C-k") #'cider-load-buffer)
     (define-key map (kbd "C-c C-l") #'cider-load-file)


Malabarba:
This behavior would make a lot of sense to me under `C-u C-c C-k`, given how tightly related it is to `C-c C-k`.
Code looks good. :+1:

jeremyheiler:
We thought through the command sequence, and we've concluded that the `C-c M-z` family makes the most sense if we update it to behave just like `C-c C-z`, but have it always load the buffer.

```
        C-c C-z =                                               switch to repl buffer
    C-u C-c C-z =                  load namespace,              switch to repl buffer
C-u C-u C-c C-z = prompt for repl, load namespace,              switch to repl buffer

        C-c M-z =                                  load buffer, switch to repl buffer
    C-u C-c M-z =                  load namespace, load buffer, switch to repl buffer
C-u C-u C-c M-z = prompt for repl, load namespace, load buffer, switch to repl buffer
```
Using `C-c C-k` could work, but isn't as symmetrical with regard to `C-c C-z`.

```
        C-c C-k =                                  load buffer
    C-u C-c C-k =                  load namespace  load buffer, switch to repl buffer
C-u C-u C-c C-k = prompt for repl, load namespace, load buffer, switch to repl buffer

```

jeremyheiler:
I have updated the function to accept universal arguments a la `C-c C-z`.

expez:
> This behavior would make a lot of sense to me under `C-u C-c C-k`

:+1: 

This also reduces the cognitive load of yet another keybinding, but just barely :/

Malabarba:
>         C-c M-z = load buffer, switch to repl buffer

If I'm loading a file and switching to the REPL as a single action, I almost certainly want the namespace to be loaded on the REPL. 

>    C-u C-c M-z = load buffer, load namespace, switch to repl buffer

This looks much more useful than the other one. 

> Using C-c C-k could work, but isn't as symmetrical with regard to C-c C-z.

It's good to maintain symmetry where we can, but limiting the number of keybinds is important too. Cider is already starting to run out of these. ;-)

I like this proposal:

        C-c C-k = load buffer
    C-u C-c C-k = load buffer,                  load namespace  switch to repl buffer
C-u C-u C-c C-k = load buffer, prompt for repl, load namespace, switch to repl buffer

jeremyheiler:
> If I'm loading a file and switching to the REPL as a single action, I almost certainly want the namespace to be loaded on the REPL. 

A counter point: if you're working in `user` or some other namespace you've already loaded, and want to load the code from a different namespace that it already has required, and switch to the REPL in one action. Also, if you're working in `user`, theirs the benefit of having `clojure.repl` automatically referred.

I agree the more common case should be shorter. However, it isn't with `C-u C-c C-k`. The only benefit is that it's based off an existing command.

When considering `C-u C-c C-k`, it felt that it overcomplicated `cider-load-buffer` by adding the REPL and namespace switching. Also, the interactive part would have to consider that `buffer` is an optional argument. However, that could be easy to do; I'm not sure.

README.md
```diff
@@ -814,6 +814,8 @@ Keyboard shortcut                    | Description
 <kbd>C-c M-n</kbd>                   | Switch the namespace of the REPL buffer to the namespace of the current buffer.
 <kbd>C-c C-z</kbd>                   | Switch to the relevant REPL buffer. Use a prefix argument to change the namespace of the REPL buffer to match the currently visited source file.
 <kbd>C-u C-u C-c C-z</kbd>           | Switch to the REPL buffer based on a user prompt for a directory.
+<kbd>C-c M-z</kbd>                   | Load (eval) the current buffer and switch to the relevant REPL buffer. Use a prefix argument to change the namespace of the REPL buffer to match the currently visited source file.
+<kbd>C-u C-u C-c M-z</kbd>           | Load (eval) the current buffer and switch to the REPL buffer based on a user prompt for a directory.
```
bbatsov: Two prefix args?

cider-interaction.el
```diff
@@ -507,6 +507,18 @@ directory."
       (message (cider--connection-info (car nrepl-connection-list))))
     connection-buffer))
 
+(defun cider-load-buffer-and-switch-to-repl-buffer (&optional arg)
```
bbatsov: Lately we've been moving away from the usage of the name `arg`. I'd suggest naming this `switch-ns` or something like this.

bbatsov:
I'm fine with the idea, but I'm not sure what the default behaviour should be. Seems there might be some point in switching the ns by default, but on the other hand this won't be consistent with the current go-to-repl behaviour.

Malabarba:
I should probably say that I don't use the REPL buffer all that much, so take my opinion with a grain of salt. :)
In any case, I don't oppose to any of the options.

README.md
jeremyheiler: See my comment above: https://github.com/clojure-emacs/cider/pull/1232#issuecomment-126799372

bbatsov:
Looking at the code behind `C-c C-z`, the variant with selecting manually a project looks pretty strange to me. I totally don't remember what was the use-case behind this and I'll probably remove it the moment #1217 is done.

bbatsov:
OK, so `C-u C-u C-c C-z` is no more. You can update the PR to reflect this.

jonpither:
Background on `C-u C-u C-z C-z` is in #350, the origin living in Hugo's question of "What happens when there are multiple REPLs for a single project".

I would agree with simplifying it away, unless the original concern stands tall.

bbatsov:
> Background on C-u C-u C-z C-z is in #350, the origin living in Hugo's question of "What happens when there are multiple REPLs for a single project".

This was problematic before and will remain to be problematic now as well. Basically you'll just fallback to the default REPL, which seems reasonable to me.

bbatsov:
One more thing - now all operations are auto-dispatched via the "correct" connection and fallback to the default connection if the connection associated with a project cannot be determined. There's still some room for improvement, but overall people will have to deal with connection switching way less.

jonpither:
Cool. That was a bit of an elephant in the room in that if Cider knows the 'relevant/correct' connection, then why not always use that info for all operations. It's a pretty big change to make as it's in the core flow. If there's no problems for anyone, then excellent!

bbatsov:
Well, previous the biggest practical issue was that you could not associate a local project with a connection created with `cider-connect`. The problem with multiple connections sharing the same project still exists, although I do plan to add some subfolder project support which should solve this to some extent. The only unsolvable problem is what do if you have two connections to different hosts for the same project, but I think that's not pretty common.

bbatsov:
@jeremyheiler ping

jeremyheiler:
@bbatsov Sorry for the delay. My son was born shortly after opening this PR. :-)

I should have some time this week to fix it up.

bbatsov:
> @bbatsov Sorry for the delay. My son was born shortly after opening this PR. :-)

Congrats!

jeremyheiler:
@bbatsov The current master no longer prompts for a namespace. Is this intentional?

jeremyheiler:
@bbatsov I've pushed some changes, and rebased on from master.

bbatsov:
:+1:

PR #1233, opened 2015-08-01
fix NodeJS CLJS REPL starting

jeffdik:
I get a ClassNotFoundException for cljs.repl.node without this commit.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -178,7 +178,7 @@ REPL."
   :type '(choice (const :tag "Rhino"
                         "(cemerick.piggieback/cljs-repl (cljs.repl.rhino/repl-env))")
                  (const :tag "Node (requires NodeJS to be installed)"
-                        "(cemerick.piggieback/cljs-repl (cljs.repl.node/repl-env))")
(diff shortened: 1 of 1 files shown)


bbatsov:
The change seems harmless. I wonder if that's a problem for `rhino` as well. Btw, the commit message has to be capitalised.

Malabarba:
Yes, it's probably best to do it for all of them.

jeffdik:
Ok, I've capitalized the commit message.

I don't have any issues with starting a rhino cljs repl.  It might future-proof this method of starting a cljs repl a bit if cljs.repl.rhino is ever not loaded by default, but I have a slight preference for leaving it as is unless explicitly require'ing becomes necessary.

Malabarba:
One thing I still don't understand is why I sometimes need to require stuff before using and why sometimes I don't.

bbatsov:
No idea. It'd be nice to get some clarity on this.

jeffdik:
In this case, I think cljs.repl.rhino doesn't need to be required because it is already loaded by piggieback: https://github.com/cemerick/piggieback/blob/master/src/cemerick/piggieback.clj#L13

bbatsov:
Good to know. Thanks!

Issue #1236, opened 2015-08-04
Support changing/configuring the text in the REPL prompt


blueberry:
When I have a long nested namespace, it takes a lot of the screen width... So, instead of something.something.something.something.something.myns> i would like it to just display the last part, myns>

I asked on slack clojurians whether this is possible now, and Bozidar says it is not, so I assume he should know. He also said it should not be complicated to support, and seems reasonable, so I have hopes :)

Thanks for the great development environment!

cichli:
You can accomplish this using `cider-repl-prompt-function`:

```emacs-lisp
(setq cider-repl-prompt-function (lambda (ns)
                                   (format "%s> " (car (last (split-string ns "\\."))))))
```

I'll keep this open as a reminder to myself to add this to the README :-).

Malabarba:
I think it would be nice to have some more functions available for this, for the users who don't know how to write them.
I'll open a PR with a couple options.

Malabarba:
@blueberry The behaviour you wanted is now available with

    (setq cider-repl-prompt-function #'cider-repl-prompt-lastname)

But you may actually prefer this one:

    (setq cider-repl-prompt-function #'cider-repl-prompt-abbreviated)

PR #1237, opened 2015-08-04
[#1236] Add two more functions for use in cider-repl-prompt-function

Malabarba:


bbatsov:
Looks good. Add a couple of tests for those functions, a changelog entry and some README mention, as people are unlikely to find them otherwise.

Malabarba:
Changelog and tests added. I didn't add the functions to the readme because the variable itself is not on the readme, and @cichli mentioned he was going to do that. :)

bbatsov:
:+1:

PR #1241, opened 2015-08-07
Add support for `refresh-clear` middleware op and improve the `cider-refresh` documentation a bit

cichli:
The diff should be pretty self-explanatory - but is there a better way to handle the prefix args than just comparing to `4` and `16` like this?


cichli:
Addressed both comments, thanks :-).

bbatsov:
:+1:

PR #1245, opened 2015-08-07
[Fix #1222] Add option to apply a single face to the results overlay

### LINKED ISSUE #1222: Result overlay too prominent
I just saw the result overlay for the first time: 

My first impulse was: "God damnit, something just altered my file! Rude!"

As you can see in the screenshot the result of the evaluation in the overlay looks exactly like it would if I had written those characters in manually, is there some way we could make this a bit more visually distinctive?  

I think it would be really cool to change the prefix to be `;; =>` and then use the comment face for the result.

PS. After getting over my initial reaction I thought this was really cool!

Malabarba:
New variable, cider-ovelays-use-font-lock controls whether results overlay should be font-locked or just use a single face

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -31,10 +31,19 @@
 ;;; Customization
 (defface cider-result-overlay-face
   '((t :inherit font-lock-builtin-face))
-  "Face used to display result of debug step at point."
-  :group 'cider-debug
+  "Face used to display evaluation results at the end of line.
+Only used on the result string if `cider-ovelays-use-font-lock' is nil.
+If it is non-nil, this face is only used on the prefix (usually a \"=>\")."
   :group 'cider
   :package-version "0.9.1")
+
+(defcustom cider-ovelays-use-font-lock nil
+  "If non-nil, results overlays are font-locked as Clojure code.
+If nil, apply `cider-result-overlay-face' to the entire overlay instead of
+font-locking it."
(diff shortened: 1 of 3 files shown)


cider-overlays.el
```diff
@@ -31,11 +31,20 @@
...
   '((t :inherit font-lock-builtin-face))
-  "Face used to display result of debug step at point."
-  :group 'cider-debug
+  "Face used to display evaluation results at the end of line.
+Only used on the result string if `cider-ovelays-use-font-lock' is nil.
+If it is non-nil, this face is only used on the prefix (usually a \"=>\")."
   :group 'cider
   :package-version "0.9.1")
 
+(defcustom cider-ovelays-use-font-lock t
```
bbatsov: I think the default should be `nil`.

bbatsov: This should also be mentioned in the README (in the common config section).

Malabarba: > I think the default should be nil.

If overlays will be ON, then yes this should be nil.
If overlays will be OFF by default, then I think we can leave this as `t` (since people will have to explicitly enable overlays, they won't be surprised by it).

I just don't which of the two we want yet.

bbatsov: The overlays generated only positive feedback for far, so I think we'll leave them on by default.

Malabarba:
Readme added

PR #1246, opened 2015-08-08
[Fix #689] Port README.md to Texinfo

### LINKED ISSUE #689: Create a (TexInfo) user manual
I'd like us to create a basic user manual for CIDER, that would be similar in spirit and scope to SLIME's [user manual](https://github.com/slime/slime/blob/master/doc/slime.texi). Ideally it'd be a texinfo document (as those are easy to read/search in Emacs), but I'm open to other alternatives (like generating the texinfo from another more popular format or publishing a web-only user manual on a dedicated CIDER website).

The document should be more or less an extended version of the README, documenting in a bit more details the important functionality available to the users (ideally it might describe a bit more). Once we have an initial version ready we'll always keep it in sync with the latest development.

Unfortunately I won't have time to work on this any time soon, so I'd to ask for your assistance to complete this task. I think the user manual is one of the biggest missing features at this point.

vermiculus:
titlepage

There is still a lot of work to be done on the manual, both hard and soft:

* hard: The manual compiles to PDF just fine, but does not currently compile to
  HTML.  This is a required feature for the manual and must be fixed.
* soft: The README reads like a dumping ground for features – it was not highly
  organized or very consistent.  Work could be done to this point.

Updates to the README should be 'put on hold' until this is finished; we don't want the two to get out-of-sync :smile: Speaking of, the README should be *really* pared down and simplified as part of this move.


bbatsov:
> hard: The manual compiles to PDF just fine, but does not currently compile to HTML. This is a required feature for the manual and must be fixed.

Yeah, we also need to figure out where to publish the HTML version.

> soft: The README reads like a dumping ground for features – it was not highly organized or very consistent. Work could be done to this point.

True that. The README doesn't even describe the common CIDER workflow and this is a major pain-point for newcomers - they simply can't understand what are they supposed to do with CIDER.

> Updates to the README should be 'put on hold' until this is finished; we don't want the two to get out-of-sync  Speaking of, the README should be really pared down and simplified as part of this move.

Of course. The moment the manual is ready & published the README will be slimmed down to an abstract, installation and a reference to the manual.

Malabarba:
>Yeah, we also need to figure out where to publish the HTML version.

How about the gh-pages branch?

bbatsov:
Yeah, that's certainly one idea. Another would be a dedicated site.

Malabarba:
True. As long as the texi source is on the master branch of the usual repo, the decision of where to place the compiled html should be based on what's more practical to whoever is going to do  the actual compiling.

Malabarba:
Although I see little difference in the two options.

vermiculus:
> Although I see little difference in the two options.

The only difference I can think of is a different URL for the content. The only use I can see for this effect is to separate the manual from other gh-pages documentation.

Even if we did have a gh-pages branch, we could still separate the manual out like flycheck.org does.

bbatsov:
Probably for starters we can use the existing repo. We still need some domain for this, though. Maybe `cider-clojure.org` or something like this?

Malabarba:
`clojure-cider.org` might be easier to remember. But `clojure-emacs.github.io/cider` isn't the end of the world either. :)

bbatsov:
> But clojure-emacs.github.io/cider isn't the end of the world either. :)

It's not the end of the world and we might use it for a while, but we can't really expect people to remember something like this.

Malabarba:
I have a small bug report. I compiled the pdf, and all quotes in code are round quotes like this:

    (unless (package-installed-p ’cider)
      (package-install ’cider))

it's not very apparent here on Github, but they really jumped out at me in the PDF.

Malabarba:
I also have small issues with the style of inline code and keybinds display, but these stuff can (and should) be fixed on CSS, so it's not important now.

Beyond that, I think I caught a few more small issues below. This is by the page numbers listed on the actual pages (not the pdf reader)

Page **4**
 
1. Curly braces are missing from Clojure code. This is probably happening throughout the manual. Shouldn't we use a `verbatim` environment for code blocks? I think that would solve the indentation issue below too.

2. There's a paragraph starting with some `***`, and the paragraph itself looks a little out of place.

        ***
        ‘x.y.z’ should match the version of CIDER you’re currently using (say ‘0.7.1’).
        For snapshot releases of CIDER you should use the snapshot of the plugin as well (say
        ‘0.7.1-SNAPSHOT’).

Page **7**

- There's a `**root**` in there.

Page **8**

- There's a black box at the end of the second line. I think latex is signaling to us that this line goes beyond the right edge, which is really not a big deal so we can just disable whatever debug option provides that behaviour.

Page **12**

- There's an unindented code block near the bottom.

Page **38**

- “Index is nonexistent.”

Malabarba:
By the way, good job Sean. By the size of the manual, I can tell this conversion was probably a lot of work.

vermiculus:
> I have a small bug report. I compiled the pdf, and all quotes in code are round quotes like this:

This is not something easily fixed. I would open up a new issue for it – [it's pretty straightforward in LaTeX](http://tex.stackexchange.com/q/7735/17423), but I'll need to do some research for how to fix this in Texinfo. This isn't a *huge* deal in the PDF, IMHO.

> Curly braces are missing from Clojure code.

Yep, I was expecting `@lisp...@end lisp` to be a verbatim TeX block, but I was mistaken :) the braces needed to be escaped. This was fixed as part of fixing the HTML export.

> There's a paragraph starting with some `***`

This is 'fixed' – it used to be an hrule, but there's no good way to do that while maintaining compilability, from what I see. I'll make a note of this.

> There's a `**root**` in there.

Fixed, thanks :)

> There's a black box…

Yep. Well, technically it's TeX signaling the overfull hbox to us :) Texinfo is based on Plain (IIUC). Removing it is easy in LaTeX, but I'm not sure how the process works in Texinfo.

A better option would be to reword. Also affected is §5.7.

> "Index is nonexistent."

This is where the bulk of the work remains. Indexing is much less painful with TeX-based tools, but it's never what I'd call 'fun'. :wink:

doc/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+
+clean:
+	rm -f {logo,gratipay,paypal}.{png,gif,pdf}
+	rm -f cider.{aux,cp,fn,ky,log,pg,toc,tp,vr}
+
+logo.png:
+	convert -background none ../logo/cider-logo.svg logo.png
+
+gratipay.png:
```
bbatsov: The donation links will probably remain only in the README.

doc/cider.texi
```diff
@@ -0,0 +1,1712 @@
...
+@alias clojar = file
+@alias clojf = t
+
+@setfilename cider.info
+@settitle CIDER
+
+@copying
+CIDER, the Clojure Interactive Development Environment that Rocks.
+
+Copyright @copyright{} 2015  CIDER Development Project
```
bbatsov: CIDER Development Project? :-)

doc/Makefile
vermiculus: I was thinking the same as well. Removed.

doc/cider.texi
vermiculus: I didn't really know what else to put there at the time – should I just use the same information [from the README](https://github.com/clojure-emacs/cider#license) here? :-)

bbatsov:
Looking at flycheck's website and manual it seems to me we should do the very same thing for CIDER. I see @lunaryorn is even building versions of the manual for MELPA and MELPA stable separately.

bbatsov:
So, any progress here?

vermiculus:
Oh, sorry. I'll get back to work on this over the weekend. I'll try to figure out @lunaryorn's system while I'm at it :)

bbatsov:
Closing this due to lack of activity. We might revisit it down the road. I'll try to find some time to rework the README into something like a manual, so the process of converting it into something with chapters would be easier.

vermiculus:
Fair enough. It might be worthwhile to instead start porting the README to Org; it will have inherent structure then and we'll have some starting point for automatic Texinfo exporting. Not to mention it will be easier to work with... I've always found `markdown-mode` to be... challenging.

lunaryorn:
@vermiculus Org's export facilities for Texinfo are poor.

vermiculus:
@lunaryorn Poor today, but in my opinion Texinfo is poor by default. I think it can be made to happen, but obviously that would need to be in place first.

I really need to push that FSF paperwork through…

bbatsov:
For the time being I'd suggest focusing our efforts on improving the contents and structure of the existing README. The process has already started - see the indentation document in `doc/`. We really can't convert the README into a manual if it doesn't have the contents you'd find in a typical manual. 

In the end of the day it's not that important to have a Texinfo manual I think - a good documentation site will be just as useful for most people.

lunaryorn:
@vermiculus When it comes to documenting software, Texinfo is more expressive than Org.  Particularly, Org provides no syntax for structured documentation of identifiers (e.g. commands, functions, options, etc.) and no support for links to arbitrary places of documents other than headings.  Hence all the actual benefits of Texinfo (automated index generation, powerful cross-referencing, etc.) are lost when using Org and cannot easily be recovered without abusing (e.g. by using headlines for identifiers, etc.) or greatly extending the Org format.

Trust me, been there, done that.  I'm not using Texinfo in Flycheck because it's so much fun and such beauty.  I'm using it because nothing else does the job.  I've gone to great length to actually use Org for Flycheck's manual and gave up, because Org is a poor format for structured documentation.  

By the way, the Org Mode manual is written in… wait for it… Texinfo.

PR #1248, opened 2015-08-08
[Fix #1177] Add TAB and RET keys to the test-report buffer

### LINKED ISSUE #1177: TAB and RET navigation on *cider-test-report* buffers
In ert report buffers you can navigate with TAB and visit the specific test with RET. Would be nice to have something like that on cider as well.

Malabarba:


bbatsov:
This turned out to be simpler than expected. :-)

Malabarba:
Indeed :-)

PR #1249, opened 2015-08-09
Fix undefined function

vmfhrmfoaj:
`font-lock-ensure` is not defined in Emacs-24 and lower version.
related commit: a84e2c1


bbatsov:
Hmm, that's a bummer. I wonder if `font-lock-flush` is defined in older Emacsen. Anyways, thanks for spotting my mistake.

Issue #1197, opened 2015-07-12
Display some indication that we're waiting for a result for long-running evaluations


bbatsov:
We should display some spinning bar in the minibuffer while waiting for interactive evaluation results. Something similar might be useful for the REPL as well.

Malabarba:
I think the mode line might be more appropriate. The mode line process is specific designed for this kind of thing. 
Technically that's the property of the major mode, but I don't think clojure-mode has any plans to use the mode-line process. :-)

bbatsov:
I suggested the use of the minibuffer, because this is what `make-progress-indicator` uses. Basically my idea is to start showing something if the evaluation takes more than some preconfigured duration (e.g. 1 second) and to make it disappear when we hear back from nREPL. Not sure if we can use something standard for the modeline (obviously, rolling out our own progress indicator is on the table as well).

> Technically that's the property of the major mode, but I don't think clojure-mode has any plans to use the mode-line-process. :-)

You're quite right. :-)

Malabarba:
> Not sure if we can use something standard for the modeline (obviously, rolling out our own progress indicator is on the table as well).

Well, there's [spinner.el](https://github.com/Malabarba/spinner.el#spinnerel). But then, I'm biased, :-) so don't take as a statement that the mode-line would be definitely better.

bbatsov:
Your package looks cool. If you're willing to integrate it, using it, instead of the built-in is fine by me.

zarkone:
I'd like to try myself at Elisp and this feature looks like a good start. I've opened `cider-interaction.el` and my plan is to `advice-add` the `cider-interactive-eval`.  `advice` would launch `:before` and cancel `:after` `cider-interactive-eval` this charming `spinner.el` in minibuffer. I'm not sure about using `advice-add`: is it idiomatic for Elisp or is it suitable for Cider..? Waiting for your help, opinions and suggestions.

expez:
@zarkone It's not common to use advice for code that you own and can modify. The advice functionality is mostly used when interacting with code you don't own which does *almost* what you want it to.

zarkone:
@expez this makes sense, thanks. I've just opened Elisp *info* page, and red the same :)

Malabarba:
You can just hack away at the `cider-interactive-eval` function itself, no need for advices. 
The only thing is that evaluations are done asynchronously, so it's useless to run some code after the function. You have to look into the function that processes results.

zarkone:
[...]

PR #1250, opened 2015-08-10
Evaluation progress bar

zarkone:
#1197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1197](https://github.com/clojure-emacs/cider/issues/1197): Display some indication that we're waiting for a result for long-running evaluations.
 * [#1127](https://github.com/clojure-emacs/cider/issues/1127): Make it possible to associate a buffer with a connection (via `cider-assoc-buffer-with-connection`).
 * [#1217](https://github.com/clojure-emacs/cider/issues/1217): Add new command `cider-assoc-project-with-connection` to associate a project directory with a connection.
 * [#1248](https://github.com/clojure-emacs/cider/pull/1248): Add <kbd>TAB</kbd> and <kbd>RET</kbd> keys to the test-report buffer.
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -207,6 +207,27 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-spinner-type 'progress-bar
+  "Appearance of the evaluation spinner. 
+
+Value is a symbol. The possible values are the symbols in the
+`spinner-types' variable."
+  :type 'symbol
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-show-eval-spinner t
+  "When true, show the evaluation spinner in the mode line."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-spinner-delay 1
+  "Amount of time, in seconds, after which the evaluation spinner will be shown."
+  :type 'integer
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
 (defface cider-error-highlight-face
   '((((supports :underline (:style wave)))
      (:underline (:style wave :color "red") :inherit unspecified))
@@ -1728,6 +1749,19 @@ Clears any compilation highlights and kills the error window."
 (defvar-local cider-interactive-eval-override nil
   "Function to call instead of `cider-interactive-eval'.")
 
+(defun cider-eval-spinner-handler (eval-buffer original-callback)
+  "Return a response handler that stops the spinner and calls ORIGINAL-CALLBACK.
+EVAL-BUFFER is the buffer where the spinner was started."
+  (lambda (response)
+    ;; buffer still exists and
+    ;; we've got status "done" from nrepl
+    ;; stop the spinner
+    (when (and (buffer-live-p eval-buffer)
+               (member "done" (nrepl-dict-get response "status")))
+      (with-current-buffer eval-buffer
+        (spinner-stop)))
+    (funcall original-callback response)))
+
 (defun cider-interactive-eval (form &optional callback bounds)
   "Evaluate FORM and dispatch the response to CALLBACK.
 This function is the main entry point in CIDER's interactive evaluation
@@ -1745,9 +1779,16 @@ arguments and only proceed with evaluation if it returns nil."
                  (functionp cider-interactive-eval-override)
                  (funcall cider-interactive-eval-override form callback bounds))
       (cider--prep-interactive-eval form)
+      (when cider-show-eval-spinner
+        (spinner-start cider-eval-spinner-type nil
+                       cider-eval-spinner-delay))
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (if cider-eval-progress-bar-show
+           (cider-eval-spinner-handler
+            (current-buffer)
+            (or callback (cider-interactive-eval-handler nil end)))
+         (or callback (cider-interactive-eval-handler nil end)))
        ;; always eval ns forms in the user namespace
        ;; otherwise trying to eval ns form for the first time will produce an error
        (if (cider-ns-form-p form) "user" (cider-current-ns))
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -586,17 +586,24 @@ If NEWLINE is true then add a newline at the end of the input."
     (goto-char (point-max))
     (cider-repl--mark-input-start)
     (cider-repl--mark-output-start)
+    (when cider-show-eval-spinner
+        (spinner-start cider-eval-spinner-type nil
+                       cider-eval-spinner-delay))
     (if (and (not (string-match-p "\\`[ \t\r\n]*\\'" input))
              cider-repl-use-pretty-printing)
         (nrepl-request:pprint-eval
          input
-         (cider-repl-handler (current-buffer))
+         (cider-eval-spinner-handler
+          (current-buffer)
+          (cider-repl-handler (current-buffer)))
          (cider-current-ns)
          nil
          (1- (window-width)))
       (nrepl-request:eval
        input
-       (cider-repl-handler (current-buffer))
+       (cider-eval-spinner-handler
+        (current-buffer)
+        (cider-repl-handler (current-buffer)))
        (cider-current-ns)))))
 
 (defun cider-repl-return (&optional end-of-input)
--- a/cider.el
+++ b/cider.el
@@ -11,7 +11,7 @@
 ;; Maintainer: Bozhidar Batsov <bozhidar@batsov.com>
 ;; URL: http://www.github.com/clojure-emacs/cider
 ;; Version: 0.10.0-cvs
-;; Package-Requires: ((clojure-mode "4.2.0") (dash "2.4.1") (pkg-info "0.4") (emacs "24.3") (queue "0.1.1"))
+;; Package-Requires: ((clojure-mode "4.2.0") (dash "2.4.1") (pkg-info "0.4") (emacs "24.3") (queue "0.1.1") (spinner "1.4"))
 ;; Keywords: languages, clojure, cider
 
 ;; This program is free software: you can redistribute it and/or modify


bbatsov:
Guess @Malabarba should review this (as the author of `spinner.el`), but it seems off to me.

1. You didn't declare the dependency to `spinner.el`.
2. The docs said this had to be registered in the relevant minor/major modes (e.g. `cider-mode`, `cider-repl-mode`).

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1250](https://github.com/clojure-emacs/cider/pull/1250): Display some indication that we're waiting for a result for long-running evaluations
```
bbatsov: This sentence should end with a `.`.

cider-interaction.el
```diff
@@ -1715,14 +1715,19 @@ If `cider-interactive-eval-override' is a function, call it with the same
...
+        (eval-buffer (current-buffer)))
     (unless (and cider-interactive-eval-override
                  (functionp cider-interactive-eval-override)
                  (funcall cider-interactive-eval-override form callback bounds))
       (cider--prep-interactive-eval form)
+      (spinner-start 'progress-bar)
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (lambda (&rest args)
```
bbatsov: The changes to the callback here puzzle me.

zarkone:
thanks @bbatsov,
> it seems off to me
yes, sorry, i'm quite new to elisp and open source contributions..
>  had to be registered in the relevant minor/major modes
as i understood, only for minor modes and this is the way to make spinner buffer-local
> The changes to the callback here puzzle me.
hm, my logic: wrap callback with `lambda` which stops spinner in the buffer it was started

I'll fix `.` missing and `spinner` dependency in minutes, sorry >..<

cider-interaction.el
zarkone: I don't know what number of args this `callback`  can take, so i've decided to write this `apply` form.

cider-interaction.el
```diff
@@ -1715,14 +1715,19 @@ If `cider-interactive-eval-override' is a function, call it with the same
 arguments and only proceed with evaluation if it returns nil."
   (let ((form  (or form (apply #'buffer-substring bounds)))
         (start (car-safe bounds))
-        (end   (car-safe (cdr-safe bounds))))
+        (end   (car-safe (cdr-safe bounds)))
+        (eval-buffer (current-buffer)))
```
Malabarba: (Small thing, not spinner related.) I tend to add new let-binds second from the start, instead of last. This reduces the risk of silly merge/rebase conflicts. For instance, in this case I would add the `eval-buffer` bind between the `form` and the `start`.

Malabarba:
The spinner.el usage is correct.
Technically, this is making use of the major-mode functionality (which places the spinner in `mode-line-process`), but we agreed that cider may as well hijack `clojure-mode`'s process string, so I think that's fine.
The result of this is that the spinner will only be displayed in the mode-line of the buffer that started it. We can make it display on all cider buffers if we want by using the minor-mode functionality instead.

There are some things to consider that are not related to spinner though:

- Although that usually won't happen, cider commands can be invoked from any major-mode. So the code should check that it's either `clojure-mode` or `cider-repl-mode`.
- Is the callback always called? I believe it's called when the evaluation errors out, but I think it's *not* called if the user interrupts evaluation with `cider-interrupt`. If so, something has to be done in that case, and the only option I see is to walk through all buffers with one of our major-modes calling `spinner-stop` on them.

cider-interaction.el
```diff
@@ -1715,14 +1715,19 @@ If `cider-interactive-eval-override' is a function, call it with the same
...
     (unless (and cider-interactive-eval-override
                  (functionp cider-interactive-eval-override)
                  (funcall cider-interactive-eval-override form callback bounds))
       (cider--prep-interactive-eval form)
+      (spinner-start 'progress-bar)
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (lambda (&rest args)
+           (with-current-buffer eval-buffer (spinner-stop))
```
Malabarba: The eval-buffer could be dead by the time evaluation finishes, so we need to check if it's alive.

cider.el
```diff
@@ -11,7 +11,7 @@
 ;; Maintainer: Bozhidar Batsov <bozhidar@batsov.com>
 ;; URL: http://www.github.com/clojure-emacs/cider
 ;; Version: 0.10.0-cvs
-;; Package-Requires: ((clojure-mode "4.2.0") (dash "2.4.1") (pkg-info "0.4") (emacs "24.3") (queue "0.1.1"))
+;; Package-Requires: ((clojure-mode "4.2.0") (dash "2.4.1") (pkg-info "0.4") (emacs "24.3") (queue "0.1.1") (spinner "1.3.1"))
```
bbatsov: @Malabarba Any plans to add support for multiple `package-requires` to `package.el`? Those lines are getting awfully long.

cider-interaction.el
bbatsov: I wonder if we shouldn't stop the spinner, when we get status "done" from nREPL. If I were writing this code I would have put the `spinner-stop` there.

bbatsov:
Btw, how does the spinner look right now?

cider-interaction.el
Malabarba: > I wonder if we shouldn't stop the spinner, when we get status "done" from nREPL.

Good point. 
Although in that case we're starting the spinner in cider-interaction and stopping it in nrepl.el, which feels a little messy. It also won't work if the function that calls `cider-interactive-eval` specifies an unusual callback.
I think the right solution would be to indeed check for a `done` status, but do that right here where the code is now.

(*In fact, now I remember why I didn't do this before. I started thinking of where the code would go and couldn't find a solution that made me 100% happy. The start of evaluation and the handling of results are so distant that it gets complicated.*)

Malabarba:
>Btw, how does the spinner look right now?

Right now it looks like the 5th spinner on this gif: (the one like this `[=== ]`)

cider.el
Malabarba: I'll bring it up

cider-interaction.el
bbatsov: Yeah, that's part of our code coupling problem - cider details leaking in the nrepl client and vice versa.

bbatsov:
> Although that usually won't happen, cider commands can be invoked from any major-mode. So the code should check that it's either clojure-mode or cider-repl-mode.

Which reminds me that `cider-repl-mode` should be handled separately as it uses other evaluation functions.

> Is the callback always called? I believe it's called when the evaluation errors out, but I think it's not called if the user interrupts evaluation with cider-interrupt. If so, something has to be done in that case, and the only option I see is to walk through all buffers with one of our major-modes calling spinner-stop on them.

Yeah, an error or an interrupt should stop the spinner for sure.

zarkone:
@bbatsov:  @Malabarba : thanks for responses. I think my implementation is too naive and that i do not know about some cider inner parts and interactions between cider and nrepl. I'll try to investigate this moment today, any advices for this journey? 
I've already know from github discussion that i have to 
- check buffer existing 
- check `done` status from nrepl
- stop spinner on error and manual interruption 
- add `let` bindings after first form =) 
- discuss: show spinner only in one current buffer or for all buffers? (my vote is for first)

looks a bit new and unknown for me, but I believe it is not so hard.

Malabarba:
> stop spinner on error

Errors already send the `done` status. So this should not be explicitly necessary.

zarkone:
i've made the changes and tried to throw error and call `cider-interrupt` -- the callback is working, and progressbar stops. May be i've missed the point...

zarkone:
@expez thanks. i found that when it takes the buffer it returns `#<killed buffer>` instead of `nil`

cider-interaction.el
```diff
@@ -1714,15 +1714,25 @@ positions of FORM in its buffer.
...
+      (spinner-start 'progress-bar)
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (lambda (response)
+         ;; buffer still exists and
+         ;; we've got status "done" from nrepl
+         ;; stop the spinner
+         (when (and (get-buffer (buffer-name eval-buffer))
+                    (string= "done" (car (lax-plist-get (cdr response) "status"))))
```
bbatsov: Why don't you threat this as an `nrepl-dict` (which it is)?

cider-interaction.el
```diff
@@ -1714,15 +1714,25 @@ positions of FORM in its buffer.
...
+           (with-current-buffer eval-buffer (spinner-stop)))
```
bbatsov: Put `(spinner-stop)` on a line of its own.

cider-interaction.el
```diff
@@ -1714,15 +1714,25 @@ positions of FORM in its buffer.
...
+         (let ((cb (or callback (cider-interactive-eval-handler nil end))))
+           (when cb (funcall cb response))))
```
bbatsov: Can this really be `nil`?

bbatsov:
I still don't see any support for the REPL buffer.

cider-interaction.el
zarkone: you mean, this `(let cb .... (when cb` part?

expez: Can `cb` ever be `nil`, or is the `when` redundant?

bbatsov: Seems to me it will never be `nil`.

bbatsov:
And one more thing - it'd be good to add some configuration to enable/disable this. Some users might find the spinners annoying, so let's give them the option to turn them off.

zarkone:
agreed. I also thought about spinner customization, but not sure is it really necessary

zarkone:
there is also option to run the spinner after e.g. one or two sec of evaluation

bbatsov:
> there is also option to run the spinner after e.g. one or two sec of evaluation

Yep, that'd be nice to have as well.

cider-interaction.el
Malabarba: I believe `cb` is never nil.

cider-interaction.el
Malabarba: This line should be: `(member "done" (nrepl-dict-get response "status"))`

cider-interaction.el
```diff
@@ -1714,15 +1714,25 @@ positions of FORM in its buffer.
...
       (cider--prep-interactive-eval form)
+      (spinner-start 'progress-bar)
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (lambda (response)
+         ;; buffer still exists and
+         ;; we've got status "done" from nrepl
+         ;; stop the spinner
+         (when (and (get-buffer (buffer-name eval-buffer))
```
Malabarba: This `(get-buffer (buffer-name eval-buffer))` should be `(buffer-live-p eval-buffer)`.

Malabarba:
> > there is also option to run the spinner after e.g. one or two sec of evaluation

> Yep, that'd be nice to have as well.

That should probably be added directly to spinner.el. I'll see about doing that today.

bbatsov:
> That should probably be added directly to spinner.el. I'll see about doing that today.

I was about to file a ticket for this. :-)

One more thing - we should definitely use the spinner when booting up CIDER as well.

zarkone:
thanks @Malabarba @bbatsov @expez for corrections.
I'll investigate REPL buffer support today and ship it with enable\disable spinner feature today. 
Also, waiting for @Malabarba to use native "before timeout" feature in spinner.

Malabarba:
I've pushed spinner 1.4, should be on Gelpa tomorrow.

zarkone:
@Malabarba proud to be first `delay` option user :)

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-progress-bar-face 'progress-bar
```
Malabarba: `face` has a very specific meaning in Emacs. This needs to be something else, like `type` or `appearance`.

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
...
+  "Apperance of evaluation progress bar. 
+
+List of the all available faces provided by SPINNER-TYPES variable."
```
Malabarba: I think this would be better as: "Value is a symbol. The possible values are the symbols in the `spinner-types' variable."
And the blank line above is unnecessary.

cider-interaction.el
zarkone: agreed, will replace with `-type`

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
...
+  :type 'symbol
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-progress-bar-show t
+  "When true, show evalualtion progress bar in mode line."
```
Malabarba: evaluation typo

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
...
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-progress-bar-delay 1
+  "Amount of time, after which evaluation progress bar will begin to spin."
```
Malabarba: Amount of time, in seconds,

cider-repl.el
```diff
@@ -561,6 +561,19 @@ the symbol."
                                (lambda (buffer)
                                  (cider-repl-emit-prompt buffer))))
 
+(defun eval-progress-bar-handler (eval-buffer original-callback)
```
Malabarba: This function should be in cider-interaction (because it is used by both files and cider-repl requires cider-interaction).
Also, the name of the function is lacking a prefix.

cider-repl.el
```diff
@@ -561,6 +561,19 @@ the symbol."
...
+  "Make a function which wraps nREPL evaluation handler with
```
Malabarba: Use `Return` instead of `Make`, and make sure that he first line is a full sentence (without going beyond the fill-width), and try to mention all arguments in the docstring.
For instance:

    Return a response handler that stops the spinner and calls ORIGINAL-CALLBACK.
    EVAL-BUFFER is the buffer where the spinner was started.

Malabarba:
Ok. Once the new barrage of comments has been addressed, this looks fine by me.

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-progress-bar-type 'progress-bar
+  "Apperance of evaluation progress bar.
```
bbatsov: appearance

I find the overall description of this defcustom a bit confusing.

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-progress-bar-type 'progress-bar
+  "Apperance of evaluation progress bar. 
+
+List of the all available faces provided by SPINNER-TYPES variable."
```
bbatsov: I'm guessing this sentence should say that all the possible values are in `spinner-types`. You don't use upcase for variable names, this is only for parameter names.

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
...
+  :type 'symbol
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-progress-bar-show t
```
bbatsov: Why not `cider-show-eval-progress-spinner`?

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-progress-bar-type 'progress-bar
```
bbatsov: progress-spinner or just spinner

cider-interaction.el
```diff
@@ -213,6 +213,26 @@ namespace-qualified function of zero arity."
...
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-progress-bar-show t
+  "When true, show evalualtion progress bar in mode line."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-progress-bar-delay 1
```
bbatsov: Again - progress-spinner or just spinner

cider-interaction.el
bbatsov: Same here - it's not really a progress bar, it's a spinner.

zarkone:
that's it. hope i didn't miss anything

cider-interaction.el
```diff
@@ -1703,6 +1724,19 @@ Clears any compilation highlights and kills the error window."
 (defvar-local cider-interactive-eval-override nil
   "Function to call instead of `cider-interactive-eval'.")
 
+(defun cider-eval-spinner-handler (eval-buffer original-callback)
+  "Return a response handler that stops the spinner and calls ORIGINAL-CALLBACK.
+    EVAL-BUFFER is the buffer where the spinner was started."
```
bbatsov: This should start from the beginning of the line.

cider-interaction.el
```diff
@@ -213,6 +213,27 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-eval-spinner-type 'progress-bar
+  "Appearance of evaluation spinner.
```
bbatsov: of the

cider-interaction.el
```diff
@@ -213,6 +213,27 @@ namespace-qualified function of zero arity."
...
+  "Appearance of evaluation spinner. 
+
+Value is a symbol. The possible values are the symbols in the
+`spinner-types' variable."
+  :type 'symbol
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-show-eval-spinner t
+  "When true, show evaluation spinner in mode line."
```
bbatsov: an/the evaluation

bbatsov: in the mode line

cider-interaction.el
```diff
@@ -213,6 +213,27 @@ namespace-qualified function of zero arity."
...
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-eval-spinner-delay 1
+  "Amount of time, in seconds, after which evaluation spinner will be shown."
```
bbatsov: the evaluation spinner

bbatsov:
Apart from my small remarks the code looks OK now. Just rebase on top of the current `master`, squash all commits into one and set the final commit message to `[Fix 1197] Display some indication that we're waiting for a result for long-running evaluations`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1250](https://github.com/clojure-emacs/cider/pull/1250): Display some indication that we're waiting for a result for long-running evaluations.
```
bbatsov: Here you should refer to #1197, not the actual PR.

cider-interaction.el
```diff
@@ -213,6 +213,27 @@ namespace-qualified function of zero arity."
...
+(defcustom cider-eval-spinner-type 'progress-bar
+  "Appearance of evaluation spinner. 
+
+Value is a symbol. The possible values are the symbols in the
+`spinner-types' variable."
+  :type 'symbol
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
+(defcustom cider-show-eval-spinner t
```
expez: This should probably be `cider-eval-show-spinner` since the other related customs start with `cider-eval-`

zarkone: actually yes, previous version was started with `cider-eval`, but @bbatsov suggested this variant -- i think it has the reason, this name looks more understandable

zarkone: when all vars of one feature begins with same prefix, it's convinient when investigating it with autocompletion

bbatsov: Yeah, I prefer the current name.

zarkone:
done!

bbatsov:
There's something odd here - I see 3 commits instead of one. Maybe you should try rebasing one more time?

zarkone:
yeah, i've merged with current `original`, and hashes changed after moving commits. I've made second rebase after squashing to `original` instead of merging, seems like this solved the problem

bbatsov:
:+1: Thanks!

cider-interaction.el
```diff
@@ -1745,9 +1779,16 @@ arguments and only proceed with evaluation if it returns nil."
                  (functionp cider-interactive-eval-override)
                  (funcall cider-interactive-eval-override form callback bounds))
       (cider--prep-interactive-eval form)
+      (when cider-show-eval-spinner
+        (spinner-start cider-eval-spinner-type nil
+                       cider-eval-spinner-delay))
       (nrepl-request:eval
        form
-       (or callback (cider-interactive-eval-handler nil end))
+       (if cider-eval-progress-bar-show
```
Malabarba: This name is wrong, and it's causing #1258

zarkone: sorry for my chronical inattention, and thanks @expez for fixing my mistake.

PR #1260, opened 2015-08-16
[Fix #1239] Add cider-refresh-show-log-buffer

### LINKED ISSUE #1239: Add a flag to show reloading info only in case of problems
One user complained on gitter that he finds the window useless most of the time. One idea would be to add a flag to show it only when there are some issues. I'm open to other suggestions. At any rate - the default will continue to be the current behaviour.

cichli:
Been using this for the past couple of days - definitely preferable to have it `nil` by default. Since the error buffer pops up when something goes wrong, the refresh log is only really useful to show when debugging using the output of the before/after fns.

Maybe a `cider-visit-refresh-log-buffer` command would be useful for quickly popping up the refresh log on demand - or add it as an option to the selector?

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -179,6 +179,20 @@ if the candidate is not namespace-qualified."
 
 (defconst cider-refresh-log-buffer "*cider-refresh-log*")
 
+(defcustom cider-refresh-show-log-buffer nil
+  "Controls when to display the refresh log buffer.
+
+If non-nil, the log buffer will be displayed every time `cider-refresh' is
+called.
+
+If nil, the log buffer will still be written to, but will never be
+displayed automatically.  Instead, the most relevant information will be
+displayed in the echo area."
+  :type '(choice (const :tag "always" t)
+                 (const :tag "never" nil))
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
+
 (defcustom cider-refresh-before-fn nil
   "Clojure function for `cider-refresh' to call before reloading.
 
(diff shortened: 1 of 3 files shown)


cider-interaction.el
```diff
@@ -2110,8 +2124,14 @@ opposite of what that option dictates."
...
-    (cl-flet ((log (message &optional face)
-                   (cider-emit-into-popup-buffer log-buffer message face)))
+    (cl-flet* ((log (message &optional face)
+                    (cider-emit-into-popup-buffer log-buffer message face))
+
+               (log-echo (message &optional face)
+                         (log message face)
+                         (unless cider-refresh-show-log-buffer
+                           (let ((message-truncate-lines t))
+                             (message "cider-refresh: %s" (s-trim message))))))
```
cichli: I don't know why `cl-flet*` indents so strangely. I'd almost prefer to use plain `let*` and `lambda`.

cider-interaction.el
```diff
@@ -2110,8 +2124,14 @@ opposite of what that option dictates."
...
   (nrepl-dbind-response response (out err reloading status error error-ns after before)
-    (cl-flet ((log (message &optional face)
-                   (cider-emit-into-popup-buffer log-buffer message face)))
+    (cl-flet* ((log (message &optional face)
+                    (cider-emit-into-popup-buffer log-buffer message face))
+
+               (log-echo (message &optional face)
+                         (log message face)
+                         (unless cider-refresh-show-log-buffer
+                           (let ((message-truncate-lines t))
```
expez: Didn't know about `message-truncate-lines`. Elegant solution!

cider-interaction.el
bbatsov: There's also `cl-labels`.

bbatsov:
> Since the error buffer pops up when something goes wrong, the refresh log is only really useful to show when debugging using the output of the before/after fns.

Guess you can mention this in the troubleshooting section. :-)

cichli:
I've added a note under the reloading section - the troubleshooting section seems to be more for diagnosing problems with CIDER itself (to clarify - I meant the output of the before/after fns is useful for debugging your own program :-)).

bbatsov:
> I've added a note under the reloading section - the troubleshooting section seems to be more for diagnosing problems with CIDER itself (to clarify - I meant the output of the before/after fns is useful for debugging your own program :-)).

Good point. :+1:

PR #1261, opened 2015-08-17
[Fix #1252] Fix breakage in cider-repl-clear-buffer

### LINKED ISSUE #1252: cider-repl-clear-buffer stops working
My ```projects.clj``` looks like this:

```
{:user {
        :dependencies [[com.cemerick/piggieback "0.2.1"]
                       [org.clojure/tools.nrepl "0.2.10"]]
        :plugins [[cider/cider-nrepl "0.10.0-SNAPSHOT"]]
        }}

```
I'm pretty new to clojure and cider but when I have the code in one window and ```cider-repl``` running in the other, it is nice to be able to clear the repl buffer with ```C-c M-o``` but  ```C-c M-o``` stops working after I have executed it once and the buffer has been cleared.

If I select ```cider-repl-clear-buffer``` from ```M-x``` I can see the error message:
> Wrong type argument: integer-or-marker-p nil

The following stack trace in the ```*Messages*``` buffer

>Loading /Users/paulcowan/projects/scratch/src/scratch/core.clj...
=> nil
Loading /Users/paulcowan/projects/scratch/src/scratch/core.clj...
=> nil
Quit
call-interactively: Wrong type argument: integer-or-marker-p, nil
user-error: Minibuffer window is not active
call-interactively: Wrong type argument: integer-or-marker-p, nil
Auto-saving...done
call-interactively: Wrong type argument: integer-or-marker-p, nilInvalid face attribute :foreground nil
Invalid face attribute :foreground nil [117 times]

So all I do is

1. ```cider-jack-in```
2. Write some code and ```C-c C-k``` to execute it.
3. ```C-c M-o``` and the buffer clears
4. Write some code and ```C-c C-k``` to execute it.
5. hit ```C-c M-o``` again and ```cider-repl-clear-buffer``` no longer works.

If nobody else is experiencing this error, I wonder if my workflow of using ```C-c C-k``` is normal?

expez:
After running cider-repl-clear-buffer and then loading a file creating
output (e.g. because it contains println statements) the output would
appear in the wrong place, causing the prompt to disappear.

The problem was in the function responsible for putting POINT in the right
place before emitting evaluation results into the repl buffer.  When the
prompt was the first line in the buffer,
cider-repl--end-of-line-before-input-start didn't move point at all.

cider-repl--end-of-line-before-input-start is now changed to move backward
past the cider-repl-prompt field in order to find the appropriate output
position.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -686,12 +686,8 @@ text property `cider-old-input'."
 
 (defun cider-repl--end-of-line-before-input-start ()
   "Return the position of the end of the line preceding the beginning of input."
-  (save-excursion
-    (goto-char cider-repl-input-start-mark)
-    ;; do not use `previous-line', but `line-move' with noerror
-    ;; moving up from the first line should not throw error
-    (line-move -1 t nil nil)
-    (line-end-position)))
+  (1- (previous-single-property-change cider-repl-input-start-mark 'field nil
+                                       (1+ (point-min)))))
 
 (defun cider-repl-clear-output ()
   "Delete the output inserted since the last input."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 * Renamed `cider-repl-output-face` to `cider-repl-stdout-face` and `cider-repl-err-output-face` to `cider-repl-stderr-face`.
 
 ### Bugs fixed
-
+* [#1252](https://github.com/clojure-emacs/cider/issues/1252) `cider-repl-clear-buffer` stops working in certain circumstances.
 * [#1164](https://github.com/clojure-emacs/cider/pull/1164): Fix an error in `cider-browse-ns--doc-at-point`.
 * [#1189](https://github.com/clojure-emacs/cider/issues/1189): Don't show result from automatic ns form evaluation.
 * [#1079](https://github.com/clojure-emacs/cider/issues/1079): Don't try to font-lock very long results. The maximum font-lockable result length is controlled by `cider-font-lock-max-length`.


cider-repl.el
```diff
@@ -688,10 +688,11 @@ text property `cider-old-input'."
   "Return the position of the end of the line preceding the beginning of input."
   (save-excursion
     (goto-char cider-repl-input-start-mark)
-    ;; do not use `previous-line', but `line-move' with noerror
-    ;; moving up from the first line should not throw error
-    (line-move -1 t nil nil)
-    (line-end-position)))
+    (condition-case e
```
bbatsov: I'm a bit confused by this - if we know the exact nature of the problem, why use `condition-case` here?

expez: I've pushed a new version with a regular `if`.

Malabarba:
Couldn't we replace all of that with a single  backward-char?

Malabarba:
Actually, can't the entire function be replaced with  `(1- cider-repl-input-start-mark)`?

expez:
> Actually, can't the entire function be replaced with `(1- cider-repl-input-start-mark)`?

No, input start is after the prompt.  One char less would be inside the prompt itself.  When I tried this the output appeared after the prompt.

Malabarba:
Yes sorry, I misinterpreted the function.
Then maybe following would do?

    (save-excursion
      (goto-char cider-repl-input-start-mark)
      (line-end-position 0))

`(line-end-position 0)` returns the end of the previous line, and it returns `(point-min)` if alterady at the first line.

Anyway, the current logic isn't bad, it just caught my attention because one of the `if` branches ends in a `goto-char` even though we're inside a `save-excursion`. :)

By the way, all of this logic will break if the user specifies a custom prompt function that returns a multiline string. The really foolproof thing to do would be to move backwards through the `field cider-repl-prompt` text property.
But I'm not saying it's worth the trouble. Just pointing it out in case someone runs into this problem futurely.

expez:
New version pushed, which is both shorter and has (unverified) support for multi-line prompts.

Malabarba:
Looks good. :+1:

bbatsov:
Teamwork FTW! :+1:

PR #1269, opened 2015-08-20
Fix vagrant provision failure

vigneshsarma:
A simple `vagrant up` fails. The default image seems to not
have `apt-add-repository`.

Fixed by installing `python-software-properties` before adding ppa.

This only affects local development, not travisci.

### DIFF
--- a/vagrant/provision.sh
+++ b/vagrant/provision.sh
@@ -29,7 +29,7 @@ CASK_DIR=/opt/cask-$CASK_VERSION
 CASK_ARCHIVE=https://github.com/cask/cask/archive/v$CASK_VERSION.tar.gz
 if ! [ -d "$CASK_DIR" -a -x "/$CASK_DIR/bin/cask" ]; then
     sudo rm -rf "$CASK_DIR"
-    wget -O - $CASK_ARCHIVE | sudo tar xz -C /opt
+    wget -q -O - $CASK_ARCHIVE | sudo tar xz -C /opt
     # Bring Cask into $PATH
     sudo ln -fs "$CASK_DIR/bin/cask" /usr/local/bin
 fi
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,8 +2,7 @@
 # vi: set ft=ruby :
 
 Vagrant::Config.run do |config|
-  config.vm.box = "precise64"
-  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
+  config.vm.box = "ubuntu/trusty64"
 
   config.vm.provision :shell, :path => "vagrant/provision.sh"
 end


bbatsov:
Lately I've been wondering if anyone actually used this with vagrant and I was planning to just rename it to `travis.sh`. :-)

vigneshsarma:
I was trying to set it up in local, but decided vagrant is easier. So good thing you didn't remove it!!!

Have fixed the spelling issues as well.

vigneshsarma:
By the way, `make` is also missing in the spun up vagrant vm. Can I add that also in this pull request?

vagrant/provision.sh
```diff
@@ -15,6 +15,9 @@ apt () {
 # Silence debconf
 export DEBIAN_FRONTEND='noninteractive'
 
+apt_update
+apt software-properties-common python-software-properties
```
bbatsov: Shouldn't this go under the existing `apt_update` - this will save us one `apt_update` invocation.

bbatsov:
Fine by me.

bbatsov:
We should probably use the latest Ubuntu version as well. Not sure if this is not the actual cause for the difference with travisci.

vagrant/provision.sh
vigneshsarma: Can't do that since `python-software-properties` give us the `ppa` command. Let me try upgrading to newer version of ubuntu, maybe that will solve the issues.

vigneshsarma:
Couple of things to note here. 
1. Travis is using 12.04 still, [Virtualization Environments](http://docs.travis-ci.com/user/ci-environment/#Virtualization-environments)
2. The `ubuntu/trusty64` seems to have these other requirements, so not adding those.

vigneshsarma:
Some thing else I noticed, https://github.com/clojure-emacs/cider/commit/a1f0171b4cc1269d823a3a8efc049eb9c6d937a5. 
There a new official `emacs-snapshot` ppa https://launchpad.net/~ubuntu-elisp/+archive/ubuntu/ppa we can use.

Reference https://github.com/abo-abo/lispy/blob/master/.travis.yml

Malabarba:
There's also `evm`. It's what I've always used for travis and it lets you test on any specific version, no ppa's necessary

Malabarba:
e.g., see https://github.com/Malabarba/names/blob/master/.travis.yml

Malabarba:
Although, apparently, evm doesn't do the snapshot :P

bbatsov:
I'm fine with using evm. I don't think running that lack of snapshot support is a showstopper, as I run the snapshot locally anyways. :-)

vigneshsarma:
From the example it looks like `evm` is used to test `24.1`,`24.2`, `24.3`, while you( @Malabarba ) are still using `cassou/emacs` and `ubuntu-elisp/ppa` for `24.5` and `snapshot` versions respectively. I can replicate this setup for cider. 
- Is that ok or get `24.5` from `evm` as well?
- Do you want this as a separate pull?
- How about the current changes?

Malabarba:
> Is that ok or get 24.5 from evm as well?

Ok And preferable. 

> Do you want this as a separate pull?

 Probably better 

> How about the current changes?

 I know nothing about vagrant, so I'll leave this for @bbatsov

vigneshsarma:
@Malabarba Ok so, I will open a separate PR, which uses `evm` for all versions except `snapshot` which we will pull from `ubuntu-elisp/ppa`. I will also make sure it works with `vagrant` as well as `travis`.

@bbatsov When you consider the current changes do check https://github.com/clojure-emacs/cider/pull/1269#issuecomment-133331952

bbatsov:
There's little point in using the same Ubuntu version as TravisCI - CIDER's pretty self-contained anyways. 

You can drop the first commit and capitalise the commit message for the other two commits.

PR #1271, opened 2015-08-22
New possible value for `cider-prompt-save-file-on-load'

Malabarba:


README.md
```diff
@@ -429,6 +429,9 @@ Buffer name will look like *cider-repl project-name:port*.
   the buffer being loaded, if it's modified:
 
 ```el
+;; Don't prompt nor save
+(setq cider-prompt-save-file-on-load nil)
```
bbatsov: nil -> noconfirm

bbatsov:
Not sure about the name of the config value (sounds a bit confusing to me), but that's definitely a good idea.

Malabarba:
What other value would be better? `save-without-confirming`? `always-save`?

bbatsov:
`always-save` sounds better to me. Others are welcome to voice their opinions as well.

PR #1272, opened 2015-08-24
Fix error in cider--check-required-nrepl-version

abo-abo:
* cider-interaction.el (cider--check-required-nrepl-version):
  `cider-repl-emit-interactive-stderr' requires a string, while
  `cider--insert-readme-button' returns an overlay. Remove
  `cider--insert-readme-button' to fix the error and bring back the old style.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -337,10 +337,8 @@ Signal an error if it is not supported."
     (if nrepl-version
         (when (version< nrepl-version cider-required-nrepl-version)
           (cider-repl-emit-interactive-stderr
-           (cider--insert-readme-button
-            (format "WARNING: CIDER requires nREPL %s (or newer) to work properly"
-                    cider-required-nrepl-version)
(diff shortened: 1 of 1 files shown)


bbatsov:
@Malabarba I think we should fix the actual problem, rather than removing this.

Malabarba:
Yes, I agree it's worth fixing. Possibly returning a propertized string instead of an overlay. 

I thought this worked when we first merged it. did anything change?

bbatsov:
No idea. I've never seen the warning myself. :-)

On Monday, August 24, 2015, Artur Malabarba <notifications@github.com>
wrote:

> Yes, I agree it's worth fixing. Possibly returning a propertized string
> instead of an overlay.
>
> I thought this worked when we first merged it. did anything change?
>
> —
> <https://github.com/clojure-emacs/cider/pull/1272#issuecomment-134295984>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

Issue #1266, opened 2015-08-19
C-c M-p doesn't work with dual repls


expez:
`Could not determine relevant nREPL connection, using: CLJS refactor-nrepl@localhost:51731 (Java 1.8.0_31, Clojure 1.7.0, nREPL 0.2.10)`

This was when called from a clj repl running https://github.com/clojure-emacs/cider/commit/f0256324fe74bf2dda04feec5b8a4ef3515726df

bbatsov:
Is this still a problem?

Issue #1263, opened 2015-08-18
Middleware based functionality not working after cider-jack-in-clojurescript


expez:
Working on adding cljs for refactor-nrepl I did `cider-jack-in-clojurescript` on the refactor-nrepl project.  This project is really a clj project, but there will be some cljs files in the project moving forward, for testing purposes.

None of the middleware based functionality is working for me now.  I can't access docs, I can't jump to vars and there's no completion candidates available.

```clj
:dev {:plugins [[jonase/eastwood "0.2.0"]]
      :dependencies [[org.clojure/clojure "1.7.0"]
                     [org.clojure/clojurescript "1.7.28"]
                     [com.cemerick/piggieback "0.2.1"]
                     [commons-io/commons-io "2.4"]]
      :repl-options {:nrepl-middleware [cemerick.piggieback/wrap-cljs-repl]}
```

bbatsov:
I'm guessing this is some dispatch issue similar to #1262, but this has to be verified.

bbatsov:
Btw, is this still an issue?

PR #1274, opened 2015-08-24
Simplify connection tracking code

Malabarba:
Might affect #1266 and #1263.

`cider-find-connection-buffer-for-project-directory` embraces
ambiguities instead of avoiding them. By returning the first of multiple
matching buffers, it plays nice with the sibling connections.
It also returns buffers with nil directories if nothing else matches,
but this is just to try to fail as little as possible, it has nothing to
do with siblings.

`cider-switch-to-relevant-repl-buffer` just calls
`cider-current-repl-buffer` directly. This allows it to work in sibling
connections without complaining.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -507,28 +507,28 @@ that of the namespace in the Clojure source buffer."
 (defun cider-find-connection-buffer-for-project-directory (project-directory)
   "Find the relevant connection-buffer for the given PROJECT-DIRECTORY.
 
-A check is made to ensure that all connection buffers have a project-directory
-otherwise there is ambiguity as to which connection buffer should be selected.
+If there are multiple connection buffers matching PROJECT-DIRECTORY the
+most recent one is returned.
+If no buffers match PROJECT-DIRECTORY but there are buffers with no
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm all for simplification of the current algorithm, but I find the notion of the "most recent connection". Looking at the code "most recent" is the connection that was prepended last to `cider-connections`, but some people might have trouble understanding this.

Malabarba:
We could change the doc to say "an arbitrary one is returned" instead of "most recent". Just so people don't rely on this functionality.

bbatsov:
My main concern is that from time to time someone would like to bind such an edge case to a particular project and they'll have trouble figuring this out. Right now using "make default connection" would make something the "most recent connection", but that's obviously not super clear.

Malabarba:
We could also extend the docstring with more explanation of what "recent" means and linking to `make-default-connection`.

bbatsov:
Probably that'd be the best short-term solution.

Malabarba:
Updated.

bbatsov:
:+1:

PR #1275, opened 2015-08-24
[Fix #1267] Make results overlay compatible with synchronous evaluation

### LINKED ISSUE #1267: Overlay is incompatible with synchronous evaluation
See #1257 for details.

Malabarba:
If already inside a command loop, only delete the overlay after the next command.


bbatsov:
:+1: Look good!

PR #1276, opened 2015-08-25
Fix many compile warnings

Malabarba:
Most of the remaining warnings would be fixed if we cleaned up
cider-interaction a little bit.
In the very least, there are many repl-related functions in there which
should cleanly fit into cider-repl.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -26,6 +26,7 @@
 ;;; Code:
 
 (require 'nrepl-client)
+(require 'cider-util)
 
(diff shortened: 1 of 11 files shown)


bbatsov:
> In the very least, there are many repl-related functions in there which
> should cleanly fit into cider-repl.

Definitely. Also, by reshuffling the code a bit we should be able to reduce the need for the warning suppressing variable/function declarations here & there.

PR #1284, opened 2015-08-28
Run tests against multiple versions of emacs

vigneshsarma:
This will allow us the run tests against three different versions of emacs. 
These changes were suggested during #1269

There seems to be failures related to `24.1`, `24.2`.
```Dependency clojure-mode failed to install: Package `emacs-24.3' is unavailable```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
+sudo: true
 language: emacs-lisp
 env:
-  - EMACS=emacs24
-#  - EMACS=emacs-snapshot
+  - EMACS_BINARY=emacs-24.3-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_BINARY=emacs-24.4-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_BINARY=emacs-24.5-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_BINARY=emacs-snapshot
+
 before_script:
   - sh vagrant/provision.sh
-  - make EMACS=${EMACS} elpa
+  - make elpa
 script:
-  - make EMACS=${EMACS} test
+  - make test
--- a/vagrant/provision.sh
+++ b/vagrant/provision.sh
@@ -12,16 +12,35 @@ apt () {
     sudo apt-get install -yy "$@"
 }
 
+install_emacs() {
+    # no EMACS version is specified use snapshot.
+    if [ -z $EMACS_BINARY ];then
+        export EMACS_BINARY=emacs-snapshot
+    fi
+    echo $EMACS_BINARY
+
+    if [ $EMACS_BINARY = "emacs-snapshot" ]; then
+        ppa ppa:ubuntu-elisp/ppa
+        apt_update
+        apt emacs-snapshot emacs-snapshot-el
+    else
+        apt_update
+        apt git
+        # evm install
+        sudo mkdir -p /usr/local/evm
+        sudo chown $USER: /usr/local/evm
+        curl -fsSkL https://raw.github.com/rejeep/evm/master/go | bash
+        export PATH=$HOME/.evm/bin:$PATH
+        evm install $EMACS_BINARY
+        evm use $EMACS_BINARY
+    fi
+}
+
 # Silence debconf
 export DEBIAN_FRONTEND='noninteractive'
 
-# Bring in the necessary PPAs
-ppa ppa:cassou/emacs
-apt_update
-
-# Install Emacs 24.x and Emacs snapshot
-apt emacs24 emacs24-el emacs24-common-non-dfsg \
-    emacs-snapshot emacs-snapshot-el
+install_emacs
+emacs --version
 
 # Install Cask for Emacs dependency management
 CASK_VERSION=0.7.2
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: .travis.yml [hunk 1]
```diff
@@ -1,9 +1,19 @@
...
-#  - EMACS=emacs-snapshot
+  - EMACS_VER=emacs-24.3-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_VER=emacs-24.5-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_VER=emacs-snapshot
+
+matrix:
+  allow_failures:
+    # failure due to clojure-mode requires 24.3
+    - env:
+        - EMACS_VER=emacs-24.1-bin PATH=$HOME/.evm/bin:$PATH
```
bbatsov: I would not run any tests on 24.1 and 24.2 as we'll never support them. It's just a waste of time. We care only about 24.3 to 24.5 and `emacs-snapshot`.

Cask
```diff
@@ -6,4 +6,6 @@
 (files "*.el" (:exclude ".dir-locals.el"))
 
 (development
- (depends-on "noflet"))
+ (depends-on "dash")
```
bbatsov: CIDER already depends on the extra deps you've added. What's the point?

.travis.yml
Malabarba: Besides, cider itself requires `(emacs "24.3")`.

.travis.yml
```diff
@@ -1,9 +1,19 @@
...
+        - EMACS_VER=emacs-24.2-bin PATH=$HOME/.evm/bin:$PATH
+
 before_script:
   - sh vagrant/provision.sh
-  - make EMACS=${EMACS} elpa
+  - make elpa
```
Malabarba: Is this change really OK?
Will the `cask` job executed by `make elpa` automatically detect the `$EMACS` from the environment that `make` was invoked in?

.travis.yml
```diff
@@ -1,9 +1,19 @@
+sudo: true
 language: emacs-lisp
 env:
-  - EMACS=emacs24
-#  - EMACS=emacs-snapshot
+  - EMACS_VER=emacs-24.3-bin PATH=$HOME/.evm/bin:$PATH
```
bbatsov: EMACS_VER -> EMACS_VERSION

bbatsov: Btw, this doesn't actually refer to a version, so perhaps a better name is in order.

Malabarba: EMACS_EVM_VERSION?

.travis.yml
vigneshsarma: It is working without issues for me. https://travis-ci.org/vigneshsarma/cider/jobs/77808154#L289

Cask
vigneshsarma: Was trying to fix something else, forgot to remove it before committing.

.travis.yml
vigneshsarma: How about just `VERSION`? Since this is not just `EVM` version.

bbatsov: Or `EMACS_BINARY` (at this looks like its name).

vagrant/provision.sh
```diff
@@ -12,16 +12,35 @@ apt () {
...
+
+    if [ $EMACS_VER = "emacs-snapshot" ]; then
+        ppa ppa:ubuntu-elisp/ppa
+        apt_update
+        apt emacs-snapshot emacs-snapshot-el
+    else
+        apt_update
+        apt git
+        # evm install
+        sudo mkdir -p /usr/local/evm
```
bbatsov: Do we really need `sudo` for this?

vigneshsarma: Haven't travis, but they definitely won't work on ubuntu without `sudo`. On mac `/usr/local/` does not need `sudo`.

.travis.yml
vigneshsarma: fixed in 27b1959430c42169ece892a965cee6a56ff59490

vagrant/provision.sh
Malabarba: I believe it's necessary. Last I checked `evm` was indeed limited in the sense that it needs to use that specific dir.

bbatsov:
Guess we can squash the commits together and move forward with this.

vigneshsarma:
@bbatsov Have squashed it. Was on a vacation thats why the delay.

.travis.yml
```diff
@@ -1,9 +1,12 @@
+sudo: true
 language: emacs-lisp
 env:
-  - EMACS=emacs24
-#  - EMACS=emacs-snapshot
+  - EMACS_BINARY=emacs-24.3-bin PATH=$HOME/.evm/bin:$PATH
+  - EMACS_BINARY=emacs-24.5-bin PATH=$HOME/.evm/bin:$PATH
```
bbatsov: Did you intentionally exclude 24.4 because it's pretty similar to 24.5?

bbatsov:
Please, remove the description from the commit (which is just the messages from the other commits). It's irrelevant.

vigneshsarma:
@bbatsov I had this memory that 24.4 was never released but they directly went to 24.5 or something.... Guess I was wrong. Have fixed it.

bbatsov:
:+1: Thanks!

PR #1286, opened 2015-08-28
Partially remove sibling connections features

Malabarba:
- REPL buffers are no longer marked as siblings. Instead, "siblings" are identified by REPLs in the same project.
- The smart dispatching logic is moved from cider-current-repl-buffer to cider-find-relevant-connection.
- Killing a REPL no longer offers to kill siblings.


bbatsov:
:+1: Looks good!

PR #1290, opened 2015-08-29
Connection vs repl

Malabarba:
Streamline connection logic. Dump cider-current-repl and use cider-find-relevant-connection as much as possible.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -208,7 +208,7 @@ Refreshes EWOC."
 
 (defun cider--connections-goto-connection (_ewoc data)
   "Goto the REPL for the connection in _EWOC specified by DATA."
-  (-when-let (buffer (with-current-buffer data nrepl-repl-buffer))
+  (-when-let (buffer (with-current-buffer data nrepl-connection-buffer))
     (select-window (display-buffer buffer))))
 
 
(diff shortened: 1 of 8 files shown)


Malabarba:
Note that this goes on top of #1287

bbatsov:
Overall the changes look good, although I'd like to rename `cider-find-relevant-connection` to something preferably shorter and clearer - e.g. `cider-matching/current-connection`. It's also unclear what are the benefits of having a separate function `cider-current-repl-buffer` right now.

Malabarba:
> Overall the changes look good, although I'd like to rename cider-find-relevant-connection to something preferably shorter and clearer - e.g. cider-matching/current-connection.

Yes, I was going to suggest this next. cider-current-connection sounds fine (if you don't mind alliterations :).

> It's also unclear what are the benefits of having a separate function cider-current-repl-buffer right now.

Indeed, that's why I deleted it. :)

bbatsov:
> Yes, I was going to suggest this next. cider-current-connection sounds fine (if you don't mind alliterations :).

Sounds fine.

> Indeed, that's why I deleted it. :)

As I mentioned before - it might be a good idea to have an alias for `cider-current-connection`, which would be used in scenarios where you don't care about the fact the repl buffer is a connection buffer (e.g. switching to the repl or inserting things in the repl). IMO it would allow us to write clearer code (it might be confusing that in switch-to-repl we look for our connection buffer instead of a repl buffer). I'm not 100% sure that's something we need to do, but it makes sense in my mind.

Malabarba:
> As I mentioned before - it might be a good idea to have an alias for
> cider-current-connection, which would be used in scenarios where you
> don't care about the fact the repl buffer is a connection buffer
> (e.g. switching to the repl or inserting things in the repl). IMO it
> would allow us to write clearer code. 

That's a good point.

Another option is to go the other way and just use `repl-buffer` instead of `connection` throughout the code. This might make the code even clearer (`repl-buffer` has a very obvious meaning, while `connection` is more abstract), but it's a bad idea if we plan to support other types of connections (or if we have other reasons to *want* to make the distinction).

I can go either way. Would you prefer an alias, or rewording?

Malabarba:
I've implemented the alias for now.

bbatsov:
:+1: Looks good!

> Another option is to go the other way and just use repl-buffer instead of connection throughout the code. This might make the code even clearer (repl-buffer has a very obvious meaning, while connection is more abstract), but it's a bad idea if we plan to support other types of connections (or if we have other reasons to want to make the distinction).

Well, we were doing pretty much this until your refactoring. :-) My only problem with this approach is the dual nature of the REPL buffers, therefore my preference for the alias - sometimes a REPL is used as a REPL, sometimes we rely on the fact this very same buffer acts as a connection. Not a biggie for sure, but definitely something to keep in mind.

I've been thinking lately that'd be nice to add some glossary or something to the README to explain such foundational things.

PR #1291, opened 2015-08-30
[Fix #1288] REPL eval with pprint not working

### LINKED ISSUE #1288: Eval in repl doesn't work
After the latest refactorings this broke

expez:
I don't feel 100% confident in the changes I made here.  Is the `connection` equivalent with `cider-current-repl-buffer`?

Is `cider-current-session` the right thing, or should I be using `cider-current-tooling-session`?

Re-ordering parameters like that is a pretty gross refactoring without any test coverage...Next time perhaps rename to a new function, with a generous amount of assertions, and remove the old name.  Then once we're sure the re-ordering went OK we can reclaim the old name and remove all the type assertions.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -591,8 +591,8 @@ If NEWLINE is true then add a newline at the end of the input."
     (cider-repl--mark-input-start)
     (cider-repl--mark-output-start)
     (when cider-show-eval-spinner
-        (spinner-start cider-eval-spinner-type nil
-                       cider-eval-spinner-delay))
+      (spinner-start cider-eval-spinner-type nil
+                     cider-eval-spinner-delay))
     (if (and (not (string-match-p "\\`[ \t\r\n]*\\'" input))
              cider-repl-use-pretty-printing)
         (nrepl-request:pprint-eval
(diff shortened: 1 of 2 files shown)


bbatsov:
The changes look good.

> Re-ordering parameters like that is a pretty gross refactoring without any test coverage...Next time perhaps rename to a new function, with a generous amount of assertions, and remove the old name. Then once we're sure the re-ordering went OK we can reclaim the old name and remove all the type assertions.

I have extremely little time to work on CIDER, so I dispense with any precautions in favour of actually driving it forward. If only I had more time... guess I'd start with a month of writing tests. Any refactorings without a decent test suite are like shooting in the dark.

Malabarba:
@expez 

> I don't feel 100% confident in the changes I made here. Is the connection equivalent with cider-current-repl-buffer?

Yes. `cider-current-repl-buffer` and `cider-current-connection` are aliases now. The approach is to use `current-connection` for nrepl requests, and to use `current-repl-buffer` for this that are very specific to a REPL buffer (like `switch-to-repl`).

> Is cider-current-session the right thing, or should I be using cider-current-tooling-session?

Yes, `cider-current-session` is the right thing. The tooling session isn't used for anything anymore.

PR #1292, opened 2015-08-31
Handle state

Malabarba:
This relies on https://github.com/clojure-emacs/cider-nrepl/pull/246


Malabarba:
Written in a simpler and cleaner way now. Let me know what you think.

bbatsov:
:+1: Looks good.

PR #1301, opened 2015-09-03
Cache namespaceinfo in repl buffers

Malabarba:
Must go with https://github.com/clojure-emacs/cider-nrepl/pull/249

- cider-resolve.el has a set of functions for resolving clojure symbols based on what we know in Emacs (without having to query nREPL). These can be used for performance-sensitive contexts, such as indentation and font-locking.

- This PR also displays how to apply that to font-locking. It implements dynamic font-locking of macros, functions, and vars as they are loaded. Here's an example:  
   font-lock  
   I think the macro font-locking is pretty nice and should be default. Functions and variables are debatable. Some people might find them excessive so they might be better turned off by default. But I'd like to hear opinions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1301](https://github.com/clojure-emacs/cider/issues/1301): CIDER can do dynamic font-locking of defined variables, functions, and macros. This is controlled by the `cider-font-lock-dynamically` custom option.
 * [#1271](https://github.com/clojure-emacs/cider/issues/1271): New possible value (`always-save`) for `cider-prompt-save-file-on-load`.
 * [#1197](https://github.com/clojure-emacs/cider/issues/1197): Display some indication that we're waiting for a result for long-running evaluations.
 * [#1127](https://github.com/clojure-emacs/cider/issues/1127): Make it possible to associate a buffer with a connection (via `cider-assoc-buffer-with-connection`).
--- a/README.md
+++ b/README.md
@@ -466,6 +466,15 @@ font-locked as in `clojure-mode` use the following:
 (setq cider-repl-use-clojure-font-lock t)
 ```
 
+* CIDER can syntax highlight symbols that are known to be defined. By default,
+  this is done on symbols from the `clojure.core` namespace as well as macros
+  from any namespace. If you'd like CIDER to also colorize usages of functions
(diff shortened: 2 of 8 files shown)


arrdem:
:+1: awesome! General macro font locking is something I've wanted for a while.

Now if you could get local font locking to work :wink:

bbatsov:
> Now if you could get local font locking to work

I don't quite get this. Most IDEs/editors would not font-lock locals anywhere. And there's also the problem of figuring out what's a local - compliment has some limited local inference, but something general will be pretty hard to do.

Malabarba:
I can think of pretty foolproof ways for the middleware to detect which symbols are locals in each scope. 
However, there's no easy way for the client and server to communicate such complicated details, and on top of that the emacs font locking system isn't designed to handle local symbols. 

So, not likely to happen. :-)

Malabarba:
Ok, I have added a nice `defcustom` to the PR. Now the default is to font-lock all macros (with `font-lock-builtin-face`) and to font-lock anything from `clojure.core` (with the proper face depending on whether it's a macro, function, or var).

bbatsov:
Finally had a time to go over the PR and most things look. My only real concern is the code copy-pasted from `clojure-mode` and keeping it in sync going forward.

Malabarba:
I worried about that too, but I'm not sure what could be done about it. =/

bbatsov:
Well, I guess we can live this for now. Apart from my inline comments here's a few more:

* the dynamic font-locking should definitely be in the README and the CHANGELOG
* all font-locking commits should be squashed together
* same goes for commits related to resolve

bbatsov:
You'll also have to rebase this.

Malabarba:
Actually, I'm rewriting this now.
The performance was not satisfactory, so I'll have to switch to static font-lock rules, instead of deferring to a function call.

Malabarba:
OK, now I'm much more satisfied with the performance. I think it's a little simpler as well.

bbatsov:
Yeah, looks simpler to me as well. I think you only need to update the `save-match-data`'s location and we're good to go.

Malabarba:
Removed the save-match-data completely, it's no longer necessary.
Also added cljs support and fixed a couple of kinks. I think it's ready now.

bbatsov:
OK, let's release this into the wild. :-)

Issue #715, opened 2014-08-07
Allow jumping to source file,line and column for compilation errors


tbatchelli:
When a compilation error happens, the error message from the compiler contains the file, line and column of the error. This appears on the stacktrace buffer, but I couldn't find any way to jump from the stacktrace to the actual error, as the offending source does not appear in any of the frames.

The feature request would be to identify compilation errors and get the information about the file and position within the file from the error message. Then turn that information into a link to the right buffer and position within the buffer.

Here is an example:

```

  Show: Clojure Java REPL Tooling Duplicates All  (47 frames hidden)

2. Unhandled clojure.lang.Compiler$CompilerException
   Error compiling:
   /Users/tbatchelli/dev/palletops/admin-user-crate/src/pallet/crate/admin_user/admin_user.clj:249:41

                 Compiler.java: 7137  clojure.lang.Compiler/load
                          REPL:    1  user/eval18774
                 Compiler.java: 6703  clojure.lang.Compiler/eval
                 Compiler.java: 6666  clojure.lang.Compiler/eval
                      core.clj: 2927  clojure.core/eval
                      main.clj:  239  clojure.main/repl/read-eval-print/fn
                      main.clj:  239  clojure.main/repl/read-eval-print
                      main.clj:  257  clojure.main/repl/fn
                      main.clj:  257  clojure.main/repl
                   RestFn.java: 1096  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   56  clojure.tools.nrepl.middleware.interruptible-eval/evaluate/fn
                      AFn.java:  152  clojure.lang.AFn/applyToHelper
                      AFn.java:  144  clojure.lang.AFn/applyTo
                      core.clj:  624  clojure.core/apply
                      core.clj: 1862  clojure.core/with-bindings*
                   RestFn.java:  425  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   41  clojure.tools.nrepl.middleware.interruptible-eval/evaluate
        interruptible_eval.clj:  171  clojure.tools.nrepl.middleware.interruptible-eval/interruptible-eval/fn/fn
                      core.clj: 2402  clojure.core/comp/fn
        interruptible_eval.clj:  138  clojure.tools.nrepl.middleware.interruptible-eval/run-next/fn
                      AFn.java:   22  clojure.lang.AFn/run
       ThreadPoolExecutor.java: 1145  java.util.concurrent.ThreadPoolExecutor/runWorker
[...]

PR #1309, opened 2015-09-09
Add jump to error location from the error message (in stacktrace) - Addresses #715.

turbopape:
If you hit enter or click with button 2 on the errors cause (present at error-buffer) which have informatin about line/column of the causes, you are automatically taken to the file and the exact location of that cause.
This addresses #715.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -502,6 +502,26 @@ This associates text properties to enable filtering and source navigation."
             (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
         (newline)))))
 
+(defun cider-stacktrace--create-go-to-err-button (beg message)
+  "Create a button that jumps to the relevant error.
+
+Buttons span over the region from BEG to current point.
+MESSAGE is parsed to find line, col and buffer name to jump to."
+  (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
+    (let* ((line (string-to-number (match-string 3 message)))
+           (col (string-to-number (match-string 4 message)))
+           (buf-name (car (last (split-string (match-string 2 message) "\\/")))))
+      (when buf-name
+        (make-button (+ beg 3)
+                     (point)
+                     'action (lambda (button)
+                               (let ((the-buf-window (get-buffer-window buf-name)))
+                                 (if the-buf-window
+                                     (select-window the-buf-window)
+                                   (switch-to-buffer buf-name)))
+                               (goto-line line)
+                               (move-to-column col t)))))))
+
 (defun cider-stacktrace-render-cause (buffer cause num note)
   "Emit into BUFFER the CAUSE NUM, exception class, message, data, and NOTE."
   (with-current-buffer buffer
@@ -518,8 +538,10 @@ This associates text properties to enable filtering and source navigation."
             (newline))
           ;; Detail level 1: message + ex-data
           (cider-propertize-region '(detail 1)
-            (cider-stacktrace-emit-indented
-             (propertize (or message "(No message)") 'font-lock-face message-face) indent t)
+            (let ((beg (point)))
+              (cider-stacktrace-emit-indented
+               (propertize (or message "(No message)") 'font-lock-face  message-face) indent t)
+              (cider-stacktrace--create-go-to-err-button beg message))
             (newline)
             (when data
               (cider-stacktrace-emit-indented
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1309](https://github.com/clojure-emacs/cider/issues/1309): Jump to error location when clicking on message.
 * [#1271](https://github.com/clojure-emacs/cider/issues/1271): New possible value (`always-save`) for `cider-prompt-save-file-on-load`.
 * [#1197](https://github.com/clojure-emacs/cider/issues/1197): Display some indication that we're waiting for a result for long-running evaluations.
 * [#1127](https://github.com/clojure-emacs/cider/issues/1127): Make it possible to associate a buffer with a connection (via `cider-assoc-buffer-with-connection`).
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -303,7 +303,7 @@ In order to work properly, this mode must be activated by
           (nrepl-dbind-response cider--debug-mode-response (input-type)
             ;; A debug session is an ongoing eval, but it's annoying to have the
             ;; spinner spinning while you debug.
-            (spinner-stop)
+            (when spinner-current (spinner-stop))
             (setq-local tool-bar-map cider--debug-mode-tool-bar-map)
             (add-hook 'kill-buffer-hook #'cider--debug-quit nil 'local)
             (add-hook 'before-revert-hook #'cider--debug-quit nil 'local)
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -563,6 +563,7 @@ type (Clojure or ClojureScript). If TYPE is nil, it is derived from the
 file extension."
   (cider-ensure-connected)
   (cond
+   ((cider--in-connection-buffer-p) (current-buffer))
    ((= 1 (length cider-connections)) (car cider-connections))
    (t (let* ((project-directory (clojure-project-dir (cider-current-dir)))
              (repls (and project-directory
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1973,7 +1973,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
 
 (defun cider-connected-p ()
   "Return t if CIDER is currently connected, nil otherwise."
-  (cider-default-connection 'no-error))
+  (not (null cider-connections)))
 
 (defun cider-ensure-connected ()
   "Ensure there is a cider connection present, otherwise
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1979,7 +1979,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
   "Ensure there is a cider connection present, otherwise
 an error is signaled."
   (unless (cider-connected-p)
-    (error "No active nREPL connection")))
+    (error "No active nREPL connections")))
 
 (defun cider-enable-on-existing-clojure-buffers ()
   "Enable CIDER's minor mode on existing Clojure buffers.
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -2003,11 +2003,6 @@ See command `cider-mode'."
   (unless (cider-connected-p)
     (cider-disable-on-existing-clojure-buffers)))
 
-(defun cider-parent-ns (ns)
-  "Go up a level of NS.
-For example \"foo.bar.tar\" -> \"foo.bar\"."
-  (cider-string-join (butlast (split-string ns "\\.")) "."))
-
 
 ;;; Completion
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -2374,17 +2374,17 @@ With a prefix argument QUIT-ALL the command will kill all connections
 and all ancillary CIDER buffers."
   (interactive "P")
   (cider-ensure-connected)
-  (let ((connection (cider-current-connection)))
-    (when (y-or-n-p (format "Are you sure you want to quit the current CIDER connection %s? " connection))
-     (if quit-all
-         (progn
-           (dolist (connection cider-connections)
-             (cider--quit-connection connection))
-           (message "All active nREPL connections were closed"))
-       (cider--quit-connection connection))
-     ;; if there are no more connections we can kill all ancillary buffers
-     (unless (cider-connected-p)
-       (cider-close-ancillary-buffers)))))
+  (if (and quit-all (y-or-n-p "Are you sure you want to quit all CIDER connections? "))
+      (progn
+        (dolist (connection cider-connections)
+          (cider--quit-connection connection))
+        (message "All active nREPL connections were closed"))
+    (let ((connection (cider-current-connection)))
+      (when (y-or-n-p (format "Are you sure you want to quit the current CIDER connection %s? " connection))
+        (cider--quit-connection connection))))
+  ;; if there are no more connections we can kill all ancillary buffers
+  (unless (cider-connected-p)
+    (cider-close-ancillary-buffers)))
 
 (defun cider--restart-connection (conn)
   "Restart the connection CONN."
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -2374,16 +2374,17 @@ With a prefix argument QUIT-ALL the command will kill all connections
 and all ancillary CIDER buffers."
   (interactive "P")
   (cider-ensure-connected)
-  (when (y-or-n-p "Are you sure you want to quit the current CIDER connection? ")
-    (if quit-all
-        (progn
-          (dolist (connection cider-connections)
-            (cider--quit-connection connection))
-          (message "All active nREPL connections were closed"))
-      (cider--quit-connection (cider-current-connection)))
-    ;; if there are no more connections we can kill all ancillary buffers
-    (unless (cider-connected-p)
-      (cider-close-ancillary-buffers))))
+  (let ((connection (cider-current-connection)))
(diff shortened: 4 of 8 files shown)


turbopape:
@bbatsov Have you had the time to look into this yet? Do I need to change anything in order to make this PR conform to CIDER's standards ?

Malabarba:
Hi there. Sorry noone looked into it yet. Could you start by fixing the way `let` forms are formatted? 
Also make sure you select that bit of code and hit TAB. There  are a couple of places where indentation seems a bit off.

turbopape:
Thank you for your response. I've fixed the indentation and let forms formatting as per your recommendations !

cider-stacktrace.el
```diff
@@ -518,23 +518,51 @@ This associates text properties to enable filtering and source navigation."
             (newline))
           ;; Detail level 1: message + ex-data
           (cider-propertize-region '(detail 1)
-            (cider-stacktrace-emit-indented
-             (propertize (or message "(No message)") 'font-lock-face message-face) indent t)
-            (newline)
-            (when data
-              (cider-stacktrace-emit-indented
-               (cider-font-lock-as-clojure data) indent nil)))
+                                   (let ((beg (point)))
```
expez: You have 3 nested `let`s here.  As far as I can tell you only need one.

cider-stacktrace.el
```diff
@@ -518,23 +518,51 @@ This associates text properties to enable filtering and source navigation."
...
+                                     (cider-stacktrace-emit-indented
+                                      (propertize (or message "(No message)") 'font-lock-face  message-face) indent t)
+                                     
+                                     (let* ((toks (split-string message ":")))
```
expez: This function does too much now.  I think a natural split point is to create something like a `create-go-to-err-action` function which creates the action used in the button.  If you think it's necessary that function might even delegate to another function which picks apart the message.

Malabarba:
>  I've fixed the indentation 

Sorry, there was problem in the definition of `cider-propertize-region` that caused your indentation to be all off now. This isn't your fault.
I've fixed the definition of `cider-propertize-region`  on master. Could you rebase to the current master, reevaluate `cider-util.el` and then reindent your whole code again?

> and let forms formatting as per your recommendations !

There's still one `let*` that looks wrong in there. Though maybe you could just merge the `let`s together like expez suggests.

cider-stacktrace.el
Malabarba: This let form and the ones below are very hard to read. Here's what you could use instead:

    (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
      (let ((line (string-to-number (match-string 3 message)))
            (col (string-to-number (match-string 2 message)))
            (filename (match-string 1 message))))
      ...)

turbopape: Ok guys,

I'll have a look into this If I manage some time in the afternoon...

Thank  you !

2015-09-11 11:04 GMT+01:00 Artur Malabarba <notifications@github.com>:

turbopape:
Hi Guys, Can you have a look and tell me what it looks like now ?

cider-stacktrace.el
```diff
@@ -518,8 +533,11 @@ This associates text properties to enable filtering and source navigation."
             (newline))
           ;; Detail level 1: message + ex-data
           (cider-propertize-region '(detail 1)
-            (cider-stacktrace-emit-indented
-             (propertize (or message "(No message)") 'font-lock-face message-face) indent t)
+            (let ((beg (point)))
+              (cider-stacktrace-emit-indented
+               (propertize (or message "(No message)") 'font-lock-face  message-face) indent t)
+              (print "hello!!")
```
expez: This probably shouldn't be here

cider-stacktrace.el
```diff
@@ -502,6 +502,21 @@ This associates text properties to enable filtering and source navigation."
             (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
         (newline)))))
 
+(defun create-go-to-err-action (beg message)
```
expez: This function has to be namespaced:  `cider-stacktrace--create-go-to-err-action`

cider-stacktrace.el
```diff
@@ -502,6 +502,21 @@ This associates text properties to enable filtering and source navigation."
...
+  (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
+    (let* ((line (string-to-number (match-string 3 message)))
+           (col (string-to-number (match-string 4 message)))
+           (buf-name (car (last (split-string (match-string 2 message) "\\/"))))
+           (the-buf-window (get-buffer-window buf-name)))
+      (make-button (+ beg 3)
```
expez: What happens if you replace `beg` with `(point)`?  I might be misreading the code but I looks like it would give the same result.

cider-stacktrace.el
```diff
@@ -502,6 +502,21 @@ This associates text properties to enable filtering and source navigation."
             (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
         (newline)))))
 
+(defun create-go-to-err-action (beg message)
+  (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
```
expez: I wonder if there's something we can do here in the event that the message doesn't match the expected format.  As it stands we'll get a crash when we try to use `nil` as an action in the button.  Perhaps this should be an `if` with an error in the other branch?

cider-stacktrace.el
turbopape: I use the let beg... stuff to store the point  before the message is
emitted. Then in the function I use make-button on the region from beg to
current (point). Is there a better way to do it?

expez: No, I think you're right.  Sorry about the noise.

turbopape:
I namespaced the go-to-err-action function.
Besides, I added an if condition so we are sure not to create a button if the buffer name is nil.
As per the first format pattern matching, nothing will happen as the when block protects all of the function body.
Tell me what you think of this commit ...

cider-stacktrace.el
```diff
@@ -502,6 +502,23 @@ This associates text properties to enable filtering and source navigation."
...
+    (let* ((line (string-to-number (match-string 3 message)))
+           (col (string-to-number (match-string 4 message)))
+           (buf-name (car (last (split-string (match-string 2 message) "\\/"))))
+           (the-buf-window (if buf-name
+                               (get-buffer-window buf-name))))
+      (if the-buf-window
+          (make-button (+ beg 3)
+                       (point)
+                       'action (lambda (button)
+                                 (if the-buf-window
```
expez: You now have two nested ifs testing for the same thing.

cider-stacktrace.el
```diff
@@ -502,6 +502,23 @@ This associates text properties to enable filtering and source navigation."
...
         (newline)))))
 
+(defun cider-stacktrace--create-go-to-err-action (beg message)
+  (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
+    (let* ((line (string-to-number (match-string 3 message)))
+           (col (string-to-number (match-string 4 message)))
+           (buf-name (car (last (split-string (match-string 2 message) "\\/"))))
+           (the-buf-window (if buf-name
+                               (get-buffer-window buf-name))))
+      (if the-buf-window
```
expez: It's customary to use `when` when the `if` clause doesn't have two branches

cider-stacktrace.el
Malabarba: Also, the `if` inside the lambda is unreliable. The window might no longer exist, or it might displaying another buffer by the time this lambda is called.
Getting the window for the buffer should happen inside the lambda.

turbopape:
I changed my single branch ifs with whens.
Regarding the two ifs, actually they are not testing for the same thing. buf-name could be not null, yet there is no window with that buffer name, so window-buf-name could be null and then we could not get there and jump to the error. These two tests are not the same.

cider-stacktrace.el
turbopape: Okay I get it. I'll change this.


2015-09-11 17:31 GMT+01:00 Artur Malabarba <notifications@github.com>:

turbopape:
I get buffer-window inside the lambda now. I test if it is found else I open the buffer with switch buffer. Is it right ?

cider-stacktrace.el
```diff
@@ -502,6 +502,22 @@ This associates text properties to enable filtering and source navigation."
             (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
         (newline)))))
 
+(defun cider-stacktrace--create-go-to-err-action (beg message)
```
bbatsov: Even though this is private, a docstring would still be useful.

cider-stacktrace.el
```diff
@@ -502,6 +502,22 @@ This associates text properties to enable filtering and source navigation."
...
+  (when (string-match "\\([^:]+\\):\\([^:]+\\):\\([^:]+\\):\\([^:]+\\)\\'" message)
+    (let* ((line (string-to-number (match-string 3 message)))
+           (col (string-to-number (match-string 4 message)))
+           (buf-name (car (last (split-string (match-string 2 message) "\\/")))))
+      (when buf-name
```
bbatsov: The indentation is incorrect here.

cider-stacktrace.el
bbatsov: I'm guessing that in the name it'd be better to replace "action" with "button".

bbatsov:
Guess you should also refer to this ticket https://github.com/clojure-emacs/cider/issues/715 in your commit message

cider-stacktrace.el
expez: The button isn't created here, though, but down on L511.

bbatsov: Exactly. It's the output of this function, so I think the name is confusing.

turbopape:
Here you are ! Done indenting according to @Malabarba new formatting.

cider-stacktrace.el
Malabarba: Is it?

turbopape: Oops! Didn't I save???
Le 13 sept. 2015 12:06, "Artur Malabarba" <notifications@github.com> a
écrit :

turbopape:
What about this commit @Malabarba ? cc @bbatsov .

Malabarba:
Looks good to me.

expez:
I tried to test this the other day but I'm not sure how to trigger this code path.  Care to enlighten me @turbopape?

turbopape:
Sure. You load any clj file with a deliberate error in it. When the
stacktrace pops up, you will be able to click on any error message
containing a line:col reference, which will take you to this location on
the corresponding window.

bbatsov:
The commits have to squashed together. The issue has to be referred in the
commit message.

On Sunday, September 13, 2015, Rafik NACCACHE <notifications@github.com>
wrote:

> Sure. You load any clj file with a deliberate error in it. When the
> stacktrace pops up, you will be able to click on any error message
> containing a line:col reference, which will take you to this location on
> the corresponding window.
> Le 13 sept. 2015 17:43, "Lars Andersen" <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> a écrit :
>
> > I tried to test this the other day but I'm not sure how to trigger this
> > code path. Care to enlighten me @turbopape <https://github.com/turbopape
> >?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/clojure-emacs/cider/pull/1309#issuecomment-139893637
> >.
> >
>
> —
> <https://github.com/clojure-emacs/cider/pull/1309#issuecomment-139893812>.
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

turbopape:
Ok. I'll close this PR, squash my commits and then open a new one.
Le 13 sept. 2015 17:48, "Bozhidar Batsov" <notifications@github.com> a
écrit :

> The commits have to squashed together. The issue has to be referred in the
> commit message.
>
> On Sunday, September 13, 2015, Rafik NACCACHE <notifications@github.com>
> wrote:
>
> > Sure. You load any clj file with a deliberate error in it. When the
> > stacktrace pops up, you will be able to click on any error message
> > containing a line:col reference, which will take you to this location on
> > the corresponding window.
> > Le 13 sept. 2015 17:43, "Lars Andersen" <notifications@github.com

expez:
@turbopape you can squash and force push to the branch to save yourself some keystrokes.

turbopape:
I squashed the commits. Can you verify @bbatsov @Malabarba please ?

expez:
@turbopape drop the trailing `.` in the commit message and use `[Fix #715]` as a prefix instead of `Related to Issue #715:`, that way it matches the conventions in the [contribution guidelines](https://github.com/clojure-emacs/cider/blob/master/CONTRIBUTING.md).

Thanks for seeing this through, we've gone quite a few rounds back and forth now!

cider-stacktrace.el
```diff
@@ -502,39 +506,61 @@ This associates text properties to enable filtering and source navigation."
...
-            (newline)
-            (let ((beg (point))
-                  (bg `(:background ,cider-stacktrace-frames-background-color)))
-              (dolist (frame stacktrace)
-                (cider-stacktrace-render-frame buffer frame))
-              (overlay-put (make-overlay beg (point)) 'font-lock-face bg)))
-          ;; Add line break between causes, even when collapsed.
-          (cider-propertize-region '(detail 0)
-            (newline)))))))
+                          (let ((indent "   ")
```
expez: Am I misreading this diff or is this `let` indented way too much to the right?

turbopape: Nope this let comes just under the upper newline.

turbopape: No this is indented right just under the "cause" in the line just before the deletions in the diff.

turbopape:
Done @expez. cc @Malabarba @bbatsov .

turbopape:
Any thoughts about this PR @expez @Malabarba @bbatsov ?

cider-stacktrace.el
expez: If you look at how the code was before, you'll see that it's not supposed to be indented that far in.  You really shouldn't be indenting code manually, emacs has pretty good facilities for taking such work off your hands.  

The reason the indentation is different, from regular function indentations where the arguments are indeed aligned like you've done here, is that `nrepl-dbind-response` is a macro which sets up a context (much like `with-open-file` in clojure).

turbopape: Actually I did not indent anything by hand ! I tabbed all the way in !
Anyway, I reloaded cider-util.el and re-tabbed. tell me if it is okay now ?
cc @malabarba @bbatsov

2015-09-14 11:53 GMT+01:00 Lars Andersen <notifications@github.com>:

turbopape:
And what do you think now @expez @bbatsov @Malabarba ?

expez:
Except for a missing space between `[Fix #715]` and the rest of the commit header this looks good to me.  I'll leave it to @bbatsov to merge this when he has a spare moment.

cider-stacktrace.el
```diff
@@ -41,6 +41,7 @@
 
 (defcustom cider-stacktrace-fill-column t
   "Fill column for error messages in stacktrace display.
+
```
Malabarba: There's no need to add this blank line. The same goes to the 3 below (lines 244, 308, and 485).

turbopape:
I eliminated the blank lines and added the space between the [FIX ISSUE] and the rest of the commit message. Any thoughts @bbatsov @Malabarba @expez ?

Malabarba:
Looks ok.

turbopape:
I updated the Changelog and squashed it into my PR's commit. But now, it shows me conflicts. I refetched from upstream and re-submitted, the conflicts are stille ther. Any idea how I can fix this @bbatsov @Malabarba @expez ?

bbatsov:
Looks ok. You'll have to rebase it against the current `master`. You should also change `Added` to `Add` in the message.

bbatsov:
This doesn't look right. I see a ton of extra commits in the PR.

turbopape:
Yeah. I am struggling with the rebasing with master...

turbopape:
I got lost trying to merge the new commits at master. Think I'm gonna close this PR, re-fork and work on a clean copy.

PR #1316, opened 2015-09-11
Remove gratuitous use of CL rest function

cddr:
The `rest` function appears to be part of the "cl" package which does not appear to be required by the cider-test package. Alternatively we could just require it but since this particular one is just an alias to an existing emacs function, I figured I'd just use `cdr`.


bbatsov:
Thanks!

PR #1319, opened 2015-09-12
Add troubleshooting section about package upgrades

expez:
There has been a surge of issues lately, from user's who's just updated
CIDER, which have been resolved after the user has manually deleted
CIDER related folders and then re-installed.  This adds a section in the
readme we can point to.  If we're lucky someone might even see this
prior to opening an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1270,6 +1270,13 @@ CIDER depends on `clojure-mode` and you should always update them together, as
 the latest CIDER version might depend on functionality present only in the latest
 `clojure-mode` version.
 
+### I upgraded CIDER using `package.el` and it broke
+
+The built-in package manager isn't perfect and sometimes it messes up.  If you
+just updated and encountered an error please go into the `.emacs.d/elpa`
+directory and delete any folders related to CIDER and then re-install before
+opening an issue.
+
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


Malabarba:
:+1:

README.md
```diff
@@ -1270,6 +1270,10 @@ CIDER depends on `clojure-mode` and you should always update them together, as
 the latest CIDER version might depend on functionality present only in the latest
 `clojure-mode` version.
 
+### I just upgraded CIDER and it broke
```
bbatsov: via `package.el`

README.md
```diff
@@ -1270,6 +1270,10 @@ CIDER depends on `clojure-mode` and you should always update them together, as
...
+
+The built-in package manager isn't perfect and sometimes it messes up.  If you just updated and encountered an error please go into the `.emacs.d/elpa` directory and delete any folders related to CIDER and then re-install before opening an issue.
```
bbatsov: This long line might be refilled. 

We probably need a similar entry about not restarting after upgrade - "I upgraded via package.el, but nothing changed" (or something like this)

expez:
Fixed, appended another commit about restarting emacs

bbatsov:
:+1:

PR #1327, opened 2015-09-16
[Fix #467] Improve completion in the repl

### LINKED ISSUE #467: Completion of namespace-qualified vars fails for (partial-completion initials)
When `partial-completion-mode` was deprecated in Emacs 24, I followed a recommendation from NEWS to get comparable behavior:

    (setq completion-styles '(partial-completion initials)
          completion-pcm-complete-word-inserts-delimiters t)

It does the job, and I have not had any problems with it so far, but it breaks completion of namespace-qualified vars in my fresh cider installation. There is no error message, only `No match`. Everything works fine without the above setup.

I am using CIDER 0.6.0snapshot (following master), Clojure 1.6.0-alpha3 (but it did not work with 1.5.1 either), nREPL 0.2.3 on Emacs 24.3.50, Ubuntu 12.04.4 LTS.

By the way, thanks for the great work!

expez:
When investigating #467 I noticed that the test checking if we're in a
string or not had a mostly false positves in the repl.

The previous solution relied on the parse state from point-min up to
point which I think is too large a context.  Furthermore, it's not
reasonable to expect the entire contents of the repl buffer, prompts and
all, to be valid lisp.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -60,6 +60,14 @@ Setting this to nil removes the fontification restriction."
       (file-name-directory buffer-file-name)
     default-directory))
 
+(defun cider-in-string-p ()
+  "Return true if point is in a string."
+  (nth 3 (parse-partial-sexp (beginning-of-defun) (point))))
+
+(defun cider-in-comment-p ()
+  "Return true if point is in a comment."
+  (nth 4 (parse-partial-sexp (beginning-of-defun) (point))))
+
 ;;; Text properties
 
 (defun cider-maybe-intern (name)
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1198,8 +1198,9 @@ The formatting is performed by `cider-annotate-completion-function'."
 
 (defun cider-complete-at-point ()
   "Complete the symbol at point."
-  (let ((sap (symbol-at-point)))
-    (when (and sap (not (nth 3 (syntax-ppss))) (cider-connected-p))
+  (-when-let (sap (cider-symbol-at-point))
+    (when (and (cider-connected-p)
+               (not (or (clojure-in-string-p) (clojure-in-comment-p))))
       (let ((bounds (bounds-of-thing-at-point 'symbol)))
         (list (car bounds) (cdr bounds)
               (completion-table-dynamic #'cider-complete)


expez:
Depends on https://github.com/clojure-emacs/clojure-mode/pull/322

Malabarba:
I don't love the idea of checking faces. Font-locking is lazy, the faces can be overridden, and the faces can be wrong on occasion (rarely, but it happens).
Would it work to run `(parse-partial-sexp FROM TO)`, where FROM is the end of the prompt, and TO is the point?

expez:
> Would it work to run (parse-partial-sexp FROM TO), where FROM is the end of the prompt, and TO is the point?

In the repl, yes.  Everywhere else, not really.  We could use `beginning-of-defun` other places, I guess.  Does this still sound better?

I've no idea if your concerns with using font-locking will ever be a problem in practice, but if they are then we should update the state predicates in clojure-mode too.  Any thoughts on this @bbatsov?

This test is going to run on each keystroke, but I think the two solutions are about equivalent in speed.

Malabarba:
If we do end up checking for faces, then I think we should compare against `font-lock-string-face` instead of `'font-lock-string-face`. They are equivalent unless the user has overriden the font-lock faces, in which case the former will work but the latter will fail.

bbatsov:
> Any thoughts on this @bbatsov?

I'm with @Malabarba on this - it's problematic. For instance - after his recent changes the docstring check function is now broken, because the docstring can have one of two faces. But comments have have them as well... And there are plenty of modes that interfere with the faces (e.g. whitespace-mode). We need to devise a better strategy for this. Maybe something using `syntax-ppss`?

Malabarba:
>  We could use beginning-of-defun other places, I guess. Does this still sound better?

I think clojure-mode buffers can just keep using `syntax-ppss`. This function is supposed to be fast and reliable in code buffers.
And then in the repl we call `parse-partial-sexp` from the end of prompt.

expez:
Let's discuss the other functions over here: clojure-emacs/clojure-mode#322

Once we resolve that I think this one should be good to merge.

bbatsov:
So, let's get back to the issue at hand - how is something in `clojure-mode` going to be helpful here? After all the REPL doesn't share much with `clojure-mode`.

expez:
> how is something in clojure-mode going to be helpful here?

If you don't want the two predicates in clojure mode I can move them over here, but I think they belong there because they are generally useful.

The only change in functionality is to disable completions in comments.  There's also a bugfix where everything was considered a string because we tried to parse the entire repl cteont as valid lisp.

bbatsov:
A few examples would be useful. I have a vague idea what you're talking about, but something more concrete would help.

expez:
Examples of something that wasn't working but is now?

bbatsov:
image

Seems to work for me.

expez:
Bah, I thought this was triggering all the time, but it's just when the repl gets into the right state:

```
(do (println "\"")
    (println "\"")
    (println " \" \" ")
    (println "\""))
```

Does it though.  After that completion won't work.

bbatsov:
Well, if that's the case then I believe we have to solve a different problem - perhaps providing some sane boundaries for `symbol-at-point`.

bbatsov:
image

Actually, no. Looking at the results it seems that string parsing in REPL is broken.

expez:
How is that going to fix it?  Completion is now contingent on the state of `parse-partial-sexp` across the entire repl.  That the entire repl buffer is being parsed, on each completion request,  should be cause enough to merge this in my eyes.  Luckily for us `parse-partial-sexp` is written in C....

bbatsov:
The real issue is that strings are not properly parsed and after some complicated code everything in the buffer is treated as string. Your patch doesn't address this, it just works around this.

bbatsov:
It seems to me we should somehow exclude the evaluation results from the parsing.

Malabarba:
I think they're separate problems and there won't be a unified way of fixing both. 
One problem is that we're font locking the whole buffer, whereas we should be doing it only for user input and return value (and as separate things). 
The other problem is that some internal cider logic is parsing the entire buffer instead of parsing just user input. 

We could solve them both in one go if there was a way to specify some sort of syntax delimiters in the buffer. So both font-lock and syntax-ppss would know where to stop and where to start. 
Since I'm not aware of any way to do that, I think this fix is valid 

But, I think the clojure mode side is unnecessary. Emacs core used to have in-string-p function and has deprecated it in lieu of syntax-ppss.

expez:
> It seems to me we should somehow exclude the evaluation results from the parsing.

This is what the patch does!  it limits scope to whatever is beyond the repl input mark.

bbatsov:
> This is what the patch does! it limits scope to whatever is beyond the repl input mark.

I meant limit the font-locking of the REPL mode itself (the same thing @Malabarba mentioned). The patch definitely doesn't do this. 

> One problem is that we're font locking the whole buffer, whereas we should be doing it only for user input and return value (and as separate things). 
> The other problem is that some internal cider logic is parsing the entire buffer instead of parsing just user input.

Can't argue with this.

> We could solve them both in one go if there was a way to specify some sort of syntax delimiters in the buffer. So both font-lock and syntax-ppss would know where to stop and where to start. 
> Since I'm not aware of any way to do that, I think this fix is valid
> 
> But, I think the clojure mode side is unnecessary. Emacs core used to have in-string-p function and has deprecated it in lieu of syntax-ppss.

I'd prefer the use of `syntax-ppss` myself. Most modern major modes rely heavily on it these days. Perhaps it's best to discard the clojure-mode stuff indeed.

expez:
> I'd prefer the use of syntax-ppss myself. Most modern major modes rely heavily on it these days. Perhaps it's best to discard the clojure-mode stuff indeed.

`syntax-ppss` is just a wrapper around `parse-partial-sexp` which is what the clojure-mode stuff uses.

bbatsov:
Let's just inline whatever's needed here and open a separate ticket about the breakage of the string font-locking. I definitely don't want to use something with `clojure-` in its name here.

expez:
> Let's just inline whatever's needed here [...]

done

cider-util.el
```diff
@@ -60,6 +60,14 @@ Setting this to nil removes the fontification restriction."
       (file-name-directory buffer-file-name)
     default-directory))
 
+(defun cider-in-string-p ()
+  "Return true if point is in a string."
+  (nth 3 (parse-partial-sexp (beginning-of-defun)(point))))
```
bbatsov: You've missed a space here.

cider-util.el
```diff
@@ -60,6 +60,14 @@ Setting this to nil removes the fontification restriction."
...
+
+(defun cider-in-comment-p ()
```
bbatsov: I think it'd be better if those were in `cider-repl.el`.

Malabarba: I think they're fine in cider-util. Though currently only the repl uses them, there's nothing really repl-specific about them.

expez:
Updated.  I agree with @Malabarba so I haven't moved those functions out of cider-util.el

bbatsov:
Fair enough.

PR #1329, opened 2015-09-16
[Fix #1320] Update cider-repl--state-handler with the new syntax

### LINKED ISSUE #1320: high nrepl/cider latency
I follow the cider `HEAD` via melpa, and update daily.

Sometime in the last week I have noticed a sudden increase in latency, both during the connection setup when I make my connection to a running nrepl server, and when using the repl interactively in emacs.

It now takes more than 15 seconds now between `nREPL: Direct connection established` and `Connected.`, during which emacs hangs. After that, every interaction with cider's repl takes a second or two to return.

Now I have switched back to 0.9.1 (need to work) and everything is snappy and quick again.

What additional information would help debug this?

Malabarba:
Relies on https://github.com/clojure-emacs/cider-nrepl/pull/255

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -180,8 +180,8 @@ via `cider-current-connection'.")
   "Handle the server STATE.
 Currently, this is only used to keep `cider-repl-type' updated."
   (with-demoted-errors "Error in `cider-repl--state-handler': %s"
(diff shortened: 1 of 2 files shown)


bbatsov:
You'll have to rebase this on top of the current `master`.

Malabarba:
Done

PR #1339, opened 2015-09-20
Refactors

Malabarba:
Solve all compilation warnings in cider-doc.el, and spread out cider-interaction a bit more.


bbatsov:
:+1:

PR #1341, opened 2015-09-20
Refactors

Malabarba:
A few more.

I'm moving to cider-util the generally useful functions that don't involve nrepl, and moving to cider-client the generally useful functions that rely on a connection.


expez:
No API changes here, right?  Just stuff being moved around?

Malabarba:
Yes. Don't merge yet though. More incoming.

Malabarba:
Alright, now I'm done. For me, this fixes all compiler warnings.

Malabarba:
Don't know why travis isn't running, thouhg. :P

expez:
Travis went on the fritz for me earlier today.  Guess no one's there to fix it on a Sunday, especially since 99% of the affected are non-paying customers :p

bbatsov:
I'll have to dwell for a while on some of the changes. My vision for the future was actually to get rid of `cider-interaction` and stuff most of the things there in `cider.el`. I think `cider-interaction` was a mistake to some extend (just made the code and the deps more complex). Fixing the warnings is great, but fixing the general code layout would be even better. And for `nrepl-client.el` - suppressing the warnings makes littles sense given the short-term goal to simply eliminate all the direct deps.

bbatsov:
This is one of the general weakness of Emacs Lisp - it's really easy to create some spaghetti monster with it. With so few of the extensions being bigger projects, there're no real conventions on how to organize them.

Malabarba:
Sure, sleep on this and let me know if you think any of the commits are not worth it.
Overall, my priority here was getting to a point where we could rely on the compiler warnings. That's a big advantage and it makes future refactorings much easier to do.

As for completely getting rid of cider-interaction, I'd say removing 700 lines from it is a pretty nice start. :-)
The rest will be even easier to spread out once we can rely on the compiler warnings.

bbatsov:
One more thought - I never intended `cider-util` to be a namespace for "common" cider functions. It was a place to put utility functions (really generic things like string extensions, backported stuff, etc). I definitely don't like so much things to be there and would prefer it if they were in a different file (e.g. cider-common/core/whatever).

expez:
> I never intended `cider-util` to be a namespace for "common" cider functions.  It was a place to put utility functions (really generic things like string extensions, backported stuff, etc). 

Your objections, when I put some predicates into util, the other day suddenly make sense :)  This does make good sense, though.

bbatsov:
OK, I finally went through all the commits. Have a look at my remarks.

Malabarba:
Alright, I'm relatively certain I've addressed all concerns. :-)

bbatsov:
You just need to rebase the branch.

Malabarba:
Rebased.

bbatsov:
Well, I guess the next step would be to add a check for compilation warnings in the build, while we still none.

PR #1350, opened 2015-09-24
Use elisp cywin functions. Refs #735

juergenhoetzel:
Use `cygwin-convert-file-name-to-windows' and
`cygwin-convert-file-name-from-windows' instead of executing external
"cygpath" commands.

I wasn't aware this functions exist, when creating the origin cygwin support.

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -127,14 +127,10 @@ otherwise, nil."
     name))
 
 (defvar cider-from-nrepl-filename-function
-  (if (eq system-type 'cygwin)
-      (lambda (resource)
-        (let ((fixed-resource (replace-regexp-in-string "^/" "" resource)))
-          (replace-regexp-in-string
-           "\n"
-           ""
-           (shell-command-to-string (format "cygpath --unix '%s'" fixed-resource)))))
-    #'identity)
+  (with-no-warnings
+    (if (eq system-type 'cygwin)
+        #'cygwin-convert-file-name-from-windows
+      #'identity))
   "Function to translate nREPL namestrings to Emacs filenames.")
 
 (defcustom cider-prefer-local-resources nil
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -932,14 +932,10 @@ evaluation command. Honor `cider-auto-jump-to-error'."
 ;;; Evaluation
 
 (defvar cider-to-nrepl-filename-function
-  (if (eq system-type 'cygwin)
-      (lambda (filename)
-        (->> (expand-file-name filename)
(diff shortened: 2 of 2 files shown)


bbatsov:
> Use cygwin-convert-file-name-to-windows' and cygwin-convert-file-name-from-windows' instead of executing external
> "cygpath" commands.
> 
> I wasn't aware this functions exist, when creating the origin cygwin support.

Are they autoloaded?

Also, change the title of the commit to "[#735] Use built-in cygwin functions" to match the project's conventions.

juergenhoetzel:
Done.

They are implemented in  native `C source code' (only available on Cygwin: ./src/cygw32.c): No need to (auto)load.

Malabarba:
They need to be declared to `"cygw32.c"`.

juergenhoetzel:
It is declared in w32-win.el which is loaded on startup on Cygwin.

Malabarba:
Yes, but that file is not loaded on other systems, which will cause a compiler warning. Adding a `(declare-function cygwin-convert-file-name-from-windows "cygw32.c")` prevents that warning.

juergenhoetzel:
@Malabarba 

I see. I wrapped it inside  *with-no-warnings* to prevent compilation warnings on non-cygwin systems.

cider-common.el
```diff
@@ -127,14 +127,9 @@ otherwise, nil."
...
 (defvar cider-from-nrepl-filename-function
-  (if (eq system-type 'cygwin)
-      (lambda (resource)
-        (let ((fixed-resource (replace-regexp-in-string "^/" "" resource)))
-          (replace-regexp-in-string
-           "\n"
-           ""
-           (shell-command-to-string (format "cygpath --unix '%s'" fixed-resource)))))
-    #'identity)
+  (with-no-warnings (if (eq system-type 'cygwin)
```
bbatsov: I'd add a line break before the `if`.

cider-interaction.el
```diff
@@ -932,14 +932,9 @@ evaluation command. Honor `cider-auto-jump-to-error'."
...
 (defvar cider-to-nrepl-filename-function
-  (if (eq system-type 'cygwin)
-      (lambda (filename)
-        (->> (expand-file-name filename)
-             (format "cygpath.exe --windows '%s'")
-             (shell-command-to-string)
-             (replace-regexp-in-string "\n" "")
-             (replace-regexp-in-string "\\\\" "/")))
-    #'identity)
+  (with-no-warnings (if (eq system-type 'cygwin)
```
bbatsov: Here as well.

juergenhoetzel:
Done.

PR #1351, opened 2015-09-25
Check declares

Malabarba:
This same test can easily be used to run checkdoc, we only need to make
CIDER comply with checkdoc first.


bbatsov:
So, this finds incorrect and redundand `declare-something`s? I've never encountered this check, therefore my question.

Malabarba:
Exactly. I only recently learned of it from oremacs.

bbatsov:
OK. Thanks!

PR #1353, opened 2015-09-28
Cask: Add spinner dependency

abo-abo:
I'm using CIDER from git. After pulling today, `make` no longer works because of missing `spinner`. This fixes it.

### DIFF
--- a/Cask
+++ b/Cask
@@ -6,4 +6,5 @@
 (files "*.el" (:exclude ".dir-locals.el"))
 
 (development
- (depends-on "noflet"))
+ (depends-on "noflet")
+ (depends-on "spinner"))


expez:
This should not be necessary.  `spinner` isn't a dev dependency so it should be installed along with the other deps like `clojure-mode`.

abo-abo:
`cask && make` should work always, regardless of semantics. It doesn't right now.

expez:
I'm not arguing semantics, I'm just pointing out that there's nothing special about `spinner`.  If you have  to add `spinner` as a dev dependency then you should also have to add `clojure-mode`, `queue`, `dash` etc. I think something else is wrong.

Malabarba:
I agree. The spinner dep should already have been added by the line `(package-file "cider.el")`. 

Something else seems to be wrong here.

abo-abo:
I guess it was just a glitch. Deleting `.cask` directory fixed it and `cask && make` works now.

Issue #1348, opened 2015-09-24
Ditch dash.el


Malabarba:
In Emacs 25.1, seq.el will be built-in (it's also a smaller package). So it would be nice to replace the dash dependency with seq.

There are currently 102 usages of dash functions that I could find by grepping for `[(']-[-a-z]`. About half of them `-when-let`s and `-if-let`s. The other half are functions provided by `seq`. The only exception I found was `-map-indexed`.

One way to replace the `-when/if-let`s is to drop 24.3 support and use `when/if-let`. Another is to copy the definition of `when/if-let` from `subr-x.el`. What do we think?

bbatsov:
:+1: Let's inline `when-let` and `if-let` and go forward with this.

rogergl:
I could try to help with this task.

bbatsov:
You're welcome to do so. :-)

Malabarba:
👏👏👏👏🎉🎊🎆

:-)

rogergl:
I had a first look at the source. There seems to be a subx-r version that contains if-let, when-let. The subx-r version my Emacs 24.5 uses,  does not. So inlining means copy them over ?

bbatsov:
Yep.

rogergl:
Should I copy them to cider-common.el ?

bbatsov:
No, copy them to `cider-util.el.`. You'll see we have other backported stuff there.

rogergl:
I think  prefixing them with cider, like the other function in util, would be not a very good idea ?

bbatsov:
Take this approach instead https://github.com/bbatsov/projectile/blob/master/projectile.el#L56

I don't want us to prefix them.

rogergl:
What about the other stuff from dash like -> ?

bbatsov:
`subr-x` has `thread-first` and `thread-last` - you can copy and alias them (or use them with their Elisp names, it's all the same to me).

rogergl:
Ok, than I'll use thread-first and thread-last. This makes it easier to see if there is still something used from dash.

rogergl:
-when-let, -if-let, -> and ->> are gone. Not sure how I should include the stuff from seq.el.

bbatsov:
> Not sure how I should include the stuff from seq.el.

Just add a dependency to `seq` (to the package itself, so it will work on older Emacsen) and apply your best judgement when swapping dash.el functions for their seq.el equivalents.

[...]

PR #1365, opened 2015-10-12
first fixes for #1348

rogergl:
Some problems I have encountered so far are:
1. I have no idea how to run the tests
2. The formatting is different
3. nrepl-client requires cider-util and cider-util requires nrepl-client

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 
 ### Changes
 
+* [#1348](https://github.com/clojure-emacs/cider/issues/1348): Drop the dash dependency.
 * The usage of the default connection has been reduced significantly. Now evaluations & related commands will be routed via the connection matching the current project automatically unless there's some ambiguity when determining the connection (like multiple or no matching connections). Simply put you'll no longer have to mess around much with connecting-setting commands (e.g. `nrepl-connection-browser`, `cider-rotate-default-connection`).
 * [#732](https://github.com/clojure-emacs/cider/issues/732): `cider-quit` and `cider-restart` now operate on the current connection only. With a prefix argument they operate on all connections.
 * `nrepl-log-messages` is now set to `t` by default.
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -35,6 +35,7 @@
 
 (require 'cider-repl)
 (require 'cider-client)
+(require 'cider-compat)
 
 (defconst cider-browse-ns-buffer "*Browse NS*")
 (defvar-local cider-browse-ns-current-ns nil)
@@ -100,11 +101,11 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
   (with-current-buffer (cider-popup-buffer cider-browse-ns-buffer t)
     (let ((vars (cider-sync-request:ns-vars namespace)))
       (cider-browse-ns--list (current-buffer)
-                            namespace
-                            (mapcar (lambda (var)
-                                      (format "%s"
-                                              (cider-browse-ns--properties var)))
-                                    vars))
+                             namespace
+                             (mapcar (lambda (var)
+                                       (format "%s"
+                                               (cider-browse-ns--properties var)))
+                                     vars))
       (setq-local cider-browse-ns-current-ns namespace))))
 
 ;;;###autoload
@@ -131,12 +132,12 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
 (defun cider-browse-ns--doc-at-point ()
   "Expand browser according to thing at current point."
   (interactive)
-  (-when-let (var (cider-browse-ns--var-at-point))
+  (when-let (var (cider-browse-ns--var-at-point))
     (cider-doc-lookup var)))
 
 (defun cider-browse-ns--find-at-point ()
   (interactive)
-  (-when-let (var (cider-browse-ns--var-at-point))
+  (when-let (var (cider-browse-ns--var-at-point))
     (cider-find-var current-prefix-arg var)))
 
 (defun cider-browse-ns--handle-mouse (event)
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -35,6 +35,7 @@
 (require 'cider-eldoc)
 (require 'cider-resolve)
 (require 'cider-doc)
+(require 'cider-compat)
 
 (defcustom cider-mode-line-show-connection t
   "If the mode-line lighter should detail the connection."
@@ -46,7 +47,7 @@
   "Return info for the `cider-mode' modeline.
 
 Info contains project name and host:port endpoint."
-  (-if-let (current-connection (ignore-errors (cider-current-connection)))
+  (if-let (current-connection (ignore-errors (cider-current-connection)))
       (with-current-buffer current-connection
         (concat
          (when cider-repl-type
@@ -305,13 +306,13 @@ Returns to the buffer in which the command was invoked."
 ;;; Dynamic indentation
 (defun cider--get-symbol-indent (symbol-name)
   "Return the indent metadata for SYMBOL-NAME in the current namespace."
-  (-when-let (indent
-              (nrepl-dict-get (cider-resolve-var (cider-current-ns) symbol-name)
-                              "indent"))
-    (let ((format (format ":indent metadata on ‘%s’ is unreadable! \nERROR: %%s"
-                          symbol-name)))
-      (with-demoted-errors format
-        (cider--deep-vector-to-list (read indent))))))
+  (when-let (indent
+             (nrepl-dict-get (cider-resolve-var (cider-current-ns) symbol-name)
+                             "indent"))
+            (let ((format (format ":indent metadata on ‘%s’ is unreadable! \nERROR: %%s"
+                                  symbol-name)))
+              (with-demoted-errors format
+                (cider--deep-vector-to-list (read indent))))))
 
 ;;; Dynamic font locking
 (defcustom cider-font-lock-dynamically '(macro core deprecated)
@@ -353,8 +354,8 @@ The value can also be t, which means to font-lock as much as possible."
 (defun cider--compile-font-lock-keywords (symbols-plist core-plist)
   "Return a list of font-lock rules for the symbols in SYMBOLS-PLIST."
   (let ((cider-font-lock-dynamically (if (eq cider-font-lock-dynamically t)
-                                    '(function var macro core deprecated)
-                                  cider-font-lock-dynamically))
+                                         '(function var macro core deprecated)
+                                       cider-font-lock-dynamically))
         deprecated
         macros functions vars instrumented)
     (when (memq 'core cider-font-lock-dynamically)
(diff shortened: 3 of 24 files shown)


bbatsov:
> I have no idea how to run the tests

make test

> The formatting is different

Well, that's normal. Changes require formatting changes.

> 3. nrepl-client requires cider-util and cider-util requires nrepl-client

Not sure how this is relevant.

cider-browse-ns.el
```diff
@@ -131,13 +132,13 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
 (defun cider-browse-ns--doc-at-point ()
   "Expand browser according to thing at current point."
   (interactive)
-  (-when-let (var (cider-browse-ns--var-at-point))
-    (cider-doc-lookup var)))
+  (when-let (var (cider-browse-ns--var-at-point))
+            (cider-doc-lookup var)))
```
bbatsov: This looks like some bug.

expez:
> Well, that's normal. Changes require formatting changes.

This is not OK imo.  It looks incredibly bad when `& body` isn't indented like usual.  Why don't these have indention specs?

cider-browse-ns.el
rogergl: What exactly ?

expez:
hmm they do have indentation specs.  @rogergl you need to load the macros and re-indent these files so the indentation spec is used.

rogergl:
@expez I'll must admit that I have no idea what indentation specs are and which macros to load. Is there some guide that describes it  ?

rogergl:
@bbatsov make test (running on Mac OS X) gives
```
GNU Emacs 24.5.1
Copyright (C) 2015 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
cask install
make: cask: Permission denied
make: *** [elpa] Error 1
```

expez:
> I'll must admit that I have no idea what indentation specs are. 

Below, after the docstring, is the indent spec. 
```el   
 (defmacro when-let (bindings &rest body)
      "Process BINDINGS and if all values are non-nil eval BODY.
Argument BINDINGS is a list of tuples whose car is a symbol to be
bound and (optionally) used in BODY, and its cadr is a sexp to be
evalled to set symbol's value.  In the special case you only want
to bind a single value, BINDINGS can just be a plain tuple."
      (declare (indent 1) (debug if-let))
      (list 'if-let bindings (macroexp-progn body))))
```

> [...]  and which macros to load

You must load macros you're substituting in, like `when-let` above.  Otherwise emacs won't know how to indent these forms.

rogergl:
The formatting is fixed. Thanks @expez

bbatsov:
@rogergl Did you install `cask`?

rogergl:
@bbatsov "cider-util should not have anything nrepl-related in it.: should I remove it ?

bbatsov:
> should I remove it ?

Yep.

Malabarba:
I still need to go through the whole diff, but I just wanted to say that we need to define these new macros on their own file, otherwise we might run into that problem of faulty installations again.

bbatsov:
We can always have `cider-compat.el` or something like this.

Malabarba:
Another thing. Even if there's just one binding, I much prefer we use `(if-let ((a b)) ...)` instead of `(if-let (a b) ...)`, for general consistency with `let` and with the many-bindings case. Dash's `-if-let` didn't allow this, but `if-let` does, so let's do it. :+1: 
Of course, the same goes for `when-let`.

Malabarba:
Also, if I haven't said it yet. Thanks for all this work @rogergl .

Malabarba:
> nrepl-client requires cider-util and cider-util requires nrepl-client

This might fix itself if we move the macros to cider-compat. Still, Bozhidar is right that anything nrepl dependent should be on cider-client, not cider-util.

rogergl:
Current to-dos: use (if-let ((a b)) ...) etc
Move the macros to cider-comapt.el.  Should I make this changes ?

Malabarba:
Yep. 👍

bbatsov:
It seems we're almost ready here. There are only some byte-compilation warnings that have to be fixed. 

It would be nice to wrap the PR as soon as possible, because it touches a lot of places in the code and any potential change right now would result in conflicts for the PR.

rogergl:
I could spend some hours tomorrow. cider-compat or cider-backports ? Which one is preferred ?

bbatsov:
Let's go with `cider-compat`.

rogergl:
Ok, the byte compilation error are fixed. I think the change to use (if-let ((a b)) ...) instead of (if-let (a b) ...) could be done in another migration ? Leaving just the following remaining changes:

cider-eldoc using -elem-index
cider-interaction using -each
cider-stacktrace using -repeat
cider-util -using map-indexed

bbatsov:
Great news! Well, I guess you can just squash the commits together when you feel you're ready. Use the project convention for commit message structuring for the final commit message.

rogergl:
Use the project convention for commit message structuring for the final commit message.
I think I need a little help with this, is there a guide that describes the required steps ?

What about 

    cider-eldoc using -elem-index
    cider-interaction using -each
    cider-stacktrace using -repeat
    cider-util -using map-indexed

Next iteration or moving them from dash to compat ?

rogergl:
Ok just found the contributing readme.

bbatsov:
> Next iteration or moving them from dash to compat ?

`-each` is basically `mapc`, so you've got that one covered. You can moved the rest to `cider-util` and rename them to `cider-elem-index`, etc.

rogergl:
Since there is already some stuff from dash in cider-compat, shouldn't they be moved there ?

bbatsov:
> Since there is already some stuff from dash in cider-compat, shouldn't they be moved there ?

Nope. `compat` is for compatibility with older versions of Emacs (it should have only things backported from Emacs itself). `util` is for general purpose functions that we use all over cider.

rogergl:
But none of the thing in emacs-compat is backported. The function there are  just copies from dash and subr-x ?

bbatsov:
If you copy from from `subr-x`, it's considered backported, as this is a built-in library. Things from `dash` should be go to util and be prefixed with `cider-`, as it's a 3rd party library.

bbatsov:
Btw, it seems to me everything in `compat` right now is from `subr-x`, so we're great.

Malabarba:
It looks like `cl-position` is a drop-in replacement for `-elem-index`.
Since we depend on emacs-24.3 we don't even have to depen on cl-lib to use it.

rogergl:
Dash is gone. Leaving 2 issues:

1. ) I had to copy the --each macro. I tried to convert cider--map-indexed to use maps but wasn't successful.

2.) I realy could need some help on how to turn all this commits into a valid pull request.

bbatsov:
> I had to copy the --each macro. I tried to convert cider--map-indexed to use mapc but wasn't successful.

We use it very rarely, so just expand it to not use the magic `it` variable. That's the only difference between a regular `-each` and `--each`. 

See http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

After you make the changes locally use `git push -f` to update your PR.

rogergl:
We use it very rarely, so just expand it to not use the magic it variable. 
That goes probally beyond the things I currently understand about Emacs macro programing. But I'll have a look at it tomorrow.

Malabarba:
In a couple of hours I can give you a standalone implementation of -map-indexed, using cl-mapcar.

rogergl:
BTW:  cider-map-index is only used by cider-intern-keys with is only used by cider-test

Malabarba:
(defun cider-map-indexed (f list)
      "Return a list of (F index item) for each item in LIST."
      (let ((i 0)
            (out))
        (dolist (it list (nreverse out))
          (push (funcall f i it) out)
          (setq i (1+ i)))))

rogergl:
I think cider-repeat can be replaced with make-list ?

Malabarba:
Yes. And that should allow you to get rid of `!cons` and `cider--dotimes` too.

rogergl:
I made a new pull request. One thing that needs fixing is probably the copyright notice in cider-combat...

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@
 
 ### Changes
 
+* [#1348](https://github.com/clojure-emacs/cider/issues/1348) dash is no longer neede.
+
```
bbatsov: I'd refresh this to "Drop the `dash` dependency.". Also - kill the blank line.

bbatsov:
In the commit message - remove -> Remove

You'll also have to rebase this branch of top of the current `master`.

rogergl:
You'll also have to rebase this branch of top of the current master
How does that work ? Just call git rebase master ?

bbatsov:
More or less. You'll have to update your local `master` first.

bbatsov:
Remove dash dependencies -> Remove dash dependency

You'll also have to squash the commits once again, but we're almost ready.

cider-mode.el
```diff
@@ -263,7 +264,7 @@ Returns to the buffer in which the command was invoked."
         ("Find"
          ["Find definition" cider-find-var]
          ["Find resource" cider-find-resource]
-         ["Go back" cider-pop-back])
+         ["Jump back" cider-jump-back])
```
bbatsov: Seems you resolved to conflict here incorrectly.

cider-mode.el
```diff
@@ -275,7 +276,6 @@ Returns to the buffer in which the command was invoked."
         ["Toggle var tracing" cider-toggle-trace-var]
         ["Toggle ns tracing" cider-toggle-trace-ns]
         ["Refresh loaded code" cider-refresh]
-        ["Select any CIDER buffer" cider-selector]
```
bbatsov: Here as well.

cider-mode.el
```diff
@@ -297,24 +297,22 @@ Returns to the buffer in which the command was invoked."
         ["Quit" cider-quit]
         ["Restart" cider-restart]
         "--"
-        ["View manual online" cider-open-manual]
```
bbatsov: Here as well.

cider-mode.el
```diff
@@ -297,24 +297,22 @@ Returns to the buffer in which the command was invoked."
...
+  (when-let (indent
+             (nrepl-dict-get (cider-resolve-var (cider-current-ns) symbol-name)
+                             "indent"))
     (let ((format (format ":indent metadata on ‘%s’ is unreadable! \nERROR: %%s"
                           symbol-name)))
       (with-demoted-errors format
         (cider--deep-vector-to-list (read indent))))))
 
 ;;; Dynamic font locking
-(defcustom cider-font-lock-dynamically '(macro core deprecated)
```
bbatsov: Here as well.

bbatsov:
You deleted some new code by mistake, while you were resolving your merge conflicts. Please go over your diff one more time and fix this.

rogergl:
I used Sourcetree and said keep their changes. Don't know what went wrong. I try to fix this.

Malabarba:
Don't give up! And remember, you're not wasting time, you're learning git. ;-)

rogergl:
I accidentally committed an additional file now I have to figure out how to undo this.

rogergl:
I tried again but probably messed up the commit history since I accidentally committed a cider-mode-old.el file.

bbatsov:
The code looks fine to me right now. @Malabarba can you go over it as well before we merge it?

Malabarba:
Ok, I see only one issue. Inside `cider-tests--no-auto.el` you require `cider-contrib` instead of `cider-compat`.

Malabarba:
I see many small things I'd like to fix, but they are mostly orthogonal to this PR (they were already there before) so I can do them myself after we merge this.

bbatsov:
I'll merge this and force-push to fix the contrib mistake.

Malabarba:
Thanks again @rogergl

rogergl:
@Malabarba: Thanks for given me the chance to learn more about git ;-)

bbatsov:
:-)

PR #1366, opened 2015-10-13
Add step to CljScr REPL usage

ftravers:
It wasn't fully explicit about how to get the Browser REPL working.  Missing step to launch browser, this hung me up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -894,6 +894,16 @@ Using Weasel, you can also have a browser-connected REPL.
 
 4. Open a file in your project and issue `M-x cider-jack-in-clojurescript`.
 
+5. Your ClojureScript CIDER REPL will now say:
+
+```
+user> << started Weasel server on ws://127.0.0.1:9001 >>
+<< waiting for client to connect ...
+```
+
+Now you can open your web page that includes your ClojureScript, and your REPL
+will finish connecting.
+    
 Provided that a Piggieback-enabled ClojureScript environment is active in your
 REPL session, code loading and evaluation will work seamlessly regardless of the
 presence of the `cider-nrepl` middleware. If the middleware is present then most


README.md
```diff
@@ -894,6 +894,16 @@ Using Weasel, you can also have a browser-connected REPL.
...
 4. Open a file in your project and issue `M-x cider-jack-in-clojurescript`.
 
+5. Your ClojureScript CIDER REPL will now say:
+
+```
+user> << started Weasel server on ws://127.0.0.1:9001 >>
+<< waiting for client to connect ...
+```
+
+Now you can open your web page that includes your ClojureScript, and your REPL
```
Malabarba: Maybe replace `Now you can open your web page that includes your ClojureScript` with `Now you can visit 127.0.0.1:9001 in a browser`.

bbatsov: And what about people developing server-side stuff in ClojureScript? :-)

Malabarba: I suppose there's always someone crazy enough... :P

ftravers: Ha ha!!!  I wouldn't be surprised!!!  I'm still trying to understand
exactly what the trade offs between this and figwheel are?

ftravers: Umm, http://127.0.0.1:9001, is nothing right?  That is just a websocket port for the underlying coms between the browser and the cider-repl correct?  I have just been navigating to my files via: file:///.  I didn't realize there was any sort of web-server that is setup with a cider-jack-in-clojurescript command..??

Malabarba: >Umm, http://127.0.0.1:9001, is nothing right? That is just a websocket port for the underlying coms between the browser and the cider-repl correct?

You're probably right. I've never tried weasel, and I probably just confused it with figwheel. 
Have you seen any other quick start guides on this? We should just copy what they say at this last step.

bbatsov: Yeah, probably this has to be reviewed more carefully. I haven't used neither projects and I can't really provide any meaningful input here. 

P.S. The commit message can be improved.

ftravers: I have been used to using figwheel as well, but as a friend suggested, you
lose all the goodness of CIDER then, so I've been kind of using both.  It
seems without figwheel having compiled the project, I get an error when
goog.require('cljs.core') happens...

So it seems almost have to say...also run 'lein figwheel' too.  But thats
just a kludge.

The other approach, which I've got working, it to launch a 'lein figwheel',
then cider-connect to it.  But this is quite a different approach so
probably shoudn't go here at all.  I just set it up 10 seconds ago, so
don't know if I'll find any weirdness to it...I'll report back
regardless...if it seems that, that is indeed the correct way to have a
cider-clojurescript repl, then perhaps that is what we should advise,
versus this docs recommendation about weasel?

ftravers: It doesn't allow a simultaneous clojure repl...only a cljs repl.  So now am
back to the setup where I also have figwheel running and use weasel...to me
this is the best setup I can figure out for now.

malchmih: @ftravers Figwheel can start nREPL server for you and you can connect to it from CIDER with two connections (one for clojure repl and another one for clojurescript). This way, you'll get CIDER goodness working for both clj and cljs, and a browser repl also.

Malabarba: Yeah. Last time I did some cljs dev I used figwheel with dual cider connections. You can find the instructions if Google figwheel cider.

ftravers: Can connect a single cider to figwheel fine, just hung up on how to connect
two, one for each of clj & cljs.  Not seeing anything on google about how
to do this for two sessions.  But I'm very glad it can be done...so I'll
keep hunting/thinking in mean time. :)

arichiardi: Great @ftravers! Please if you find out let us know :dancers:

Malabarba: It should be sufficient to connect twice to the server (which will give you two clj repl buffers) and only activate the cljs repl in one of the repl-buffers.

arichiardi: Well, I think it is even easier: with ```lein figwheel``` you actually have already a server up and running at the port you choose in ```project.clj```. I guess you then need to ```cider-connect``` to it and then run [```cider-figwheel-repl()```](https://github.com/bhauman/lein-figwheel/wiki/Using-the-Figwheel-REPL-within-NRepl) (scroll down a bit I contributed it, hopefully it is still working fine).

ftravers: I think this approach still has some issues.  For example it doesn't have
the smarts to know which REPL to direct commands to, for example C-c C-z
(cider-switch-to-repl-buffer) in a *.cljs buffer takes me to the wrong REPL.

So at this point I guess, I'd just go with the double REPLs that
cider-jack-in-clojurescript provides with Weasel, and just run an
additional Figwheel REPL in a console to get the live reloading
functionality.

Malabarba: > For example it doesn't have the smarts to know which REPL to direct commands to, for example C-c C-z

It should, so you're probably suffering from #1368. Hopefully it will be fixed by 0.10

malchmih: @ftravers It had the smarts until very recently. I've opened the issue @Malabarba mentioned.

bbatsov:
So, what are we going to do about this PR?

ftravers:
maybe lets just close for now...  I'm not doing front end stuff right now
so don't have a good opinion about what to do...  I'm sure I'll circle back
to this in maybe 3 weeks.

PR #1374, opened 2015-10-16
Correctly jump to real source file under Boot

### LINKED ISSUE #1370: When using Boot, cider-find-dwim jumps to a source file in tmp folder
Probably it's related to https://github.com/boot-clj/boot/issues/249

As far as I could understood from sources, ```cider``` uses absolute file path it get's from ```nrepl```. Under ```boot``` it points to the file in boot's tmp folder.

Steps to reproduce this issue:

1. Clone https://github.com/danielsz/holygrail
2. run ```boot dev``` inside project folder
3. ```cider-connect``` to it
4. Then go to https://github.com/danielsz/holygrail/blob/master/src/clj/holy_grail/core.clj#L3, place cursor under ```holy-grail.systems``` and do ```cider-find-dwim```.

A workaround that fixes this issue for me is to try to use relative file path first and fallback to absolute file path. Here is a modified function that I use:

```elisp
      ;; workaround for cider finding files in boot's temp directory
      ;; first try to find file using relative path (resource key in response)
      ;; if not found - it's probably a library and should be found using absolute path (file key in response)
      (defun cider--jump-to-loc-from-info (info &optional other-window)
        "Jump to location give by INFO.
INFO object is returned by `cider-var-info' or `cider-member-info'.
OTHER-WINDOW is passed to `cider-jamp-to'."
        (let* ((line (nrepl-dict-get info "line"))
               (file (nrepl-dict-get info "file"))
               (resource (nrepl-dict-get info "resource"))
               (name (nrepl-dict-get info "name"))
               (local-buffer (and resource
                                  (not (cider--tooling-file-p resource))
                                  (cider-find-file resource)))
               (buffer (or local-buffer (and file
                                             (not (cider--tooling-file-p file))
                                             (cider-find-file file)))))
          (if buffer
              (cider-jump-to buffer (if line (cons line nil) name) other-window)
            (error "No source location"))))

```

I don't know if it's the right repository for this issue. Maybe it should be fixed in cider-nrepl.

env0der:
Should fix #1370 

Having this patch locally, I've tried jumping to definition in both Leningen and Boot projects and did not encounter any issues.


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ when in buffer that's not visiting a file (e.g. a REPL buffer).
 * Tunneled ssh connection now deals correctly with the ssh password request.
 * [#1026](https://github.com/clojure-emacs/cider/issues/1026): The full `(ns ...)` form for the current buffer is now sent with all source-tracking eval requests, to fix ClojureScript compatibility.
 * [#1033](https://github.com/clojure-emacs/cider/issues/1033): Removed erroneous underlining from stacktrace frames and disabled frame filters in the error buffer.
+* [#1370](https://github.com/clojure-emacs/cider/issues/1370): Correctly jump to source files when connected to repl runned by Boot.
```
Malabarba: runned  -> run

env0der: Thanks!

cider-interaction.el
```diff
@@ -678,6 +678,12 @@ which nREPL uses for temporary evaluation file names."
   (let ((fname (file-name-nondirectory file-name)))
     (string-match-p "^form-init" fname)))
 
+(defun cider--find-file-if-not-tooling (file-name)
+  "Returns a buffer visiting the file URL if it exists and is a real source file, or nil otherwise."
```
Malabarba: This line looks too long. It should be short enough that it doesn't break if you hit `M-q`.

Malabarba:
The code looks good. I'm not familiar enough with boot to know that this would work, but I can take your word for it. :+1: 

You have some conflict with master, so you'll have to rebase.

env0der:
@Malabarba, done. It turned out that my master was very outdated, so rebasing took a while.

cider-common.el
```diff
@@ -159,6 +160,13 @@ If no local or remote file exists, return nil."
 (declare-function archive-extract "arc-mode")
 (declare-function archive-zip-extract "arc-mode")
 
+(defun cider--find-file-if-not-tooling (file-name)
```
Malabarba: I just had a look at `cider-find-file`, and I saw that it uses `url` for the argument name. We should probably keep that consistent.
I think we should change `file-name` to `url`, and then use the following docstring:

      "Return a buffer visiting file URL if it is a real source file.
    If URL is nil, or corresponds to a tooling file, return nil.
    See `cider-find-file' for the possible URL formats."

env0der: Done. Thanks for suggestion.

Malabarba:
Ok, squash the commits together and add `[Fix #1370] ` to the start of the commit message.
Then we wait for the others to :+1:.

cider-common.el
```diff
@@ -77,11 +77,12 @@ On failure, read a symbol name using PROMPT and call CALLBACK with that."
 INFO object is returned by `cider-var-info' or `cider-member-info'.
 OTHER-WINDOW is passed to `cider-jamp-to'."
   (let* ((line (nrepl-dict-get info "line"))
-         (file (nrepl-dict-get info "file"))
+         (absolute-file-name (nrepl-dict-get info "file"))
+         (relative-file-name (nrepl-dict-get info "resource"))
```
bbatsov: Why would we care about this? I find it quite confusing.

bbatsov:
I'm kind of puzzled that boot interferes with the metadata at all, as this definitely doesn't seem right. I totally don't understand why we should care about the `resource` key for instance. AFAIK it's role is to simply start the server - after that `load-file` should behave as usual. While I'd love to solve you're problem I don't generally accept fixes without a solid explanation.

bbatsov:
@env0der ping :-)

env0der:
@bbatsov during the compilation boot moves source files to a temporary directory (see picture on https://github.com/boot-clj/boot/wiki/Filesets#the-task-model).

```cider-find-dwim``` calls info function in nrepl which eventually adds information about file to the response (https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/info.clj#L171). As you can see, the ```:file``` key in response gets the absolute path and the ```:resource``` key gets relative path of the file.

The absolute file path comes from ```clojure.java.io/resource``` (https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/info.clj#L151) which finds resource in the class-path. And class-path contains those temp directories with sources instead of original ones (don't ask me why, I don't know). The original original directories are in ```fake.class.path``` (https://github.com/boot-clj/boot/issues/249).

I don't think that ```cider``` or ```nrepl``` should know about that ```fake.class.path``` on ```boot``` side. So what I propose is to try to find the file using the relative path at first (would work for project source files) and fallback to absolute path in case file could not be found (would work for libraries). I've been using this solution for about a week and it works for me.

**Comment by env0der** [comment]
I'm totally new to ```boot```, ```nrepl``` and ```cider``` so my understanding of how it works altogether may very wrong :)

**Comment by bbatsov** [comment]
@micha @alandipert can you chime in on this as well?

**Comment by alandipert** [comment]
I can speak to why boot copies input files to a temporary directory, if that's helpful:

* Tasks might remove files from the Fileset and we don't want the files being processed to actually be removed from the project
* Fileset contents are content-addressed and the underlying files must be immutable

So, the copy step makes the first immutable Fileset value in the succession of immutable Filesets that a boot pipeline is.

Because of this there is added indirection between files on the classpath and files on the disk.  Any file you can see on the classpath came from some weird tempdir, but its absolute path on the classpath might map to a project-relative path if the file in question is a project file.  What @env0der proposes looks good to me.

File: cider-common.el (hunk 4 above)
Comment (bbatsov) [inline cider-common.el]: Now that I read more about the problem at hand I think this code has to be commented to some extent, so that future readers won't be puzzled by it. I guess the comment should also refer to the PR's discussion.

File: cider-common.el [hunk 5]
```diff
@@ -77,11 +77,12 @@ On failure, read a symbol name using PROMPT and call CALLBACK with that."
...
          (name (nrepl-dict-get info "name"))
-         (buffer (and file
-                      (not (cider--tooling-file-p file))
-                      (cider-find-file file))))
+         (buffer (or
+                  (cider--find-file-if-not-tooling relative-file-name)
```
bbatsov: If this is relative to the project's root or it's src folder how would it work correctly without a temp binding to for `default-directory`?

cider-common.el
```diff
@@ -159,6 +160,14 @@ If no local or remote file exists, return nil."
 (declare-function archive-extract "arc-mode")
 (declare-function archive-zip-extract "arc-mode")
 
+(defun cider--find-file-if-not-tooling (url)
```
bbatsov: I'm not super fond of the name of this function. The fact we're ignoring tooling files isn't so important that it would justify putting it in the function's name.

cider-common.el
env0der: I don't know. For me it works probably because I also user ```projectile```. Is there a way to determine project's root in cider without projectile?

cider-common.el
env0der: What would be your suggestion for a better name?

cider-common.el
env0der: Ok, will do that

cider-common.el
bbatsov: `cider--find-file` would be enough I guess. I'm also not sure `url` is a good param name, as I believe it's just a simple file path.

cider-common.el
bbatsov: `clojure-project-root` or something like that. I'm not sure how projectile factor in all of this. Is something like `M-.` working for you with boot? If I could see some actual relative paths I'd have a better idea what's the best course of action.

cider-common.el
Malabarba: I think that'd be confusing with `cider-find-file`. I'm fine with the current function name, considering it kinda describes exactly what the function does. :-)

bbatsov: > I'm fine with the current function name, considering it kinda describes exactly what the function does. :-)

`if-not-tooling-and-has-url`... :-) I've never written a name like this in my life. 

We already use similar names prefixed as private here and there. For me it's apparent they are lower-level helpers. I'm open to other suggestions.

env0der: how about ```cider--find-source-file``` ?

bbatsov: Sounds good.

cider-common.el
env0der: Ah, crap, the returned relative path is like this: ```myproject/foo/bar.clj``` but the file itself is in ```src/clj/myproject/foo/bar.clj```. The absolute path is ```file:/Users/anton/.boot/cache/tmp/Users/anton/Projects/myproject/1yoq/7o2s6s/myproject/foo/bar.clj```.

It was finding the files correctly because they were opened already and ```cider-find-file``` looks in open buffers as well. Any idea how to find the file without hardcoding ```src/clj/``` prefix?

bbatsov: This exactly what I expected, when I raised my concern. I think `getResource()` should work for the source files, as they are on the classpath. I'm guessing there's also a way to query boot about it's source folder.

env0der: Ok, I did some investigation and in order to get the absolute path of an original source file we need to do something like this:

```clojure
(defn boot-absolute-file-path [file-name]
  ;; fake.class.path under boot contains the original directories with source files, see https://github.com/boot-clj/boot/issues/249
  (let [class-path (clojure.string/split (System/getProperty "fake.class.path") #":")
        class-path-urls (map #(new java.net.URL (str "file:" % "/")) class-path)
        class-loader (new java.net.URLClassLoader (into-array java.net.URL class-path-urls))]
    (.getFile (clojure.java.io/resource file-name class-loader))))
```

Probably this should go to ```cider-nrepl```. @bbatsov what do you think about this? Should I send a PR to ```cider-nrepl``` with such feature?

Then in ```cider``` we must somehow determine if the project is using ```boot```. What would be the best option to do this?

bbatsov: I think that adding this to `cider-nrepl` is the way to go indeed. If you simply check for the `.boot/cache/tmp` there when setting the file-info values you can set the proper paths directly and drop all the custom logic from the Emacs Lisp side.

env0der: Ok, I will implement it there. Indeed, we can avoid changes in Emacs package if we implement everything in ```cider-nrepl```. I think this PR can be closed then.

bbatsov: OK. Looking forward to the other PR.

PR #1376, opened 2015-10-17
Subscribe to server out upon connection

Malabarba:
Related to clojure-emacs/cider-nrepl#266.
Goes with clojure-emacs/cider-nrepl#267.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -481,6 +481,11 @@ In case `default-directory' is non-local we assume the command is available."
     (catch Throwable _ \"not installed\"))"
    (cider--check-middleware-compatibility-callback (current-buffer))))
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good to me. I'm wondering if we should expose the ability to unsubscribe a session in CIDER. Probably few people would use this, so it doesn't make much sense. On the other hand it would make it possible to toggle between the old and new behaviour. This is definitely not important, I'm just jotting down some thoughts.

Malabarba:
> Looks good to me. I'm wondering if we should expose the ability to unsubscribe a session in CIDER. 

If somebody asks for it, we can probably just tell them to do something like this

    (cider-nrepl-send-unhandled-request '("op" "out-unsubscribe"))

PR #1377, opened 2015-10-17
Make cider-emit-into-popup-buffer compatible with Emacs 24.3 and lower

### LINKED ISSUE #1375: Symbol's function definition is void: add-face-text-property
In Emacs 24.3.1, run `cider-refresh` to get the above error

Yuhta:
Fix #1375


bbatsov:
:+1: Thanks for spotting and fixing this!

Issue #1373, opened 2015-10-16
Support gradle and maven project in cider-jack-in


bbatsov:
Right now we properly detect only `boot` and `lein` projects, but some people are actually using `maven` and `gradle` adding the relevant project type detection logic and nREPL boot commands shouldn't be hard.

Related to https://github.com/clojure-emacs/refactor-nrepl/issues/124

//cc @kotarak

bbatsov:
Basically, to make some progress on this we need info regarding how maven and gradle users typically boot the nREPL server.

RadicalZephyr:
For gradle at least this is probably a good place to start looking: https://github.com/clojuresque

kotarak:
I'm working on the gradle support. I created a pull request for this: https://github.com/clojure-emacs/cider/pull/1380. (Please let me know if the flow is different. I usually don't work work with PRs.)

I'm also working on the gradle plugin to provide the necessary interface from the "other" side.

kotarak:
One question in particular, which is open for discussion in the PR, is the order of preference.

https://github.com/kotarak/cider/commit/9a22027d72530608b8266f7e0d5c3861edbb957b#diff-00bd1ee3b6816bf7f502bc824ec4a6e9R422

bbatsov:
As I mentioned in the PR - there should be no order of preference; we should simply detect all possible project types and prompt the use to pick one. 

You'll have to update `clojure-project-root` in `clojure-mode` as well.

kotarak:
Updated PR with sqashed commit: [PR #1391](https://github.com/clojure-emacs/cider/pull/1391).

bbatsov:
Gradle support has been merged. I'll close this for now as it's unclear now many people are using Maven with CIDER and I don't want us to just fill the backlog with noise.

PR #1380, opened 2015-10-21
Add basic gradle support

kotarak:
Judging from where "lein" and "boot" is used, this should be a first guess to add gradle support for `cider-jack-in`. This is not ready for pull, yet. I'm not sure the changes are sufficient and complete.

https://github.com/clojure-emacs/cider/issues/1373

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -413,19 +413,20 @@ Use `cider-ps-running-nrepls-command' and `cider-ps-running-nrepl-path-regexp-li
     (-distinct paths)))
 
 (defun cider-project-type ()
-  "Determine the type, either leiningen or boot, of the current project.
-If both project file types are present, prompt the user to choose."
+  "Determine the type, either leiningen, boot or gradle, of the current project.
+If more than one project file types are present, prompt the user to choose."
   (let* ((default-directory (clojure-project-dir (cider-current-dir)))
-         (lein-project-exists (file-exists-p "project.clj"))
-         (boot-project-exists (file-exists-p "build.boot"))
-         (gradle-project-exists (file-exists-p "build.gradle")))
-    ;; FIXME: multiple-choice
-    (cond ((and lein-project-exists boot-project-exists)
-           (completing-read "Which command should be used? "
-                            '("lein" "boot") nil t "lein"))
-          (lein-project-exists "lein")
-          (boot-project-exists "boot")
-          (gradle-project-exists "gradle"))))
+         (choices (delq nil
+                        (mapcar (lambda (candidate)
+                                  (when (file-exists-p (cdr candidate))
+                                    (car candidate)))
+                                '(("lein" . "project.clj")
+                                  ("boot" . "build.boot")
+                                  ("gradle" . "build.gradle"))))))
+    (if (< 1 (length choices))
+        (completing-read "Which command shoud be used? " choices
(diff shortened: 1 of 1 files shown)


cider.el
```diff
@@ -106,6 +106,20 @@ version from the CIDER package or library.")
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-gradle-command
+  "gradle"
+  "The command used to execute Gradle."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.9.0"))
```
Malabarba: This should be 0.10.0

cider.el
```diff
@@ -162,19 +176,22 @@ Sub-match 1 must be the project path.")
   "Check if the command matching PROJECT-TYPE is present."
   (pcase project-type
     ("lein" 'cider--lein-present-p)
-    ("boot" 'cider--boot-present-p)))
+    ("boot" 'cider--boot-present-p)
+    ("gradle" 'cider--gradle-present-p)))
```
Malabarba: This is just personal preference, but I would have added this line before the lein entry to minimise the diff. Not sure if others agree, though.

cider.el
```diff
@@ -424,6 +444,14 @@ In case `default-directory' is non-local we assume the command is available."
       (executable-find cider-boot-command)
       (executable-find (concat cider-boot-command ".exe"))))
 
+(defun cider--gradle-present-p ()
+  "Check if `cider-gradle-command' is on the `exec-path'.
+
```
Malabarba: I'm not a fan of empty second lines in short docstrings.

bbatsov:
@kotarak ping :-)

kotarak:
@bbatsov Pong!

Searching my way through elisp and emacs. I shortly tested this. It seems to work for me with the new version of the clojuresque plugin 2.0.0. (I launched to latter to the official gradle plugin site, but it doesn't work, yet. :-/)

I also opened a [required PR for clojure-mode](https://github.com/clojure-emacs/clojure-mode/pull/340).

cider.el
kotarak: I went with this existing project style.

cider.el
kotarak: Fixed.

bbatsov:
Looks OK - just squash the commits together and update the changelog.

cider.el
```diff
@@ -396,16 +413,20 @@ Use `cider-ps-running-nrepls-command' and `cider-ps-running-nrepl-path-regexp-li
...
-          (lein-project-exists "lein")
-          (boot-project-exists "boot"))))
+         (choices (delq nil
+                        (mapcar (lambda (candidate)
+                                  (when (file-exists-p (cdr candidate))
+                                    (car candidate)))
+                                '(("lein" . "project.clj")
+                                  ("boot" . "build.boot")
+                                  ("gradle" . "build.gradle"))))))
+    (if (< 1 (length choices))
```
bbatsov: I'd normally flip the condition here as you usually say "if the length of choices is more than 1". I wonder is `seq.el` has predicates like `seq-one-p` and `seq-several-p`. Those would be pretty neat. :-)

kotarak:
Messed up sqashing.

PR #1383, opened 2015-10-25
Add a keybinding for cider-jack-in-clojurescript

edvorg:
Hi. I suggest to add binding `C-c M-J` to `jack-in-clojurescript` in addition to `jack-in` (`C-c M-j`).
It might be more convenient for newbies and will make feeling that cider is easy to use out-of-the-box.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -517,6 +517,7 @@ process buffer."
 (eval-after-load 'clojure-mode
   '(progn
      (define-key clojure-mode-map (kbd "C-c M-j") #'cider-jack-in)
+     (define-key clojure-mode-map (kbd "C-c M-J") #'cider-jack-in-clojurescript)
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm fine with adding a keybinding, although I'm not sure I like this this one. 

The commit message should be "Add a keybinding for cider-jack-in-clojurescript".

@Malabarba @expez @cichli Ideas for the keybinding would be welcome!

edvorg:
Fixed commit message.

Malabarba:
I'm fine with the keybind. :+1:

bbatsov:
Guess there are no objections/alternative ideas.

expez:
Yeah, we're running out of attractive real estate for sure, but I've been missing this keybinding myself so I think this is a win.

Issue #1385, opened 2015-10-27
Cider checks nil process when reusing a buffer


udalrich:
When the clojure server is stopped by calling `(clojure.tools.nrepl.server/stop-server the-server)`,  `cider-find-reusable-repl-buffer` throws an error because `get-buffer-process` returns nil and this is passed to `process-live-p`. 

This can be fixed by checking for a nil process around line 248 in cider.el:

      (if exact-buff
          (if (process-live-p (get-buffer-process exact-buff))
              (when (y-or-n-p

becomes

      (if exact-buff
          (if (and (get-buffer-process exact-buff)
                   (process-live-p (get-buffer-process exact-buff)))
              (when (y-or-n-p

expez:
Thanks! :+1:

expez:
What's your emacs version @udalrich?  I don't think this fix was necessary.  `process-live-p` handles `nil` just fine on `24.5.1`

Malabarba:
@expez On 24.3, a nil argument to `process-live-p` meant “use the current buffer's process”, which then throws an error when the current buffer has no process.

Malabarba:
Fortunately, we don't need the process-live-p at all. get-buffer-process apparently never returns dead processes. 
This information is not on the docstring, but it's on the info manual:

    -- Function: get-buffer-process buffer-or-name
         This function returns a nondeleted process associated with the
         buffer specified by BUFFER-OR-NAME.  If there are several processes
         associated with it, this function chooses one (currently, the one
         most recently created, but don’t count on that).  Deletion of a
         process (see ‘delete-process’) makes it ineligible for this
         function to return.
    
         It is usually a bad idea to have more than one process associated
         with the same buffer.
    
              (get-buffer-process "*shell*")
                   ⇒ #<process shell>
    
         Killing the process’s buffer deletes the process, which kills the
         subprocess with a ‘SIGHUP’ signal (*note Signals to Processes::).

udalrich:
I was using 24.3.

Troy

PR #1388, opened 2015-10-28
Refactor cider-find-reusable-repl-buffer

expez:
- Remove unneeded liveness test for processes.
- Extract local function to toplevel function.
- Rewrite to favor if-let.

Related to #1385.

Could use a few more eyes on this.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -217,6 +217,16 @@ should be the regular Clojure REPL started by the server process filter."
              "code" cider-cljs-repl)
        (cider-repl-handler (current-buffer))))))
 
+(defun cider--select-zombie-buffer (repl-buffers)
+  "Return a zombie buffer from REPL-BUFFERS, or nil if none exists."
+  (when-let ((zombie-buffs (seq-remove #'get-buffer-process repl-buffers)))
+    (when (y-or-n-p
+           (format "Zombie REPL buffers exist (%s).  Reuse? "
+                   (mapconcat #'buffer-name zombie-buffs ", ")))
+      (if (= (length zombie-buffs) 1)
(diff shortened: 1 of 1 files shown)


cider.el
```diff
@@ -217,6 +217,18 @@ should be the regular Clojure REPL started by the server process filter."
              "code" cider-cljs-repl)
        (cider-repl-handler (current-buffer))))))
 
+(defun cider--select-zombie-buffer (repl-buffers)
+  "Return a zombie buffer from REPL-BUFFERS, or nil if none exists."
+  (when-let ((zombie-buffs (seq-remove
+                            (lambda (buff) (get-buffer-process buff))
```
Malabarba: a.k.a., `#'get-buffer-process` :-)

cider.el
```diff
@@ -225,39 +237,23 @@ and has no process, return it.  If the process is alive, ask the user for
...
+  (if-let ((repl-buffers (cider-repl-buffers))
+           (exact-buff (seq-find
+                        (lambda (buff)
+                          (with-current-buffer buff
+                            (or (and endpoint
+                                     (equal endpoint nrepl-endpoint))
+                                (and project-directory
+                                     (equal project-directory nrepl-project-dir)))))
+                        repl-buffers)))
+      (if-let ((process (get-buffer-process exact-buff)))
```
Malabarba: Not sure if I'm reading this right, but looks like `process` is not used here, so it could just an `if`.

Malabarba:
Much more readable :+1:

bbatsov:
:+1: Looks good to me.

PR #1395, opened 2015-11-02
Fix read-directory invocation the *right* way.

### LINKED ISSUE #1356: Weird prompt when associating connection with directory
This is the prompt that I get when I `cider-connect` and chose to associate the connection with a directory (the cursor is between those two `//`):

    Project directory: ~/path/to/cider-nreplh//home/artur/path/to/cider-nrepl/

I should say that the prompt works. I can just hit `RET` and the directory used will be `/home/artur/path/to/cider-nrepl/`. But it's certainly not optimal.

I'm guessing this is due to #1326. @Peeja do you have this too?

Peeja:
Improves #1326.
Fixes #1356.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -154,7 +154,7 @@ such a link cannot be established automatically."
   (interactive)
   (cider-ensure-connected)
   (let ((conn-buf (or connection (cider-read-connection "Connection: ")))
-        (project-dir (or project (read-directory-name "Project directory: " nil (clojure-project-dir) nil (clojure-project-dir)))))
+        (project-dir (or project (read-directory-name "Project directory: " (clojure-project-dir)))))
     (when conn-buf
       (with-current-buffer conn-buf
         (setq nrepl-project-dir project-dir)))))


bbatsov:
Are you sure? Isn't this where we started from? :-)

Peeja:
Close. :)

```diff
- (read-directory-name "Project directory: " nil (clojure-project-dir))
+ (read-directory-name "Project directory: " (clojure-project-dir))
```

bbatsov:
OK, I've read the docstring again and your fix seems reasonable. Just drop the `.` from the commit message's title.

Peeja:
Okay, that seems odd to me, but if that's the style guide… :)

bbatsov:
Take a look here http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Peeja:
That's my personal commit message style guide too, but it doesn't say anything about periods, does it? It just says to limit the subject line to 50 characters, unless I'm missing something.

Malabarba:
It doesn't say it exploicitly, but the summary line in the example doesn't end in a period.

Peeja:
I suppose. Personally, I prefer to end my sentences with periods in all cases, but I'm happy to defer to the project's standards.

bbatsov:
The first line of a commit message is a title, not a sentence.

Peeja:
Huh, is it weird that reading that makes me feel better about the whole thing? You've convinced me. Body text has periods, headlines don't. If it's good enough for the NY Times, it's good enough for me. :+1:

PR #1396, opened 2015-11-03
[Fix #1324] Don't font-lock local variables

### LINKED ISSUE #1324: False positives when highlighting vars in core

Malabarba:
This detects local variables as part of font-locking, and records them
to a cider-locals text property.

Note this is a bit of a hack, and has known issues, but a proper
solution would involve implementing a reader of Clojure code.

Known issues:

1. False positive for multi-arity functions. The args of all possible
   arities are gathered and applied to all scopes in the function.
2. False positives when a destructuting map has an :or clause.

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -428,6 +428,117 @@ namespace itself."
       (with-no-warnings
         (font-lock-fontify-buffer)))))
 
+
+;;; Detecting local variables
+(defun cider--read-locals-from-next-sexp ()
+  "Return a list of all locals inside the next logical sexp."
+  (save-excursion
+    (ignore-errors
+      (clojure-forward-logical-sexp 1)
+      (let ((out nil)
+            (end (point)))
+        (forward-sexp -1)
+        ;; FIXME: This returns locals found inside the :or clause of a
+        ;; destructuring map.
+        (while (search-forward-regexp "\\_<[^:&]\\(\\sw\\|\\s_\\)*\\_>" end 'noerror)
+          (push (match-string-no-properties 0) out))
+        out))))
+
+(defun cider--read-locals-from-bindings-vector ()
+  "Return a list of all locals inside the next bindings vector."
+  (save-excursion
+    (ignore-errors
+      (cider-start-of-next-sexp)
+      (when (eq (char-after) ?\[)
+        (forward-char 1)
+        (let ((out nil))
+          (setq out (append (cider--read-locals-from-next-sexp) out))
+          (while (ignore-errors (clojure-forward-logical-sexp 3)
+                                (unless (eobp)
+                                  (forward-sexp -1)
+                                  t))
+            (setq out (append (cider--read-locals-from-next-sexp) out)))
+          out)))))
+
+(defun cider--read-locals-from-arglist ()
+  "Return a list of all locals in current form's arglist(s)."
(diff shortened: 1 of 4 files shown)


expez:
Does this actually solve #1324, though?  I don't see anything here about excluding the ns from syntax highlighting.

That's actually the only case that's bothered me, because it's considered bad form to shadow stuff found in `core`.  From that perspective the previous behavior is there to highlight the names in core and prevent you from writing unidiomatic code.

Malabarba:
No, you're right, it doesn't fix your issue. 😆
I'll have to add a commit to it.

Malabarba:
Done.

bbatsov:
I'm not sure if that'd be helpful, but you can also check the code @alexander-yakushev is using for a similar purpose in `compliment`. 

I was also wondering if we might push such info via `track-state`, but the false positives will be visible until something is actually evaluated that way.

Malabarba:
> I was also wondering if we might push such info via track-state, but the false positives will be visible until something is actually evaluated that way.

I could see us using the state tracker for that. It would involve

1. Using the instrumenter to read every piece of code evaluated by the user and detect the local variables inside each def
2. Save information about this in the defined var's metadata
3. Let the state tracker transmit that information

However, the complexity of that is a little more than I am comfortable pushing so close to a release. :-)

cider-mode.el
```diff
@@ -314,6 +314,7 @@ Returns to the buffer in which the command was invoked."
       (with-demoted-errors format
         (cider--deep-vector-to-list (read indent))))))
 
+
```
bbatsov: Why the extra blank line?

bbatsov: Nevermind. It's probably a page break.

cider-mode.el
```diff
@@ -428,6 +440,114 @@ namespace itself."
       (with-no-warnings
         (font-lock-fontify-buffer)))))
 
+
+;;; Detecting local variables
+(defun cider--read-locals-from-next-sexp ()
+  "Return a list of all symbols inside the next logical sexp."
```
bbatsov: symbols or locals?

cider-mode.el
```diff
@@ -428,6 +440,114 @@ namespace itself."
...
+before point."
+  (while (search-forward-regexp "(\\(ns\\_>\\|def\\|fn\\|for\\b\\|loop\\b\\|with-\\|do[a-z]+\\|\\([a-z]+-\\)?let\\b\\)"
+                                end 'noerror)
+    (goto-char (match-beginning 0))
+    (let ((sym (match-string 1))
+          (sexp-end (save-excursion
+                      (or (ignore-errors (forward-sexp 1)
+                                         (point))
+                          end))))
+      (if (equal sym "ns")
```
bbatsov: Guess the special handling of `ns` forms should be explained.

bbatsov:
Btw, not sure if that's a good idea or not, but now that we know the locals we can add font-locking for them. Or we can just leave them without font-locking. Just mentioning an idea. In a way they are the only thing using the default face anyways. :-)

Malabarba:
> now that we know the locals we can add font-locking for them. 

Yep, I'm thinking of that too. But that's a feature, not a bug, so I'll restrain myself until things are quiet again. :)

PR #1397, opened 2015-11-04
[Fix #1297] C-c M-n should act on both repls 

### LINKED ISSUE #1297: C-c M-n should act on both repls in cljc files.
It works as expected in `clj` and `cljs` files but when in a `cljc` file I expected both repls to change.

It seems this only acts on the *current-connection* in cljc files.

expez:
In a cljc or cljx buffer hitting `C-c M-n` to switch the repl ns will now act on both repls so they are in sync.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -28,6 +28,8 @@
 (require 'spinner)
 (require 'nrepl-client)
 (require 'cider-common)
+(require 'cider-util)
+(require 'clojure-mode)
 
 (require 'cider-compat)
 (require 'seq)
@@ -124,11 +126,15 @@ Also close associated REPL and server buffers."
 (defvar-local cider-repl-type nil
   "The type of this REPL buffer, usually either \"clj\" or \"cljs\".")
 
-(defun cider-find-connection-buffer-for-project-directory (project-directory &optional all-connections)
-  "Return the most appropriate connection-buffer for the given PROJECT-DIRECTORY.
+(defun cider-find-connection-buffer-for-project-directory (&optional project-directory all-connections)
+  "Return the most appropriate connection-buffer for the current project.
+
 By order of preference, this is any connection whose directory matches
-PROJECT-DIRECTORY, followed by any connection whose directory is nil,
+`clojure-project-dir', followed by any connection whose directory is nil,
 followed by any connection at all.
+
+If PROJECT-DIRECTORY is provided act on that project instead.
+
 Only return nil if `cider-connections' is empty (there are no connections).
 
 If more than one connection satisfy a given level of preference, return the
@@ -140,7 +146,9 @@ precedence over other connections associated with the same project.
 
 If ALL-CONNECTIONS is non-nil, the return value is a list and all matching
 connections are returned, instead of just the most recent."
-  (let ((fn (if all-connections #'seq-filter #'seq-find)))
+  (let ((project-directory (or project-directory
+                               (clojure-project-dir (cider-current-dir))))
+        (fn (if all-connections #'seq-filter #'seq-find)))
     (or (funcall fn (lambda (conn)
                       (when-let ((conn-proj-dir (with-current-buffer conn
                                                   nrepl-project-dir)))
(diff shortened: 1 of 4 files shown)


cider-repl.el
```diff
@@ -768,11 +770,19 @@ namespace to switch to."
...
-  (if (and ns (not (equal ns "")))
+  (when (or (not ns) (equal ns ""))
+    (error "No namespace selected"))
+  (let* ((connections (list (cider-current-connection)))
+         (connections (if (or (derived-mode-p 'clojurec-mode)
+                              (derived-mode-p 'clojurex-mode))
+                          (cons
+                           (cider-other-connection (cider-current-connection))
+                           connections)
+                        connections)))
```
Malabarba: Kinda minor nitpick, but I think it reads a little better if you remove the second binding and use this inside the `let` body.

    (when (or (derived-mode-p 'clojurec-mode)
              (derived-mode-p 'clojurex-mode))
      (push (cider-other-connection (cider-current-connection))
            connections))

Malabarba: Oh, and it needs to account for the fact that (cider-other-connection (cider-current-connection)) might be nil.

expez:
Rewritten to address all comments.  Thanks guys :)

Malabarba:
:+1:

cider-client.el
```diff
@@ -220,6 +226,15 @@ file extension."
     ;; TODO: Add logic to dispatch to a matching Clojure/ClojureScript REPL based on file type
     (car cider-connections)))
 
+(defun cider-other-connection (connection)
+  "Return the first connection of another type than CONNECTION or nil."
```
bbatsov: Here it should say the connection has the same project.

cider-client.el
```diff
@@ -783,10 +798,15 @@ endpoint and Clojure version."
   "Extract the essential properties of CONN-BUFFER."
   (with-current-buffer conn-buffer
     (list
+     :type cider-repl-type
      :host (car nrepl-endpoint)
      :port (cadr nrepl-endpoint)
      :project-dir nrepl-project-dir)))
 
+(defun cider--connection-type (conn-buffer)
+  "Get CONN-BUFFER's type."
```
bbatsov: I'd mention that the possible types are Clojure and ClojureScript.

cider-repl.el
```diff
@@ -761,18 +761,26 @@ text property `cider-old-input'."
...
                          (completing-read "Switch to namespace: "
                                           (cider-sync-request:ns-list))
                        (cider-current-ns))))
-  (if (and ns (not (equal ns "")))
+  (when (or (not ns) (equal ns ""))
+    (user-error "No namespace selected"))
+  (cider-nrepl-request:eval (format "(in-ns '%s)" ns)
+                            (cider-repl-switch-ns-handler
+                             (cider-current-connection)))
+  (when (or (derived-mode-p 'clojurec-mode) (derived-mode-p 'clojurex-mode))
```
bbatsov: I sense this check will be used for other operations as well, so it can be made a predicate method.

cider-client.el
Malabarba: I'd rather just point to the docstring of `cider-repl-type`. Less duplication.

bbatsov:
This also needs a changelog entry.

Sorry about the extra comments - I was interrupted while going though the code the previous time.

cider-repl.el
```diff
@@ -761,18 +761,26 @@ text property `cider-old-input'."
 (defun cider-repl-set-ns (ns)
   "Switch the namespace of the REPL buffer to NS.
 
-If invoked in a REPL buffer the command will prompt you for the name of the
+If called from a cljc or cljx buffer act on both the Clojure and
+ClojureScript REPL if there are more than one REPL present.
```
bbatsov: if there is more than one

cider-client.el
bbatsov: Fine by me.

expez: Went with @Malabarba's suggestion

cider-repl.el
expez: Naming this function was a bitch, I called it `cider--cljc-or-cljx-buffer-p` and placed it in `cider-util.el`.

I also added a bunch of pagebreaks in that file at section already delimited with `;;;`.  I use [helm-pages](https://github.com/david-christiansen/helm-pages) so this made my life easier.

bbatsov:
:+1:

PR #1399, opened 2015-11-06
[Fix #1299] Eval in both REPLs for cljc and cljx 

### LINKED ISSUE #1299: Eval should dispatch to both repls in cljc files
This used to work but I couldn't get the clj buffer to sync with the `cljc` file using `C-c C-k`.

expez:
This changes the load file request to dispatch to `cider-other-connection` in addition to `cider-current-connection`, when the file is a cljc or cljx file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 
 ### Changes
 
+* [#1299]https://github.com/clojure-emacs/cider/issues/1299 <kbd>C-c C-k</kbd> and <kbd> C-c C-l</kbd> now dispatch to both the Clojure and ClojureScript REPL (in the same project) when called from a `.cljc` or `.cljx` file.
 * [#1397](https://github.com/clojure-emacs/cider/issues/1297) <kbd>C-c M-n</kbd> now changes the ns of both the Clojure and ClojureScript REPL (in the same project) when called from a cljc or cljx file.
 * [#1348](https://github.com/clojure-emacs/cider/issues/1348): Drop the dash dependency.
 * The usage of the default connection has been reduced significantly. Now evaluations & related commands will be routed via the connection matching the current project automatically unless there's some ambiguity when determining the connection (like multiple or no matching connections). Simply put you'll no longer have to mess around much with connecting-setting commands (e.g. `nrepl-connection-browser`, `cider-rotate-default-connection`).
--- a/cider-client.el
+++ b/cider-client.el
@@ -218,9 +218,7 @@ from the file extension."
               ;; OW, find one matching the extension of current file.
               (let ((type (or type (file-name-extension (or (buffer-file-name) "")))))
                 (or (seq-find (lambda (conn)
-                                (equal (with-current-buffer conn
-                                         cider-repl-type)
-                                       type))
+                                (equal (cider--connection-type conn) type))
                               project-connections)
                     (car project-connections)
                     (car cider-connections)))))))
@@ -431,21 +429,27 @@ If the nREPL request lacks a session `cider-current-session' is added."
       request
     (append request (list "session" (cider-current-session)))))
 
-(defun cider-nrepl-send-request (request callback)
+(defun cider-nrepl-send-request (request callback &optional connection)
   "Send REQUEST and register response handler CALLBACK.
 REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
 \"par1\" ... ).
-Return the id of the sent message."
-  (nrepl-send-request (cider--ensure-session request) callback (cider-current-connection)))
+If CONNECTION is provided dispatch to that connection instead of
+the current connection.
 
-(defun cider-nrepl-send-sync-request (request &optional abort-on-input)
+Return the id of the sent message."
+  (nrepl-send-request (cider--ensure-session request) callback
+                      (or connection (cider-current-connection))))
+
+(defun cider-nrepl-send-sync-request (request &optional abort-on-input connection)
   "Send REQUEST to the nREPL server synchronously.
 Hold till final \"done\" message has arrived and join all response messages
 of the same \"op\" that came along and return the accumulated response.
 If ABORT-ON-INPUT is non-nil, the function will return nil
 at the first sign of user input, so as not to hang the
 interface."
-  (nrepl-send-sync-request (cider--ensure-session request) (cider-current-connection) abort-on-input))
+  (nrepl-send-sync-request (cider--ensure-session request)
+                           (or connection (cider-current-connection))
+                           abort-on-input))
 
 (defun cider-nrepl-send-unhandled-request (request)
   "Send REQUEST to the nREPL server and ignore any responses.
@@ -613,17 +617,21 @@ thing at point."
 ;;; Requests
 
 (declare-function cider-load-file-handler "cider-interaction")
-(defun cider-request:load-file (file-contents file-path file-name &optional callback)
+(defun cider-request:load-file (file-contents file-path file-name &optional connection callback)
   "Perform the nREPL \"load-file\" op.
 FILE-CONTENTS, FILE-PATH and FILE-NAME are details of the file to be
-loaded. If CALLBACK is nil, use `cider-load-file-handler'."
+loaded.
+
+If CONNECTION is nil, use `cider-current-connection'.
+If CALLBACK is nil, use `cider-load-file-handler'."
   (cider-nrepl-send-request (list "op" "load-file"
                                   "session" (cider-current-session)
                                   "file" file-contents
                                   "file-path" file-path
                                   "file-name" file-name)
                             (or callback
-                                (cider-load-file-handler (current-buffer)))))
+                                (cider-load-file-handler (current-buffer)))
+                            connection))
 
 
 ;;; Sync Requests
--- a/cider-client.el
+++ b/cider-client.el
@@ -421,14 +421,6 @@ Signal an error if it is not supported."
   (unless (cider-nrepl-op-supported-p op)
     (error "Can't find nREPL middleware providing op \"%s\".  Please, install (or update) cider-nrepl %s and restart CIDER" op (upcase cider-version))))
 
-(defun cider--ensure-session (request)
-  "Make sure REQUEST has session.
-
-If the nREPL request lacks a session `cider-current-session' is added."
-  (if (seq-contains request "session")
-      request
-    (append request (list "session" (cider-current-session)))))
-
 (defun cider-nrepl-send-request (request callback &optional connection)
   "Send REQUEST and register response handler CALLBACK.
 REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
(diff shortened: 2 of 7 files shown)


expez:
The build result doesn't make sense to me. I'm not even sure what the difference between the two builds are.

bbatsov:
There's just one shortcoming of the current implementation - the responses from the two requests will be interleaved, meaning most people will probably not realize what's going on. @Malabarba and I mentioned this yesterday as the biggest issue we need to tackle - how to handle the 2 return values that we'd get from operations like this one (or eval on multiple connections).

bbatsov:
> The build result doesn't make sense to me. I'm not even sure what the difference between the two builds are.

Which build result?

cider-interaction.el
```diff
@@ -1356,10 +1359,18 @@ unloaded."
...
+      (dolist (connection (list (cider-current-connection) other-connection))
+        (cider-request:load-file
+         (cider-file-string filename)
+         (funcall cider-to-nrepl-filename-function (cider--server-filename filename))
+         (file-name-nondirectory filename)
+         connection))
+    (cider-request:load-file
+     (cider-file-string filename)
+     (funcall cider-to-nrepl-filename-function (cider--server-filename filename))
+     (file-name-nondirectory filename)))
```
Malabarba: I think this whole diff should be implemented as a macro that checks for `(cider--cljc-or-cljx-buffer-p)`. This can then be used on the already-merged namespace switching too. Here's an example of how it would work:

    (cider-do-connections (conn)
      (cider-nrepl-request:eval (format "(in-ns '%s)" ns)
                                (cider-repl-switch-ns-handler conn)
                                conn))

This will make it easy for us to customize this behavior, implement a toggle variable, etc.

Malabarba:
Ok. I like this feature more than I initially did. I think dispatching load-file to both repls is a good default.
There are just 2 points I'd like to bring up:

1. There should be a macro (`cider-do-connections`?) that checks if runs `body` once for each relevant connection. See my comment here: https://github.com/clojure-emacs/cider/pull/1399/files#r44129568
2. Whenever we mess with stuff that refactors connections, things tend to break. I'd be slightly more comfortable if we merged this after the next release. But maybe I'm just being a coward. :-)

bbatsov:
> Whenever we mess with stuff that refactors connections, things tend to break. I'd be slightly more comfortable if we merged this after the 0.10 release. But maybe I'm just being a coward. :-)

There will definitely be some period of bug fixing after the initial release, so I wouldn't worry that much about this. At some point we'll have to invest into better tests for the connection-related functionality, though. :-)

expez:
`cider-do-connections` is a good idea, but there's the issue of whether the body acts on a file or a buffer.  In this pr we're dispatching differently based on a file but in the set-ns code, from yesterday, we were acting on buffers.

I don't think I'll have time to make these changes this week.  If anyone else wants to take this across the finish line that would be great, if not I'll probably have some more time for CIDER next week.

I also think this needs a bit more manual testing.

Malabarba:
> cider-do-connections is a good idea, but there's the issue of whether the body acts on a file or a buffer. 

I don't see why files vs buffers is a meaningful distinction here, however.... 

Once we starting doing this for local operations (eq, eval-last-sexp), the macro will have to take a keyword that indicates that we also  care about local context. So we could also use a keyword to indicate we only care about the file extension, and not the major mode.

expez:
> I don't see why files vs buffers is a meaningful distinction here, however....

You said:

> I think this whole diff should be implemented as a macro that checks for `(cider--cljc-or-cljx-buffer-p)`. 

And that made sense in terms of `set-ns` because that function acts in the context of a buffer, but `cider-load-file` takes `filename` is an input argument and so does not.

So if the macro had the signature you proposed, then it would just be a `dolist` on the connections and thus not contain the test for whether we want to dispatch to multiple REPLs or not.

Most of our functions act in the context of a buffer, though.  So I think this macro is going to make good sense in the next PR.

expez:
This feature is now blocked on https://github.com/clojure-emacs/cider/issues/1400

I can't test this feature because the middleware is constantly setting the repl type to the wrong value.

Malabarba:
Oh god. I just realised I think I edited your comment instead of posting a comment. 😱sorry about that! 
I've no idea how that happened

expez:
Aight, once I can test this feature I will rewrite `cider-load-file` to work in terms of `cider-load-buffer` instead of the other way around.  That way we can branch on the mode in the buffer and not worry about file extensions.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 
 ### Changes
 
+* [#1299]https://github.com/clojure-emacs/cider/issues/1299 <kbd>C-c C-k</kbd> and <kbd> C-c C-l</kbd> now dispatch to both the Clojure and ClojureScript REPL (in the same project) when called from a cljc or cljx file.
```
bbatsov: `.cljc` or `.cljx`

cider-client.el
```diff
@@ -591,17 +595,21 @@ thing at point."
...
 
 (declare-function cider-load-file-handler "cider-interaction")
-(defun cider-request:load-file (file-contents file-path file-name &optional callback)
+(defun cider-request:load-file (file-contents file-path file-name &optional connection callback)
   "Perform the nREPL \"load-file\" op.
 FILE-CONTENTS, FILE-PATH and FILE-NAME are details of the file to be
-loaded. If CALLBACK is nil, use `cider-load-file-handler'."
+loaded.
+
+If CONNECTION is nil, use `cider-current-connection'
```
bbatsov: A `.` is missing here. This comment is also a bit misleading as this function doesn't really do a `nil` check for the connection (although the function it invokes does it).

bbatsov:
One small remark about the commit message title - it's a bit unclear in it which eval is being referred to there. Today for moment I decided you've implemented interactive eval operations for cljc files as well. :-)

expez:
> One small remark about the commit message title - it's a bit unclear in it which eval is being referred to there.

The current patch affects both `C-c C-k` and `C-c C-l`.  We should definitely do the same for `C-c C-c`.  I'm not sure about `C-x C-e`.

bbatsov:
> I'm not sure about C-x C-e.

Don't see why we shouldn't do it for it as well.

expez:
Rewritten in terms of `cider-do-connections`

expez:
What can I do about this `gensym` problem the linter is complaining about?

Malabarba:
Use make-symbol instead.

bbatsov:
Apart from my small remarks the code looks good to me.

expez:
Testing this is blocked on https://github.com/clojure-emacs/cider/issues/1400

bbatsov:
@Malabarba I'm guessing you still can't reproduce #1400?

Malabarba:
No. I thought  @expez had tracked down the problem to some exceptions when switching namespace. Isn't that the case?

bbatsov:
So, @expez what's the status here?

expez:
I can't test it due to #1400 but #1400 can't be reproduced.  Perhaps someone else could test this so we could get it merged?

bbatsov:
Just rebase this and let's have it merged. I doubt it will break anything.

cider-client.el
```diff
@@ -209,20 +209,18 @@ from the file extension."
...
                 (or (seq-find (lambda (conn)
-                                (equal (with-current-buffer conn
-                                         cider-repl-type)
-                                       type))
-                              (append repls cider-connections))
-                    (car repls)
-                    (car cider-connections)))))))
+                                (equal (cider--connection-type conn) type))
+                              project-connections)
+                    (car project-connections)))))))
```
Malabarba: There should still be a `(car cider-connections)` as a last clause here. Otherwise this function might return nil.

cider-client.el
```diff
@@ -209,20 +209,18 @@ from the file extension."
       (cond
        ((cider--in-connection-buffer-p) (current-buffer))
```
Malabarba: I know you didn't touch this line, but I just noticed this clause prevents the `type` argument from being used. This, in turn, means that `(other-connection)` won't work inside a connection buffer.

cider-common.el
```diff
@@ -224,5 +224,20 @@ existing file ending with URL has been found."
       (16 t) ; empty empty
       (_ nil))))
 
+(defmacro cider-do-connections (var &rest body)
+  "For each appropriate connection, bind it to VAR, and execute BODY.
+
+The appropriate connections are found by inspecting the current buffer.  If
+the buffer is associated with a .cljc or .cljx file, BODY will be executed
+multiple times."
+  (declare (indent 1))
```
Malabarba: also `(debug (symbolp body))`

Malabarba:
Don't merge yet. I tested it and it seems to be breaking the state tracker. @expez, do you have any local work on this, or can I rebase it to current master?

expez:
Go ahead and rebase.

Malabarba:
I've found what the problem was. The “connection” and “session” specified in the load-file request didn't match (mostly because of -ensure-session), so everything got messed up.

Most relevantly, the state-tracker was sending information about the cljs session to the clj connection, which is why the repl-type info went wrong.

Overall, I think we should start purging `current-session` from our code as well. Due to our current structure, it is redundant information at best, and a bug source at worst.

Malabarba:
I changed do-connections to a function, map-connections.

Sadly, we don't yet live in a world where inter-file macros are super
safe, so this is going to cause the least headache in the future.

Malabarba:
The tests are failling due to byte-comp warnings from the new message-logging.
I was going to fix it myself but I'm not even sure what needs to be done (because it's catching a real bug) so I'll leave it to you @bbatsov. The issue is that some places call `nrepl-log-message` with a single argument (and worse, this argument is a string, not a request or a response so this code is guaranteed to barf when it's reached).

bbatsov:
> I was going to fix it myself but I'm not even sure what needs to be done (because it's catching a real bug) so I'll leave it to you @bbatsov.

Damn, I totally forgot about this. I know what you're referring to - an odd usage of the logging regarding some errors.

bbatsov:
> I've found what the problem was. The “connection” and “session” specified in the load-file request didn't match (mostly because of -ensure-session), so everything got messed up.
> 
> Most relevantly, the state-tracker was sending information about the cljs session to the clj connection, which is why the repl-type info went wrong.
> 
> Overall, I think we should start purging current-session from our code as well. Due to our current structure, it is redundant information at best, and a bug source at worst.

Purge it how? It's appended requests that don't have it, but there you can't automatically have there the tooling session if you want to use it. I was thinking lately because of the middleware exceptions that maybe the middleware requests should operate on the tooling session.

bbatsov:
Ah, looking at the change you've made we can remove most usages of it I guess. That was my idea with appending the session automatically actually, but I guess it makes more sense to do so where the connection is explicitly known.

Malabarba:
Yes. My thoughts were to just replace current-connection with nil everywhere, so we could still specify a connection if we ever want the tooling connection. 
But I was a little scared of doing a wide refactor right before release. So I opted to make the minimal change that fixed the bug.

bbatsov:
> My thoughts were to just replace current-connection with nil everywhere, so we could still specify a connection if we ever want the tooling connection.

You meant `current-session` and `current-tooling-session` I guess. :-)

Yeah, let's focus on fixing everything else that's important right now. I addressed the failing build in the morning and will try to refine/update the documentation in time for the release. As the next two days will be busy for me at work I doubt I'll be able to work or more bug fixes/features before the release.

expez:
Thanks for taking this one across the finish line @Malabarba!

PR #1402, opened 2015-11-07
Update test-report buffer when tests pass

jeffh:
If the user has the test-report buffer open. Update the buffer to show
the tests pass instead of keeping previously failed errors. This is less
confusing than seeing previously failed errors despite tests passing.


Malabarba:
LGTM 👍

bbatsov:
:+1:

PR #1408, opened 2015-11-10
[Fix #1407] Jack-in with existing connections

### LINKED ISSUE #1407: `cider-jack-in` doesn't handle correctly existing connections for the same project
This is likely related to the refactoring @expez did a while back.

```
(if (and (get-buffer-process exact-buff)
               (y-or-n-p (format "REPL buffer already exists (%s).  \
Do you really want to create a new one? "
                                 exact-buff)))
          (or (cider--select-zombie-buffer repl-buffers) 'new)
        exact-buff)
```

No matter how you answer the question a new connection gets created. One more illustration how susceptible the project is to serious regressions.

expez:
The branch where the user is asked whether to create a new REPL in the
face of an existing REPL erroneously ignored the user's answer and
always created a new REPL.

The return value of this code branch, has been changed to be nil, as
described in the docstring.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -261,11 +261,11 @@ be reused."
                                 (and project-directory
                                      (equal project-directory nrepl-project-dir)))))
                         repl-buffers)))
-      (if (and (get-buffer-process exact-buff)
-               (y-or-n-p (format "REPL buffer already exists (%s).  \
+      (if (get-buffer-process exact-buff)
+          (when (y-or-n-p (format "REPL buffer already exists (%s).  \
 Do you really want to create a new one? "
-                                 exact-buff)))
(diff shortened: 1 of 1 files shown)


expez:
This function has grown to be a real beast with tons of branches described in the docstring.  Given the lack of tests we should try to avoid such functions in the future.

bbatsov:
Or start writing tests. :-)

Malabarba:
Or both. :-)

bbatsov:
> Or both. :-)

Indeed. If we don't backfill the tests they are never going to write themselves. I'll definitely try to work more on them in the future (and on the general testing strategy for the project).

PR #1410, opened 2015-11-10
[Fix #1405] Always update test-report but pop buffer only as appropriate

### LINKED ISSUE #1405: test-report pops up even when all tests pass if the report buffer already exists
The desired behavior would be for the report buffer to pop only on failed tests, while green simply prints the status to the minibuffer. 

When `*cider-test-report*` does not exist and a test run is green, the behavior is as just described.
However, if a test fails and report buffer appears, if I bury the buffer instead of killing it,  it
pops up again when I rerun the (fixed) tests - even if they are all green.

Fabricationatorishnicatorist:
closes #1405


bbatsov:
:+1: Thanks!

PR #1424, opened 2015-11-18
Add cider-fn-insert-arglist and hookup with :exit-function of company-mode

ozjongwon:
I've written auto complete arglist to help my bad memory. It can be useful to other cider users as well.

Ex)
```
;;; _ cursor position
;;; [thing] possible autocomplete
(defrec_[ord]  <enter>) ;; will produce:
(defrecord name [& fields] & opts+specs)

(defrecord_) ;; won't produce anything.
```

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -542,6 +542,14 @@ The formatting is performed by `cider-annotate-completion-function'."
            (ns (cider-completion--get-candidate-ns symbol)))
       (funcall cider-annotate-completion-function type ns))))
 
+(defun cider-fn-insert-arglist (&rest ignore)
+  "Ignore IGNORE."
+  (when-let ((args (first (cider-eldoc-arglist (cider-symbol-at-point)))))
+    (save-excursion
+      (dolist (arg args)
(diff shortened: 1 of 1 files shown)


expez:
I'd much rather just hit <kbd>C-h</kbd> to get a reminder about the signature of `defrecord, in a transient buffer.

I think this is preferable because you get the desired reminder without having to cleanup the reminder which has been inserted into the buffer.

bbatsov:
I don't really like this as well. You still get the docsig in the minibuffer while doing completion and as @expez mentioned pressing `C-h` is pretty convenient. There's also https://github.com/expez/company-quickhelp

Malabarba:
@ozjongwon I have the same memory problem as you, but I don't like the idea of having to delete some stuff that the completion inserted for me.

It would make more sense to force a regular eldoc display after completion is done. This way the information is displayed on the minibuffer, without the user having to clean it up.

bbatsov:
I'm open to ideas for improvements of the current behaviour, but I don't like the proposal in this PR, so I'll close it.

PR #1426, opened 2015-11-18
Can't use bound variable in if-let's else clause

Fabricationatorishnicatorist:
Introduced in refactor addressing #1385, scuppers `cider-jack-in` for me.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -252,22 +252,22 @@ and has no process, return it.  If the process is alive, ask the user for
 confirmation and return 'new/nil for y/n answer respectively.  If other
 REPL buffers with dead process exist, ask the user if any of those should
 be reused."
-  (if-let ((repl-buffers (cider-repl-buffers))
-           (exact-buff (seq-find
-                        (lambda (buff)
-                          (with-current-buffer buff
-                            (or (and endpoint
-                                     (equal endpoint nrepl-endpoint))
-                                (and project-directory
-                                     (equal project-directory nrepl-project-dir)))))
-                        repl-buffers)))
-      (if (get-buffer-process exact-buff)
-          (when (y-or-n-p (format "REPL buffer already exists (%s).  \
+  (let ((repl-buffers (cider-repl-buffers)))
+    (if-let ((exact-buff (seq-find
+                          (lambda (buff)
+                            (with-current-buffer buff
+                              (or (and endpoint
+                                       (equal endpoint nrepl-endpoint))
+                                  (and project-directory
+                                       (equal project-directory nrepl-project-dir)))))
+                          repl-buffers)))
+        (if (get-buffer-process exact-buff)
+            (when (y-or-n-p (format "REPL buffer already exists (%s).  \
 Do you really want to create a new one? "
-                                  exact-buff))
-            'new)
-        exact-buff)
-    (or (cider--select-zombie-buffer repl-buffers) 'new)))
+                                    exact-buff))
+              'new)
+          exact-buff)
+      (or (cider--select-zombie-buffer repl-buffers) 'new))))
 
 ;;;###autoload
 (defun cider-jack-in (&optional prompt-project cljs-too)


Fabricationatorishnicatorist:
I've restarted Travis once but the failure persists, Looks unrelated to the patch.

expez:
Hi

Thanks for spotting this and opening a PR.  One minor thing, could you update the commit message to describe what was done? e.g "Fix reference to unbound variable"

cider.el
```diff
@@ -267,7 +267,7 @@ Do you really want to create a new one? "
                                   exact-buff))
             'new)
         exact-buff)
-    (or (cider--select-zombie-buffer repl-buffers) 'new)))
+    (or (cider--select-zombie-buffer (cider-repl-buffers)) 'new)))
```
bbatsov: I'd still extract this to a binding in some outer `let`, to remove the extra function invocation.

bbatsov:
And you should address my remark about calling the same function twice.

Fabricationatorishnicatorist:
Done.

How does a bug that outright breaks cider get into master and stay there for three weeks? How can I be the only one affected? Don't you dogfood?

bbatsov:
> How does a bug that outright breaks cider get into master and stay there for three weeks? How can I be the only one affected? Don't you dogfood?

I rarely run more than one REPL and as you've noticed we don't really have many tests. As no one else noticed it I'm guessing most people don't run many REPL as well. :-)

expez:
> How does a bug that outright breaks cider get into master and stay there for three weeks? How can I be the only one affected? Don't you dogfood?

We're dogfooding the `master` branch using melpa, just like you are.  Most likely it would've taken months, if not more, for me to execute the failing code path.  This is why it's so valuable when people take the time to report issues like this.

bbatsov:
The failure on the snapshot is super strange. Maybe there's some upstream problem...

Anyways, thanks for spotting and fixing the issue @Fabricationatorishnicatorist!

Fabricationatorishnicatorist:
@expez, I'm not pointing fingers - I'm just baffled. Doesn't every cider user run `cider-jack-in` 20x times a day? what am I doing differently that made me the only one affected?

bbatsov:
`cider-jack-in` was working totally fine for me.

Malabarba:
@Fabricationatorishnicatorist Your remark made me curious. (Of course everybody runs through this function at least once a day)
So I edebuged the function and stepped through it. Over here, there's no error in using `repl-buffers` inside the `else` clause. So now I'm not sure whether your commit actually fixed something or just hid a deeper issue.

Malabarba:
For instance, the following snippet evaluates fine for me:

    (if-let ((x 1)
             (y nil))
        10
      (list x y))
    ;; => (1 nil)

bbatsov:
Haha. I was just doing the same thing - the original code works just fine. I'll revert the patch. It seemed really odd that the bindings won't be visible in the `else`. Should have tested it earlier, but as usual I'm crazy busy.

Fabricationatorishnicatorist:
Now I'm even more confused.

I'm running on emacs git, could that be it?

bbatsov:
@Malabarba and me are using it as well.

Malabarba:
@Fabricationatorishnicatorist did you try the snippet I pasted?
if that snippet fails for you, try `find-function` on `if-let` and see where it's coming from.

Fabricationatorishnicatorist:
Well, the patch definitly fixes the problem for me and undoing it brings it back.

Your snippet returns the same result as given. WTF.

bbatsov:
`edebug` the code.

Fabricationatorishnicatorist:
I did when I was chasing it down, and with `toggle-debug-on-error`. I got a `repl-buffers unknown variable` backtrace.

Fabricationatorishnicatorist:
The thing is I came across this bug a couple of weeks ago, and just patched  it locally thinking you'll hit and fix this soon enough, Then today I upgraded cider and it came back, so I thought the hell with this, redid the patch and opened a PR. And now I've removed cider then reinstalled and I can't repro.

I have no idea what the my emacs is doing. But my apologies for the noise, this must be something weird and local on my end.

PR #1431, opened 2015-11-20
Give better advice for *print-length* setting

hanshuebner:
The README.md suggested that it would be a good idea to globally set `*print-length*` through the `:global-vars` key.  I followed the advice and have been bitten by that several times, because a non-`nil` `*print-length*` setting can cause several libraries (for example alembic or anything clojurescript related) to be incorrectly compiled.  In theory, everyone should make sure that the printer related dynamic variables are in a sane state when relying on the printer output to be something specific, but in practice this is hard to do, also because there is no macro like CL's `with-standard-io-syntax` that binds all printer dynamic variables to a sane state.

This PR updates README.md so that it suggests using `:repl-options` instead of `:global-vars`, which will not interfere with compilation of dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,11 +371,11 @@ and it expects `clojure.pprint` to have been required already
 Accidentally printing large objects can be detrimental to your
 productivity. Clojure provides the `*print-length*` var which, if set,
 controls how many items of each collection the printer will print. You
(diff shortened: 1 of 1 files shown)


bbatsov:
Seems reasonable to me. Any objections to merging this?

Malabarba:
None from me

PR #1447, opened 2015-12-03
Add cider-pprint-fn option

cichli:
Client bits for clojure-emacs/cider-nrepl#247.


bbatsov:
There are some bytecomp issues.

cichli:
Thanks; I also missed passing `pprint-fn` through in `cider-test-stacktrace-for`.

btw, from my limited testing, `fipp` is 5-10x faster than `clojure.pprint/pprint` - I'm looking forward to dropping 1.5/1.6 support so we can pull it in as the default :-).

README.md
```diff
@@ -1205,6 +1206,43 @@ Completion annotations can be disabled by setting
...
+This function must be resolvable by CIDER at the time it is called (i.e. its
+containing namespace must have already been required).
+
+CIDER will bind `*print-length*`, `*print-level*`, `*print-meta*`, and
+`clojure.pprint/*print-right-margin*` when calling the pretty-printing
+function - the function you provide is expected to respect these options.
+
+Fortunately, users of Leiningen can use the `:injections` feature to easily
+ensure the above conditions hold; for example, to use
+[fipp](https://github.com/brandonbloom/fipp), add the following to your `:user`
```
expez: Presumably this only makes sense in the `:repl` profile

cichli: @expez `lein repl` isn't the only way to start a REPL though - at work we embed a REPL in every application, and connect to them after doing `lein run`. Happy to change it if you think this is an overly-niche case but I think a lot of people REPL into e.g. QA/live environments where the `:repl` profile isn't applied

cichli:
Just pushed again to _actually_ fix the free variable error :-).

README.md
expez: You embed a REPL, but do you embed a repl running CIDER's middleware?  If not, this won't work, right?

Malabarba:
New one :-)
> cider-interaction.el:161:1:Error: defcustom for `cider-pprint-fn' fails to specify containing group

We should definitely figure out a way to get a list of all warnings in that build, instead of stopping at the first.

README.md
cichli: Yeah, we embed a REPL with `cider-nrepl` and `refactor-nrepl` :-).

gtrak: I embed a stock REPL, optionally with CIDER-nrepl on classpath with a maven
profile, loaded via dynamic require.

cichli:
> cider-interaction.el:161:1:Error: defcustom for `cider-pprint-fn' fails to specify containing group

:facepalm:

Should be fine now.

README.md
bbatsov: Maybe you can just extend the docs with some of the points you discussed here.

cider-client.el
```diff
@@ -483,23 +483,39 @@ If NS is non-nil, include it in the request."
    (cider-current-session)
    ns))
 
-(defun cider--nrepl-pprint-eval-request (input session &optional ns right-margin)
+(defcustom cider-pprint-fn nil
+  "Sets the function to use when pretty-printing evaluation results.
+
+If nil, `clojure.pprint/pprint` will be used.
```
Malabarba: The closing quote should be a '. Same throughout the docstrings.

cider-client.el
```diff
@@ -483,23 +483,39 @@ If NS is non-nil, include it in the request."
...
+
+Must be a namespace-qualified function of one argument.  If the function
+cannot be resolved, an exception will be thrown.
+
+The function is assumed to respect the contract of `clojure.pprint/pprint`
+with respect to the bound values of `*print-length*`, `*print-level*`,
+`*print-meta*`, and `clojure.pprint/*print-right-margin*`."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.10.0"))
```
Malabarba: 0.10.1 (or maybe 0.11?)

bbatsov: 0.11.0 - hopefully we're going to start releasing more often. :-) Smaller releases for the win.

cider-client.el
cichli: Isn't that only for emacs-lisp functions/variables? i.e. things that should be linkable in `help-mode`?

cichli:
Thanks all; I've bumped the version of the custom variable to `0.11.0` for now and have fleshed out the README bits around using `:injections`. I have two ideas for making this integration a bit more easy:

* PRs for `fipp` and `puget` to make them use `clojure.pprint/*print-right-margin*` if `:width` is not provided (this will obviate the need for the wrapper fn, so `cider-pprint-fn` can simply be set to `fipp.edn/pprint`)
* Once we've dropped support for Clojure 1.5 and 1.6, we can inline `fipp` and use that as the default pretty-printer (possibly having `pprint` and `fipp` as choices for the custom variable), obviating the need to ensure that `fipp.edn` has been required

cider-client.el
Malabarba: Ah, you're right. Then I think I would either use double quotes (the value is a string after all) or no quotes at all. :-)

bbatsov:
:+1:

PR #1448, opened 2015-12-03
Fix broken link to "Debugging" section

Peeja:
Looks like this Readme heading has changed its name.


bbatsov:
:+1: Thanks for spotting this!

PR #1455, opened 2015-12-08
[Fix #1452] Fix broken ANSI coloring in the repl

### LINKED ISSUE #1452: ANSI coloring in REPL buffer is broken
cider

### How to reproduce

1. Interactively eval the following code:

```clj
  (println "foo")
  (println "\u001b[31mbar\u001b[0m")
  (println "baz")
```

2. Then you will see the following output in REPL buffer:

```
foo  <-- default face
bar  <-- red face
baz  <-- red face(!)
user>
```

### Cause

This is because new output of interactive-eval extends an overlay for ANSI coloring created before.

The following is a part of output of <kbd>C-u C-x =</kbd> at 'r' of 'bar' after evaluating `(println "\u001b[31mbar\u001b[0m")`:

```
There is an overlay here:
 From 1 to 4
  face                 (foreground-color . "#fb4934")
  modification-hooks   (ansi-color-freeze-overlay)
```

Then, the following is new one after evaluating `(println "baz")`:

```
There is an overlay here:
 From 1 to 8
  face                 (foreground-color . "#fb4934")
  modification-hooks   (ansi-color-freeze-overlay)
```

It cannot be probably said that it is due to CIDER because I suspect this is ansi-color's problem.
I think `ansi-color-freeze-overlay` [should prevent overlays from being extended](https://github.com/emacs-mirror/emacs/blob/acfb5cd0353406784f085ddb6edfb0d0587048c8/lisp/ansi-color.el#L488-L514) at the first place.
I've actually confirmed it works by modifying ansi-color to use `insert-behind-hooks` instead of `modification-hooks`.

However, I think it is not so difficult to introduce a workaround for this on CIDER's side.
For example, it seems to work for me to use `insert` instead of `insert-before-markers` [here](https://github.com/clojure-emacs/cider/blob/5cd56ce76bdc4b0c50962ce626521133fb9e6edb/cider-repl.el#L471) though I'm not sure whether this cause other problems.
(BTW, cider-repl-output-start/cider-repl-output-end seems not to work but it is another issue.)

I'm a novice at elisp (and CIDER internals, of course), so I would be happy if someone investigate this problem further.

rfkm:
(Temporary?) fix for #1452

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -484,7 +484,19 @@ If BOL is non-nil insert at the beginning of line."
     (let ((pos (cider-repl--end-of-line-before-input-start))
           (string (replace-regexp-in-string "\n\\'" "" string)))
       (cider-repl--emit-output-at-pos (current-buffer) string face pos t)
-      (ansi-color-apply-on-region pos (point-max)))))
+      (ansi-color-apply-on-region pos (point-max))
+
+      ;; If the output has a trailing overlay created by ansi-color, it would be
+      ;; extended by the following outputs. To avoid this, ansi-color adds
+      ;; `ansi-color-freeze-overlay' to the `modification-hooks', but it never
+      ;; seems to be called. By using `insert-behind-hooks' instead, we can make
+      ;; it work. For more details, please see
+      ;; https://github.com/clojure-emacs/cider/issues/1452
+      (dolist (ov (overlays-at (1- (cider-repl--end-of-line-before-input-start))))
+        ;; Ensure ov is created by ansi-color
+        (when (and (member #'ansi-color-freeze-overlay (overlay-get ov 'modification-hooks))
+                   (not (member #'ansi-color-freeze-overlay (overlay-get ov 'insert-behind-hooks))))
+          (push #'ansi-color-freeze-overlay (overlay-get ov 'insert-behind-hooks)))))))
 
 (defun cider-repl-emit-interactive-stdout (string)
   "Emit STRING as interactive output."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and try to associate the created connection with this project automatically.
 * [#1466](https://github.com/clojure-emacs/cider/issues/1466): Correctly font-lock pretty-printed results in the REPL.
 * [#1475](https://github.com/clojure-emacs/cider/pull/1475): Fix `args-out-of-range` error in `cider--get-symbol-indent`.
 * [#1479](https://github.com/clojure-emacs/cider/pull/1479): Make paredit and `cider-repl-mode` play nice.
+* [#1452](https://github.com/clojure-emacs/cider/issues/1452): Prevent ANSI color overlays in the REPL buffer from being extended.
 
 ## 0.10.0 / 2015-12-03
 


cider-repl.el
```diff
@@ -484,7 +484,14 @@ If BOL is non-nil insert at the beginning of line."
     (let ((pos (cider-repl--end-of-line-before-input-start))
           (string (replace-regexp-in-string "\n\\'" "" string)))
       (cider-repl--emit-output-at-pos (current-buffer) string face pos t)
-      (ansi-color-apply-on-region pos (point-max)))))
+      (ansi-color-apply-on-region pos (point-max))
+
+      ;; Workaround for https://github.com/clojure-emacs/cider/issues/1452
```
bbatsov: I'd add more of the explanation of why this is needed here.

cider-repl.el
```diff
@@ -484,7 +484,14 @@ If BOL is non-nil insert at the beginning of line."
...
+      (dolist (ov (overlays-at (1- (cider-repl--end-of-line-before-input-start))))
+        ;; Ensure ov is created by ansi-color
+        (when (member #'ansi-color-freeze-overlay (overlay-get ov 'modification-hooks))
+          (unless (member #'ansi-color-freeze-overlay (overlay-get ov 'insert-behind-hooks))
```
bbatsov: I'd probably fold the `when` and the `unless`.

bbatsov:
Apart from my small remarks the code looks OK to me. @Malabarba you're our resident overlays expert - any feedback from you?

Malabarba:
What does `(overlays at -1)` yield? If it throws an error then line 490 could be dangerous. 

> @Malabarba you're our resident overlays expert - any feedback from you?

The overlay code looks solid. But I must say I didn't understand why the modification-hook used by ansi-color isn't enough. Shouldn't it also be triggered whenever insert-behind-hook is triggered?

rfkm:
@bbatsov Thank you for the review! I'll fix them later.

@Malabarba
> What does (overlays at -1) yield? If it throws an error then line 490 could be dangerous.

Just returns nil.


> But I must say I didn't understand why the modification-hook used by ansi-color isn't enough. Shouldn't it also be triggered whenever insert-behind-hook is triggered?

I tested the behavior with the following snippet:

```elisp
(with-temp-buffer
  (insert "foo")
  (let ((ov (make-overlay 1 4)))
    (overlay-put ov 'modification-hooks
                 (list (lambda (ov &rest args)
                         (message "hook %s" args))))
    (message "ov %s" ov)
    (goto-char 4)
    (insert-before-markers "a")
    (message "ov %s" ov)))
```
In this case, the hook is never called even though the overlay has been extended (tested with 24.4/24.5).

[The manual](http://www.gnu.org/software/emacs/manual/html_node/elisp/Overlay-Properties.html) says `modification-hooks` is called "if any character within the overlay is changed or if text is inserted *strictly* within the overlay".
I'm not sure what *strictly* exactly means, but it might mean text is inserted within the range of the overlay which is not extended yet.

Curiously, [the comments in `ansi-color-make-extent`](https://github.com/emacs-mirror/emacs/blob/acfb5cd0353406784f085ddb6edfb0d0587048c8/lisp/ansi-color.el#L499) says, "we use the `insert-behind-hook`'" even though `modification-hooks` is actually used.
On the other hand, [`ansi-color-freeze-overlay`'s docstring](https://github.com/emacs-mirror/emacs/blob/acfb5cd0353406784f085ddb6edfb0d0587048c8/lisp/ansi-color.el#L507) says, "This function can be used for the `modification-hooks`".

rfkm:
I'm sorry I'm late.
I have just pushed some changes.

The test has been failed, but this seems not to be due to my modification.

bbatsov:
Seems you'll have to rebase this branch on top of the current `master`.

rfkm:
Done

bbatsov:
@Malabarba are you happy with the current state of the patch?

Malabarba:
Yes. Sorry for the delay.

bbatsov:
@rfkm Rebase this and I'll have it merged.

rfkm:
Rebased. Thanks :-)

bbatsov:
:+1:

Issue #1099, opened 2015-05-12
Rewrite nrepl callback handling


vspinu:
This is a continuation of conversation at #1098

vspinu:
> One idea was to drop buffer argument in handlers and provide a macro that would create a handler with current buffer and current connection stored in its closure. 

The lexical approach is awkward with respect to the composition of the handlers that @cichli proposed. So, I guess, this leaves us with the dynamic approach only.

bbatsov:
Btw, I understand why we need the connection, but why do we need the current/another buffer? Not everything acts on it, so maybe the handlers should just obtain the buffers they need. Passing some buffer to all handlers always seems wrong to me.

vspinu:
> but why do we need the current/another buffer?  Not everything acts on it, so maybe the handlers should just obtain the buffers they need. 

I see at least two reasons for that. 

Most importantly, this is useful for "global" handlers that CIDER or the USER can insert into the chain of handlers. Most common example is caching.  As with  `(setq cider-buffer-ns ns)` in current `nrepl-make-response-handler` which is always called when `value` is present. 

Secondly, if you don't bind the buffer automatically then the handler has to do that itself (all eval handlers for example). This means that you cannot simply define a callback and then store only it's name in `nrepl-pending-requests`. You will still need to define a lambda-generator, call it in the code and hen store the whole calllback lambda in the request hash table. I am pretty sure if we store the current buffer we will reduce the use of closures as callbacks almost entirely.

I think the general idea is to store a complete context of the invocation and it's rather independent whether the handler wants to use it or not. For completeness I would also include the current point in this list (a marker actually). Examples are error handlers that must invoke a jump, and would like to push mark or xref-mark before hand.  It might be useful for asyncronuous completion as well. Imagine asyncronous eldoc which would inhibit message display if the point moved in the meanwhile.

A more general idea would be to have `nrepl-defhandler` macro which would allow to store arbitrary context. You would call it like `(nrepl-defhandler my-handler ((buf (current-bufer)) (mode major-mode)) &rest body)` but that basically means re-implementing closures which you can store nicely in the `nrepl-pending-requests` structure. Not sure it's worth the bang.

bbatsov:
I know why it's done as it is. My problem is that there are plenty of handlers that just ignore the "attached" buffer and currently this isn't even an optional param. 

[...]

PR #1467, opened 2015-12-12
Correctly font-lock pretty-printed results at the REPL

cichli:
Adding another argument to `nrepl-make-response-handler` is regrettable, but we have #1099 logged for untangling that. I'm going to take a look at fixing that soon - it will make it easier to solve the problem where `out` and `pprint-out` responses are interleaved (e.g. in the face of concurrency or laziness).


bbatsov:
> I'm going to take a look at fixing that soon - it will make it easier to solve the problem where out and pprint-out responses are interleaved (e.g. in the face of concurrency or laziness).

Excellent! IMO this is one of the ugliest parts on the nrepl client and reworking it would make the entire codebase easier to comprehend/extend. This was on my radar for 0.10, but I never got any time to work on it. Hopefully you'll manage to get the ball rolling this time around.

PR #1476, opened 2015-12-19
Update "Using Boot" setup instructions

fromheten:


Malabarba:
Is there a particular reason for this? In general I don't think the readme should tell people to use the snapshot.

bbatsov:
Yeah, the readme should not promote the snapshots.

fromheten:
The reason was that I updated Cider, and it complained that cider-nrepl was out of date. Sorry, I should have written that in the description.

bbatsov:
Yeah, but you've updated to the snapshot release (probably from melpa) and in general we avoid promoting its usage. This is why the readme always suggests the latest stable version in setup examples.

PR #1477, opened 2015-12-19
Check major mode first when choosing connection

fajpunk:
Emacs version: `24.5.1`
`; CIDER 0.11.0snapshot (package: 20151214.1327) (Java 1.8.0_05, Clojure 1.7.0, nREPL 0.2.12)`

`cider-repl-type` does not initially get set differently for the two
repls created by `cider-jack-in-clojurescript`.  If the major mode of
the current buffer is definitely Clojure or ClojureScript, use the
matching REPL connection for that type, if there is one open.  If the
major mode is not either of those modes, only then determine the
appropriate connection by looking at `cider-repl-type`.

Thanks to @rfkm for the suggestion for the change
(https://github.com/clojure-emacs/cider/commit/003adaa11c7b417486f58a7265e85c5e4c6d188b#commitcomment-15082109)!

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -197,6 +197,14 @@ such a link cannot be established automatically."
   (cider-ensure-connected)
   (kill-local-variable 'cider-connections))
 
+(defun cider-connection-type-for-buffer ()
+  "Return the matching connection type (clj or cljs) for the current buffer."
+  (cond
+   ((derived-mode-p 'clojurescript-mode) "cljs")
+   ((derived-mode-p 'clojure-mode) "clj")
+   (cider-repl-type)
+   (t "clj")))
+
 (defun cider-current-connection (&optional type)
   "Return the REPL buffer relevant for the current Clojure source buffer.
 A REPL is relevant if its `nrepl-project-dir' is compatible with the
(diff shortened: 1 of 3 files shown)


cider-client.el
```diff
@@ -197,6 +197,13 @@ such a link cannot be established automatically."
   (cider-ensure-connected)
   (kill-local-variable 'cider-connections))
 
+(defun cider-connection-type-for-buffer ()
+  "Return the matching connection type (clj or cljs) for the current buffer"
+  (cond ((derived-mode-p 'clojurescript-mode) "cljs")
+        ((derived-mode-p 'clojure-mode) "clj")
+        (cider-repl-type cider-repl-type)
```
Malabarba: The second one is redundant.

Malabarba:
Thanks for this. Indeed, I messed up with that logic because I forgot that the repl type defaults to clj, not nil. 

Other than that comment, it LGTM.

cider-client.el
```diff
@@ -197,6 +197,13 @@ such a link cannot be established automatically."
   (cider-ensure-connected)
   (kill-local-variable 'cider-connections))
 
+(defun cider-connection-type-for-buffer ()
+  "Return the matching connection type (clj or cljs) for the current buffer"
```
bbatsov: This sentence should end with a `.`.

cider-client.el
```diff
@@ -197,6 +197,13 @@ such a link cannot be established automatically."
...
+  (cond ((derived-mode-p 'clojurescript-mode) "cljs")
```
bbatsov: Add a line break after `cond`.

bbatsov:
I've added a few remarks. You should also squash related commits together.

fajpunk:
Thanks to both of you for the feedback!  I have addressed the remarks and squashed all of the commits.

bbatsov:
:+1: Thanks!

PR #1487, opened 2015-12-27
[Fix #1482] Clear nREPL sessions when connection is closed

### LINKED ISSUE #1482: Cannot re-connect after repl quits unexpectedly
Thanks for the awesome project! In recent versions of CIDER, I have to restart Emacs to connect to an `nrepl` after the previous `lein` process died.

## STR

1. `lein repl`
1. `cider-connect`
1. In the repl terminal window, `^D`
   `nREPL: Connection closed unexpectedly (connection broken by remote peer)`
1. re-start the repl
1. `cider-connect`
1. Re-use zombie REPL buffers? Yes

## Expected result

I can re-evaluate the buffer

## Actual result

I see a beach ball (on OSX) and Emacs becomes unresponsive. Evaluating the buffer doesn't work.

`*Messages*` shows ```nrepl-send-sync-request: Sync nREPL request timed out (op clone id 2)```

## System

```
; CIDER 0.11.0snapshot (package: 20151220.156) (Java 1.8.0_51, Clojure 1.5.1, nREPL 0.2.10)
```

In this test I'm using leiningen with an ancient version of Clojure, but the same problem exists with `boot` and Clojure 1.7.

Not that I'm *not* using cider-nrepl-middleware because it slows down startup of a largish projects by many seconds.

In tests the problem doesn't occur if I `cider-quit` before exiting the external `lein repl` (but you sometimes don't have that luxury).

rfkm:
Fix for #1482.
Following @Malabarba's suggestion, I put the clearing sessions code into nrepl-client-sentinel.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -615,6 +615,7 @@ and kill the process buffer."
              (substring message 0 -1)))
   (when (equal (process-status process) 'closed)
     (when-let ((client-buffer (process-buffer process)))
+      (nrepl--clear-client-sessions client-buffer)
       (with-current-buffer client-buffer
         (run-hooks 'nrepl-disconnected-hook)
         (when (buffer-live-p nrepl-server-buffer)
@@ -799,6 +800,12 @@ values of *1, *2, etc."
       (with-current-buffer conn-buffer
         (setq nrepl-ops ops)
         (setq nrepl-versions versions)))))
+
+(defun nrepl--clear-client-sessions (conn-buffer)
+  "Clear client nREPL sessions in CONN-BUFFER."
+  (with-current-buffer conn-buffer
+    (setq nrepl-session nil)
+    (setq nrepl-tooling-session nil)))
 
 (define-obsolete-function-alias 'nrepl-close 'cider--close-connection-buffer "0.10.0")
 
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1: Thanks!

nrepl-client.el
```diff
@@ -800,6 +801,12 @@ values of *1, *2, etc."
         (setq nrepl-ops ops)
         (setq nrepl-versions versions)))))
 
+(defun nrepl--clear-client-sessions (conn-buffer)
+  "Clear client nREPL sessions in CONN-BUFFER."
```
Malabarba: Can we expand this docstring a bit? How about:

Clear information about nREPL sessions in CONN-BUFFER. 
CONN-BUFFER is a connection buffer which contained a (presumably dead) REPL connection.

rfkm: @Malabarba 
Oops, sorry for the delay.
Looks great! Since I'm not a very good English writer, such a suggestion is very helpful for me, thanks ;).
Is it better to send another PR for this?  Or would you commit the fix directly?

expez:
@rfkm PR welcome with the suggestion docstring changes :)

Malabarba:
It's been difficult for me to find time at the computer lately, so a PR would be faster. :-)

rfkm:
OK, I'll send it later :)

PR #1488, opened 2015-12-27
Delete overlays when clearing REPL buffer

rfkm:
I've noticed that zombie overlays for ANSI coloring in the REPL buffer seem to impact its performance.

For example, evaluate the following code:

```clj
(dotimes [_ 100]
  (println "\u001b[31mfoo\u001b[0m"))
```

Then, type <kbd>C-u C-c C-o</kbd> to clear the repl buffer.

Finally, you can confirm empty overlays are still there by evaluating the following elisp code in the REPL buffer:

```elisp
(length (overlays-in (point-min) (point-max))) ; => 100
```

In my case, the number of zombie overlays reached more than 20000 after long working and REPL buffer's response was very slow.
In this PR, I tweaked `cider-repl-clear-output` and `cider-repl-clear-buffer` to remove overlays.


bbatsov:
:+1: Nice catch!

cichli:
Very nice catch indeed, thanks for your recent contributions @rfkm!

rfkm:
Glad to hear that! CIDER is one of the most important OSS for me, so I'm very happy to help.
Thanks for the great project! :-)

Issue #228, opened 2013-01-23
Bad performance in nrepl-show-maximum-output


acieroid:
When printing lots of lines, the performances of nrepl are really bad (emacs hang out for a long time, depending how long is the *nrepl* buffer). This seems to come from nrepl-show-maximum-output:

````elisp
 (dotimes (i 1000)
   (with-current-buffer nrepl-nrepl-buffer
     (insert-before-markers "foo\n")
     (nrepl-show-maximum-output)))
````
Takes 0.7827s to run on a clean buffer (that has just been C-c C-o'd), while
````elisp
 (dotimes (i 1000)
   (with-current-buffer nrepl-nrepl-buffer
     (insert-before-markers "foo\n")))
````
only takes 0.0169s.

The more the *nrepl* buffer is long, the more time it will take. For example, with a ~13k lines buffer, it takes around 7s while only 0.018s without nrepl-show-maximum-output.

This affects the performance of clojure codes that prints lots of stuff (eg. `(dotimes [i 1000] (println "foo"))`) to the point that you might have to kill emacs because it does not respond anymore.

sbocq:
The issue is also present in CIDER.

I ran ELP on `CIDER 0.9.0snapshot (package: 20150315.528)` and here is what it reports when parsing large XML files with `data.zip.xml`:

Function Name |	Call Count | 	Elapsed Time | 	Average Time
-------------|-------------|-----------------|--------------
cider-repl--show-maximum-output|3|49.264890923|16.421630307
cider-repl-emit-result|1|33.385848477|33.385848477
cider-repl-clear-output|1|22.934797576|22.934797576
cider-repl--end-of-line-before-input-start|1|22.931481792|22.931481792
cider-repl-emit-prompt|1|15.849498747|15.849498747
cider-eldoc|2|0.857376006|0.428688003
cider-eldoc-info-in-current-sexp|2|0.857124579|0.4285622895
cider-eldoc-beginning-of-sexp|2|0.8566141789|0.4283070894
cider-repl-return|1|0.035156144|0.035156144
cider-repl--send-input|1|0.034885998|0.034885998
cider-current-ns|7|0.0004598360|6.569...e-05
cider-symbol-at-point|2|0.000410108|0.000205054
cider-repl--insert-prompt|1|0.000250591|0.000250591
cider-repl-previous-prompt|1|0.000247797|0.000247797
cider-repl--find-prompt|1|0.000227403|0.000227403
[...]

PR #1500, opened 2016-01-01
Use text properties instead of overlays for ANSI coloring

### LINKED ISSUE #1452: ANSI coloring in REPL buffer is broken

rfkm:
### TL;DR

This PR improves the performance of REPL buffers by using text properties instead of overlays for ANSI coloring.
Also, this fixes #1452 while removing the hacky workaround introduced by #1455.

---

I recently sent some PRs related to overlays in the REPL buffer,
but I've finally realized we should avoid using overlays for ANSI coloring in the first place.
It's simply because overlays are pretty slow. I improved the performance a bit in #1488, but it's not radical.

So, in this PR, I remove all code related to overlays and use text properties instead.

I measured the performance by calling `cider-repl--emit-interactive-output` repeatedly.
The complete benchmark code is [here](https://gist.github.com/rfkm/c28f0e35965bcebe3e37), and the version of my Emacs is `GNU Emacs 24.4.1 (x86_64-apple-darwin14.0.0, NS apple-appkit-1343.14)`.

active

As this figure shows, the overlay version (current master) is getting slower and slower as the number of lines in the REPL buffer increases.

total

Total time the overlay version spent on inserting 1000 lines is about 5 seconds. As for the text property version, it's about 0.6 seconds.

If no window displays the REPL buffer, it is much faster in the both cases:

hidden

total_hidden

---

I keep my changes in #1488 as is because they are harmless, but I can remove them if it is necessary.

I also wonder whether I should remove my old changelog entries related to overlays.
Since this PR removes old PRs' changes, it looks a bit strange that these changelog entries appears in the same release.

P.S.

While I had not intended that, the PR #1455 seems to have improved the performance significantly ;)

active_old
(This is the worst case, that is, all outputs have a trailing overlay.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and try to associate the created connection with this project automatically.
 
 ### Changes
 
+* [#1500](https://github.com/clojure-emacs/cider/pull/1500): Improve the performance of REPL buffers by using text properties instead of overlays for ANSI coloring.
 * `cider-current-connection` considers major mode before `cider-repl-type`.
 * `cider-inspect` now operates by default on the last sexp. Its behavior can be altered via prefix arguments.
 * Requires Clojure(Script) 1.7 or newer.
@@ -24,9 +25,8 @@ and try to associate the created connection with this project automatically.
 * [#1466](https://github.com/clojure-emacs/cider/issues/1466): Correctly font-lock pretty-printed results in the REPL.
 * [#1475](https://github.com/clojure-emacs/cider/pull/1475): Fix `args-out-of-range` error in `cider--get-symbol-indent`.
 * [#1479](https://github.com/clojure-emacs/cider/pull/1479): Make paredit and `cider-repl-mode` play nice.
-* [#1452](https://github.com/clojure-emacs/cider/issues/1452): Prevent ANSI color overlays in the REPL buffer from being extended.
+* [#1452](https://github.com/clojure-emacs/cider/issues/1452): Fix wrong ANSI coloring in the REPL buffer.
 * [#1486](https://github.com/clojure-emacs/cider/issues/1486): Complete a partial fix in stacktrace font-locking.
-* [#1488](https://github.com/clojure-emacs/cider/pull/1488): Delete zombie overlays in the REPL buffer.
 * [#1482](https://github.com/clojure-emacs/cider/issues/1482): Clear nREPL sessions when a connection is closed.
 * [#1435](https://github.com/clojure-emacs/cider/issues/1435): Improve error display in cider-test.
 * [#1379](https://github.com/clojure-emacs/cider/issues/1379): Fix test highlighting at start of line.
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -469,13 +469,13 @@ If BOL is non-nil insert at the beginning of line."
           (goto-char position)
           ;; TODO: Review the need for bol
           (when (and bol (not (bolp))) (insert-before-markers "\n"))
-          (cider-propertize-region `(font-lock-face ,output-face
-                                                    rear-nonsticky (font-lock-face))
-            (insert-before-markers string)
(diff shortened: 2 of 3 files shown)


Malabarba:
Makes sense to me. Overlays are indeed very slow in large numbers. 
Thanks for the very thorough analysis. :-)

Do we have tests for this ansi coloring feature?

bbatsov:
> I also wonder whether I should remove my old changelog entries related to overlays.
> Since this PR removes old PRs' changes, it looks a bit strange that these changelog entries appears in the same release.

Yeah, you should remove them.

Overall - impressive work. I was amazed by your comprehensive analysis of the problem!
Performance optimization work is critical, but we've always ignored it to some extent so far. Maybe you can help us with tickets like:

* https://github.com/clojure-emacs/cider/issues/228
* https://github.com/clojure-emacs/cider/issues/1115

> Do we have tests for this ansi coloring feature?

Pretty sure we don't. A few tests would be nice to have.

Malabarba:
Also, I forgot to mention, but the code is good to merge for me (though some tests would be nice).

bbatsov:
Yeah, it's good to merge for sure - I just want us to update the changelog accordingly as well.

rfkm:
Sorry for the delay.
I've just pushed some changes.
Since this PR also fixes #1452, I kept the changelog entry about #1452 though I removed the mention of overlays.
Instead, I've added summary of this PR on top of this thread.
Also, I added a test for this.

> Performance optimization work is critical, but we've always ignored it to some extent so far. Maybe you can help us with tickets like:

I've already noticed `cider-repl--show-maximum-output` could be slow under certain conditions, but I hope one of the causes has been already eliminated.
One of the causes I noticed is `(recenter -1)` that `cider-repl--show-maximum-output` calls could be pretty slow if there are many overlays.
Also, I've noticed eldoc could be very slow in a large REPL buffer. As for this, I might send a PR in the near future.
Anyway, I'll try to ivenstigate their issues :)

bbatsov:
Great! Thanks a lot! :-)

Malabarba:
Thanks indeed!

PR #1501, opened 2016-01-01
Improve docstring of `nrepl--clear-client-sessions`

rfkm:
Thanks for your suggestion @Malabarba ! (#1487)

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -802,7 +802,7 @@ values of *1, *2, etc."
         (setq nrepl-versions versions)))))
 
 (defun nrepl--clear-client-sessions (conn-buffer)
-  "Clear client nREPL sessions in CONN-BUFFER."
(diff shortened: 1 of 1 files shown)


nrepl-client.el
```diff
@@ -802,7 +802,9 @@ values of *1, *2, etc."
         (setq nrepl-versions versions)))))
 
 (defun nrepl--clear-client-sessions (conn-buffer)
-  "Clear client nREPL sessions in CONN-BUFFER."
+  "Clear information about nREPL sessions in CONN-BUFFER.
+CONN-BUFFER is a connection buffer which contained a (presumably dead) REPL
```
bbatsov: Or simply - `CONN-BUFFER refers to a (presumably) dead connection, which we can eventually reuse.

rfkm:
Updated.

Issue #1361, opened 2015-10-06
Moving in the repl buffer


timur87:
First of all, thank you for all the effort to create this cool development environment. Till now it has been working flawless but I have been experiencing some difficulties nowadays. It's expected as I'm using a snapshot version but sill wanted to inform you about this.

I'm having issues when I want to move the curser in a REPL buffer. EMACS starts to hang and it gets really annoying, I'm just guessing but it's an issue related to matching the start and end tags of clojure constructs. 

I'm using CIDER 0.10.0snapshot (package: 20151005.453) (Java 1.8.0_45, Clojure 1.6.0, nREPL 0.2.7) on GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21)

If somehow I can provide more information please let me know.

Malabarba:
1. `M-x toggle-debug-on-quit` 
2. Reproduce the issue
3. Hit `C-g` while Emacs is hanged
4. Show us the backtrace you get

bbatsov:
> M-x toggle-debug-on-quit
> Reproduce the issue
> Hit C-g while Emacs is hanged
> Show us the backtrace you get

We should probably stuff this in the "troubleshooting" section.

timur87:
Debugger entered--Lisp error: (quit)
  scan-sexps(1662676 -1)
  forward-sexp(-1)
  #[0 "\301\302\303\304\305\306\307\300!\310\"\311\312%D\313\302\314\301EDC\217\210`\315\316!\210`W\205(\300\211\242T\240\262\2050\202\207" [(6229) err funcall make-byte-code 0 "`\301\302!\210\301\303!\210`W\205\300\303\240\207" vconcat vector [forward-sexp -1 1] 3 "\n\n(fn)" error #[257 "\300\207" [nil] 2 "\n\n(fn ##)"] forward-sexp -1] 8 "\n\n(fn)"]()
  funcall(#[0 "\301\302\303\304\305\306\307\300!\310\"\311\312%D\313\302\314\301EDC\217\210`\315\316!\210`W\205(\300\211\242T\240\262\2050\202\207" [(6229) err funcall make-byte-code 0 "`\301\302!\210\301\303!\210`W\205\300\303\240\207" vconcat vector [forward-sexp -1 1] 3 "\n\n(fn)" error #[257 "\300\207" [nil] 2 "\n\n(fn ##)"] forward-sexp -1] 8 "\n\n(fn)"])
[...]

PR #1502, opened 2016-01-04
Suppress eldoc when the current sexp seems to be too large

rfkm:
This PR restricts the range of searching the beginning of the current sexps for eldoc.

Current version of eldoc could be slow in a large buffer because `cider-eldoc-beginning-of-sexp` could traverse the whole buffer.
Such a situation is especially prone to happen in the REPL buffer.

In the following example, if the cursor is located at the right of the prompt,
`cider-eldoc-beginning-of-sexp` calls `(forward-sexp -1)` about 15 times (`user> -> line -> a -> is -> ...`).

```
This is a line
This is a line
This is a line
user>
```

When the REPL buffer has 4k lines and 35k sexp-like objects (this means `(forward-sexp -1)` will be called 35k times),
`cider-eldoc-beginning-of-sexp` takes about 4 seconds in my environment.
With my fix, it takes less than 0.001 seconds.

This might be related to #1361.

### DIFF
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -42,6 +42,9 @@
 (defvar cider-extra-eldoc-commands '("yas-expand")
   "Extra commands to be added to eldoc's safe commands list.")
 
+(defvar cider-eldoc-max-num-sexps-to-skip 30
+  "The maximum number of sexps to skip while searching the beginning of current sexp.")
+
 (defvar-local cider-eldoc-last-symbol nil
   "The eldoc information for the last symbol we checked.")
 
@@ -89,7 +92,8 @@ POS is the index of current argument."
 (defun cider-eldoc-beginning-of-sexp ()
   "Move to the beginning of current sexp.
 
-Return the number of nested sexp the point was over or after."
+Return the number of nested sexp the point was over or after.  Return nil
+if the maximum number of sexps to skip is exceeded."
   (let ((parse-sexp-ignore-comments t)
         (num-skipped-sexps 0))
     (condition-case _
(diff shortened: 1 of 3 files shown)


cider-eldoc.el
```diff
@@ -107,14 +111,18 @@ Return the number of nested sexp the point was over or after."
               (let ((p (point)))
                 (forward-sexp -1)
                 (when (< (point) p)
-                  (setq num-skipped-sexps (1+ num-skipped-sexps))))))
+                  (setq num-skipped-sexps (unless (and cider-eldoc-max-num-sexps-to-skip
```
bbatsov: There should probably be comments here explaining why this is necessary (for posterity's sake).

bbatsov:
This seems like a good solution for now, although long-term it seems to me that we should improve the process of sexp boundaries detection in the REPL.

rfkm:
Updated.

Yeah, I had been going to do so with narrowing at first, but I found it difficult to detect the boundary strictly because we might sometimes want eldoc to work with sexp in a log message in a REPL buffer.

bbatsov:
> because we might sometimes want eldoc to work with sexp in a log message in a REPL buffer.

I don't know - it seems to me that if something isn't REPL input we can safely ignore it in terms of eldoc.

cap10morgan:
Do we need to add something to the changelog for this?

Malabarba:
I think the proper way to do this would be to narrow from input-start to end-of-buffer, and then find the start of current sexp by running syntax-ppss. If point is before input-start then just don't do eldoc (or maybe do it by narrowing from beginning of line).

This might still be slow if the user pastes some huge sexp into the REPL (unlikely) but we can get around that by checking the size of the narrowed buffer.

Malabarba:
But, in any case, thanks for looking into this. :-)

PR #1508, opened 2016-01-09
[Fix #1507] Configure show-maximum-output with a defcustom

### LINKED ISSUE #1507: would like show-maximum-output to be optional
I understand the appeal of "show-maximum-output" and that it should be the default setting, but I'd like to be able to easily disable it.  I'd like to add a defcustom variable for it, and have cider-repl--show-maximum-output check the variable and do nothing if it's set to nil.

johnv02139:
Add a defcustom to control the behavior of cider-repl--show-maximum-output.  The default behavior is unchanged; this commit changes nothing unless the user customizes the variable to nil, in which case, the repl buffer will not re-center itself regardless of its position.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -89,6 +89,19 @@ When set to nil the buffer will only be created."
   :type 'boolean
   :group 'cider-repl)
 
+(defcustom cider-repl-always-show-max-output-in-buffer t
+  "Controls whether the REPL buffer displays the maximum output.
+
+When set to t (the default), if the REPL buffer contains more lines than the
+size of the window, the buffer is automatically re-centered upon completion
+of evaluating an expression, so that the bottom line of output is on the
+bottom line of the window.
+
+If this is set to nil, no re-centering takes place."
+  :type 'boolean
+  :group 'cider-repl
+  :package-version '(cider . "0.11.0"))
+
 (defcustom cider-repl-use-pretty-printing nil
   "Control whether the results in REPL are pretty-printed or not.
 The `cider-toggle-pretty-printing' command can be used to interactively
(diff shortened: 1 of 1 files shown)


cider-repl.el
```diff
@@ -89,6 +89,14 @@ When set to nil the buffer will only be created."
   :type 'boolean
   :group 'cider-repl)
 
+(defcustom cider-repl-use-show-max-output t
+  "Controls whether the REPL buffer displays the maximum output.
```
bbatsov: I get the point of the change, but I think that both the name of the defcustom and the description will be confusing for most people. In a way `show-maximum-output` is just an implementation detail.

cider-repl.el
```diff
@@ -89,6 +89,14 @@ When set to nil the buffer will only be created."
...
+
+To show the maximum output, the end of the buffer is moved to the bottom
+of the window.  When set to nil, the buffer is not re-centered."
+  :type 'boolean
+  :group 'cider-repl)
```
bbatsov: This should have a `:package-version` as well.

bbatsov:
This change should be mentioned in the changelog.

Malabarba:
WRT variable name, we could call it `cider-repl-scroll-on-output`.

bbatsov:
> WRT variable name, we could call it cider-repl-scroll-on-output.
>  Show all checks

:+1: Yeah, that's a much more meaningful name. Now idea who came up with a name like `comint-show-maximum-output`...

bbatsov:
@johnv02139 ping :-)

bbatsov:
I updated and merged this myself.

PR #1517, opened 2016-01-16
Implement the :here debug action

Malabarba:
This is the counterpart to https://github.com/clojure-emacs/cider-nrepl/pull/287


bbatsov:
The changes look good to me.

Malabarba:
Now that it's merged, I realise that `jump-here` might be a more intuitive name.

bbatsov:
Or `go-here`. Well, we can always change this. :-)

PR #1520, opened 2016-01-20
Complete Bowie lyric in cider-words-of-inspiration

Robsteranium:
It's a pretty trivial change but I think the juxtaposition of colours in "red shoes" and "the blues" is what makes this lyric so memorable!


bbatsov:
:+1: No argument from me. Lately I've been thinking about other ways to pay him tribute in the words of inspiration - e.g. "We can be heroes, we just need a REPL.".... or something like this.

PR #1521, opened 2016-01-20
Make auto scroll work with repl buffer in non-selected frames

rfkm:
I noticed `cider-repl--show-maximum-output` doesn't work with non-selected frames.
(Sometimes, I move a repl buffer to sub-monitor, creating a new frame.)

This PR should make it work with non-selected frames.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -392,7 +392,7 @@ This will not work on non-current prompts."
 (defun cider-repl--show-maximum-output ()
   "Put the end of the buffer at the bottom of the window."
   (when (eobp)
-    (let ((win (get-buffer-window (current-buffer))))
+    (let ((win (get-buffer-window (current-buffer) t)))
       (when win
         (with-selected-window win
           (set-window-point win (point-max))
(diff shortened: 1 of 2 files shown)


cider-repl.el
```diff
@@ -392,7 +392,7 @@ This will not work on non-current prompts."
 (defun cider-repl--show-maximum-output ()
   "Put the end of the buffer at the bottom of the window."
   (when (eobp)
-    (let ((win (get-buffer-window (current-buffer))))
+    (let ((win (get-buffer-window (current-buffer) 'visible)))
```
Malabarba: Any reason not to use `t` instead of `visible`?

rfkm: As far as I understand, `get-buffer-window` ignores hidden frames if `visible` is passed.
I feel like the auto scroll feature is meaningless if the repl buffer is invisible.

bbatsov: invisible == minimized?

rfkm: Yes, I think so.

bbatsov: Maybe we should make this a defcustom if some people would like to get auto-scrolling for invisible frames as well.

rfkm: Or just pass `t`?
I feel like https://github.com/clojure-emacs/cider/pull/1508 is fine grained enough as customizability.
In any case, I'm ready to update this PR. ;-)
What shall I do?

bbatsov: Ok, let's just go with `t`.

rfkm:
Done :)

bbatsov:
:+1:

PR #1526, opened 2016-01-24
Make local Javadoc lookup paths configurable

plexus:
If I do `cider-javadoc` on e.g. `javafx.stage.Stage`, it results in an elisp call to `(browse-url "javafx.stage.Stage")`, which doesn't even open a browser tab, presumably because `browse-url` doesn't know what protocol that relative URL has.

If I do the same for e.g. `java.lang.String` it results in a full url to the Oracle website, so that works. I tried to figure out where in cider-nrepl that URL is coming from, but I only find code there to build a relative URL, so I may be missing something.

What I really want is to have the JDK and JavaFX javadocs locally, and have CIDER open the right file from those local docs.

This commits adds a customizable search path for Javadoc directories, so you can easily do just that. If it doesn't find the file in any of the configured locations, it falls back to the old behavior.

I may be missing something here, or this may not be the best approach, so it's more a starting point for discussion. This is my first PR here, I'm happy to take any feedback into consideration.

```
; CIDER 0.11.0snapshot (package: 20160121.1400) (Java 1.8.0_66, Clojure 1.7.0, nREPL 0.2.12)
```

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -86,6 +86,11 @@
   :group 'cider-docview-mode
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-docview-javadoc-path ()
+  "List of directories where CIDER looks for Javadoc HTML files."
+  :type '(repeat string)
+  :group 'cider-docview-mode
+  :package-version '(cider . "0.7.0"))
 
 
 ;; Faces
@@ -181,9 +186,13 @@
   (when symbol-name
     (cider-ensure-op-supported "info")
     (let* ((info (cider-var-info symbol-name))
-           (url (nrepl-dict-get info "javadoc")))
-      (if url
-          (browse-url url)
+           (url (nrepl-dict-get info "javadoc"))
+           (f-expand (lambda (dir) (concat (file-name-as-directory dir) url)))
+           (candidates (mapcar f-expand cider-docview-javadoc-path))
+           (files (remove-if-not 'file-exists-p candidates))
+           (target (car `(,@files ,url))))
+      (if target
+          (browse-url target)
         (user-error "No Javadoc available for %s" symbol-name)))))
 
 (defun cider-javadoc (arg)


cider-doc.el
```diff
@@ -86,6 +86,11 @@
   :group 'cider-docview-mode
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-docview-javadoc-path ()
```
bbatsov: Use `nil` instead of `()` - it a time-honored convention of Elispers.

cider-doc.el
```diff
@@ -86,6 +86,11 @@
...
+  "List of directories where CIDER looks for Javadoc HTML files."
+  :type '(repeat string)
+  :group 'cider-docview-mode
+  :package-version '(cider . "0.7.0"))
```
bbatsov: 0.7.0 -> 0.11.0

cider-doc.el
```diff
@@ -181,9 +186,13 @@
   (when symbol-name
     (cider-ensure-op-supported "info")
     (let* ((info (cider-var-info symbol-name))
-           (url (nrepl-dict-get info "javadoc")))
-      (if url
-          (browse-url url)
+           (url (nrepl-dict-get info "javadoc"))
+           (f-expand (lambda (dir) (concat (file-name-as-directory dir) url)))
```
bbatsov: We don't use `f.el`, so this has to be replaced with something like `expand-file-name`.

cider-doc.el
```diff
@@ -181,9 +186,13 @@
...
+           (candidates (mapcar f-expand cider-docview-javadoc-path))
```
bbatsov: Here as well.

cider-doc.el
```diff
@@ -181,9 +186,13 @@
...
+           (files (remove-if-not 'file-exists-p candidates))
```
bbatsov: We're not using `cl.el`, so this should be ideally replaced with `seq-remove` or something similar.

cider-doc.el
bbatsov: Oh, you're naming a function `f-expand` - judging from my confusing that's not a good idea. :-) In general you should use `expand-file-name` instead of `concat` when building a file name.

bbatsov:
I'm a bit too tired to think in details about this right now, but I'll ping @jeffvalk to share his thoughts when he has time to look into the PR. At a glance it seems to me that this is something we'd better solve on the nREPL side, by relying on Java's own conventions about local javadoc.

jeffvalk:
@plexus If you add your local javadoc directory as a resource on the classpath, everything should work as you're expecting.

For reference, the `cider-nrepl` code that resolves this is [here](https://github.com/clojure-emacs/cider-nrepl/blob/a98f23614591395b33e74200b57001496c64453a/src/cider/nrepl/middleware/info.clj#L204).

plexus:
Oh that's great, thanks for the tip! And @bbatsov thanks for the review, my elisp-fu still has a long way to go :)

bbatsov:
We should probably document this, as the same question pops up occasionally about source lookups for 3rd party jars. Maybe we can even directly suggest this when we can't find the source/javadoc for something?

plexus:
It is actually already documented https://github.com/clojure-emacs/cider#basic-configuration

> If you are targeting the JVM and prefer a local copy of the JDK API documentation over Oracle's official copy (e.g., for JavaSE 8), per nREPL's javadoc-info logic (accurate as of 29 Dec 2014), you can arrange your project to include the root path of the local API doc (i.e., where the index.html is located) to be included on your classpath (i.e., where the doc HTML files can be located by clojure.java.io/resource). For example, for Leiningen, with the local API path being /usr/share/doc/java/api/, put the following line in project.clj:
> 
> `:dev {:resource-paths ["/usr/share/doc/java/api/"]}`
> 
> or the following line in $HOME/.lein/profiles.clj:
> 
> `:user {:resource-paths ["/usr/share/doc/java/api/"]}`
> 
> More detail can be found here.

Really sorry for the noise, I should have had a better look at the docs.

PR #1535, opened 2016-01-31
Suggest just using `.lein/profiles.d/repl.clj` for installation

hyPiRion:
After some feedback from technomancy/leiningen#2074, I realised that it may be easier for newcomers to Clojure and CIDER to just use a default provided `:repl` profile. Leiningen supports putting different profiles in `.lein/profiles.d` (for example `.lein/profiles.d/repl.clj`), and it may be an idea to provide steps to just download a profile file and put it in the right place.

I'm not sure if the way I've written this makes it easier or not, but at least now you know about the option. (I will also take a look into and see if we can make it easier to support more "automatic" kind of setups for tools that use Leiningen.)

Pinging @phillord for this, since he brought it up in the first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,8 +245,28 @@ Much of CIDER's functionality depends on the presence of CIDER's own
 
 #### Using Leiningen
 
-Use the convenient plugin for defaults, either in your project's
-`project.clj` file or in the :repl profile in `~/.lein/profiles.clj`.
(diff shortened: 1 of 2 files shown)


phillord:
Commented at https://github.com/technomancy/leiningen/issues/2074

Apologies -- I seem to have pushed the comments over several issues!

hyPiRion:
No problem – I'll close this so that we can discuss this over fewer issues =)

PR #1536, opened 2016-01-31
Add `eval-top-level-to-comment` command.

phillord:
As discussed ages ago!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### New features
 
-* Change default value of `cider-overlays-use-font-lock` to `t`. Unlike before, a value of `t`, causes `cider-result-overlay-face` is to be prepended to the font-lock faces (instead of just not being used).
+* [#1169](https://github.com/clojure-emacs/cider/pull/1169): New command `cider-eval-defun-to-comment`.
+* Change default value of `cider-overlays-use-font-lock` to `t`. Unlike before, a value of `t`, causes `cider-result-overlay-face` is to be prepended to the font-lock faces (instead of just not being used). 
 * `cider-result-overlay-face` default value changed to a background and a box, so it can be prepended to other faces without overriding the foreground.
 * [#1518](https://github.com/clojure-emacs/cider/pull/1518): Add `cider-dynamic-indentation` defcustom, to disable dynamic indent functionality.
 * Font-lock traced vars.
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -678,6 +678,24 @@ can be used to display the evaluation result."
                                  (cider-emit-interactive-eval-err-output err))
                                '()))
 
+(defun cider-eval-print-with-comment-handler (buffer location comment-prefix)
+  "Make a handler for evaluating and printing commented results in BUFFER.
+
+LOCATION is the location at which to insert.
+COMMENT-PREFIX is the comment prefix to use."
+  (nrepl-make-response-handler buffer
+                               (lambda (buffer value)
+                                 (with-current-buffer buffer
+                                   (save-excursion
+                                     (goto-char location)
+                                     (insert (concat comment-prefix
+                                                     value "\n")))))
+                               (lambda (_buffer out)
+                                 (cider-emit-interactive-eval-output out))
+                               (lambda (_buffer err)
+                                 (cider-emit-interactive-eval-err-output err))
+                               '()))
+
 (defun cider-popup-eval-out-handler (&optional buffer)
   "Make a handler for evaluating and printing stdout/stderr in popup BUFFER.
 
@@ -1049,6 +1067,21 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     ;; seems like the sexp is valid, so we can safely kill it
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
+
+(defun cider-eval-defun-to-comment (loc)
+  "Evaluate the \"top-level\" form and insert result as comment at LOC.
+
+With a prefix arg, LOC, insert before the form, otherwise afterwards."
+  (interactive "P")
+  (let ((defun-at-point (cider-defun-at-point))
+        (insertion-point
+         (if loc
+             (cider-defun-at-point-start-pos)
+           (cider-defun-at-point-end-pos))))
+    (cider-interactive-eval defun-at-point
+                            (cider-eval-print-with-comment-handler
+                             (current-buffer) insertion-point
+                             ";; => "))))
 
 (declare-function cider-switch-to-repl-buffer "cider-mode")
 
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -185,6 +185,7 @@ Returns to the buffer in which the command was invoked."
     (define-key map (kbd "C-x C-e") #'cider-eval-last-sexp)
     (define-key map (kbd "C-c C-e") #'cider-eval-last-sexp)
     (define-key map (kbd "C-c C-w") #'cider-eval-last-sexp-and-replace)
+    (define-key map (kbd "C-c M-;") #'cider-eval-defun-to-comment)
     (define-key map (kbd "C-c M-e") #'cider-eval-last-sexp-to-repl)
     (define-key map (kbd "C-c M-p") #'cider-insert-last-sexp-in-repl)
     (define-key map (kbd "C-c C-p") #'cider-pprint-eval-last-sexp)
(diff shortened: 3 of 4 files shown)


cider-interaction.el
```diff
@@ -1050,6 +1050,22 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
 
+(defun cider-eval-top-level-to-comment ()
+  "Evaluate the \"top-level\" form at point and insert result as pre-comment."
+  (interactive)
+  (let ((defun-at-point (cider-defun-at-point))
+        (point-at-defun (cider-defun-at-point-start-pos)))
+    (goto-char point-at-defun)
```
expez: I think we should wrap all this in a `save-excursion` I don't like it when commands move point when they don't have to.

Malabarba:
This looks a little roundabout. Why does it need to goto-char and insert two separate times?

cider-interaction.el
```diff
@@ -1050,6 +1050,22 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
 
+(defun cider-eval-top-level-to-comment ()
+  "Evaluate the \"top-level\" form at point and insert result as pre-comment."
```
bbatsov: Why a pre-comment? Maybe there should be some option to choose either a pre or a post comment?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#1169] New command `cider-eval-top-level-to-comment`.
```
bbatsov: There should also be a link to the ticket.

cider-interaction.el
```diff
@@ -1050,6 +1050,22 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
 
+(defun cider-eval-top-level-to-comment ()
```
bbatsov: I'd suggest this be renamed to `cider-eval-defun-to-comment`.

Malabarba:
Also, is there a reason to insert above? When you see code samples the results comment is usually below the snippet.

cider-interaction.el
expez: > I'd suggest this be renamed to `cider-eval-defun-to-comment`

That's not a very good name, I think.  Only elispers or common lispers know what a `defun` is.  Furthermore, the purpose isn't to evaluate defn-forms but toplevel forms like `(+ 1 1)` and insert the result as a comment.

expez:
I agree that having the result above the form is non-standard.  In many examples you also see a marker signifying evaluation like:

```clj
(+ 1 1)
;;=> 2
```

I prefer this style, but I'm probably not going to use this feature a lot so I'm good either way.

cider-interaction.el
bbatsov: Maybe it's not, but all similar functions use the term `defun` and I'm not going to rename them. Consistency is important.

cider-interaction.el
phillord: Tried this. If you surround the whole thing in save-excursion then  it prints

    results-of-eval;; =>

rather than 

  ;; => results of-eval

Why? Don't know. I am guessing that the call-back `cider-eval-print-handler` happens asynchronously? save-excursion has already emptied?

phillord:
Why comments before? Because comments always come before the thing to which they refer. That's how I've always seen it anyway.

phillord:
Why the two `goto-char`s --- I don't understand the semantics of the functions I am using at the moment, as I said above -- surround the whole thing in save-excursion and you get a different result.

In this case, we move the form forward first, then add the eval result (otherwise, we loose the end of the eval result.

There are other ways, of course.

cider-interaction.el
phillord: Think you are doing this backward. Why is `cider-eval-defun-at-point` bound to "Eval top-level sexp at point" in the menu bar?

Apart from that inconsistency, clojure does have a `defun`. Still I guess you know this:-)

phillord:
Oh, yeah, wanted to ask. At the moment, this misbehaves when the eval errors. `cider-eval-last-sexp-and-replace` calls `nrepl-sync-request:eval` first, and then `cider-interactive-eval`. Does this not eval the form twice?

cider-interaction.el
bbatsov: I've named it differently in the menu for the sake of people not familiar with the implementation (and Emacs traditions). This doesn't change the fact that the accepted convention for such names is `defun` and not `top-level`.

phillord: Okay, as you want.

phillord:
All changed.

All  the insertion logic is in the callback now. point is preserved, errors are handled, comments after or before on prefix arg and generally everything is cool and froody.

cider-interaction.el
```diff
@@ -664,7 +664,7 @@ can be used to display the evaluation result."
                                    (funcall nrepl-err-handler)))))
 
 (defun cider-eval-print-handler (&optional buffer)
-  "Make a handler for evaluating and printing result in BUFFER."
+  "Make a handler for evaluating and printing result in BUFFER at LOCATION."
```
Malabarba: You changed this docstring, but I think you meant to add this to the function below.

cider-interaction.el
```diff
@@ -1050,6 +1063,22 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
 
+(defun cider-eval-defun-to-comment (loc)
+  "Evaluate the \"top-level\" form at point and insert result as comment.
+
+If the prefix arg is positive, insert after the form.
+If the prefix arg is negative, insert before the form."
+  (interactive "p")
```
Malabarba: As is, this inserts after by default, and inserts before only with a negative prefix. 
I think it would make more sense to insert before with any prefix at all. (you can use a capital `"P"` for this).

phillord: That was the intention. Everyone seemed to think that after was normal and before was weird. Who am I to argue? So I moved the default.

cider-interaction.el
phillord: Oh dear. I keep forgetting to check the overall diff.

cider-interaction.el
Malabarba: Yes. I agree with that. I'm not suggesting to change the default, I'm just suggesting to do this:

```
(interactive "P")
(let ((defun-at-point (cider-defun-at-point))
      (point-at-defun
       (if loc
           (cider-defun-at-point-start-pos) ; <- this will be used only if a prefix is provided
         (cider-defun-at-point-end-pos)))) ; <- this is the default
  ...)
```

Malabarba: Anyway, this is a small detail. The PR LGTM.

phillord: Oh, yes, see the point. Have done.

cider-interaction.el
```diff
@@ -1050,6 +1067,21 @@ If invoked with a PREFIX argument, print the result in the current buffer."
     (backward-kill-sexp)
     (cider-interactive-eval last-sexp (cider-eval-print-handler))))
 
+(defun cider-eval-defun-to-comment (loc)
```
bbatsov: LOC should appear somewhere in the docstring.

cider-interaction.el
```diff
@@ -1050,6 +1067,21 @@ If invoked with a PREFIX argument, print the result in the current buffer."
...
+  "Evaluate the \"top-level\" form at point and insert result as comment.
+
+With a prefix arg, insert before the form, otherwise afterwards."
+  (interactive "P")
+  (let ((defun-at-point (cider-defun-at-point))
+        (point-at-defun
```
bbatsov: This is either the start or end of the defun - why is the variable named `point-at-defun`?

bbatsov:
You'll have to rebase on top of the current master. Please drop the `.` from the commit message's title.

phillord:
All done. Travis failure seems to be an existing problem with master.

cider-interaction.el
bbatsov: This is still not addressed.

cider-interaction.el
```diff
@@ -678,6 +678,23 @@ can be used to display the evaluation result."
                                  (cider-emit-interactive-eval-err-output err))
                                '()))
 
+(defun cider-eval-print-with-comment-handler (buffer location comment)
```
bbatsov: comment -> comment-prefix

cider-interaction.el
phillord: Will be this evening now...

phillord: Complete now

PR #1549, opened 2016-02-07
Obsolete many functions

Malabarba:
This doesn't change functionality, it only tries to consolidate some redundant code.
I've made this into a PR because it obsoletes quite a few functions, so don't hesitate to let me know if I've gone too far. :-)


bbatsov:
The changes look good to me. I'd mention that this consolidation is also a good opportunity to add a few unit tests here and there. Just saying'. :-)

Malabarba:
:-)

PR #1550, opened 2016-02-07
Require cider-test in cider-repl

r0man:
When loading cider the following error occurs:

`Symbol's value as variable is void: cider-test-menu`

This happens because the `cider-repl.el` file uses the symbol
`cider-test-menu` from the `cider-test.el` file, but doesn't require
it. This patch fixes the problem.


bbatsov:
:+1:

Issue #1534, opened 2016-01-31
Expose api for injecting additional middlewarep deps


expez:
Once https://github.com/clojure-emacs/cider/issues/1531 gets solved CIDER users won't have to manually add the cider-nrepl plugin.  It would be nice if CIDER exposed an API for additional things to be added to this bootstrap process.  clj-refactor would benefit from transparent inclusion of its middleware, as would packages like clojure-flycheck aka squiggly-clojure.

Malabarba:
Agreed. My suggestion would be a variable somewhat like this:

```el
(defvar cider-repl-dependencies
  `((cider-nrepl ,(upcase cider-version)))
  "List of Clojure dependencies for starting a REPL.
Each item is a list whose first element is the artifact name (a
symbol) and second element is the version (a string).")
```

benedekfazekas:
we might need different vars for leiningen and boot as their CLI api has slightly different semantics.

in detail for leiningen we need to provide a list of dependenices (for nrepl 0.2.12) and plugins (cider-nrepl and refactor-nrepl). however, for boot we need list of dependcies and list of middlewares (or we could use the latter for leiningen too perhaps if we decide not to use the lein plugins provided...)

Malabarba:
Then it probably makes sense to add a third element to that list identifying the middleware name. 

The difference in semantics (or the difference in how much information is necessary) should be handled by cider. The plugins just need to provided all necessary information, without bothering with which interface needs what.

bbatsov:
Do we need to do something here? I guess everyone can just modify the defcustoms.

benedekfazekas:
They are (def)vars so they don't show in customize buffers. But I don't
think there is anything here to do. Clj-refactor actually using this system
already

bbatsov:
One less ticket then. :-)

PR #1552, opened 2016-02-09
Inject cider REPL dependencies at cider-jack-in

### LINKED ISSUE #1531: Cider should inject its own dependencies/plugins automatically at cider-jack-in
Currently, cider complains when the dependencies are wrong in profile.clj. I understand the reasons for this, but it would be nice if Cider offered the ability to update your profile for you.

Another possibility would be for cider to check whether pomegranate is present and if not, inject the correct dependencies. Don't know if that is possible with plugins, but worth thinking about.

This would be great for me when I am switching Emacs' -- sometimes using melpa-stable, sometimes melpa normal. Then I have to continually re-edit my profile.clj. Painful.

benedekfazekas:
So users don't have to fiddle with profiles.clj and the like, see
details in #1531 Supports both leiningen and boot. Additionally there is
a defcustom `cider-skip-repl-dependencies-at-jack-in` to switch this
functionality off (defaults to nil); and an extension point
defun (`cider-add-repl-dependencies`) so other tools like clj-refactor
can inject to their own dependencies, middlewares

Fix #1531 and #1534


cider.el
```diff
@@ -165,6 +165,11 @@ This variable is used by `cider-connect'."
   :type 'boolean
   :version "0.9.0")
 
+(defcustom cider-skip-repl-dependencies-at-jack-in nil
```
expez: Why use a negation here?  Why not call it `cider-inject-dependencies` or something with a default of `t`?

benedekfazekas: I guess to stress the fact that we default to inject. but I am easy, happy to rename.

bbatsov: I agree with @expez.

bbatsov:
This should also be in the README and the changelog. In particular the README's installation section will need some updating.

cider.el
```diff
@@ -177,6 +182,15 @@ Sub-match 1 must be the project path.")
 (defvar cider-host-history nil
   "Completion history for connection hosts.")
 
+(defvar cider-repl-dependencies
+  `((("org.clojure/tools.nrepl" "0.2.12"))
+    (("cider/cider-nrepl" ,(upcase cider-version)))
+    ("cider.nrepl/cider-middleware"))
```
expez: I suggest we split this var in two: `cider-repl-dependencies` and `cider-repl-middleware`

benedekfazekas: thinking to leave this as is but rather split `cider-add-repl-dependencies` into `add-middlewares` and `add-dependencies`

cider.el
benedekfazekas: done

benedekfazekas:
@bbatsov changelog and README is done

cider.el
expez: Why introduce complexity (var which holds two types of data) when simplification also makes it easier to work with?

bbatsov: I agree with @expez. It's confusing right now.

cider.el
```diff
@@ -177,6 +182,15 @@ Sub-match 1 must be the project path.")
...
+  "List of Clojure artifacts for starting the REPL.
+First item is a list of dependencies where elements are lists of artifact name and version.
+Second item is a list of leiningen plugins where elements are lists of artifact name and version.
+Third item is a list of nrepl middleware names.")
```
bbatsov: nrepl -> nREPL

cider.el
```diff
@@ -204,6 +218,70 @@ Sub-match 1 must be the project path.")
     ("boot" cider-boot-parameters)
     ("gradle" cider-gradle-parameters)))
 
+(defun boot-command-prefix (dependencies)
```
bbatsov: Please add docstrings and unit tests for these 4 functions.

cider.el
```diff
@@ -165,6 +165,11 @@ This variable is used by `cider-connect'."
   :type 'boolean
   :version "0.9.0")
 
+(defcustom cider-inject-dependencies-at-jack-in t
+  "When nil, do not inject repl dependencies (most likely nrepl middlewares) at jack in time."
+  :type 'boolean
+  :version "0.11.0")
```
Malabarba: This should be `:package-version '(cider "0.11.0")`. 
(yes, the other defcustoms in this file are wrong too).

Malabarba: I mean `:package-version '(cider . "0.11.0")`

bbatsov: You beat me to this. :-) It's hard to do code reviews, while you're on a call. :-)

README.md
```diff
@@ -218,12 +218,20 @@ Stable by adding this to your Emacs initialization:
   [GNU ELPA repository](https://elpa.gnu.org/). It's the only package repository
   enabled by default in Emacs and you should not disable it!**
 
-### Setting up CIDER's nREPL middleware
+### CIDER's nREPL middleware
 
 Much of CIDER's functionality depends on the presence of CIDER's own
-[nREPL middleware][cider-nrepl].
+[nREPL middleware][cider-nrepl]. When `cider-jack-in` (<kbd>C-c M-j</kbd>) is used CIDER modifies the CLI starting up the REPL to inject its own dependencies. Both Leiningen and Boot powered projects are supported.
```
bbatsov: Modifies the CLI? I thought we just passed a few params using the CLI.

README.md
```diff
@@ -218,12 +218,20 @@ Stable by adding this to your Emacs initialization:
...
 
-#### Using Leiningen
+**`profiles.clj` or `profile.boot` don't need to be modified anymore for the above usecase!**
+
+If this behaviour is not preferred set `cider-inject-dependencies-at-jack-in` to nil. At `cider-jack-in` the CLI command is always shown in the minibuffer (or see it in the `*Messages*` buffer).
```
bbatsov: which behaviour is "this behaviour"? :-) I get what you mean, but the users might not get it.

README.md
```diff
@@ -218,12 +218,20 @@ Stable by adding this to your Emacs initialization:
...
+
+If a standalone REPL is preferred and `cider-connect` is used adding the middlewares are still needed as CIDER can not control dependencies of an independent REPL.
```
bbatsov: cannot
independent REPL -> embedded nREPL

cider.el
```diff
@@ -165,6 +165,11 @@ This variable is used by `cider-connect'."
   :type 'boolean
   :version "0.9.0")
 
+(defcustom cider-inject-dependencies-at-jack-in t
+  "When nil, do not inject repl dependencies (most likely nrepl middlewares) at jack in time."
```
bbatsov: nREPL

cider.el
```diff
@@ -204,6 +218,70 @@ Sub-match 1 must be the project path.")
...
+    (seq-concatenate
+     'list
+     (seq-map (lambda (dep) (format "update-in :dependencies conj \"[%s \\\"%s\\\"]\"" (car dep) (cadr dep))) (car repl-deps))
+     (seq-map (lambda (plugin) (format "update-in :plugins conj \"[%s \\\"%s\\\"]\"" (car plugin) (cadr plugin))) (cadr repl-deps)))
+    " -- ")
+   " -- "
+   params))
+
+(defun cider-add-repl-dependencies (dependencies lein-plugins middlewares)
+  "Extension point to add REPL dependencies to `cider-repl-dependencies' so they are injected from the CLI at cider jack in."
```
bbatsov: ``cider-jack-in'`

README.md
benedekfazekas: hm well, I guess by CLI I meant the specific command (with params and all) which is quite hairy for lein (as you need to conj/assoc stuff into the project map) and quite long (but nicer) for boot. I am happy to iterate on this text but if you want to propose your own version don't hold it back ;)

Malabarba: How about this?

> When `cider-jack-in` (<kbd>C-c M-j</kbd>) is used, CIDER takes care of that automatically.

Most users don't care what's happening underneath. Those who do can read the stuff below.

cider.el
bbatsov: all the params should appear uppercased in the docstring.

cider.el
```diff
@@ -204,6 +218,70 @@ Sub-match 1 must be the project path.")
...
+       ,(if (not lein-plugins)
+            orig-plugins
+          (seq-concatenate 'list orig-plugins lein-plugins))
+       ,(if (not middlewares)
+            orig-middlewares
+          (seq-concatenate 'list orig-middlewares middlewares))))))
+
+(defun cider-jack-in-dependencies (params project-type)
+  "Modify the params so that REPL dependencies as set in `cider-repl-dependencies' are injected from the CLI according to the used `project-type'.
+Eliminates the need for hacking profiles.clj or the boot script for supporting cider with its nrepl middleware."
```
bbatsov: Once again - the params should be denoted using the Elisp conventions.

bbatsov:
Just to be clear, as it's kind of hard to figure out what the code does without tests - this simply passes more params to `lein` and `boot`, right? 

I think it should also account for `cider-jack-in-clojurescript` as there we have additional params (e.g. piggieback).

README.md
Malabarba: I think we can reduce both sentences to this:

> If you don't want `cider-jack-in` to inject dependencies automatically, set `cider-inject-dependencies-at-jack-in` to nil. 

The part about minibuffer/Messages buffer doesn't seem super relevant to the current instructions.

README.md
Malabarba: I would recommend:

> If a standalone REPL is preferred, you need to invoke `cider-connect` (instead of `jack-in`) and you'll need to manually add the dependencies to your Clojure project (explained in the following section).

cider.el
Malabarba: Me too.

cider.el
Malabarba: If we're going to split the variable into 3, maybe we don't even need this function anymore.

cider.el
Malabarba: Oh, and we should definitely mark these variables as risky-local.

cider.el
```diff
@@ -204,6 +218,70 @@ Sub-match 1 must be the project path.")
     ("boot" cider-boot-parameters)
     ("gradle" cider-gradle-parameters)))
 
+(defun boot-command-prefix (dependencies)
+  (concat
+   (mapconcat
+    'identity
+    (seq-map (lambda (dep) (format "-d %s:%s" (car dep) (cadr dep))) dependencies)
```
Malabarba: This `seq-map` can be removed by replacing that `'identity` above with this `(lambda ...)`.
Same thing for the ones below.

cider.el
```diff
@@ -204,6 +218,70 @@ Sub-match 1 must be the project path.")
...
+    " ")
+   " "))
+
+(defun boot-repl-task-params (params middlewares)
+  (replace-regexp-in-string
```
Malabarba: I don't quite understand this. Why do we need to call `replace-regexp-in-string` on a string we're building ourselves?

Malabarba:
Thanks for working on this Benedek. 

I agree that the functions that manipulate the parameters need tests. Right now, it's a little hard to guess what they're supposed to do, so a couple of tests for each of them are sure to be helpful in the future.

cider.el
benedekfazekas: so split into 2 or 3?

bbatsov: 3.

cider.el
benedekfazekas: I may be missing something but `params` can be set by the user to any arbitrary string.

bbatsov:
@benedekfazekas Btw, just to be clear - this will take precedence over something in `profiles.clj`, right?

benedekfazekas:
> Just to be clear, as it's kind of hard to figure out what the code does without tests - this simply passes more params to lein and boot, right?

if you jack in you should see what it generates to start up the repl. if there is additional params (for cljs): code only modifies params so it should be fine. but you are right I deffo need to test it.

> Btw, just to be clear - this will take precedence over something in profiles.clj, right?

I assumed so but noted: need to explicitly test this too.

benedekfazekas:
the documentation https://github.com/technomancy/leiningen/blob/master/doc/PROFILES.md#declaring-profiles is not decisive.

bbatsov:
Guess you'll have to test it and see what happens.

cider.el
bbatsov: I've fixed the other incorrect attributes in this file.

benedekfazekas:
thanks for the review, I *think* I made all the changes requested.

Tested precedence: CLI params win over `~/lein/profiles.clj`.

Postponed syncing with master as master build seems to be broken atm.

benedekfazekas:
it seems I have the same test failure as master. suppose I still wait until that is fixed on master...

bbatsov:
> Postponed syncing with master as master build seems to be broken atm.

Actually it seems the fucked up tests are broken. I'm 100% sure that the code I wrote is correct, but the I'm not nearly as sure about the odd logic of the failing 2 tests. I'll likely just delete them...

cider.el
```diff
@@ -177,6 +182,18 @@ Sub-match 1 must be the project path.")
 (defvar cider-host-history nil
   "Completion history for connection hosts.")
 
+(defvar cider-jack-in-dependencies
+  '(("org.clojure/tools.nrepl" "0.2.12"))
+  "Clojure artifacts for starting the REPL. List of dependencies where elements are lists of artifact name and version.")
```
bbatsov: All 3 docstring start in more or less the same manner. This looks like some mistake to me.

cider.el
```diff
@@ -177,6 +182,18 @@ Sub-match 1 must be the project path.")
...
+
+(defvar cider-jack-in-lein-plugins
+  `(("cider/cider-nrepl" ,(upcase cider-version)))
+  "List of Clojure artifacts for starting the REPL. List of Leiningen plugins where elements are lists of artifact name and version.")
+
+(defvar cider-jack-in-nrepl-middlewares
+  '("cider.nrepl/cider-middleware")
```
bbatsov: This is not a middleware name - it's actually a list of all of CIDER's middlewares. The docstring seems misleading.

cider.el
benedekfazekas: Haha that was intentional. I guess my semi-conscious attempt to group them. happy to remove the first sentence if you'd like.

cider.el
benedekfazekas: what is it then? can't really call it `cider-jack-in-nrepl-middlewares-lists`, or?

Malabarba: I think he just complained about the docstring, not the variable name.

Malabarba: Maybe change to something like this:

"List of Clojure variable names.
Each of these Clojure variables should hold a vector of nREPL middlewares."

bbatsov: Yeah, that's what I meant.

test/cider-tests.el
```diff
@@ -655,3 +655,17 @@
...
   (should (equal (cider-project-name "path/to/project/") "project")))
+
+(ert-deftest cider-inject-jack-in-dependencies ()
+  (should (string= (cider-inject-jack-in-dependencies "repl :headless" "lein")
+                   "update-in :dependencies conj \"[org.clojure/tools.nrepl \\\"0.2.12\\\"]\" -- update-in :plugins conj \"[cider/cider-nrepl \\\"0.11.0-SNAPSHOT\\\"]\" -- repl :headless"))
+  (should (string= (cider-inject-jack-in-dependencies "repl -s wait" "boot")
+                   "-d org.clojure/tools.nrepl:0.2.12 -d cider/cider-nrepl:0.11.0-SNAPSHOT repl -m cider.nrepl/cider-middleware -s wait")))
+
+(ert-deftest cider-inject-jack-in-dependencies-add-refactor-nrepl ()
+  (setq cider-jack-in-lein-plugins (seq-concatenate 'list cider-jack-in-lein-plugins '(("refactor-nrepl" "2.0.0"))))
```
bbatsov: Use `let` instead of `setq`. Those are dynamic variables and you don't want the tests to affect the defaults.

test/cider-tests.el
```diff
@@ -655,3 +655,17 @@
   (should (equal (cider-project-name "") "-"))
   (should (equal (cider-project-name "path/to/project") "project"))
   (should (equal (cider-project-name "path/to/project/") "project")))
+
+(ert-deftest cider-inject-jack-in-dependencies ()
+  (should (string= (cider-inject-jack-in-dependencies "repl :headless" "lein")
```
bbatsov: I'd suggest stubbing the version and having one test for a snapshot version and one test for a stable version.

bbatsov:
:+1:

benedekfazekas:
Woot! :)

phillord:
Thank you all. This will have a big impact on the usability of cider, I think.

benedekfazekas:
I hope you are right. As I said on twitter thx for bringing this up.

Issue #1541, opened 2016-02-04
IDO/helm support for apropos


martintrojer:
Scenario; looking for a function somewhere in the codebase. All I have is the name.

`M-x cider-apropos` does what I want, but I would prefer the results to feed ido/helm as I am typing. Then when a var is selected, a buffer is opened at the correct line.

Also discussed here #1059

bbatsov:
Yeah, we should cook something like this. Btw, you can add different action keys to `ido` so there can be options to go to the source, view the doc, conj.io, etc. 

We'll implement something using completing-read and the apropos results and people will be able to pick by themselves what completion system is used.

xiongtx:
Proof of concept for `cider-apropos` using Helm:

```emacs-lisp
(defun cider-def-source--apropos (query)
  (helm-build-sync-source (concat "CIDER apropos: '" query "'")
    :action '(("Lookup doc" . cider-doc-lookup))
    :candidates (mapcar (lambda (dict)
                          (nrepl-dbind-response dict (name type doc)
                            (let* ((label (capitalize (if (string= type "variable") "var" type)))
                                   (help (concat "Display doc for this " (downcase label))))
                              (with-temp-buffer                                
                                (cider-propertize-region (list 'apropos-symbol name
                                                               'action 'cider-apropos-doc
                                                               'help-echo help)
                                  (insert-text-button name 'type 'apropos-symbol)
                                  (insert "\n") 
                                  (insert-text-button label 'type (intern (concat "apropos-" type)))
                                  (insert ": ") 
                                  (let ((beg (point)))
                                    (insert doc "\n")
                                    (fill-region beg (point-max)))) 
                                (cons (buffer-string)
                                      name)))))
                        (cider-sync-request:apropos query))
    :multiline t))

(helm :sources (cider-def-source--apropos "def")
      :buffer "*Helm CIDER apropos*")
```

bbatsov:
> Would it be possible to have a function that takes a name like clojure.core/defn and jump to the source file?

It already exists - `cider-find-var`.

bbatsov:
I've just added a couple of new commands - `cider-apropos-select` and `cider-apropos-documentation-select`. Play with them and let me know what you think.

martintrojer:
This works great.

One suggestion, I'd love to give a default selection for `cider-apropos-act-on-symbol`. In my case `find-def`.

bbatsov:
[...]

Issue #1411, opened 2015-11-11
Add some Hydras to improve discoverability


expez:
The [hydra](https://github.com/abo-abo/hydra) package is pretty easy to use and it can make a huge difference to usability.

xiongtx:
How should this be done?

* A single file of hydras would have to `require` almost all the other files in the project. Is that a good idea?
* Separate files of hydra could result in a lot of files. 
    * Maybe they can go into their own `/hydra` directory?
* Putting hydras in existing files is probably a bad idea because it spreads the dependency on the `hydra` package throughout the code base.

xiongtx:
Here's something that could be included in a hypothetical `cider-hydra` file:

```emacs-lisp
(require 'hydra)
(require 'cider-doc)
(require 'cider-apropos)
(require 'cider-grimoire)


;;; Documentation

(defcustom cider-hydra-show-docstring nil
  "If t, show docstrings for commands in CIDER hydras.")

(defun cider-hydra--docstring-first-line (symbol)
  "Get the first line of SYMBOL's docstring."
  (let ((doc (documentation symbol)))
    (substring doc 0 (string-match "\n" doc))))

(defun cider-hydra--show-docstring-maybe (symbol)
  "Maybe get and return the first line of SYMBOL's docstring.

The docstring is returned if `cider-hydra-show-docstring' is non-nil. A
newline character and some spaces are prepended to the docstring for
formatting."
  (if cider-hydra-show-docstring
      (concat "\n" (make-string 6 ?\s) (cider-hydra--docstring-first-line symbol))
    ""))

(defhydra cider-hydra-doc (:color blue)
  "
CIDER Documentation                                                [_q_] quit
^^---------------------------------------------------------------------------
[_d_] cider-doc %s(cider-hydra--show-docstring-maybe 'cider-doc)

Javadoc
^^---------------------------------------------------------------------------
[_j_] cider-javadoc %s(cider-hydra--show-docstring-maybe 'cider-javadoc)

Apropos
^^---------------------------------------------------------------------------
[_a_] cider-apropos %s(cider-hydra--show-docstring-maybe 'cider-apropos)
[_A_] cider-apropos-documentation %s(cider-hydra--show-docstring-maybe 'cider-apropos-documentation)

Grimoire
^^---------------------------------------------------------------------------
[_r_] cider-grimoire  %s(cider-hydra--show-docstring-maybe 'cider-grimoire)
[...]

PR #1554, opened 2016-02-11
Add hydras for CIDER.

xiongtx:
Added some CIDER hydras based on the [refcard](https://github.com/clojure-emacs/cider/blob/master/doc/cider-refcard.pdf).

**Please ignore the branch name**. This is for #1411, not #1541.

### DIFF
--- a/cider-hydra.el
+++ b/cider-hydra.el
@@ -1,0 +1,263 @@
+;;; cider-hydra.el --- Hydras for CIDER -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2016 Tianxiang Xiong
+
+;; Author: Tianxiang Xiong <tianxiang.xiong@gmail.com>
+;; Keywords: convenience, help
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This file defines some hydras (pop-up menus of commands with common
+;; prefixes) for CIDER. For more information on the `hydra' package, see
+;; https://github.com/abo-abo/hydra
+
+;; Hydras serve several important purposes: discovery, memorization, and
+;; organization.
+
+;; - Discovery
+
+;;   - Grouping related commands together under a common prefix and
+;;     displaying them in a single menu facilitates discovery.
+
+;;   - For example, if a user wants to know about CIDER's documentation
+;;     commands, they could bring up a hydra that includes commands like
+;;     `cider-doc', `cider-javadoc', etc, some of which may be new to them.
+
+;; - Memorization
+
+;;   - Hydras serve as a memory aid for the user. By grouping related
+;;     commands together, the user has less need to memorize every command;
+;;     knowing one, she can find the others.
+
+;; - Organization
+
+;;   - The process of creating hydras can aid in organizing code. This
+;;     gives both developers and users a better overview of what the
+;;     project is can or cannot do.
+;;
+;;   - Thus, each hydra is like a section of a quick-reference card. In
+;;     fact, many of the hydras here are inspired by the CIDER refcard:
+;;     https://github.com/clojure-emacs/cider/blob/master/doc/cider-refcard.pdf
+
+;;; Code:
+
+;; For documentation
+(require 'cider-apropos)
+(require 'cider-doc)
+(require 'cider-grimoire)
+(require 'hydra)
+
+;; For loading and evaluation
+(require 'cider-macroexpansion)
+
+;; For debugging and testing
+(require 'cider-test)
+
+;; For REPL
+(require 'cider-client)
+(require 'cider-interaction)
+(require 'cider-mode)
+(require 'cider-repl)
+
+;;; Customize
+
+(defgroup cider-hydra nil
+  "Hydras for CIDER."
+  :prefix "cider-hydra-"
+  :group 'cider)
+
+(defcustom cider-hydra-show-docstring nil
+  "If non-nil, show docstrings for commands in CIDER hydras."
+  :type 'boolean)
+
+;;; Functions to get symbol docstrings
+
+(defun cider-hydra--docstring-first-line (symbol)
+  "Get the first line of SYMBOL's docstring."
+  (let ((doc (documentation symbol)))
+    (substring doc 0 (string-match "\n" doc))))
+
+(defun cider-hydra--show-docstring-maybe (symbol)
+  "Maybe get and return the first line of SYMBOL's docstring.
+
+The docstring is returned if `cider-hydra-show-docstring' is non-nil. A
+newline character and some spaces are prepended to the docstring for
+formatting."
+  (if cider-hydra-show-docstring
+      (concat "\n" (make-string 6 ?\s) (cider-hydra--docstring-first-line symbol))
+    (identity "")))
+
+;;; Documentation
+
+(defhydra cider-hydra-doc (:color blue)
+  "
+CiderDoc                                                           [_q_] quit
+^^---------------------------------------------------------------------------
+[_d_] cider-doc %s(cider-hydra--show-docstring-maybe #'cider-doc)
+
+JavaDoc
+^^---------------------------------------------------------------------------
+[_j_] cider-javadoc %s(cider-hydra--show-docstring-maybe #'cider-javadoc)
+
+CIDER Apropos
+^^---------------------------------------------------------------------------
+[_a_] cider-apropos %s(cider-hydra--show-docstring-maybe #'cider-apropos)
+[_A_] cider-apropos-documentation %s(cider-hydra--show-docstring-maybe #'cider-apropos-documentation)
+
+Grimoire
+^^---------------------------------------------------------------------------
+[_r_] cider-grimoire  %s(cider-hydra--show-docstring-maybe #'cider-grimoire)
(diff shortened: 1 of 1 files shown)


bbatsov:
I was just reminded that we need update the refcard (recently we made changes to the test command mappings). :-)

There's one thing we need to consider - this file obviously depends on the `hydra` package, but CIDER itself doesn't depend on it. The way I see it we can have a separate repo `cider-hydra` where we can put the hydras, but I'm guessing there are other options. @expez How did you organize this in `clj-refactor.el`?

We'll also need an entry in the documentation for this and ideally some screenshots (or a short animated gif), showcasing this functionality.

xiongtx:
Yes, how to include the hydras is [an issue I've raised](https://github.com/clojure-emacs/cider/issues/1411#issuecomment-182598466).

[clj-refactor](https://github.com/clojure-emacs/clj-refactor.el/blob/master/clj-refactor.el) depends directly on the `hydra` package. Looking at the code, `hydra` seems to be the main way to interact with `clj-refactor`.

Malabarba:
> Having cider-hydra as a different package may be OK.

I'm fine with a separate package, but we can also do this:

    (defmacro cider-defhydra (&rest args)
      `(eval-after-load 'hydra
         '(defhydra ,@args)))

This way we only use hydras if the user installs the package.

> I'm also not sure if we should have a consistent hydra style across projects. 

CIDER and clj-refactor organize their commands in different ways. I think we're better off focusing on what makes the most sense to us (of course, we can look at their hydras for inspiration 😊).

> CIDER hangs without jacking in

That's probably because of the sync request. We should `-ensure-connected`, but that has nothing to do with this PR.
Or did you mean that this *only* happens when called from a hydra?

cider-hydra.el
```diff
@@ -0,0 +1,263 @@
...
+
+(defun cider-hydra--show-docstring-maybe (symbol)
+  "Maybe get and return the first line of SYMBOL's docstring.
+
+The docstring is returned if `cider-hydra-show-docstring' is non-nil. A
+newline character and some spaces are prepended to the docstring for
+formatting."
+  (if cider-hydra-show-docstring
+      (concat "\n" (make-string 6 ?\s) (cider-hydra--docstring-first-line symbol))
+    (identity "")))
```
Malabarba: a.k.a.: `""`. ;-)

cider-hydra.el
```diff
@@ -0,0 +1,263 @@
...
+

+;;; Documentation
+
+(defhydra cider-hydra-doc (:color blue)
+  "
+CiderDoc                                                           [_q_] quit
```
Malabarba: Personally, I think these Header-like titles are overkill because:
1. Each "group" only has 1 or 2 entries. 
2. Each entry already has the title in the command name.

Malabarba:
TBH, I had a different notion of what our hydras would be. 

I imagined them more compact, with multiple keys per line. 
I also thought they'd take the place of our regular keymaps, like `C-c C-d` would invoke the documentation hydra, and `C-c ,` (or `C-c C-t` now) would invoke the tests hydra.

I'm not saying the proposed structure is bad, though.

cider-hydra.el
```diff
@@ -0,0 +1,263 @@
...
+  ("h" cider-grimoire-web nil)
+  ;; Quit hydra
+  ("q" nil nil))
+

+;;; Loading and evaluation
+
+(defhydra cider-hydra-eval (:color blue)
+  "
+CIDER Load                                                         [_q_] quit
```
Malabarba: Because our hydras are one-shot (blue, instead of red) I think we don't need to advertise a `quit` command. Good-old C-g will work as usual.

cider-hydra.el
```diff
@@ -0,0 +1,263 @@
...

+;;; Debugging and testing
+
+(defhydra cider-hydra-debug (:color blue)
+  "
+CIDER Eval                                                         [_q_] quit
+^^---------------------------------------------------------------------------
+[_x_] cider-eval-defun-at-point %s(cider-hydra--show-docstring-maybe #'cider-eval-defun-at-point)
+
+Debugging
```
Malabarba: We should Call this Tracing. It's confusing with the debugger.

bbatsov:
> That's probably because of the sync request. We should -ensure-connected, but that has nothing to do with this PR.

A ton of interactive commands are missing `(cider-ensure-connected)` checks. They should hang though, but raise some nil exception. 

> clj-refactor
> 
> clj-refactor depends directly on the hydra package. Looking at the code, hydra seems to be the main way to interact with clj-refactor.

clj-refactor depends on a ton of packages, which IMO is not very wise. I've always been a believer that a package should be completely useful without the bell and whistles provided by something like `multiple-cursors` and `hydra` for instance. You'll notice that CIDER is a much bigger project with just a handful of dependencies (and most of them are GNU ELPA).

> Consistency of hydras
> 
> I'm also not sure if we should have a consistent hydra style across projects. These things are a crapshoot as far as I can see. clj-refactor does things a certain way; does that mean we should do it that way as well?

I'd love some consistency, but afaik @magnars and @expez don't care much for it. The keybindings employed by `clj-refactor` are not exactly common Emacs practice. 

I'll create a `cider-hydra` repo and you can file this PR there.

bbatsov:
The new repo is here https://github.com/clojure-emacs/cider-hydra

bbatsov:
You can use this README https://github.com/bbatsov/super-save as the template for the new project's README. 

P.S. I'm not contributing much to the discussion about the structure of the hydras, because I've yet to try the hydra package.

xiongtx:
@Malabarba 

> That's probably because of the sync request. We should -ensure-connected, but that has nothing to do with this PR. Or did you mean that this only happens when called from a hydra?

No, this happens generally, as I mentioned in the [previous comment](https://github.com/clojure-emacs/cider/pull/1554#issuecomment-182747999). **The hydras have nothing to do with the hanging**; all the hydras do is provide a new way of invoking the commands.

> TBH, I had a different notion of what our hydras would be.

Yes, this is a first attempt based on the refcard. We can break the hydras apart, format them differently, include/exclude commands, etc. It'd be good if we can get more feedback from users, but that's an issue with software development in general.

You may have been imagining something more like `clj-refactor`'s hydras. This ties into the issue of consistency, which is not primarily about how we format the hydra strings, but **what the purpose of the hydras are**.

* Are they the primary means of interacting with CIDER? 
  * If so, they should be more comprehensive. There are lots of commands that are not included.
* Are they trying to abstract away the Emacs details?
  * If so, we should use English words instead of command names (e.g. `Run tests` instead of `cider-test-run-tests`).
* Are they training wheels for new users to learn more about CIDER, with the expectation that they will transition away from the hydras and learn to invoke commands directly?
  * If so, we *should* display the command names, possibly along with keybindings and the first docstring line, to quicken familiarity with CIDER's capabilities.

xiongtx:
@bbatsov 


> A ton of interactive commands are missing (cider-ensure-connected) checks. They should hang though, but raise some nil exception.

@bbatsov They *should* hang? What do you mean? Invoking a command should not render Emacs unusable.

> clj-refactor depends on a ton of packages, which IMO is not very wise.

@bbatsov It's probably a good idea not to have a project dependency on an end-user oriented package like [multiple-cursors](https://github.com/magnars/multiple-cursors.el), but `hydra` is more like a library.

Nonetheless, I think you're right. Not having a package manager or CPAN-like central respository for most of Emacs' 30+ year existence has created a culture of everyone making their packages "batteries-included", with all the good and bad (inconsistently re-inventing the wheel being the most obvious "bad") that comes from that.

I'll push to the new [cider-hydra](https://github.com/clojure-emacs/cider-hydra) repo when I get a chance. We can work from there.

bbatsov:
> They should hang? What do you mean? Invoking a command should not render Emacs unusable.

Should not hang. The amount of silly mistakes I make while typing is always legendary...

cider-hydra.el
xiongtx: I agree.

cider-hydra.el
xiongtx: We can get ride of the header lines, but what if we have more commands? 

For something like `eval`, in a different hydra, there could be ~5-10 commands. Maybe that's a sign we should break the hydras apart into smaller hydras.

cider-hydra.el
xiongtx: I disagree. An obvious `quit` option is useful, especially to new users.

It's also possible that not all hydras *should* be blue. Some of the commands that operate on regions or the last sexp may make more sense as [pink](https://github.com/abo-abo/hydra#color) hydras.

cider-hydra.el
xiongtx: I agree.

We should also move the lone `cider-eval-defun-at-point` into the `eval` hydra. I included it here because that's how the refcard grouped it, but upon closer scrutiny it doesn't seem to belong.

Malabarba:
@xiongtx

> Yes, this is a first attempt based on the refcard. We can break the hydras apart, format them differently, include/exclude commands, etc. It'd be good if we can get more feedback from users, but that's an issue with software development in general.

I think we can start by basing them on our keymaps.

You may have been imagining something more like `clj-refactor`'s hydras. This ties into the issue of consistency, which is not primarily about how we format the hydra strings, but **what the purpose of the hydras are**.

> * Are they the **primary means of interacting with CIDER**? 
>   * If so, they should be more comprehensive. There are lots of commands that are not included.

That's fine. There are lots of commands not bound to keymaps too. They don't have to be. Some commands were added for the convenience of more advanced users with specific needs. They don't have to clutter other stuff.

> * Are they trying to **abstract away the Emacs details**?
>   * If so, we should use English words instead of command names (e.g. `Run tests` instead of `cider-test-run-tests`).

Yes, I was thinking of this too. We should do this.

> * Are they **training wheels for new users to learn more about CIDER, with the expectation that they will transition away from the hydras** and learn to invoke commands directly?
>   * If so, we *should* display the command names, possibly along with keybindings and the first docstring line, to quicken familiarity with CIDER's capabilities.

I think they can be decent training wheels without the docstrings, and then you don't even need the expectation that they will transition away (hydras can have a delay so they act as a normal keymap if you type the key fast enough).

I would prefer the docstring to be available on-demand (like if you type `h` inside a hydra), but that will probably be very troublesome to implement, as it would change the entire structure of the hydra. 
We can start small, with just some concise hydras that completely replace the current keymaps.

xiongtx:
Based on what @Malabarba has said, I'll spend some time designing the hydras before I submit them again.

bbatsov:
@xiongtx Fine by me. Submit this PR to the new project when ready.

PR #1558, opened 2016-02-12
Update kbd for running test in current buffer

annapawlicka:
`C-c ,` is no longer supported - this PR updates its README with the new kbd.


bbatsov:
:+1:

PR #1569, opened 2016-02-15
Position point correctly when debugging forms with metadata

grammati:
`cider--debug-move-point` would position the point incorrectly when
debugging a form with a metadata map attached via the reader, because
the call to `down-list` would descend into the metadata map.


Malabarba:
Thanks! I'm surprised I hadn't run into this yet.

PR #1575, opened 2016-02-17
Try a bit harder to find a boot executable

Malabarba:


bbatsov:
Who's using `boot.sh`?

Malabarba:
I was. 😁
boot.sh is the name of the executable you download from their site. Maybe I was the only one who didn't rename it.

bbatsov:
Fair enough then.

PR #1581, opened 2016-02-20
Add instructions for using Figwheel with jack-in

magnars:
Figwheel is very popular and should probably have its own section in the Readme. :)


benedekfazekas:
+1

Malabarba:
LGTM 👍

bbatsov:
:+1:

bbatsov:
Btw, it'd be nice if someone added similar instructions for boot. I assume the bulk of the setup is the same.

benedekfazekas:
i may be wrong but afaik you don't really need figwheel if you run with boot as it has very similar features for cljs.

PR #1590, opened 2016-02-26
Fixes #1578 in windows process-buffer can be nil

### LINKED ISSUE #1578: nrepl-server-filter called, even though process buffer is dead
Trying to debug #390. This may or may not be related:

I often see an error when I kill CIDER in windows using `cider-quit`: 

````
Debugger entered--Lisp error: (error "Selecting deleted buffer"))
nrepl-server-filter(#<process nrepl-server> "^CTerminate batch job (Y/N)? ")
````

This happens inside of `nrepl-server-filter`, specifically the line:

````
(with-current-buffer (process-buffer process))
````

TBH my knowledge of process buffers and sentinels etc is a bit naive. But it seems to me that CIDER kills the process buffer, yet this server-filter is still hanging around to intercept output from the nREPL process which is still on-going, in particular the output "Terminate batch job (Y/N)", which always happens on windows when you try to exit from batch jobs (.bat files).

So my question is, a) can you help me understand how this nrepl-server-filter is being called for a process with a dead corresponding process buffer, and b) could CIDER have some defensive code? I.e. don't call `with-current-buffer` on a potentially dead buffer.

### PR DESCRIPTION
**Pull request by jonpither**
Adding a simple when condition, so that the spurious error in windows goes away.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1110,30 +1110,33 @@ the port, and the client buffer."
 
 (defun nrepl-server-filter (process output)
   "Process nREPL server output from PROCESS contained in OUTPUT."
-  (with-current-buffer (process-buffer process)
-    ;; auto-scroll on new output
-    (let ((moving (= (point) (process-mark process))))
-      (save-excursion
-        (goto-char (process-mark process))
-        (insert output)
-        (set-marker (process-mark process) (point)))
-      (when moving
-        (goto-char (process-mark process))
-        (when-let ((win (get-buffer-window)))
-          (set-window-point win (point))))))
-  ;; detect the port the server is listening on from its output
-  (when (string-match "nREPL server started on port \\([0-9]+\\)" output)
-    (let ((port (string-to-number (match-string 1 output))))
-      (message "nREPL server started on %s" port)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Malabarba** [comment]
LGTM

File: nrepl-client.el [hunk 1]
```diff
@@ -1111,30 +1111,32 @@ the port, and the client buffer."
...
-        (let* ((client-proc (nrepl-start-client-process nil port process))
-               (client-buffer (process-buffer client-proc)))
-          (setq nrepl-client-buffers
-                (cons client-buffer
-                      (delete client-buffer nrepl-client-buffers)))
-
-          (when (functionp nrepl-post-client-callback)
-            (funcall nrepl-post-client-callback client-buffer)))))))
+  ;; In Windows this can be false:
+  (when (buffer-live-p (process-buffer process))
```
bbatsov: You can extract `(process-buffer)` in a let binding (it's used again on the next line).

bbatsov:
Please, change the commit message to "[Fix #1578] Add a guard for missing process-buffer"

jonpither:
Changes made.

bbatsov:
:+1: Thanks!

PR #1591, opened 2016-02-26
Conditional break

grammati:
Incomplete, but this is the cider code to accompany https://github.com/clojure-emacs/cider-nrepl/pull/303


Malabarba:
Looks like you have an accidental commit by @bbatsov in there. It probably just means he rewrote history on you and now you'll have to rebase. ;-)

Malabarba:
I'm not totally happy with this interface. It's ok for the cider-nrepl code to try to be friendly and do-the-right-thing when the user adds the metadata to the `defn` instead of the body.

However, I'm not supper happy with this being our only exposed interface. It only works with `defn`, and users are almost guaranteed to get exceptions when they try to use it with something else, lika a `defmethod`, `defmacro`, `proxy`, etc.

I would prefer something that inserted the `^{...}` in the right place. It doesn't have to be a “fake” string like we do with the `#dbg` (which is also something I'm thiking of changing). I'd be fine with actually inserting this string into the buffer.

cider-interaction.el
```diff
@@ -1121,11 +1121,18 @@ command `cider-debug-defun-at-point'."
                             "  \nWould you like to read the Feature Request?"))
       (browse-url "https://github.com/clojure-emacs/cider/issues/1416"))
     (user-error "The debugger does not support ClojureScript"))
-  (cider-interactive-eval
-   (concat (if debug-it "#dbg\n")
-           (cider-defun-at-point))
-   nil
-   (cider-defun-at-point 'bounds)))
+  (let* ((conditional? (equal debug-it '(16)))
```
bbatsov: Names ending with `?` are not common in Elisp.

bbatsov:
Apart from @Malabarba's comments - this functionality should also be mention in the changelog and the README.

grammati:
@Malabarba, thanks for reviewing!

Good point about `deftest`, `defmethod`, etc. It seems there needs to be a way to extend the magic behavior of `defn` so that conditions just work as users would expect, on various types of def-producing forms. That special logic could be in cider, or in cider-nrepl. If we care about making cider-nrepl front-end agnostic, so that other editors can use the debugger, it seems to me that it's better to put it in the backend. I'm not sure how much of a concern that is, since as far as I know only cider uses the cider-nrepl debugger so far.

If you actually insert metadata as text into the buffer, do you envision having a quick and easy way to undo that insertion? I can see myself accidentally pushing code to git with the inserted conditions.

Also, I often meta-dot my way into functions inside third-party jars, and then debug functions that I find there. These buffers are read-only. I would have to make them writable to insert breaks and conditions. That's why I stuck with the approach of just prepending the `#dbg ^{...}` to the form before evaling it.

Malabarba:
> If you actually insert metadata as text into the buffer, do you envision having a quick and easy way to undo that insertion? I can see myself accidentally pushing code to git with the inserted conditions.

We can put an overlay on top of it. The overlay would use some very hard-to-miss face, and when you delete any of the characters in the overlay then the whole thing would delete itself. 
I think that would address both issues. 

>Also, I often meta-dot my way into functions inside third-party jars, and then debug functions that I find there. These buffers are read-only.

We can inhibit-read-only for this. 

This is all up for discussion, of course. If @bbatsov or @expez would like to pitch in too.

bbatsov:
@Malabarba's suggestion sounds good to me.

Malabarba:
I've pushed an example of what I meant, should be on Melpa soon. This doesn't do conditional breakpoints yet. It just changes the current `C-u C-M-x` behavior to insert a litteral (and fragile) `#dbg` in the buffer.

bbatsov:
Any progress here?

Malabarba:
I should say, this is probably going to move slowly, as I want to get in right. 
Currently, I think I'll revert my recent change, and then I'll set it so that a double prefix inserts a #dbg and a conditional on the sexp at point (similar to Chris' original suggestion, but inserting at point, instead of before the function).

grammati:
I would still like the debugger to "do the right thing" when given a conditional on a top-level form, but I have realized that my original approach of moving the condition down in the post-macroexpanded form is more complicated than I originally thought.

Overall I see 4-ish options:

1. Do nothing, close this PR. Require the user to type in conditionals in the right place.
2. Keep this PR, send conditionals prefixed to the top-level form. Make the debugger "dwim" by moving the conditions down to the right place in the form:
  - a) before macroexpansion
  - b) after macroexpansion
3. Write elisp code to figure out the right place in the top-level form and insert the `^{:break/when ...}` literally into the buffer (if I understood correctly, this is what @Malabarba was suggesting above).

My original PR to cider-nrepl had a naive implementation of 2b, but it was removed. That's good, because I now realize that it really didn't work in any but the simplest of cases.

Example: moving the metadata post-macroexpansion transformed this:

```clojure
^{:break/when (pos? x)}
(defn foo [x]
  x)
```

into this:

```clojure
(def foo
  (fn* ([x]
    ^{:break/when (pos? x)}
    x)))
```

which is fine, and worked in this case.

But when it gets even a little more complicated, such as destructuring in the argument, it transformed this:

```clojure
^{:break/when (pos? x)}
(defn foo [[x]]
  x)
```

into this:

```clojure
(def foo
 (fn*
   ([p__20146]
     ^{:break/when (pos? x)}
     (let* [vec__20147 p__20146 x (nth vec__20147 0 nil)]
       x))))
```

This fails, because `x` is not in scope yet where the condition has been placed. In this case we could probably deduce that the `let*` form was introduced by the macroexpansion of the the `fn`, and not typed in by the user, and move the condition down one more level.

Also, what about multi-arity functions? The original PR would just plop the conditional into each of the bodies, transforming this:

```clojure
^{:break/when (pos? x)}
(defn bar
  ([] (bar 3))
  ([x] x))
```

into this:

```clojure
(def bar
  (fn* 
    ([] ^{:break/when (pos? x)} (bar 3)) 
    ([x] ^{:break/when (pos? x)} x)))
```

which fails because there is no `x` in the arity-0 case.

Alternatively, we could try to move the condition down before macroexpansion (option 2a above). This would probably be reasonably easy to do for well known forms like `defn`. But the number of possible toplevel forms is open-ended - how could we know what to do with `defblah` or `defthingy` forms?

At least when working with the post-macroexpaned form, there is a closed world of possible forms to deal with - everything expands into the primitives of clojure: `def*`, `fn*`, `deftype*`, etc.

So overall, I'm not sure it's worth the complication in the back end, but I'm happy to spend some time (when I find some) to explore further.

grammati:
Ha, @Malabarba, I should have just waited for you to reply! I'm +1000 on that. Just placing the point where you need the conditional is so much easier than doing complicated shenanigans to move around metadata. Good call.

Malabarba:
Glad we're all on board. 
I'm fine with also doing option 2a, it would just need to be a little more general (by applying to any form, not just defs) and more careful (not throwing exceptions if a local is not defined).

bbatsov:
I'd just like to remind everyone that I plan to release 0.12 before Clojure/west and it'd be nice if we wrapped this before the release.

Malabarba:
I've implemented a version of this already under the double prefix argument. This PR can be closed. 

I'd love to hear opinions and suggestions on the current solution. I don't have any further enhancements planned for 0.12.

bbatsov:
@Malabarba Guess you should also update the documentation regarding this.

PR #1603, opened 2016-03-06
Exclude Clojure dependency when injecting deps

arrdem:
This change allows users to make use of Clojure forks, without having CIDER clobber everything when starting a REPL.

Maybe a more general structure such as a `cider-excluded-dependencies` var is in order, but it's worth talking about and this is what I'm rolling locally.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -183,6 +183,12 @@ This variable is used by `cider-connect'."
   "Regexp list to extract project paths from output of `cider-ps-running-nrepls-command'.
 Sub-match 1 must be the project path.")
 
+(defcustom cider-jack-in-exclusions nil
+  "List of artifact names which will be excluded when injecting dependencies."
+  :type '(repeat string)
+  :group 'cider
+  :package-version '(cider . "0.12.0"))
+
 (defvar cider-host-history nil
   "Completion history for connection hosts.")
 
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm fine with the change, but let's really introduce something like `cider-excluded-dependencies`, as I'm not fond of hardcoded stuff.

bbatsov:
Btw, feel free to add some notes to the readme about running cider with clojure forks. That will be useful.

cider.el
```diff
@@ -183,6 +183,12 @@ This variable is used by `cider-connect'."
   "Regexp list to extract project paths from output of `cider-ps-running-nrepls-command'.
 Sub-match 1 must be the project path.")
 
+(defcustom cider-jack-in-exclusions nil
+  "List of artifact names which will be excluded when injecting dependencies."
+  :type '(string)
```
Malabarba: This should be `:type '(repeat string)`.

cider.el
```diff
@@ -258,11 +264,20 @@ string is quoted for passing as argument to an inferior shell."
...
 (defun cider--list-as-lein-artifact (list)
   "Return an artifact string described by the elements of LIST.
 LIST should have the form (ARTIFACT-NAME ARTIFACT-VERSION).  The returned
 string is quoted for passing as argument to an inferior shell."
-  (shell-quote-argument (format "[%s %S]" (car list) (cadr list))))
+  (shell-quote-argument
+   (format "[%s %S%s]"
+           (car list)
+           (cadr list)
+           (cider--list-as-lein-exclusions cider-jack-in-exclusions))))
```
Malabarba: Since we're adding functionality to this function, could we extend the tests we have for this?

Malabarba:
Besides the 2 comments, I'm good with the PR.

bbatsov:
There should also be a couple of unit tests for this to check the command strings are generated properly.

arrdem:
Updated PR to fix `:type`. Sorry I'm newish to emacs hacking generally, but how would I go about testing with a custom binding?

arrdem:
So after stopping and thinking about this, this PR isn't needed. If a user has provided the `:exclusions` key in their `project.clj`, then those exclusions will apply both to the injected `tools.nrepl` dependency, and to the injected `cider/cider-nrepl` dependency. `cider/cider-nrepl` does not impose a dependency on Clojure, although it does check that `org.clojure/clojure` exists in the project's dependencies. Patching the elisp client like this won't therefore enable CIDER to work with Jaunt or whatever. Changes to `cider/cider-nrepl` are required instead. Consequently closing.

bbatsov:
Btw, isn't your other PR lein specific?

arrdem:
Yes the successor PR is Leiningen specific, however that is sufficient. For boot, cider-nrepl does not itself depend on Clojure and expects an acceptable version to be provided. The only reason that a change is required for Leiningen support is that the Leiningen plugin does do project.clj validation to check for the `org.clojure/clojure` artifact. No such check is performed with boot, so no compatibility patch seems to be required there.

Issue #800, opened 2014-09-14
Can't install cider using Cask / Melpa


nlamirault:
i try to install cider using Cask / Melpa : 

```
$ cask install
....
Dependency cider failed to install: Package `queue-0.1.1' is unavailable
Output:
```

No package named 'queue' is available from http://melpa.milkbox.net/#/
Any idea ?

PR #1605, opened 2016-03-06
Set instrumented and traced faces for terminal

grammati:
cider-instrumented-face and cider-traced-face used a :box attribute,
which does not display in terminal-mode emacs.

Use an underline and a background color instead.

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -347,14 +347,16 @@ The value can also be t, which means to font-lock as much as possible."
   :group 'cider)
 
 (defface cider-instrumented-face
-  '((t :box (:color "#c00" :line-width -1)))
+  '((((type graphic)) :box (:color "#c00" :line-width -1))
+    (t :underline t :background "#800"))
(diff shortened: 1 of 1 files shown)


Malabarba:
Thanks @grammati. But these kind of things can actually be specified as part of the defface spec.
For instance, the instrumented face should be

    (defface cider-instrumented-face
      '((((type graphic))
         :box (:color "#c00" :line-width -1))
        (t :underline t :background "#800"))
      "Face used to mark code being debugged."
      :group 'cider-debug
      :group 'cider

And similarly for the traced face.
I know that's a little confusing (and a lot of parentheses), but it's a little better than using a separate conditional. :)

grammati:
Thanks for the tip, @Malabarba. I pushed an update.

Malabarba:
LGTM. :+1: 
I'll let @bbatsov merge.

bbatsov:
:+1: Thanks!

PR #1610, opened 2016-03-08
Stop using reserved C-h keybinding for grimoire

grammati:
`C-c C-d C-h` was bound to `cider-grimoire-web`, but that should display the keybindings under `C-c C-d`.

`C-c C-d h` also calls `cider-grimoire-web`, and is the only documented binding for it.


Malabarba:
I'm in favor. :+1:

bbatsov:
Oh, fuck. My mistake. However, this can't be fixed by just removing `C-h`. The other binding should be changed as well - we have `x` and `C-x` binding for all doc commands for consistency. Some people prefer not having to list their finger from control, therefore the current dual keybindings.

bbatsov:
I'm guessing good alternatives would be `C-e`, `C-r`, `C-c`, `C-v` - something on this side of the keyboard.

bbatsov:
Also amusingly - I managed to choose a poor keybinding for this twice in a row. It used to be `C-c C-d C-g` before (which suffers from similar problems).

expez:
I prefer <kbd>C-r</kbd> as it's slightly easier to remember g**r**imoire

bbatsov:
Yeah, but it turns out we already use it for the non-web grimoire (guess for the very same reason). :-)

grammati:
How about w, for "web"?

bbatsov:
Fine by me.

Issue #1420, opened 2015-11-14
Show stacktrace buffers for sync requests


bbatsov:
Right now when something goes wrong in `nrepl-send-sync-request` we just dump the stacktrace in the REPL buffer and switch to it. Obviously, this is less than ideal. The reason why we're not showing our custom pretty stacktraces is that for cider middleware requests (which are generally done synchronously) `*e` doesn't get bound to the last exception and our stacktrace printing middleware doesn't work.

bbatsov:
Just to add a bit more details here. I think the root cause of the issue is that we decided not to associate cider's custom middleware requests with a session. As I said I'm contemplating what's the best course of action and I definitely wouldn't mind some suggestions. @cemerick What would you recommend us to do?

Malabarba:
Could we write a middleware that wraps around all others and sets `*e` when appropriate?

bbatsov:
> Could we write a middleware that wraps around all others and sets *e when appropriate?

This idea also came to my mind, but this would also require changes to the way we request the stacktrace from the Elisp code. And I was also wondering if we might need more such data in the future. So maybe it makes sense to leverage the tooling session with the middleware... Anyways, if I had really great ideas - I would have implemented something by now. I'm definitely open to any suggestion/idea whatever. 

P.S. If we do some wrapper middleware it'd be nice if it did exception handling in general, so we'd avoid deadlocks on unhandled exceptions. Something like this would eliminate the need for the current notion of "sync timeout".

bbatsov:
@Malabarba Another relatively simple idea would be require the session middleware in all the middleware we have and to set some different var (e.g. `*e-middleware*`) there, so we won't clobber the evaluation exceptions. Then the stacktrace requests will be checking for both.

bbatsov:
@jeffvalk You might take a look at this, if you'd like.

cichli:
I think the ideal solution to this would be one that doesn't rely on checking `*e` at all - since it could be altered from underneath us by other evaluations, and we obviously don't want to be altering it ourselves.

If we had a way of getting nREPL to convey exception values to our middleware, it could then work generically for either exceptions thrown by an eval or exceptions thrown by other middlewares. I'll open a JIRA with a proposed solution and link to it here :-).

jeffvalk:
Haven't looked at this closely, but at one point I did bake up a middleware that wraps `eval` and returns errors and compiler warnings as data. If that might be useful, I'll resurrect it.

[...]

PR #1617, opened 2016-03-16
Try to handle CIDER mistakes more gracefully

sanjayl:
Not ready yet, but thought I'd see what you guys thought before going any further

Before submitting a PR make sure the following things have been done:

Thanks!

Unless specifically coded into the sync request, errors result in
stacktraces being thrown onto the REPL and focus changing to the REPL
buffer. These changes just log the exception info into a buffer and
notify the user of this in the message area. Easiest way to see how this
works is to select a region with unbalanced parens and then call a
cider-format-region-edn.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -965,16 +965,11 @@ sign of user input, so as not to hang the interface."
     (when (member "done" status)
       (when-let ((ex (nrepl-dict-get response "ex"))
                  (err (nrepl-dict-get response "err")))
-        ;; non-eval requests currently don't set the *e var
-        ;; which is required by the stacktrace middleware
-        ;; so we have to handle them differently until this is resolved
-        (if (member "eval-error" status)
-            (funcall nrepl-err-handler)
-          ;; dump the stacktrace in the REPL
-          ;; TODO: This has to be replaced with rendering of the
-          ;; standard stacktrace buffer
-          (cider-repl-emit-interactive-stderr err)
-          (switch-to-buffer-other-window connection)))
+        ;; Non-eval requests currently don't set the *e var
+        ;; but instead get stored in util/storage, the storage-key
+        ;; lets us retrieve these errors. If there's no storage-key,
+        ;; then this will end up returning *e.
+        (funcall nrepl-err-handler (nrepl-dict-get response "storage-key")))
       (when-let ((id (nrepl-dict-get response "id")))
         (with-current-buffer connection
           (nrepl--mark-id-completed id)))
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -811,13 +811,20 @@ into a new error buffer."
     (cond (class (cons response causes))
           (status (cider--render-stacktrace-causes causes)))))
 
-(defun cider-default-err-op-handler ()
-  "Display the last exception, with middleware support."
+(defun cider-default-err-op-handler (&optional storage-key)
+  "Display an exception, with middleware support.
+If optional argument STORAGE-KEY is not given, will return the most recent
+exception bound to *e; this binding typically happens during a non-sync
+`eval` op.  Optional argument STORAGE-KEY will look up a stored exception
+instead of trying to use the *e bound exception; this is useful when the
+error occurred due to a sync request op which typically do not bind *e."
   ;; Causes are returned as a series of messages, which we aggregate in `causes'
   (let (causes)
     (cider-nrepl-send-request
      (append
       (list "op" "stacktrace" "session" (cider-current-session))
+      (when storage-key
+        (list "storage-key" storage-key))
       (when (cider--pprint-fn)
         (list "pprint-fn" (cider--pprint-fn)))
       (when cider-stacktrace-print-length
@@ -830,11 +837,15 @@ into a new error buffer."
        ;; after it has been handled, so it's fine to set it unconditionally here
        (setq causes (cider--handle-stacktrace-response response causes))))))
 
-(defun cider-default-err-handler ()
+(defun cider-default-err-handler (&optional storage-key)
   "This function determines how the error buffer is shown.
-It delegates the actual error content to the eval or op handler."
+It delegates the actual error content to the eval or op handler.
+Optional argument STORAGE-KEY will be passed to the op handler to help
+retrieve a stored exception instead of trying to use the *e bound
+exception; this is useful when the error occurred due to a sync request op
+which typically do not bind *e."
   (if (cider-nrepl-op-supported-p "stacktrace")
-      (cider-default-err-op-handler)
+      (cider-default-err-op-handler storage-key)
     (cider-default-err-eval-handler)))
 
 (defvar cider-compilation-regexp


nrepl-client.el
```diff
@@ -1336,6 +1333,26 @@ The default buffer name is *nrepl-error*."
       (set-window-point win (point-max)))
     (setq buffer-read-only t)))
 
+;;TODO: IMPROVE ON THIS, pull out buffer name into a variable, make formatting serious, etc.
+(defun log-cider-error (response)
+  "Log errors caught by CIDER if you don't want to spam the REPL with stacktraces"
+  (with-current-buffer (get-buffer-create "*cider-error-log*")
+    (setq buffer-read-only nil)
```
Malabarba: Don't set this variable. Let-bind `inhibit-read-only` to t instead.

Malabarba:
Thanks. :+1:

Could you show an example of what this would look like? And what's the current behaviour? 

Also, the function names need to be namespaced.

sanjayl:
Yeah sure, here's a screencast of the before:
https://youtu.be/lRfk5ic6euY
and afterwards with the PR:
https://youtu.be/X3zrGdlmHhE

Any preference on where the error management should live? I just put them in the nrepl-client code because that was proximal to where the errors were coming into the system

bbatsov:
Note that the real problem in a way is this one - https://github.com/clojure-emacs/cider/issues/1420

Ideally people would get stacktrace buffers when using sync nREPL requests and there won't be a need for another different error buffer. I'll go over your code a bit later.

sanjayl:
Interesting...I'll try and make a version where everything goes through the standard stacktrace buffer. 

I think that it'll be beneficial to ultimately have some kind of tiered error reporting system with varying degrees of interaction, attention, and logging. This would require doing some analysis of the stack traces at the middleware level and then have some standard hierarchy of error messages to use in the reply's `status` slot

sanjayl:
Here's a demonstration of error reporting using a stacktrace buffer (the vanilla one, not the fancy one that relies on the stacktrace middleware). It was pretty much just a two line change in the existing code.
https://youtu.be/L4TS77nZMq8

Not a huge fan of it for this particular problem, but would be a nice option to have for sync request errors in general. 

In order to get it to use the fancy stacktrace middleware, I was planning on keeping a cache of the caught java.lang.Exception's and stuffing a key to that cache into the reply packet we send back to the CIDER side. That way the CIDER side can request a StackTrace analysis after the fact without needing *e to be bound and it wouldn't matter if *e got changed out from underneath us or anything like that. (The error catching refactorings/abstractions that we've been discussing would be pretty useful for this to hide all this complexity)

bbatsov:
> Here's a demonstration of error reporting using a stacktrace buffer (the vanilla one, not the fancy one that relies on the stacktrace middleware). It was pretty much just a two line change in the existing code.
> https://youtu.be/L4TS77nZMq8

This video is private. :-)

sanjayl:
oops, my bad, give it a try again:
https://www.youtube.com/watch?v=L4TS77nZMq8

johngit22:
Thanks!

bbatsov:
Looks ok. 

> Not a huge fan of it for this particular problem, but would be a nice option to have for sync request errors in general.

Why not? I'd rather use something generic instead of having some special buffer type for middleware errors. I see the point of your initial solution, but it seems to me that's it's a bit excessive (and I don't like that the error buffer combines different data, that while useful debugging is likely going to confuse most users). I do agree, however, that it makes sense to display some message in the minibuffer saying that an "internal cider error occurred" or something like this in such situations, as obviously just dumping/showing a stacktrace is not terribly informative either. 

> In order to get it to use the fancy stacktrace middleware, I was planning on keeping a cache of the caught java.lang.Exception's and stuffing a key to that cache into the reply packet we send back to the CIDER side. That way the CIDER side can request a StackTrace analysis after the fact without needing *e to be bound and it wouldn't matter if *e got changed out from underneath us or anything like that. (The error catching refactorings/abstractions that we've been discussing would be pretty useful for this to hide all this complexity)

Yeah, I guess something like this might work.

nrepl-client.el
```diff
@@ -1336,6 +1333,26 @@ The default buffer name is *nrepl-error*."
       (set-window-point win (point-max)))
     (setq buffer-read-only t)))
 
+;;TODO: IMPROVE ON THIS, pull out buffer name into a variable, make formatting serious, etc.
+(defun log-cider-error (response)
+  "Log errors caught by CIDER if you don't want to spam the REPL with stacktraces"
```
bbatsov: Btw, you should check out https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html (or use `M-x checkdoc`).

nrepl-client.el
```diff
@@ -1336,6 +1333,26 @@ The default buffer name is *nrepl-error*."
       (set-window-point win (point-max)))
     (setq buffer-read-only t)))
 
+;;TODO: IMPROVE ON THIS, pull out buffer name into a variable, make formatting serious, etc.
+(defun log-cider-error (response)
```
bbatsov: Functions in Emacs should always have names prefixed with their package - e.g. `nrepl-log-error`. There shouldn't be cider references in nrepl-client, as in theory it's usable without cider (although in really they're quite coupled presently).

sanjayl:
Thanks for the pointers and documentation resources on the Elisp code! I'll definitely update with better organized and documented code once I have a more concrete idea of what to implement and what parts of the system it will touch. Ideally I'd like to pull all of the error handling related code out into a different namespace but I'm worried to make drastic changes since I don't understand Elisp or the CIDER side of things well at all, so I'll probably just start small.

> Why not? I'd rather use something generic instead of having some special buffer type for middleware errors. I see the point of your initial solution, but it seems to me that's it's a bit excessive (and I don't like that the error buffer combines different data, that while useful debugging is likely going to confuse most users). I do agree, however, that it makes sense to display some message in the minibuffer saying that an "internal cider error occurred" or something like this in such situations, as obviously just dumping/showing a stacktrace is not terribly informative either.

Yeah, that's a good point...I don't have any compelling arguments for my case to be honest and the examples I put together are more just for brainstorming rather than a final destination. The next prototype that I'm planning will let you coarsely tune the error response, so maybe we'll be able to figure out what the best direction would be by getting some real world usage under our belts.

I will say that in light of the [bug report in CIDER-nREPL that prompted this branch](https://github.com/clojure-emacs/cider-nrepl/issues/313), it would be nice if bug reports would contain the request packet, reply packet, AND the stacktrace instead of just the stacktrace so that it would be a bit easier to replicate the problem. You're right that it is confusing though, so I'll have to think more about how to balance UI with QA needs.

Thanks again for the input!

sanjayl:
Here's the demo of a sync-request error displayed in the nice stacktrace-op dependent buffer as you mentioned:

https://youtu.be/RsYtKBZniVs

I just added:

1. A tiny util ns to basically mock out a cache
2. Modified the `u/err-info` function so that it automatically registers every exception into the cache and stuffs a key into the response packet. 
3. An op in the stacktrace middleware to look up an Exception from the cache instead of using `*e`

From there, I basically just used the same internals that we use on the CIDER side for handling the eval-type errors. Actually, there's very little that's new in these changes at all, just a couple lines here and there that are different.

I've pushed the CIDER-nREPL [code to my repo under the store-stacktrace branch](https://github.com/clojure-emacs/cider-nrepl/compare/master...sanjayl:store-stacktrace). It's very very (embarrassingly) rudimentary, so I didn't want to send over a PR, but let me know if it's going in the right direction and I'll try to polish it up and get it out soon.

Thanks!

sanjayl:
Actually, now that I think about it, we don't even need the separate Stacktrace op, could just pass in an optional storage-key...If no storage key passed in, return *e. If key passed in, return stored stacktrace. If we did it that way, there'd be even fewer overall changes

bbatsov:
> Here's the demo of a sync-request error displayed in the nice stacktrace-op dependent buffer as you mentioned:
> 
> https://youtu.be/RsYtKBZniVs

Looks good!

> Actually, now that I think about it, we don't even need the separate Stacktrace op, could just pass in an optional storage-key...If no storage key passed in, return *e. If key passed in, return stored stacktrace. If we did it that way, there'd be even fewer overall changes

Sounds like a good plan to me.

> I will say that in light of the bug report in CIDER-nREPL that prompted this branch, it would be nice if bug reports would contain the request packet, reply packet, AND the stacktrace instead of just the stacktrace so that it would be a bit easier to replicate the problem. You're right that it is confusing though, so I'll have to think more about how to balance UI with QA needs.

Indeed. Finding the right balance is always hard. On a related note - the UI of the messages log itself should eventually be improved - e.g. it might be nice to be able to filter out certain message types, highlight requests and responses, font-lock errors differently, and fix the alignment of dictionaries... If only we had time. :-)

sanjayl:
Submitting a different pull request which doesn't require the whole "storage" mess. This PR can be rejected.

PR #1634, opened 2016-03-25
Fix bug in cider-expected-ns; make tests run again

aiba:
Oops, my previous PR sorted the seq incorrectly.  This PR fixes the bug and adds a test case for it.

I also discovered the `string-prefix-p` function, which makes the classpath prefix test simpler.

Side note 1: I think 3f345bfd8 broke `make test`, so this fixes the tests also.

Side note 2: is there an elisp seq equivalant to clojure's `sort-by`?  Such a function would make things a little simpler.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -675,12 +675,12 @@ active nREPL connection."
       (let* ((path (or path (file-truename (buffer-file-name))))
              (relpath (thread-last (cider-sync-request:classpath)
                         (seq-map
-                         (lambda (p)
-                           ;; when path starts with p
-                           (when (string-match (rx-to-string `(: bos ,p)) path)
-                             (substring path (length p)))))
+                         (lambda (cp)
+                           (when (string-prefix-p cp path)
+                             (substring path (length cp)))))
                         (seq-filter #'identity)
-                        (seq-sort #'string<)
+                        (seq-sort (lambda (a b)
+                                    (< (length a) (length b))))
                         (car))))
         (when relpath
           (thread-last (substring relpath 1) ; remove leading /
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -782,8 +782,9 @@
     (should-error (cider-find-ns) :type 'user-error)))
 
 (ert-deftest cider-expected-ns ()
-  (noflet ((cider-ensure-connected () t)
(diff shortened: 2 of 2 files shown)


bbatsov:
> Side note 1: I think 3f345bf broke make test, so this fixes the tests also.

My bad.

> Side note 2: is there an elisp seq equivalant to clojure's sort-by? Such a function would make things a little simpler.

Don't think so. But we can always ask @NicolasPetton for this. :-)

NicolasPetton:
Isn't `seq-sort` enough? (Or I don't understand what Clojure's `sort-by` does)

bbatsov:
It allows you write more expressive code - https://www.conj.io/store/v1/org.clojure/clojure/1.8.0/clj/clojure.core/sort-by

e.g. `(sort-by length ["alal" "bala" "portokala"])`

NicolasPetton:
Bozhidar Batsov <notifications@github.com> writes:

> It allows you write more expressive code - https://www.conj.io/store/v1/org.clojure/clojure/1.8.0/clj/clojure.core/sort-by
>
> e.g. `(sort-by length ["alal" "bala" "portokala"])`

I see, it's pretty neat!

I'm afraid that an Elisp would not be as elegant though, as you'd have
to provide a compare fn most of the times (comparison functions being
different depending on types).

What about something like:

(seq-sort-by #'seq-length #'> myseq)

aiba:
@NicolasPetton yes, it would be awesome to have a `seq-sort`, even if it required a comparator arg.  In this PR (where `foo` is a list of strings), it would allow us to write:

```elisp
(seq-sort-by #'length #'< foo)
```

Instead of

```elisp
(seq-sort (lambda (a b)
            (< (length a) (length b)))
          foo)
```

I even kinda like the comparator arg because it makes it explicit whether we're sorting ascending or descending.

NicolasPetton:
`seq` 2.14 has it!

bbatsov:
Nice!

On 29 March 2016 at 10:29, Nicolas Petton <notifications@github.com> wrote:

> seq 2.14 has it!
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/clojure-emacs/cider/pull/1634#issuecomment-202755196>
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

PR #1636, opened 2016-03-28
cider-auto-test-mode

Malabarba:
A minor mode that runs tests after every successful load-file op.
I've been using this a lot lately, and I'm quite happy with it now.

It might make sense to also let the user run tests after every eval op (instead of just load-file). But I actually like that I can play around small code snippets with `C-x C-e` and `C-M-x` without accidentally reruning all tests, so I'm leaving this as is for now.

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -533,21 +533,23 @@ This uses the Leiningen convention of appending '-test' to the namespace name."
 (declare-function cider-emit-interactive-eval-output "cider-interaction")
 (declare-function cider-emit-interactive-eval-err-output "cider-interaction")
 
-(defun cider-test-execute (ns &optional tests)
+(defun cider-test-execute (ns &optional tests silent)
   "Run tests for NS, which may be a keyword, optionally specifying TESTS.
 
 This tests a single NS, or multiple namespaces when using keywords `:project',
 `:loaded' or `:non-passing'.  Optional TESTS are only honored when a single
 namespace is specified.  Upon test completion, results are echoed and a test
 report is optionally displayed.  When test failures/errors occur, their sources
-are highlighted."
+are highlighted.
+If SILENT is non-nil, suppress all messages other then test results."
   (cider-test-clear-highlights)
   (cider-map-connections
    (lambda (conn)
-     (if (and tests (= (length tests) 1))
-         ;; we generate a different message when running individual tests
-         (cider-test-echo-running ns (car tests))
-       (cider-test-echo-running ns))
+     (unless silent
+       (if (and tests (= (length tests) 1))
+           ;; we generate a different message when running individual tests
+           (cider-test-echo-running ns (car tests))
+         (cider-test-echo-running ns)))
      (cider-nrepl-send-request
       (list "op"     (cond ((stringp ns)         "test")
                            ((eq :project ns)     "test-all")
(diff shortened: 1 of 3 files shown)


cider-test.el
```diff
@@ -646,6 +646,29 @@ is searched."
...
+
+;;;###autoload
+(define-minor-mode cider-auto-test-mode
+  "Toggle automatic testing of Clojure files.
+
+When enabled this reruns tests every time a Clojure file is loaded.
+Only runs tests corresponding to the loaded file's namespace and does
+nothing if no tests are defined or if the file failed to load."
+  nil nil nil
+  :global t
```
bbatsov: Whi is this global?

bbatsov:
> A minor mode that runs tests after every successful load-file op.
> I've been using this a lot lately, and I'm quite happy with it now.

Looks good!

> It might make sense to also let the user run tests after every eval op (instead of just load-file). But I actually like that I can play around small code snippets with C-x C-e and C-M-x without accidentally reruning all tests, so I'm leaving this as is for now.

Yeah, I think this should be a defcustom.

Btw, the final commit seems unrelated to the PR.

cider-test.el
```diff
@@ -646,6 +646,29 @@ is searched."
             (cider-test-execute ns (cdr def))
           (message "No test at point"))))))
 
+;;; Auto-test mode
+(defun cider--test-silently ()
+  "Like `cider-test-run-tests', but with less feedback.
+Only notify the user if there actually were any tests to run and only after
```
bbatsov: I'm kind of confused about this description - shouldn't this say you'll see the results only if something failed? I might be missing something, though.

cider-test.el
Malabarba: I don't know. Why would it be local? :-)

cider-test.el
Malabarba: No, it displays a message if the tests were successful. Otherwise you wouldn't know if the tests are finished or if they're hanging. 

The point here is just to not bother the user if there are no tests to run.

Malabarba:
Yes, looks like I got my git messed up and pushed a wrong commit.

cider-test.el
bbatsov: I thought people would be supposed to enable it using `cider-mode-hook`, but if you think this is a better idea - it's fine by me. Just mention the mode somewhere in the manual.

bbatsov:
One more thing - mention this in the manual.

Malabarba:
> One more thing - mention this in the manual.

How do I do that? :-)

bbatsov:
Just edit the files under /doc. The manual will be updated automatically.

On Wednesday, 30 March 2016, Artur Malabarba <notifications@github.com>
wrote:

> One more thing - mention this in the manual.
>
> How do I do that? :-)
>
> —
> You are receiving this because you commented.
> <https://github.com/clojure-emacs/cider/pull/1636#issuecomment-203459679>
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

bbatsov:
Guess you can squash the final 3 commits together and we're good to go.

bbatsov:
It also seems you'll have to rebase on top of the current `master`.

Malabarba:
All done. :+1:

bbatsov:
Hehe - this was fast! :-)

arichiardi:
I just discovered `cider-auto-test-mode` and I wanted to thank you! This is an awesome feature!

Malabarba:
:-)

arichiardi:
Now I just need to add a `defcustom` in order to avoid the load-file confirmation :smile:

Malabarba:
Is that a joke? :-)

arichiardi:
Lol, sorry now, it was half a question. :) When I do `C-c C-l` it asks me the file name every time and I checked a bit cider and I did not see how to disable it. Is it possible? If not, I can patch!

Malabarba:
Sorry, I think I sounded a bit rude. :-)
Yes, there is already a variable for doing exactly that, that's why I wasn't sure if you were joking or really asking. `cider-prompt-save-file-on-load` ;-)

arichiardi:
Ahah no probs, I sounded a bit silly in asking indeed, thanks a lot for the hint!

Issue #1352, opened 2015-09-25
Run checkdoc


Malabarba:
It would be nice to add checkdoc to our travis builds, but first we need to actually pass checkdoc. This means running checkdoc on each source file and making the edits suggested by it.
We don't need to do all files in one go. Fixing up a file should be enough for a PR.

Pretty much anyone can call `M-x checkdoc` in a file and follow the instructions, so this is as low as a fruit can hang.

Malabarba:
If anyone takes this up, please set

    (setq checkdoc-package-keywords-flag nil)
    (setq checkdoc-arguments-in-order-flag nil)

so as to avoid these (irrelevant) warnings.

bpacreau:
Hi Malabarba,

With @g12l, we are willing to help you on this issue. We're currently working on it and we're going to open our first PR soon

bbatsov:
Sounds good!

Malabarba:
That's great to hear! :)

g12l:
Some files are already ok with checkdoc:
  -cider-scratch.el
  -cider-classpath.el
  -cider-resolve.el
  -cider-browse-ns.el
  -cider-compat.el
  -cider-selector
  -cider-eldoc
  -cider-macroexpansion.

With @bpacreau we corrected simple errors in cider-grimoire.el and cider-popul.el (PRs). We need some time for the other files and we will probably ask some questions on the gitter channel.

bbatsov:
Roger that! :-)

expez:
@g12l @bpacreau Thanks for taking this on,! :+1:

Malabarba:
And feel free to ask away!

PR #1637, opened 2016-03-29
Documentation follows checkdoc style.

cskksc:

Thanks!

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -7,6 +7,8 @@
   (fill-column . 80)
   (sentence-end-double-space . t)
   (emacs-lisp-docstring-fill-column . 75)
+  (checkdoc-symbol-words . ("top-level" "major-mode" "macroexpand-all" "print-level" "print-length"))
+  (checkdoc-package-keywords-flag)
   (checkdoc-arguments-in-order-flag))
  (emacs-lisp-mode
   (bug-reference-bug-regexp . "#\\(?2:[[:digit:]]+\\)")))
--- a/cider-apropos.el
+++ b/cider-apropos.el
@@ -46,7 +46,14 @@
   (cider-doc-lookup (button-get button 'apropos-symbol)))
 
 (defun cider-apropos-summary (query ns docs-p include-private-p case-sensitive-p)
-  "Return a short description for the performed apropos search."
+  "Return a short description for the performed apropos search.
+
+QUERY can be a regular expression list of space-separated words
+\(e.g take while) which will be converted to a regular expression
+\(like take.+while) automatically behind the scenes.  The search may be
+limited to the namespace NS, and may optionally search doc strings
+\(based on DOCS-P), include private vars (based on INCLUDE-PRIVATE-P),
+and be case-sensitive (based on CASE-SENSITIVE-P)."
   (concat (if case-sensitive-p "Case-sensitive " "")
           (if docs-p "Documentation " "")
           (format "Apropos for %S" query)
@@ -104,8 +111,8 @@
 QUERY can also be a list of space-separated words (e.g. take while) which
 will be converted to a regular expression (like take.+while) automatically
 behind the scenes.  The search may be limited to the namespace NS, and may
-optionally search doc strings, include private vars, and be case
-sensitive."
+optionally search doc strings (based on DOCS-P), include private vars
+\(based on PRIVATES-P), and be case-sensitive (based on CASE-SENSITIVE-P)."
   (interactive
    (cons (read-string "Search for Clojure symbol (a regular expression): ")
          (when current-prefix-arg
@@ -132,7 +139,14 @@ sensitive."
 
 ;;;###autoload
 (defun cider-apropos-select (query &optional ns docs-p privates-p case-sensitive-p)
-  "Similar to `cider-apropos', but presents the results in a completing read."
+  "Similar to `cider-apropos', but presents the results in a completing read.
+
+Show all symbols whose names match QUERY, a regular expression.
+QUERY can also be a list of space-separated words (e.g. take while) which
+will be converted to a regular expression (like take.+while) automatically
+behind the scenes.  The search may be limited to the namespace NS, and may
+optionally search doc strings (based on DOCS-P), include private vars
+\(based on PRIVATES-P), and be case-sensitive (based on CASE-SENSITIVE-P)."
   (interactive
    (cons (read-string "Search for Clojure symbol (a regular expression): ")
          (when current-prefix-arg
--- a/cider-client.el
+++ b/cider-client.el
@@ -103,7 +103,7 @@ If the list is empty and buffer-local, return the global value."
 
 (defun cider-make-connection-default (connection-buffer)
   "Make the nREPL CONNECTION-BUFFER the default connection.
-Moves CONNECTION-BUFFER to the front of `cider-connections'."
+Moves CONNECTION-BUFFER to the front of variable `cider-connections'."
   (interactive (list (if (cider--in-connection-buffer-p)
                          (current-buffer)
                        (user-error "Not in a REPL buffer"))))
@@ -115,7 +115,7 @@ Moves CONNECTION-BUFFER to the front of `cider-connections'."
 
 (declare-function cider--close-buffer "cider-interaction")
 (defun cider--close-connection-buffer (conn-buffer)
-  "Close CONN-BUFFER, removing it from `cider-connections'.
+  "Close CONN-BUFFER, removing it from variable `cider-connections'.
 Also close associated REPL and server buffers."
   (let ((buffer (get-buffer conn-buffer)))
     (setq cider-connections
@@ -147,10 +147,11 @@ followed by any connection at all.
 
 If PROJECT-DIRECTORY is provided act on that project instead.
 
-Only return nil if `cider-connections' is empty (there are no connections).
+Only return nil if variable `cider-connections' is empty,
+i.e there are no connections.
 
 If more than one connection satisfy a given level of preference, return the
-connection buffer closer to the start of `cider-connections'.  This is
+connection buffer closer to the start of variable `cider-connections'.  This is
 usally the connection that was more recently created, but the order can be
 changed.  For instance, the function `cider-make-connection-default' can be
 used to move a connection to the head of the list, so that it will take
@@ -507,7 +508,7 @@ Do nothing if `cider-show-eval-spinner' is nil."
                      cider-eval-spinner-delay))))
 
 (defun cider-eval-spinner-handler (eval-buffer original-callback)
-  "Return a response handler that stops the spinner and calls ORIGINAL-CALLBACK.
+  "Return a response handler to stop the spinner and call ORIGINAL-CALLBACK.
 EVAL-BUFFER is the buffer where the spinner was started."
   (lambda (response)
     ;; buffer still exists and
@@ -1041,7 +1042,8 @@ INPUT, SESSION, and NS are passed to `nrepl--eval-request'."
 The request is dispatched via CONNECTION and SESSION.
 If NS is non-nil, include it in the request.
 RIGHT-MARGIN specifies the maximum column width of the
-pretty-printed result, and is included in the request if non-nil."
+pretty-printed result, and is included in the request if non-nil.
+PPRINT-FN generates the pretty-printed result."
   (cider-nrepl-request:eval input callback ns nil nil
                             (cider--nrepl-pprint-request-plist right-margin pprint-fn)))
 (make-obsolete 'cider-nrepl-request:pprint-eval
--- a/cider-common.el
+++ b/cider-common.el
@@ -42,9 +42,16 @@ prompt if that throws an error."
   :package-version '(cider . "0.9.0"))
 
 (defun cider--should-prompt-for-symbol (&optional invert)
+  "Return the value of the variable `cider-prompt-for-symbol'.
+Optionally invert the value, if INVERT is truthy."
   (if invert (not cider-prompt-for-symbol) cider-prompt-for-symbol))
 
 (defun cider-prompt-for-symbol-function (&optional invert)
+  "Prompt for symbol if funcall `cider--should-prompt-for-symbol' is truthy.
+Otherwise attempt to use the symbol at point for the command, and only
+prompt if that throws an error.
+
+INVERT is used to invert the semantics of the function `cider--should-prompt-for-symbol'."
   (if (cider--should-prompt-for-symbol invert)
       #'cider-read-symbol-name
     #'cider-try-symbol-at-point))
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -100,7 +100,7 @@ configure `cider-debug-prompt' instead."
   :package-version '(cider . "0.9.1"))
 
 (defcustom cider-debug-print-level 10
-  "print-level for values displayed by the debugger.
+  "The print-level for values displayed by the debugger.
 This variable must be set before starting the repl connection."
   :type '(choice (const :tag "No limit" nil)
                  (integer :tag "Max depth" 10))
@@ -108,7 +108,7 @@ This variable must be set before starting the repl connection."
   :package-version '(cider . "0.10.0"))
 
 (defcustom cider-debug-print-length 10
-  "print-length for values displayed by the debugger.
+  "The print-length for values displayed by the debugger.
 This variable must be set before starting the repl connection."
   :type '(choice (const :tag "No limit" nil)
                  (integer :tag "Max depth" 10))
@@ -136,7 +136,7 @@ This variable must be set before starting the repl connection."
     (message "No currently instrumented definitions")))
 
 (defun cider--debug-response-handler (response)
-  "Handle responses from the cider.debug middleware."
+  "Handles RESPONSE from the cider.debug middleware."
   (nrepl-dbind-response response (status id causes)
     (when (member "enlighten" status)
       (cider--handle-enlighten response))
@@ -307,7 +307,7 @@ In order to work properly, this mode must be activated by
             (add-hook 'kill-buffer-hook #'cider--debug-quit nil 'local)
             (add-hook 'before-revert-hook #'cider--debug-quit nil 'local)
             (unless (consp input-type)
-              (error "debug-mode activated on a message not asking for commands: %s" cider--debug-mode-response))
+              (error "Activated debug-mode on a message not asking for commands: %s" cider--debug-mode-response))
             ;; Integrate with eval commands.
             (setq cider-interactive-eval-override
                   (apply-partially #'cider--debug-lexical-eval
@@ -348,7 +348,7 @@ In order to work properly, this mode must be activated by
 (define-key cider--debug-mode-map "h" #'cider-debug-move-here)
 
 (defun cider--debug-remove-overlays (&optional buffer)
-  "Remove CIDER debug overlays from BUFFER if `cider--debug-mode' is nil."
+  "Remove CIDER debug overlays from BUFFER if variable `cider--debug-mode' is nil."
   (when (or (not buffer) (buffer-live-p buffer))
     (with-current-buffer (or buffer (current-buffer))
       (unless cider--debug-mode
@@ -418,6 +418,8 @@ message."
 (defconst cider--debug-buffer-format "*cider-debug %s*")
 
 (defun cider--debug-trim-code (code)
+  "Remove whitespace and reader macros from the start of the CODE.
+Return trimmed CODE."
   (replace-regexp-in-string "\\`#[a-z]+[\n\r[:blank:]]*" "" code))
 
 (declare-function cider-set-buffer-ns "cider-mode")
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -73,8 +73,11 @@ navigate to this buffer."
   :group 'cider)
 
 (defcustom cider-auto-jump-to-error t
-  "When non-nil automatically jump to error location during interactive compilation.
-When set to 'errors-only, don't jump to warnings."
+  "Control the cursor jump behaviour in compilation error buffer.
+
+When non-nil automatically jump to error location during interactive
+compilation.  When set to 'errors-only, don't jump to warnings.
+When set to nil, don't jump at all."
   :type '(choice (const :tag "always" t)
                  (const errors-only)
                  (const :tag "never" nil))
@@ -442,6 +445,8 @@ Invert meaning of `cider-prompt-for-symbol' if PREFIX indicates it should be."
     (nrepl-dict-get "path")))
 
 (defun cider--find-ns (ns &optional other-window)
+  "Find the file containing NS's definition.
+Optionally open it in a different window if OTHER-WINDOW is truthy."
   (if-let ((path (cider-sync-request:ns-path ns)))
       (cider-jump-to (cider-find-file path) nil other-window)
     (user-error "Can't find %s" ns)))
@@ -614,8 +619,9 @@ The handler simply inserts the result value in BUFFER."
 (defun cider--emit-interactive-eval-output (output repl-emit-function)
   "Emit output resulting from interactive code evaluation.
 
-The output can be send to either a dedicated output buffer or the current REPL buffer.
-This is controlled via `cider-interactive-eval-output-destination'."
+The OUTPUT can be sent to either a dedicated output buffer or the current
+REPL buffer.  This is controlled by `cider-interactive-eval-output-destination'.
+REPL-EMIT-FUNCTION emits the OUTPUT."
   (pcase cider-interactive-eval-output-destination
     (`output-buffer (let ((output-buffer (or (get-buffer cider-output-buffer)
                                              (cider-popup-buffer cider-output-buffer t))))
@@ -738,7 +744,6 @@ Returns the position at which PROPERTY was found, or nil if not found."
 
 (defun cider-jump-to-compilation-error (&optional _arg _reset)
   "Jump to the line causing the current compilation error.
-
 _ARG and _RESET are ignored, as there is only ever one compilation error.
 They exist for compatibility with `next-error'."
   (interactive)
@@ -1219,6 +1224,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
 
 (defun cider-insert-region-in-repl (start end &optional arg)
   "Insert the curent region in the REPL buffer.
+START and END represent the region's boundaries.
 If invoked with a prefix ARG eval the expression after inserting it."
   (interactive "rP")
   (cider-insert-in-repl
@@ -1240,7 +1246,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
 
 (defun cider-enable-on-existing-clojure-buffers ()
   "Enable CIDER's minor mode on existing Clojure buffers.
-See command `cider-mode'."
+See `cider-mode'."
   (interactive)
   (add-hook 'clojure-mode-hook #'cider-mode)
   (dolist (buffer (cider-util--clojure-buffers))
@@ -1249,7 +1255,7 @@ See command `cider-mode'."
 
 (defun cider-disable-on-existing-clojure-buffers ()
   "Disable `cider-mode' on existing Clojure buffers.
-See command `cider-mode'."
+See `cider-mode'."
   (interactive)
   (dolist (buffer (cider-util--clojure-buffers))
     (with-current-buffer buffer
@@ -1513,8 +1519,8 @@ of the buffer into a formatted string."
 (defun cider--format-region (start end formatter)
   "Format the contents of the given region.
 
-START and END are the character positions of the start and end of the
-region.  FORMATTER is a function of one argument which is used to convert
+START and END represent the region's boundaries.
+FORMATTER is a function of one argument which is used to convert
 the string contents of the region into a formatted string."
   (let* ((original (buffer-substring-no-properties start end))
          (formatted (funcall formatter original))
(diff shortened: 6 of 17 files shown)


bbatsov:
That's one big cleanup! :-) Thanks! I'll go over it and get back to you with some feedback soon.

cskksc:
Start fixing #1352

Cider files:

- [X] cider-apropos.el
- [X] cider-browse-ns.el
- [X] cider-classpath.el
- [X] cider-client.el
- [X] cider-common.el
- [X] cider-compat.el
- [X] cider-debug.el
- [X] cider-doc.el
- [X] cider-eldoc.el
- [X] cider-grimoire.el
- [ ] cider-inspector.el
 11 missing function docstrings

- [ ] cider-interaction.el
  5 missing function docstrings

- [X] cider-macroexpansion.el

- [X] cider-mode.el
- [X] cider-overlays.el
- [X] cider-popup.el
- [X] cider-repl.el
- [X] cider-resolve.el
- [X] cider-scratch.el
- [X] cider-selector.el
- [X] cider-stacktrace.el
- [ ] cider-test.el
  A lot of warnings about "tests should be renamed to test". May be we need to disable some flag ?
Need to look into this further.

- [X] cider-util.el
- [ ] cider.el
  3 missing function docstrings

- [X] nrepl-client.el

I'm gonna try and write docstrings for the missing function definitions. But it would be great if someone familiar with the codebase does some of the non-trivial ones.

Ill change the commit message to match `[Fix ...]` as a final step before merging into master, once we are happy with the changes. I feel that multiple updates spam notifications on the issue.

/cc @bbatsov @Malabarba @g12l

cider-apropos.el
```diff
@@ -46,7 +46,14 @@
   (cider-doc-lookup (button-get button 'apropos-symbol)))
 
 (defun cider-apropos-summary (query ns docs-p include-private-p case-sensitive-p)
-  "Return a short description for the performed apropos search."
+  "Return a short description for the performed apropos search.
```
cskksc: Duplicate of https://github.com/cskksc/cider/blob/feature/checkdoc/cider-apropos.el#L109.

cider-apropos.el
```diff
@@ -132,7 +139,14 @@ sensitive."
 
 ;;;###autoload
 (defun cider-apropos-select (query &optional ns docs-p privates-p case-sensitive-p)
-  "Similar to `cider-apropos', but presents the results in a completing read."
+  "Similar to `cider-apropos', but presents the results in a completing read.
```
cskksc: Duplicate of https://github.com/cskksc/cider/blob/feature/checkdoc/cider-apropos.el#L109.

cider-client.el
```diff
@@ -457,8 +458,8 @@ Refreshes EWOC."
   (interactive)
   (cider--ewoc-apply-at-point #'cider--connections-goto-connection))
 
-(defun cider--connections-goto-connection (_ewoc data)
-  "Goto the REPL for the connection in _EWOC specified by DATA."
+(defun cider--connections-goto-connection (ewoc data)
+  "Goto the REPL for the connection in EWOC specified by DATA."
```
cskksc: `checkdoc` has problems with variables starting with `_` or only `_`. It cannot recognise that the docstring contains the variable.

cider-debug.el
```diff
@@ -348,7 +348,7 @@ In order to work properly, this mode must be activated by
 (define-key cider--debug-mode-map "h" #'cider-debug-move-here)
 
 (defun cider--debug-remove-overlays (&optional buffer)
-  "Remove CIDER debug overlays from BUFFER if `cider--debug-mode' is nil."
+  "Remove CIDER debug overlays from BUFFER if command `cider--debug-mode' is nil."
```
cskksc: This has been done in a couple of different places. `mode` is not allowed as a prefix. It has to be something from `(function,command,variable,option or symbol)`. I think `command` or `function` is a workable substitute.

cider-debug.el
```diff
@@ -418,6 +418,7 @@ message."
 (defconst cider--debug-buffer-format "*cider-debug %s*")
 
 (defun cider--debug-trim-code (code)
+  "Return trimmed CODE."
```
cskksc: I'm not really sure what it does. I assume it cleans up the buffer if it contains any left overs from the previous runs, by removing any chars or blank space. Let me know what should be correct docstring.

cider-interaction.el
```diff
@@ -736,10 +741,10 @@ Returns the position at which PROPERTY was found, or nil if not found."
     (when (and (not (= p (point-min))) (not (= p (point-max))))
       p)))
 
-(defun cider-jump-to-compilation-error (&optional _arg _reset)
+(defun cider-jump-to-compilation-error (&optional arg reset)
   "Jump to the line causing the current compilation error.
 
-_ARG and _RESET are ignored, as there is only ever one compilation error.
+ARG and RESET are ignored, as there is only ever one compilation error.
```
cskksc: Same `_` thing.

cider-debug.el
Malabarba: Actually, in this case I was referring to the variable.

cider-macroexpansion.el
```diff
@@ -109,29 +109,29 @@ This variable specifies both what was expanded and the expander.")
 
 ;;;###autoload
 (defun cider-macroexpand-1 (&optional prefix)
-  "Invoke \\=`macroexpand-1\\=` on the expression preceding point.
-If invoked with a PREFIX argument, use \\=`macroexpand\\=` instead of
-\\=`macroexpand-1\\=`."
+  "Invoke `macroexpand-1' on the expression preceding point.
```
cskksc: Why is the `\\=` necessary ? The docstring looks same without them, in `describe-function`.

cider-selector.el
```diff
@@ -46,7 +46,7 @@ DESCRIPTION is a one-line description of what the key selects.")
   "If non-nil use `switch-to-buffer-other-window'.")
 
 (defun cider--recently-visited-buffer (mode)
-  "Return the most recently visited buffer whose `major-mode' is derived from MODE.
+  "Return the most recently visited buffer, deriving its `major-mode' from MODE.
```
cskksc: Writing it this way reduces the string length to <80. :stuck_out_tongue:

cider-stacktrace.el
```diff
@@ -442,7 +442,13 @@ it wraps to 0."
     (back-to-indentation)))
 
 (defun cider-stacktrace-jump (&optional arg)
-  "Like `cider-find-var', but uses the stack frame source at point, if available."
+  "Like `cider-find-var', but use the stack frame source at point, if available.
+
+Prompt according to prefix ARG and `cider-prompt-for-symbol'.
```
cskksc: Duplicate of https://github.com/cskksc/cider/blob/feature/checkdoc/cider-client.el#L752

cider-test.el
```diff
@@ -212,7 +212,13 @@
           (goto-char pos))))))
 
 (defun cider-test-jump (&optional arg)
-  "Like `cider-find-var', but uses the test at point's definition, if available."
+  "Like `cider-find-var', but use the test at point's definition, if available.
+
+Prompt according to prefix ARG and `cider-prompt-for-symbol'.
```
cskksc: Duplicate of https://github.com/cskksc/cider/blob/feature/checkdoc/cider-client.el#L752

cider-client.el
Malabarba: If the argument is unused, it should start with a `_`. If the argument is used, then it shouldn't.

From the code snippet I can see here on my phone I can't tell which is the case here. But it is important we respect that otherwise we'll get byte compiler warnings that will break the tests.

cider.el
```diff
@@ -241,6 +242,7 @@ string is quoted for passing as argument to an inferior shell."
   (concat "-d " (shell-quote-argument (format "%s:%s" (car list) (cadr list)))))
 
 (defun boot-command-prefix (dependencies)
+  "Return a list of boot artifact strings created from DEPENDENCIES."
```
cskksc: I'm not sure if this is the best description of the function.

Malabarba:
Thanks a million. That's a lot of stuff. ☺
I'll have a look too.

nrepl-client.el
```diff
@@ -610,7 +611,7 @@ older requests with \"done\" status."
                         (gethash id nrepl-completed-requests))))
       (if callback
           (funcall callback response)
-        (error "nREPL: No response handler with id %s found" id)))))
+        (error " nREPL: No response handler with id %s found" id)))))
```
cskksc: The spaces are added before `nREPL` as checkdoc complains if any messages start with lower-case letters.

cider-client.el
cskksc: I've done this in at-least 2 places I think. Ill see if I can find some checkdoc setting to fix this.

nrepl-client.el
Malabarba: I'd rather replace `nREPL:` with `[nREPL]`.

cider-test.el
Malabarba: I think I would prefer if the whole docstring were just:

      "Like `cider-find-var', but use the test at point's definition, if available.
    The prefix ARG and `cider-prompt-for-symbol' decide whether to
    prompt and whether to use a new window, as in `cider-find-var'."

This way the user can just click through to see information, and we don't have to duplicate information everywhere. The same goes for other such duplicate docstringsl.

cider-util.el
```diff
@@ -91,7 +91,7 @@ If BUFFER is provided act on that buffer instead."
 
 ;;; Thing at point
 (defun cider-defun-at-point (&optional bounds)
-  "Return the text of the top-level sexp at point.
+  "Return the text of the `top-level` sexp at point.
```
Malabarba: Did checkdoc ask for this?

Malabarba: Ah, I see. I guess checkdoc was suggesting that you quote the variable-name.
Still, this is not a variable name, it's an english expression, and should not be quoted.

cider-util.el
```diff
@@ -220,7 +220,7 @@ PROP is the name of a text property."
   "A list of buffers for different major modes.")
 
 (defun cider--make-buffer-for-mode (mode)
-  "Return a temp buffer using major-mode MODE.
+  "Return a temp buffer using `major-mode` MODE.
```
Malabarba: Again, not a variable name and should not be quoted.

cider-util.el
```diff
@@ -178,7 +178,7 @@ Can only error if SKIP is non-nil."
           (cons (pop plist) (cider-intern-keys plist)))))
 
 (defmacro cider-propertize-region (props &rest body)
-  "Execute BODY and add PROPS to all the text it inserts.
+  "Execute BODY and add PROPS to all the text it will insert.
```
Malabarba: > to all the inserted text

cider-client.el
cskksc: I found a way to ignore the `_` arguments. Its mentioned in the `checkdoc.el` commentary. So a sample docstring would look like this
```
(defun hello (_name)
  "; checkdoc-params: (_name)
Hello world."
  (message "Hello World"))
```

Thats good enough right ? I'll update the PR soon, along with the other changes you mentioned. Have you had the time to review the remaining files ?

cider-util.el
cskksc: To avoid checkdoc warnings about a hyphenated word that is also a symbol, it suggests adding them to `checkdoc-symbol-words` as documented [here](https://github.com/emacs-mirror/emacs/blob/7dba17670f42cdaa73424a2bf60925ef49671fad/lisp/emacs-lisp/checkdoc.el#L141). I've tested it out with words we've seen here, `major-mode` and `top-level` and it works. Maybe we should we create a `checkdoc.el` which would contain all these cider specific checkdoc settings and also the flags mentioned on the issue page. Thoughts ?

cider-client.el
Malabarba: I don't understand why you need this. My checkdoc doesn't complain if I do:

```el
(defun hello (_name)
  "Hello world."
  (message "Hello World"))
```

cider-debug.el
Malabarba: It removes whitespace and reader macros from the *start* of the code.

Malabarba: By the waym, this is referring to the string `CODE`. It has nothing to do with the buffer.

cider-interaction.el
Malabarba: Just replace the entire docstring with:

      "Jump to the line causing the current compilation error.
    Optional arguments exist for compatibility with `next-error'."

cider-macroexpansion.el
Malabarba: They'll look different in Emacs 25.
The reason it used a double \` (instead of a \` and a ') is because we are referring to Clojure functions, not Elisp symbols. And, in Emacs 25, those \` will get converted to **left**-round-quotes, which looks plain wrong on the **right**-side of a word. 

However, I'm fine with writing it the way you have written (even though it doesn't represent an Elisp symbol).

Malabarba: ACtually, if you make this change, Emacs is going to link to its own `macroexpand` function. I think that's bad, so maybe we keep it the way it was.

cider.el
Malabarba: I think it's fine. :+1:

Malabarba:
> I'm gonna try and write docstrings for the missing function definitions. But it would be great if someone familiar with the codebase does some of the non-trivial ones.

> Ill change the commit message to match [Fix ...] as a final step before merging into master, once we are happy with the changes. I feel that multiple updates spam notifications on the issue.

Don't feel obligated to do this. It's perfectly fine to make such a huge change over several PRs. In fact, it's probably safer if you focus on finalizing the current PR (without any more additions), so that we can merge it before we start getting merge conflicts. 
In that case, instead of writing `[Fix #...]` just write `[#...]`.

cskksc:
I agree, this patch is big enough. I've made the changes. I don't know why the `_args` dont work on my setup. Are you using emacs 25? I have 24.5. Also checkdoc is now failing for some words like `top-level`, `major-mode` which we intend to be used as plain english. What do you think about adding a file containing these changes ? And also the ones you mentioned on the issue page. I think it would be useful when running checkdoc through travis.

Malabarba:
> What do you think about adding a file containing these changes ? 

I'm not sure I understand. Do you mean a separate `cider-*.el` file where we have made the extra changes checkdoc is requesting? How would that help with travis?

cskksc:
No. So currently checkdoc is failing because we have not quoted words like `top-level`, `major-mode` etc. We have to specify somewhere that these are normal words along with being elisp things. We do that by adding them to `checkdoc-symbol-words`. I was thinking about doing these things in `cider-checkdoc.el`, which would be loaded when running checkdoc related tests.

Malabarba:
Ah, I see.
We will eventually need a `cider-checkdoc.el` file for running this build, but that's not where you want to set the `checkdoc-symbol-words` variable. 
Instead, you can simply add the following line

    (checkdoc-symbol-words . ("top-level" "major-mode"))

to our `.dir-locals.el` file. This line should be immediately above the line with `checkdoc-symbol-words`.

cskksc:
@Malabarba `dir-locals` is nice. Dint know about it before. I've added the vars there. This PR is ready I think. Let me know if something needs to change. Ill submit another one with the remaining docs and a test runner :-)

Malabarba:
LGTM. 

@bbatsov, WDYT?

bbatsov:
@cskksc Remove the `.` from the commit message's title.

.dir-locals.el
```diff
@@ -7,6 +7,8 @@
   (fill-column . 80)
   (sentence-end-double-space . t)
   (emacs-lisp-docstring-fill-column . 75)
+  (checkdoc-symbol-words . ("top-level" "major-mode" "macroexpand-all"))
```
bbatsov: I just use "major mode" instead of "major-mode". Seem more English to me in general.

bbatsov: @lunaryorn Any idea if flycheck will ever honour those?

cider-apropos.el
```diff
@@ -56,7 +63,7 @@
             " (public symbols only)")))
 
 (defun cider-apropos-highlight (doc query)
-  "Return the DOC string propertized to highlight QUERY matches."
+  "Return the DOC string propertized to highlight QUERY match."
```
bbatsov: Remove this change. The original docstring is correct, the new version is unclear. There are *multiple* matches.

cider-apropos.el
```diff
@@ -140,7 +154,7 @@ sensitive."
                    (if (string= ns "") nil ns))
                  (y-or-n-p "Search doc strings? ")
                  (y-or-n-p "Include private symbols? ")
-                 (y-or-n-p "Case-sensitive? ")))))
+                 (y-or-n-p "Case sensitive? ")))))
```
bbatsov: case-sensitive is written with a hyphen - if you need to add an extra word this is not the solution :-)

cider-client.el
```diff
@@ -507,7 +508,7 @@ Do nothing if `cider-show-eval-spinner' is nil."
                      cider-eval-spinner-delay))))
 
 (defun cider-eval-spinner-handler (eval-buffer original-callback)
-  "Return a response handler that stops the spinner and calls ORIGINAL-CALLBACK.
+  "Return a response handler to stop the spinner & call ORIGINAL-CALLBACK.
```
bbatsov: I'd rather drop the callback part on the next line. I hate the use of ampersands in docs.

cider-common.el
```diff
@@ -42,9 +42,16 @@ prompt if that throws an error."
...
+  "Return the value of the variable `cider-prompt-for-symbol'.
+Optionally invert the value, if INVERT is truthy."
   (if invert (not cider-prompt-for-symbol) cider-prompt-for-symbol))
 
 (defun cider-prompt-for-symbol-function (&optional invert)
+  "Prompt for symbol if funcall `cider--should-prompt-for-symbol' is truthy.
+Otherwise attempt to use the symbol at point for the command, and only
+prompt if that throws an error.
+
+INVERT is used to invert the semantics of the function `cider--should-prompt-for-symbol'"
```
bbatsov: The final `.` is missing here.

cider-debug.el
```diff
@@ -100,7 +100,7 @@ configure `cider-debug-prompt' instead."
   :package-version '(cider . "0.9.1"))
 
 (defcustom cider-debug-print-level 10
-  "print-level for values displayed by the debugger.
+  "The `print-level' for values displayed by the debugger.
```
bbatsov: As we're talking about Clojure's `*print-level*` here I don't think this is the best change - seems like we're referring to an Elisp symbol.

cider-debug.el
```diff
@@ -108,7 +108,7 @@ This variable must be set before starting the repl connection."
   :package-version '(cider . "0.10.0"))
 
 (defcustom cider-debug-print-length 10
-  "print-length for values displayed by the debugger.
+  "The `print-length' for values displayed by the debugger.
```
bbatsov: Same here.

cider-debug.el
```diff
@@ -418,6 +418,8 @@ message."
 (defconst cider--debug-buffer-format "*cider-debug %s*")
 
 (defun cider--debug-trim-code (code)
+  "Return trimmed CODE.
```
bbatsov: I'd swap those two sentences.

cider-interaction.el
```diff
@@ -73,8 +73,10 @@ navigate to this buffer."
   :group 'cider)
 
 (defcustom cider-auto-jump-to-error t
-  "When non-nil automatically jump to error location during interactive compilation.
-When set to 'errors-only, don't jump to warnings."
+  "Control the cursor jump behaviour in compilation error buffer.
+
+When non-nil automatically jump to error location during interactive
+compilation.  When set to 'errors-only, don't jump to warnings."
```
bbatsov: `nil` is not mention here at all.

cider-interaction.el
```diff
@@ -272,11 +274,11 @@ not used as initial input (input is left empty)."
 ;;; Utilities
 
 (defun cider--clear-compilation-highlights ()
-  "Remove compilation highlights."
+  "Remove compilation highlight's."
```
bbatsov: Nope. :-) The original was correct, but checkdoc is a bit dumb.

cider-interaction.el
```diff
@@ -272,11 +274,11 @@ not used as initial input (input is left empty)."
...
   (remove-overlays (point-min) (point-max) 'cider-note-p t))
 
 (defun cider-clear-compilation-highlights (&optional arg)
-  "Remove compilation highlights.
+  "Remove compilation highlight's.
```
bbatsov: Same here.

cider-interaction.el
```diff
@@ -442,6 +444,8 @@ Invert meaning of `cider-prompt-for-symbol' if PREFIX indicates it should be."
     (nrepl-dict-get "path")))
 
 (defun cider--find-ns (ns &optional other-window)
+  "Find the file containing NS.
```
bbatsov: NS's definition

cider-interaction.el
```diff
@@ -738,9 +743,7 @@ Returns the position at which PROPERTY was found, or nil if not found."
 
 (defun cider-jump-to-compilation-error (&optional _arg _reset)
   "Jump to the line causing the current compilation error.
-
-_ARG and _RESET are ignored, as there is only ever one compilation error.
-They exist for compatibility with `next-error'."
+Optional arguments exist for compatibility with `next-error'."
```
bbatsov: I prefer the old version of this message.

cider-interaction.el
```diff
@@ -1218,7 +1221,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
   (cider-insert-in-repl (cider-defun-at-point) arg))
 
 (defun cider-insert-region-in-repl (start end &optional arg)
-  "Insert the curent region in the REPL buffer.
+  "Insert the curent region marked by START and END in the REPL buffer.
```
bbatsov: marked -> defined

cider-interaction.el
```diff
@@ -614,8 +618,9 @@ The handler simply inserts the result value in BUFFER."
 (defun cider--emit-interactive-eval-output (output repl-emit-function)
   "Emit output resulting from interactive code evaluation.
 
-The output can be send to either a dedicated output buffer or the current REPL buffer.
-This is controlled via `cider-interactive-eval-output-destination'."
+The OUTPUT can be sent to either a dedicated output buffer or the current
+REPL buffer.  This is controlled by `cider-interactive-eval-output-destination'.
+REPL-EMIT-FUNCTION formats the OUTPUT."
```
bbatsov: Does it really format it?

cider-interaction.el
```diff
@@ -1248,7 +1251,7 @@ See command `cider-mode'."
       (cider-mode +1))))
 
 (defun cider-disable-on-existing-clojure-buffers ()
-  "Disable `cider-mode' on existing Clojure buffers.
+  "Disable command `cider-mode' on existing Clojure buffers.
```
bbatsov: It seems to me that this sounds odd. This is why I intentionally ignore those warnings. Perhaps we should just disable them.

cider-interaction.el
```diff
@@ -1524,13 +1527,13 @@ the string contents of the region into a formatted string."
       (insert indented))))
 
 (defun cider-format-region (start end)
-  "Format the Clojure code in the current region."
+  "Format the Clojure code in the current region marked by START and END."
```
bbatsov: I'd rather have a second sentence saying "START and END represent the region's boundaries." everywhere.

cider-mode.el
```diff
@@ -375,7 +375,7 @@ The value can also be t, which means to font-lock as much as possible."
     value))
 
 (defun cider--compile-font-lock-keywords (symbols-plist core-plist)
-  "Return a list of font-lock rules for the symbols in SYMBOLS-PLIST."
+  "Return a list of font-lock rules for the symbols in SYMBOLS-PLIST and CORE-PLIST."
```
bbatsov: Probably we should explain here what those do.

bbatsov: The params I mean.

cider-stacktrace.el
```diff
@@ -308,8 +308,8 @@ hidden count."
       (goto-char pos))))
 
 (defun cider-stacktrace-cycle-cause (num &optional level)
-  "Update element NUM of `cider-stacktrace-cause-visibility', optionally to LEVEL.
-If LEVEL is not specified, its current value is incremented. When it reaches 3,
+  "Set element NUM of `cider-stacktrace-cause-visibility', optionally to LEVEL.
```
bbatsov: Retain the word "update" here.

cider-stacktrace.el
```diff
@@ -442,7 +442,9 @@ it wraps to 0."
     (back-to-indentation)))
 
 (defun cider-stacktrace-jump (&optional arg)
-  "Like `cider-find-var', but uses the stack frame source at point, if available."
+  "Like `cider-find-var', but use the test at point's definition, if available.
```
bbatsov: test at point?

cider-test.el
```diff
@@ -212,7 +212,9 @@
           (goto-char pos))))))
 
 (defun cider-test-jump (&optional arg)
-  "Like `cider-find-var', but uses the test at point's definition, if available."
+  "Like `cider-find-var', but use the test at point's definition, if available.
```
bbatsov: Not sure I like the change uses -> use

cider-test.el
```diff
@@ -212,7 +212,9 @@
...
+The prefix ARG and `cider-prompt-for-symbol' decide whether to
```
bbatsov: decide -> determine

cider.el
```diff
@@ -611,7 +613,8 @@ In case `default-directory' is non-local we assume the command is available."
                                cider-required-nrepl-version)))
 
 (defun cider--check-middleware-compatibility-callback (buffer)
-  "A callback to check if the middleware used is compatible with CIDER."
+  "A callback to check if the middleware used is compatible with CIDER.
+Use connection at BUFFER."
```
bbatsov: Connection at buffer? I don't get this.

nrepl-client.el
```diff
@@ -1252,7 +1253,8 @@ Set this to nil to prevent truncation."
   (delete-overlay button))
 
 (defun nrepl--expand-button-mouse (event)
-  "Expand the text hidden under overlay BUTTON."
+  "Expand the text hidden under overlay button.
+EVENT gives the button postition on window."
```
bbatsov: position

bbatsov:
I finally went through the PR myself and I've added a bunch of remarks. Frankly I've intentionally didn't fix some of the things @cskksc, as I think they add 0 value. Unfortunately flycheck doesn't honour dir-locals, as it runs outside the source dir, so I didn't bother customzing those. 

I'd suggest reducing the scope of the PR only to meaningful changes and dropping things like "command" prefixes and odd rewordings to fit something in 80 columns.

cider-test.el
Malabarba: Should I link you to the style guide? :-)

bbatsov: I get the point of the change, but this wording simply sounds odd. To properly use imperative further rewording is needed here IMO.

bbatsov: E.g. No docstring should start with "Like"...

.dir-locals.el
lunaryorn: @bbatsov When someone steps up to address https://github.com/flycheck/flycheck/issues/741.

cpitclaudel:
@bbatsov I should have the Flycheck thing fixed soon :)

bbatsov:
Fantastic news!

On Saturday, 2 April 2016, Clément Pit--Claudel <notifications@github.com>
wrote:

> @bbatsov <https://github.com/bbatsov> I should have the Flycheck thing
> fixed soon :)
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/clojure-emacs/cider/pull/1637#issuecomment-204762427>
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

Malabarba:
I only disagree WRT 80 columns. If a line is longer than that it will get when I open the help buffer on my laptop.

Malabarba:
*it will get truncated

cider-mode.el
cskksc: I'm don't understand the complete function. It decides different highlighting rules for core/non-core symbols ? Would this be ok ?

`"CORE-PLIST contain symbols either clojure.core or cljs.core, given by ```cider-resolve-core-ns.'
SYMBOLS-PLIST contain symbols from other namespaces."`

cskksc:
@bbatsov Sorry for the delay. I have changed most of the things. Even I thought that many of checkdoc's rules are too strict and don't add any value in this case. But I made those changes anyway because I am a n00b in elisp, and assumed it was just me; so checkdoc rules should not be wrong :-). There are loads of other things not touched in this PR. `checkdoc` complains on every occurrence of the word `tests` in `cider-test.el`.
We should definitely disable the plural thing and `command/variable` before we start enforcing checkdoc through travis. But I think a separate PR would be best for making all these changes and adding tests around it.

cider-stacktrace.el
```diff
@@ -442,7 +442,9 @@ it wraps to 0."
     (back-to-indentation)))
 
 (defun cider-stacktrace-jump (&optional arg)
-  "Like `cider-find-var', but uses the stack frame source at point, if available."
+  "Find definition for stack frame at point, if available.
+The prefix ARG and `cider-prompt-for-symbol' decide whether to
+prompt and whether to use a new window. Similar to `cider-find-var'"
```
cskksc: Is this ok ?

bbatsov: Yeah. But you have to have 2 spaces before `Similar` and you forgot the final `.`.

cider-test.el
```diff
@@ -212,7 +212,9 @@
           (goto-char pos))))))
 
 (defun cider-test-jump (&optional arg)
-  "Like `cider-find-var', but uses the test at point's definition, if available."
+  "Find definition for test at point, if available.
+The prefix ARG and `cider-prompt-for-symbol' decide whether to
+prompt and whether to use a new window. Similar to `cider-find-var'"
```
bbatsov: Same here.

cider-stacktrace.el
cskksc: Done.

cider-mode.el
bbatsov: We font-lock core symbols as built-ins. I'm guessing this is why they are treated differently here. @Malabarba can confirm this. Also core symbols are unlikely to change, so I think we cached their font-locking.

bbatsov:
I'll merge this in its present state. We can keep working on it afterwards.

Thanks for your work, @cskksc!

cskksc:
@bbatsov @Malabarba Thanks a lot! My first cider contribution :-)

bbatsov:
Hopefully not the last! ;-)

Malabarba:
Thank you indeed!

PR #1640, opened 2016-03-29
Update seq dependency to 2.14

NicolasPetton:
Updating seq.el to v 2.14 has the following benefits:

- The 2.x branch of seq now provides the same API for Emacs 25.1, Emacs master and Emacs 24.x.
- This version adds `seq-map-indexed` and `seq-sort-by`, which removes the need for `cider-map-indexed`.


bbatsov:
:+1: Thanks!

PR #1655, opened 2016-04-08
Doc improvements

Malabarba:
Most important change is that doc buffer splits arglists into several lines.

Instead of this:

    org.httpkit.client/post
    ([url & [opts callback]] [url & [callback]])
      Issues an async HTTP POST request. See `request` for details.

We get this:

    org.httpkit.client/post
     [url & [opts callback]]
     [url & [callback]]
      Issues an async HTTP POST request. See `request` for details.

Which I find much easier to read.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ an individual test using `C-c C-t t`.
 
 ### Changes
 
+* Doc buffer splits arglists into several lines.
 * Changed the face of the words “Macro” and “Special form” in the doc buffer to be easier to see.
 * Display multi-line eval overlays at the start of the following line. It looked weird that these overlays started on the middle of a line, but then folded onto the start of following lines.
 * [#1627](https://github.com/clojure-emacs/cider/issues/1627): Align the terminology used by `cider-test` with the one used by lein and boot (use the terms `assertion` and `test`).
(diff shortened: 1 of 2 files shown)


cider-doc.el
```diff
@@ -402,7 +402,7 @@ Tables are marked to be ignored by line wrap."
         (when (or forms args)
           (emit (cider-font-lock-as-clojure (or forms args))))
         (when (or special macro)
-          (emit (if special "Special Form" "Macro") 'font-lock-comment-face))
+          (emit (if special "Special Form" "Macro") 'font-lock-variable-name-face))
```
bbatsov: I'm guessing this is dependant on the theme.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ an individual test using `C-c C-t t`.
 
 ### Changes
 
+* Doc buffer splits arglists into several lines
```
bbatsov: Add a `.` here for consistency with the other entries.

bbatsov:
I like the change.

cider-doc.el
Malabarba: Yes, but the point was just to change it to something other than `comment`. Comment face is pretty hidden on a lot of themes.

In the end it's personal preference. I can revert if desired.

PR #1657, opened 2016-04-08
Display var docstrings in tooltips

Malabarba:


bbatsov:
👍 Nice!

bbatsov:
Btw, see the build failure about `x-gtk-use-system-tooltips`.

Issue #202, opened 2012-12-29
nrepl-eval-print-last-expression could not eval certain expressions


yuanmai:
If you type C-x C-e at the end of (with quotes): "redirect_uri=http%3A%2F%2Fb%2Fa"
it has no response.

PR #1658, opened 2016-04-10
Simplify middleware version compatibility check

sanjayl:

Thanks!

This mostly addresses CIDER-nREPL issue #202
https://github.com/clojure-emacs/cider-nrepl/issues/202. We can now
check for CIDER -> CIDER-nREPL compatibility without having to perform
an eval request and set up a callback handler to process that request.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -606,29 +606,18 @@ In case `default-directory' is non-local we assume the command is available."
                                "Can't determine nREPL's version.\nPlease, update nREPL to %s."
                                cider-required-nrepl-version)))
 
-(defun cider--check-middleware-compatibility-callback (buffer)
-  "A callback to check if the middleware used is compatible with CIDER.
-BUFFER specifies the connection buffer to be used."
-  (nrepl-make-response-handler
-   buffer
-   (lambda (_buffer result)
-     (let ((middleware-version (read result)))
-       (unless (and middleware-version (equal cider-version middleware-version))
-         (cider-repl-manual-warning "installation/#ciders-nrepl-middleware"
-                                    "CIDER's version (%s) does not match cider-nrepl's version (%s). Things will break!"
-                                    cider-version middleware-version))))
-   '()
-   '()
-   '()))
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -607,29 +607,20 @@ In case `default-directory' is non-local we assume the command is available."
...
-      (require 'cider.nrepl.version)
-      (:version-string @(resolve 'cider.nrepl.version/version))
-    (catch Throwable _ \"not installed\"))"
-   (cider--check-middleware-compatibility-callback (current-buffer))))
+  "CIDER frontend/backend compatibility check.
+Retrieve the underlying connection's CIDER-nREPL version and checks if the
+middleware used is compatible with CIDER.  If not, will display a warning
+message in the REPL area."
+  (let* ((version-dict        (nrepl-aux-info "cider-version" (cider-current-connection)))
+         (middleware-version* (nrepl-dict-get version-dict "version-string"))
```
bbatsov: This reminds me that we can simply add support for a `default-value` to `nrepl-dict-get` (same as with Clojure maps). That'd be trivial and would yield much nicer code in similar situations.

sanjayl: Do you want identical semantics to Clojure's `get` (ie, if map `contains?` key, and its assigned value is `nil`, return `nil` regardless of a default value being supplied) or just a simple "return the default value if the `nrepl-dict-get` returns `null`"?

bbatsov: I was thinking of the former, but probably the simpler implementation would be fine as well.

sanjayl: Since the elisp `get-hash` also employs the first behavior, it probably would be less confusing overall if we did it the more complicated way. Let me work on that separately so I don't cluttering this PR. In the meantime, I've simplified my code to `(or (nrepl-dict-get ....) default)`

sanjayl:
(test failures are on master, don't think they're related to this branch)

bbatsov:
👍

Issue #1650, opened 2016-04-03
Replace ert with buttercup


bbatsov:
ert gets the job done, but it does little to enforce good test structure and to help with code reuse. [buttercup](https://github.com/jorgenschaefer/emacs-buttercup) is an RSpec-like test framework which seems way more capable (and modern).

cskksc:
I'd like to give this a try. I have not used `buttercup` before,  but have worked quite a lot with rspec, so I think I can work this out. I'll submit a PR soon, with some of the tests changed, and then lets see if my level to elisp knowledge is sufficient to tackle this.

bbatsov:
Great!

Malabarba:
🎉🎉🎊🎊

phillord:
If I may be permitted a plug, I've been working an http://github.com/phillord/assess which provides a set of stateless predicates (and ert-explainer functions, although you are perhaps less interested in that).

bbatsov:
@cskksc Do you plan to finish the migration? Now we're in a nasty place using two systems and should try to kill ert as soon as possible.

cskksc:
@bbatsov Sorry for the delay. Only 1 file `cider-tests` is pending. I'll finish that in the next 2 days.

bbatsov:
Great!

On 26 April 2016 at 20:17, Chaitanya Koparkar <notifications@github.com>
wrote:

> @bbatsov <https://github.com/bbatsov> Sorry for the delay. Only 1 file
> cider-tests is pending. I'll finish that in the next 2 days.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/clojure-emacs/cider/issues/1650#issuecomment-214959196>
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

vspinu:
I am late to the party but all I have noticed so far that buttercup is close to impossible to use interactively. 

With ert I could run each test interactively, run all all loaded tests and step-through with debugger on each test. With buttercup nothing of that seem to be available. There is one interactive function `buttercup-run-at-point` which does nonsence with cider tests (splits windows in N buffers, resets buffers states, doesn't report failed tests correctly).  

How do you use it correctly?  Run full suite after each test?

dpsutton:
[...]

PR #1659, opened 2016-04-10
Change some tests to use buttercup

cskksc:

Thanks!


cskksc:
Start fixing #1650 
A bulk of the tests are still not migrated. I'll submit separate PR's for some of the larger files.

test/cider-apropos-tests.el
```diff
@@ -1,20 +1,17 @@
...
+(require 'buttercup)
 (require 'cider)
 (require 'cider-apropos)
 
-(ert-deftest cider-apropos-not-connected ()
-  (noflet ((cider-connected-p () nil))
-    (should-error (cider-apropos "test") :type 'user-error)))
+(describe "cider-apropos"
+  (it "raises user-error when cider is not connected."
+    (expect (lambda () (cider-apropos "test")) :to-throw 'user-error))
```
Malabarba: Is this lambda really necessary? Doesn't it work if you just write the code to be executed?

cskksc: I believe it is necessary for the `to-throw` matcher. Otherwise the error is not handled and the tests fail with `error: (user-error "`nil' needs an active nREPL connection")`.

test/cider-apropos-tests.el
```diff
@@ -1,20 +1,17 @@
-(require 'ert)
-(require 'noflet)
+(require 'buttercup)
 (require 'cider)
 (require 'cider-apropos)
 
-(ert-deftest cider-apropos-not-connected ()
-  (noflet ((cider-connected-p () nil))
```
cskksc: > Otherwise they may fail if some other tests happened to create a connection (or if you run them interactively, instead of from a script).

Ah, I dint know about this. Initially I had added the mocks using `fset`, but the tests seemed to work even without them. I also removed the `noflet` mocks and ran the old tests, and they worked too.
WDYT is better, to add mocks similar to `noflet`, or set the connections to nil ?

bbatsov: buttercup has it's builtin mocking capabilities. I prefer to mock the functions themselve instead of let-binding the connections.

test/cider-error-parsing-tests.el
```diff
@@ -1,77 +1,84 @@
...
-      (should (string= (nth 0 info) "/some/test/file/core.clj"))
-      (should (= (nth 1 info) 31))
-      (should (equal (nth 2 info) nil))
-      (should (equal (nth 3 info) 'cider-error-highlight-face)))))
+(describe "cider-extract-error-info"
+  (before-all
+    (fset 'file-name (lambda (info) (nth 0 info)))
+    (fset 'line-num (lambda (info) (nth 1 info)))
+    (fset 'col-num (lambda (info) (nth 2 info)))
+    (fset 'face (lambda (info) (nth 3 info))))
```
cskksc: Done. I believe the `:var` would create the functions only for the scope of `describe` and won't affect the environment.

test/cider-apropos-tests.el
cskksc: Done.

bbatsov:
You'll have to update the travis config to run the buttercup tests and the ert tests. Otherwise we can't make a staged migration to buttercup.

test/cider-error-parsing-tests.el
Malabarba: > Done. I believe the :var would create the functions only for the scope of describe and won't affect the environment.

I have no idea if that's true or not, so I'll trust your judgement. :-)

cskksc:
Done. Have squashed everything into a single commit and rebased against the latest master.

test/cider-error-parsing-tests.el
cskksc: Basing my assumption on the `let` form being used in the [source](https://github.com/jorgenschaefer/emacs-buttercup/blob/master/buttercup.el#L361)

cskksc:
I've changed some more files, but din't include them here because this PR has gotten big enough. Will submit follow up PR's.

bbatsov:
> Also, should we mention the migration somewhere in the contributing guide, or somewhere else ?

After the migration is ready we can add something about this in the docs. Hopefully the migration won't take long - we don't have many tests.

Malabarba:
Cool. Thanks for the effort and patience. :-)

cskksc:
Thanks @Malabarba 🎉

PR #1660, opened 2016-04-11
Debugger: step-in

grammati:
This is the front-end part of step-in for the debugger.

See: https://github.com/clojure-emacs/cider-nrepl/pull/335


cider-debug.el
```diff
@@ -582,11 +583,14 @@ is a coordinate measure in sexps."
           ;; But we can create a temp buffer if that fails.
           (or out
               (when create-if-needed
-                (cider--initialize-debug-buffer
-                 code ns original-id
-                 (if (and line column)
-                     "you edited the code"
-                   "your tools.nrepl version is older than 0.2.11"))
+                (condition-case nil
```
bbatsov: This bit of code would benefit from some comment explaining its purpose.

bbatsov:
You'll also have to update the manual and the changelog when you're ready with the final UI for this feature.

grammati:
This seems to work, in combination with the PR for step-in in cider-nrepl.

My elisp-fu is not strong, so I would appreciate careful review :) thanks

If it looks OK, I'll add something to the changelog and the docs.

cider-debug.el
```diff
@@ -185,8 +185,8 @@ This variable must be set before starting the repl connection."
 
 
 ;;; Minor mode
-(defvar-local cider--debug-mode-commands-alist nil
-  "Alist from keys to debug commands.
+(defvar-local cider--debug-mode-commands-dict nil
+  "nrepl-dict from keys to debug commands.
```
bbatsov: Prefix this with an `An`.

cider-debug.el
```diff
@@ -214,16 +214,17 @@ Each element of LOCALS should be a list of at least two elements."
...
-          (concat
-           (mapconcat (lambda (x) (put-text-property 0 1 'face 'cider-debug-prompt-face x) x)
-                      command-list
-                      " ")
-           "\n")))
+  (format (propertize "%s\n" 'face 'default)
+          (apply #'concat
+                 (nrepl-dict-map (lambda (char cmd)
+                                   (when-let ((pos (search char cmd)))
+                                     (put-text-property pos (+ 1 pos) 'face 'cider-debug-prompt-face cmd))
```
bbatsov: There's a function called `1+`. :-)

cider-debug.el
```diff
@@ -313,13 +310,13 @@ In order to work properly, this mode must be activated by
...
             ;; Set the keymap.
-            (let ((alist (mapcar (lambda (k) (cons (string-to-char k) (concat ":" k)))
-                                 ;; `here' needs a special command.
-                                 (seq-difference input-type '("here")))))
-              (setq cider--debug-mode-commands-alist alist)
-              (dolist (it alist)
-                (define-key cider--debug-mode-map (vector (upcase (car it))) #'cider-debug-mode-send-reply)
-                (define-key cider--debug-mode-map (vector (car it)) #'cider-debug-mode-send-reply)))
+            (nrepl-dict-map (lambda (char cmd)
+                              (when (not (string= char "h")) ; `here' needs a special command.
```
bbatsov: when + not == unless

cider-debug.el
```diff
@@ -394,12 +391,14 @@ message."
...
                   (ignore-errors
-                    (cdr (assq (downcase last-command-event)
-                               cider--debug-mode-commands-alist))))
+                    (nrepl-dict-get cider--debug-mode-commands-dict
+                                    (downcase (string last-command-event)))))
                 nil
                 (ignore-errors
                   (let ((case-fold-search nil))
                     (string-match "[[:upper:]]" (string last-command-event))))))
+  (when (not (string= ":" (substring command 0 1)))
```
bbatsov: unless
There's also `string-prefix-p`.

bbatsov:
Apart from my small remarks, the code looks OK.

bbatsov:
Your changes broken one unit test and the branch doesn't seem to be in sync with `master`.

grammati:
Hmm, now travis is saying "function `search' from cl package called at runtime". Is there a different function I should use for "String.indexOf"?

bbatsov:
You can use `cl-search` or `seq-position`. Generally we prefer to use `seq` functions.

grammati:
OK, I finally got the tests to pass!

cider-debug.el
Malabarba: I think we can discard this whole change if we just replace that `(find-buffer-visiting file)` with a 

    (ignore-errors
      (cider--jump-to-loc-from-info response)
      (current-buffer))

CHANGELOG.md
```diff
@@ -25,6 +26,7 @@
 
 ### Changes
 
+* Changed the data structure used to send available debug commands. Breaking change to the debugger if not updated in sync with the corresponding change to cider-nrepl.
```
bbatsov: Not sure if this is worth mentioning as most people won't notice it at all.

cider-debug.el
bbatsov: In simplicity I trust. :-)

grammati:
OK, I made the suggested changes. This simplified code for opening the buffer works great, thanks @Malabarba .

Malabarba:
It looks good to me!

bbatsov:
👍

PR #1662, opened 2016-04-11
Add default value to nrepl-dict-get

sanjayl:

Thanks!

Make the nrepl-dict hashes act similarly to the Clojure maps by giving
them an optional default return value. If the provided key is not found
in the dict, then the optional default value will be returned. Notably,
the nil value is allowed as both a key and a value in nrepl-dict's.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -301,12 +301,22 @@ Bind the value of the provided KEYS and execute BODY."
   "Return t if nREPL dict DICT is empty."
   (null (cdr dict)))
 
-(defun nrepl-dict-get (dict key)
-  "Get from DICT value associated with KEY.
-If dict is nil, return nil."
+(defun nrepl-dict-contains (dict key)
+  "Return nil if nREPL dict DICT doesn't contain KEY.
+If DICT does contain KEY, then a non-nil value is returned.  Due to the
+current implementation, this return value is the tail of DICT's key-list
+whose car is KEY.  Comparison is done with `equal'."
+  (member key (nrepl-dict-keys dict)))
+
+(defun nrepl-dict-get (dict key &optional default)
+  "Get from DICT value associated with KEY, optional DEFAULT if KEY not in DICT.
+If dict is nil, return nil.  If DEFAULT not provided, and KEY not in DICT,
+return nil.  If DICT is not an nREPL dict object, an error is thrown."
   (when dict
     (if (nrepl-dict-p dict)
-        (lax-plist-get (cdr dict) key)
+        (if (nrepl-dict-contains dict key)
+            (lax-plist-get (cdr dict) key)
+          default)
       (error "Not an nREPL dict object: %s" dict))))
 
 (defun nrepl-dict-put (dict key value)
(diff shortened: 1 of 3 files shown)


sanjayl:
Once these changes get verified, I'll make the change to the utilizing client as discussed prior.

(A couple of test failures that should be unrelated to this code are present)

nrepl-client.el
```diff
@@ -301,12 +301,22 @@ Bind the value of the provided KEYS and execute BODY."
...
   (null (cdr dict)))
 
-(defun nrepl-dict-get (dict key)
-  "Get from DICT value associated with KEY.
-If dict is nil, return nil."
+(defun nrepl-dict-contains-p (dict key)
+  "Return t if nREPL dict DICT contains KEY.
+Akin to to `clojure.core/contains?`."
+  (if (member key (nrepl-dict-keys dict))
+      't
```
bbatsov: There's no need to quote `t` - it's special.

bbatsov:
Apart from my small remark, the changes look good.

bbatsov:
I've just fixed the broken build.

nrepl-client.el
```diff
@@ -301,12 +301,22 @@ Bind the value of the provided KEYS and execute BODY."
   "Return t if nREPL dict DICT is empty."
   (null (cdr dict)))
 
-(defun nrepl-dict-get (dict key)
-  "Get from DICT value associated with KEY.
-If dict is nil, return nil."
+(defun nrepl-dict-contains-p (dict key)
+  "Return t if nREPL dict DICT contains KEY.
+Akin to to `clojure.core/contains?`."
```
expez: an extra `to` here

nrepl-client.el
Malabarba: There's no need for the `if` either. The body of the function would be more idiomatic as just `(member key (nrepl-dict-keys dict))`.

bbatsov: But it won't return `t` or `nil` then.

Malabarba: A lot of Elisp predicates don't worry about converting non-nil values to t, they just return whatever is most convenient. 
Maybe I'm misremembering this. I can check it out when I'm back on the computer. 

Besides, you're adding code and hiding information without getting any benefit (the code that uses the return value can just use it as if it were t/nil).

bbatsov: I get your point and I've had this argument many times for various languages. :-) For me predicates deal with booleans, but I can I'm fine with dropping the `if` here.

Malabarba: I struggle with this in my head every time I see people convert to truthy in Clojure code. But in that case I just silently resign and accept that it's customary to do it in Clojure. :-)

sanjayl: I think that all the type predicates return `t` or `nil`, as do almost all the `seq-` predicate functions. However, the collection membership functions usually return a truthy object (including `seq-contains`). I was trying to make this function behave like `clojure.core/contains?`, but since this is elisp and not clojure, I'm going to change the name to `nrepl-dict-contains` and return truthy in order to be in line with similar functions (`member`, `seq-contains`, `plist-member`, etc).

PR #1663, opened 2016-04-11
Tag and remove cider-test overlays only

### LINKED ISSUE #1643: Running tests clears all overlays in the buffer
## Expected behavior

I have a Clojure test buffer with overlays.  Launching tests will not touch these overlays and do the necessary bookkeeping for CIDER's overlays.

## Actual behavior

All overlays are removed.  [This](https://github.com/clojure-emacs/cider/blob/master/cider-test.el#L501-L502) looks like the culprit.

- Open a Clojure test buffer
- Enable a minor mode that adds its own overlays to the current buffer (I'm using [column-enforce-mode](https://github.com/jordonbiondo/column-enforce-mode/) and a buffer with long lines)
- `C-c ,`

### CIDER version information

```
; CIDER 0.10.0snapshot (package: 20151125.2330) (Java 1.8.0_74, Clojure 1.7.0, nREPL 0.2.12)
```

### Emacs version

Emacs 24.5.1

### Operating system

Arch Linux

wasamasa:

Thanks!

Fixes #1643

---

I couldn't spot any tests for `cider-test.el`, so I didn't add any.  Can't run tests at the moment, but will do later.

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -33,6 +33,7 @@
 (require 'cider-popup)
 (require 'cider-stacktrace)
 (require 'cider-compat)
+(require 'cider-overlays)
 
 (require 'button)
 (require 'easymenu)
@@ -468,13 +469,13 @@ The optional arg TEST denotes an individual test name."
           (search-forward "(" nil t)
           (let ((beg (point)))
             (forward-sexp)
-            (let ((overlay (make-overlay beg (point))))
-              (overlay-put overlay 'font-lock-face (cider-test-type-face type))
-              (overlay-put overlay 'type type)
-              (overlay-put overlay 'help-echo message)
-              (overlay-put overlay 'message message)
-              (overlay-put overlay 'expected expected)
-              (overlay-put overlay 'actual actual))))))))
+            (cider--make-overlay beg (point) 'cider-test
+                                 'font-lock-face (cider-test-type-face type)
+                                 'type type
+                                 'help-echo message
+                                 'message message
+                                 'expected expected
+                                 'actual actual)))))))
 
 (defun cider-find-var-file (ns var)
   "Return the buffer visiting the file in which the NS VAR is defined.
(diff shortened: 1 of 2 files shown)


cider-test.el
```diff
@@ -469,6 +469,7 @@ The optional arg TEST denotes an individual test name."
           (let ((beg (point)))
             (forward-sexp)
             (let ((overlay (make-overlay beg (point))))
```
Malabarba: Could you refactor this code to use `cider--make-overlay` instead?
That function already adds a `cider-type` property that serves the same purpose as `category` here. (We can also change that function to use `category` instead of `cider-type` if that's more idiomatic).

cider-test.el
```diff
@@ -507,7 +508,10 @@ Or nil if not found."
        (dolist (var (nrepl-dict-keys vars))
          (when-let ((buffer (cider-find-var-file ns var)))
            (with-current-buffer buffer
-             (remove-overlays)))))
```
Malabarba: You can just do this here: `(remove-overlays (point-min) (point-max) 'category 'cider-test)`.

cider-test.el
wasamasa: That would indeed be more idiomatic, but not really part of this fix.  What I'm more worried about is whether this catches all overlays this code adds to the Clojure buffer.

Malabarba: Yes, it should catch everything. But you don't have to do it now. It's mostly tangential to this PR so I can do it after we merge.

wasamasa:
I've updated the code as suggested and ran the tests.  To my disappointment, nothing broke.

Malabarba:
:smile: 
Now you just need to add a `(require 'cider-overlays)` at the top of that file.

cider-test.el
```diff
@@ -507,7 +507,8 @@ Or nil if not found."
...
+             (remove-overlays (point-min) (point-max)
+                              'cider-type 'cider-test)))))
```
bbatsov: 'cider-type?

wasamasa: That's how an overlay made with `cider--make-overlay` is tagged, yes.

bbatsov: OK, guess I missed this.

wasamasa:
Thanks!  Time to upgrade to a non-stable release I guess...

bbatsov:
Guess so. :-)

PR #1679, opened 2016-04-14
Fix force-step operations in the debugger

grammati:
Two operations, `:here` and `:out` currently support a `force?`
parameter, which causes the debugger to skip breakpoints in other
instrumented functions until reaching the specified point.

They are bound to uppercase letters, `H` and `O` respectively. Neither
was working previously, because the `force?` parameter was not being
sent (in the case of `:here`), or was being sent incorrectly (in the
case of `:out`).

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -343,7 +343,10 @@ In order to work properly, this mode must be activated by
     (when nrepl-ongoing-sync-request
       (ignore-errors (exit-recursive-edit)))))
 
+;;; Bind the `:here` command to both h and H, because it behaves differently if
+;;; invoked with an uppercase letter.
 (define-key cider--debug-mode-map "h" #'cider-debug-move-here)
+(define-key cider--debug-mode-map "H" #'cider-debug-move-here)
 
 (defun cider--debug-remove-overlays (&optional buffer)
   "Remove CIDER debug overlays from BUFFER if variable `cider--debug-mode' is nil."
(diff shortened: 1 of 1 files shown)


cider-debug.el
```diff
@@ -344,6 +344,7 @@ In order to work properly, this mode must be activated by
       (ignore-errors (exit-recursive-edit)))))
 
 (define-key cider--debug-mode-map "h" #'cider-debug-move-here)
+(define-key cider--debug-mode-map "H" #'cider-debug-move-here)
```
bbatsov: Same command, different keybinding?

cider-debug.el
```diff
@@ -693,9 +696,12 @@ TARGET is inside it.  The returned list is suitable for use in
...
         (goto-char starting-point)))))
 
-(defun cider-debug-move-here ()
+(defun cider-debug-move-here (&optional force)
   "Skip any breakpoints up to point."
-  (interactive)
+  (interactive (list
+                (ignore-errors
+                  (let ((case-fold-search nil))
+                    (string-match "[[:upper:]]" (string last-command-event))))))
```
bbatsov: What's this supposed to do?

grammati: I copied this part from `cider-debug-mode-send-reply`, which all other debug commands (other than `:here`) are bound to. The `force` argument is true only if the command was called with an uppercase key.

bbatsov: Got it. Guess we can extract this code to a small helper function to remove the duplication.

Also we should mention this behaviour in the manual, as I'm pretty sure right now you and @Malabarba are the only people aware this exists. :-)

cider-debug.el
Malabarba: Yes. The command detects whether it was invoked with shift.

bbatsov: I understood this when I got a bit farther into the code. Might be useful to add a comment here, though.

grammati:
I added a comment and deleted some unnecessary code that I missed the first time.

bbatsov:
@grammati and what about the duplicated code in `(interactive)`?

grammati:
I moved it into a function, `uppercase-command`.

cider-debug.el
```diff
@@ -383,6 +386,12 @@ In order to work properly, this mode must be activated by
      ["List locals" cider-debug-toggle-locals :style toggle :selected cider-debug-display-locals])
     ["Customize" (customize-group 'cider-debug)]))
 
+(defun uppercase-command ()
```
bbatsov: Rename this to `cider--uppercase-command-p` or something like this. There are no namespace in Elisp and we have to be careful with unprefixed names. `-p` means `?` in Elisp.

grammati: Done!

PR #1690, opened 2016-04-18
Continue ert to buttercup migration

cskksc:
The file `cider-tests.el` is a huge one. Still working on it. Also, does it make sense to use `undercover` to check test coverage ? Not for the % number, but it could show up a lot of places where the tests are missing.


Malabarba:
Note that you've got an old version of the debug tests. In particular, the first test is not going to pass.

bbatsov:
For consistency it might be a good idea to have every migrated test file as a separate commit in the PR.

cskksc:
Done. @Malabarba I rebased with the latest master at the last moment and forgot to run the tests 😞.

Does `declare-function` work differently on emacs 24.3 ? The travis test failed for just that emacs version. 
https://travis-ci.org/clojure-emacs/cider/jobs/124124560#L1933

relevant file: https://github.com/clojure-emacs/cider/blob/master/cider-selector.el#L159
Any hints on fixing it ?

Malabarba:
`declare-function` is only used to fix compiler warnings, so it's not going to fix any errors. (We do have a build that checks for warnings, but that is only one build and it runs agains 24.5).
The problem is probably something else

bbatsov:
Guess some require is missing, although I don't understand why this is working on the other versions.

cskksc:
Ok. Guess I'm gonna have to install 24.3 and debug it there 😛 . I'll do this today.

bbatsov:
@cskksc Any progress?

cskksc:
I've managed to replicate it locally. You were right, requiring `cider-selector` fixed it, but now a related test is failing on travis. Will definitely finish this today.

test/cider-selector-tests.el
```diff
@@ -7,69 +7,59 @@
...
-          (should (not (equal (current-buffer) b1)))
-          (cider-invoke-selector-method-by-key ?m)
-          (should (equal (current-buffer) b1)))))))
+(describe "cider-seletor-method-c"
+  (it "switches to most recently visited clojure-mode buffer"
+    (cider--test-selector-method ?c 'clojure-mode "*testfile*.clj")))
+
+(describe "cider-seletor-method-e"
+  (it "switches to most recently visited emacs-lisp-mode buffer"
+    (kill-buffer "*scratch*")
```
cskksc: This is a bit of a workaround, but I couldn't replicate the travis emacs behaviour locally. The most recently selected elisp buffer there, was always `*scratch*`. I tried, `set-buffer`, `with-current-buffer`, and `set-current-buffer`.

cskksc:
@bbatsov Done :-)

test/cider-selector-tests.el
bbatsov: Hmm, that's odd.

PR #1691, opened 2016-04-18
[Fix #1659] Add fringe indicators for evaluated forms

Malabarba:
Out of all the fringe bitmap options, these 4 are the ones that seem most appropriate.
Not sure which one to use though.
screenshot-2016-04-18--194945-0300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New Features
+
+* Fringe indicators highlight which sexps have been loaded. Disable it with `cider-use-fringe-indicators`.
+
 ### Changes
 
 * Signal an error sooner if the user misconfigured `cider-known-endpoints`.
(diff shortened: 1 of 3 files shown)


bbatsov:
I like option 1 and 4 the most.

expez:
Does the face also work as a setting? E.g. if I set the face to `nil` it won't draw anything?

Malabarba:
@expez No. We'll have to make a variable for that.

bbatsov:
I'm guessing we'll have one for evaluations which resulted in an error. An maybe one for evals that produced warnings (not sure about this one).

We should also mention this in the changelog.

Malabarba:
I've only implemented the evaluation indicator for now. I'd like to do the error too eventyually.
There's also a custom option to disable this.

vspinu:
What was the motivation for this? I always start with loading the entire buffer, thus all forms end up having this fringe marker. To me it seems like a useless functionality which also interferes with other packages (#2018). 

I am looking into fixing #2018 but I think for the long run it would be better to entirely remove this.

Issue #1561, opened 2016-02-13
Color symbols in the ns-browser like with cider-font-lock-dynamically


Malabarba:
As the title says, font-lock vars, macros, and functions in the ns-browser as they would be in a Clojure buffer.

PR #1695, opened 2016-04-19
Font-lock vars, macros, and functions in the ns-browser

cskksc:
For #1561 
I still have to write the tests, but does this make sense ?

Before vs after:
ns-browser

Also, should `nrepl-dict*` be their own thing ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
 
 * Use `cider-apropos-select` instead of `cider-apropos` in `cider-apropos-documentation-select`.
 
+* [#1561] Use an appropriate font-lock-face for variables, macros and functions in
+the ns-browser.
+
+
 ## 0.12.0 (2016-04-16)
 
 ### New Features
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -39,6 +39,7 @@
 (require 'cider-client)
 (require 'cider-compat)
 (require 'cider-util)
+(require 'nrepl-client)
 
 (defconst cider-browse-ns-buffer "*cider-ns-browser*")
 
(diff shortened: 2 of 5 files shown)


Malabarba:
I noticed that you're running `cider-sync-request:info` for each var. Does that make the browser slow to load for large namespaces like `clojure.core`?

cskksc:
Good point. Its too slow, ~17 secs for `clojure.core`. I remember @bbatsov saying something about cider caching `clojure.core` and `cljs.core` font-locks. That needs to be used here. I'll see if I can find that.
But this runs in under a second for most namespaces.

| ns             |  time |
| :---           | :---: |
| clojure.test   |  0.98 |
| zookeeper      |   0.7 |
| clojure.set    |  0.35 |
| clojure.string |   0.5 |

Malabarba:
Indeed, I think we should use the cache instead. The REPL buffer
contains a variable called `cider-repl-ns-cache`. Inside it you will
find the desired information (arglist/macro) for a lot of namespaces
(not just clojure.core), and it should be 1000s of times faster to
access.

## Option 1

You can get a plist that maps var-names to their metadata for all vars
in a namespace, by calling `(cider-resolve-ns-symbols "the.ns.name")`.
Then you can get the metadata for a specific var by running 
`(lax-plist-get the-plist var-name)`.

## Option 2

You can also get the metadata of one var at a time (similar to what
you're doing now, but without going through the middleware) by doing
`(cider-resolve-var "the.ns.name" "the-var-name")`.

cskksc:
Thanks. This is awesome!

| ns             |  time |
| :---           | :---: |
| clojure.core   |  0.16 |
| zookeeper      |   0.03 |

One problem we have with this is that, some of the namespaces returned by `cider-sync-request:ns-list` are not present in the cache. They're on the classpath, but not loaded in the repl. So the font-locks for those are not correct. How can the cache be updated in such a case ? I saw that the fn `cider-repl--state-handler` sets the cache, but after which response is this callback invoked, or after all of them?

Malabarba:
Yeah, the state_tracker middleware automatically tracks namespaces written or required by the user, but not other namespaces. There's no way to manually update the cache right now. But I can fix that afterwards. For now, I think that proper font-locking for some namespaces is better than for none, so maybe we can merge. 
WDYT @bbatsov ?

bbatsov:
Guess we can have some special face for things that are not yet resolved. Seems like a reasonable middle ground to me.

Or we can fallback to a separate eval for things not in the cache. Btw, if we can batch the evals together they should be slow. A single nREPL request should take pretty little time as most of the overhead you got originally was from the separate requests. We need a middleware op that returns the info for a bunch of vars.

cskksc:
Sounds good. I'll add it there. Right now the default is `font-lock-type-face`, which is the font-lock-face for namespaces. I'll change it to `font-lock-function-name-face`. Things are definitely more likely to be a function than anything else. Just need to add some tests and this PR can be merged I think. I'll raise another one with the nrepl fix.

Malabarba:
We can also change the op used by browse-ns. 
Since we're already fetching a list of all vars, might as well return some metadata with it.

Why not add an op `ns-vars-with-meta`, that is like `ns-vars` but returns a map from symbols to metadata (filtered with `cider.nrepl.middleware.track-state/relevant-meta`). This way we'd have all information from one place.

bbatsov:
> Why not add an op ns-vars-with-meta, that is like ns-vars but returns a map from symbols to metadata (filtered with cider.nrepl.middleware.track-state/relevant-meta). This way we'd have all information from one place.

👍

cskksc:
I've added the tests for this. Can be merged IMO.
@Malabarba  Nice! It'll certainly save one extra middleware call. I'll get started on this soon.

bbatsov:
@cskksc You should mention this changes in the changelog.

cskksc:
Done :-). I've just added it in bugs fixed, not sure if that's right. It isn't completely fixed yet.
Should I close this one and file another issue with the nrepl integration or continue on this one ?

bbatsov:
> Should I close this one and file another issue with the nrepl integration or continue on this one ?

I'm fine either way.

PR #1697, opened 2016-04-19
Inject minimal clojure version as a dependency.

phillord:

Thanks!

Haven't done all of that -- this is just a starter-for-ten, as it explains the idea better than raising an issue. Happy to add all of it, if you think the idea is sound.

Given that cider requires 1.7, when not inject it? At the moment, if you open a project which requires less than 1.7, the middleware fails to load, and cider fails with an unhelpful error message. I don't want to require a latter version of clojure in my project (although I don't mind running inside one).

Other options: make the error message better; change cider-nrepl from a provided dependency (which I think would cause lein to load 1.7 when cider-nrepl is injected. That's all I can think of.

Cider-nrepl has a provided dependency of Clojure 1.7. Projects can now
have a Clojure version lower than 1.7 therefore fail to run cider
correctly, unless this dependency is added. Injection means that cider
can support these projects without requiring them to update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Fringe indicators highlight which sexps have been loaded. Disable it with `cider-use-fringe-indicators`.
 * New command: `cider-inspect-last-result`.
 * `cider-cljs-lein-repl` now also supports figwheel.
+* Option `cider-jack-in-auto-inject-clojure` enables the user to specify a
+  version of Clojure for CIDER. This allows the user to override the version
+  used in a project, particular if it is lower than minimum required for CIDER.
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -220,8 +220,14 @@ namespace-qualified function of zero arity."
   "Face used to highlight compilation warnings in Clojure buffers."
   :group 'cider)
 
+(defconst cider-clojure-artifact-id "org.clojure/clojure"
+  "Artifact identifier for Clojure.")
+
 (defconst cider-minimum-clojure-version "1.7.0"
   "Minimum supported version of Clojure.")
+
+(defconst cider-latest-clojure-version "1.8.0"
+  "Latest supported version of Clojure.")
 
 (defconst cider-required-nrepl-version "0.2.12"
   "The minimum nREPL version that's known to work properly with CIDER.")
--- a/cider.el
+++ b/cider.el
@@ -229,6 +229,23 @@ Sub-match 1 must be the project path.")
 (cider-add-to-alist 'cider-jack-in-dependencies
                     "org.clojure/tools.nrepl" "0.2.12")
 
+(defcustom cider-jack-in-auto-inject-clojure nil
+  "Version of clojure to auto-inject into REPL.
+
+If nil, do not inject clojure into the REPL.  If `latest', inject
+`cider-latest-clojure-version', which should approximate to the most recent
+version of clojure.  If `minimal', inject `cider-minimum-clojure-version',
+which will be the lowest version cider supports.  If a string, use this as
+the version number.  If it is a list, the first element should be a string,
+specifying the artifact ID, and the second element the version number."
+  :type '(choice (const :tag "None" nil)
+                 (const :tag "Latest" 'latest)
+                 (const :tag "Minimal" 'minimal)
+                 (string :tag "Specific Version")
+                 (list :tag "Artifact ID and Version"
+                       (string :tag "Artifact ID")
+                       (string :tag "Version"))))
+
 (defvar cider-jack-in-lein-plugins nil
   "List of Leiningen plugins where elements are lists of artifact name and version.")
 (put 'cider-jack-in-lein-plugins 'risky-local-variable t)
@@ -288,6 +305,24 @@ string is quoted for passing as argument to an inferior shell."
    " -- "
    params))
 
+(defun cider-add-clojure-dependencies-maybe (dependencies)
+  "Return DEPENDENCIES with an added Clojure dependency if requested.
+
+See also `cider-jack-in-auto-inject-clojure'."
+  (if cider-jack-in-auto-inject-clojure
+      (if (consp cider-jack-in-auto-inject-clojure)
+          (cons cider-jack-in-auto-inject-clojure dependencies)
+        (cons (list cider-clojure-artifact-id
+                    (cond
+                     ((stringp cider-jack-in-auto-inject-clojure)
+                      cider-jack-in-auto-inject-clojure)
+                     ((eq cider-jack-in-auto-inject-clojure 'minimal)
+                      cider-minimum-clojure-version)
+                     ((eq cider-jack-in-auto-inject-clojure 'latest)
+                      cider-latest-clojure-version)))
+              dependencies))
+    dependencies))
+
 (defun cider-inject-jack-in-dependencies (params project-type)
   "Return PARAMS with injected REPL dependencies.
 These are set in `cider-jack-in-dependencies', `cider-jack-in-lein-plugins' and
@@ -298,11 +333,13 @@ dependencies."
   (pcase project-type
     ("lein" (cider-lein-jack-in-dependencies
              params
-             cider-jack-in-dependencies
+             (cider-add-clojure-dependencies-maybe
+              cider-jack-in-dependencies)
              cider-jack-in-lein-plugins))
     ("boot" (cider-boot-jack-in-dependencies
              params
-             cider-jack-in-dependencies
+             (cider-add-clojure-dependencies-maybe
+              cider-jack-in-dependencies)
              cider-jack-in-lein-plugins
              cider-jack-in-nrepl-middlewares))
     ("gradle" params)
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -584,6 +584,34 @@
                      "-d org.clojure/tools.nrepl\\:0.2.12 -d cider/cider-nrepl\\:0.10.2 repl -m cider.nrepl/cider-middleware -s wait"))
     (should (string= (cider-inject-jack-in-dependencies "--no-daemon clojureRepl" "gradle") "--no-daemon clojureRepl"))))
 
+(ert-deftest cider-inject-clojure-dependency ()
+  (should
+   (equal
+    `((,cider-clojure-artifact-id ,cider-minimum-clojure-version))
+    (let ((cider-jack-in-auto-inject-clojure 'minimal))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (equal
+    `((,cider-clojure-artifact-id "bob"))
+    (let ((cider-jack-in-auto-inject-clojure "bob"))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (equal
+    `((,cider-clojure-artifact-id ,cider-latest-clojure-version))
+    (let ((cider-jack-in-auto-inject-clojure 'latest))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (equal
+    '(("Hello, I love you" "won't you tell me your name"))
+    (let ((cider-jack-in-auto-inject-clojure
+           '("Hello, I love you" "won't you tell me your name")))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (eq
+    'nil
+    (let ((cider-jack-in-auto-inject-clojure nil))
+      (cider-add-clojure-dependencies-maybe nil)))))
+
 (ert-deftest cider-inject-jack-in-dependencies-add-refactor-nrepl ()
   (let ((cider-jack-in-lein-plugins '(("refactor-nrepl" "2.0.0") ("cider/cider-nrepl" "0.11.0")))
         (cider-jack-in-nrepl-middlewares '("refactor-nrepl.middleware/wrap-refactor" "cider.nrepl/cider-middleware")))
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -72,6 +72,11 @@ If you don't want `cider-jack-in` to inject dependencies automatically, set
 `cider-inject-dependencies-at-jack-in` to `nil`. Note that you'll have to setup
 the dependencies yourself (see the section below), just as in CIDER 0.10 and older.
 
+CIDER can also inject a Clojure dependency into your project, which is useful,
+for example, if your project defaults to an older version of Clojure than that
+supported by the CIDER middleware. Set `cider-jack-in-auto-inject-clojure`
+appropriately to enable this.
+
 If a standalone REPL is preferred, you need to invoke `cider-connect` (instead
 of `cider-jack-in`) and you'll need to manually add the dependencies to your
 Clojure project (explained in the following section).


expez:
@arrdem, will this ruin your day when working on [jaunt](https://github.com/jaunt-lang/jaunt)?

arrdem:
@expez offhand iunno, I'll throw this on my list and test this in the morning. I'm gonna guess that it will.

As a general matter I suppose I'm opposed to this. While 1.6 -> 1.7 is very very minor, in principle that CIDER is injecting non-obvious dependencies (here a language core update) and thus overriding whatever `lein deps :tree` may show is bound to be a source of confusion. The error messages in `cider-nrepl` which I added for when an inappropriate version of Clojure is found should be more than sufficient for someone checking `*messages*` to debug. The whole notion is that the onus is on the user to provide an appropriate "clojure" version whether that be 1.7, 1.8, some edition of `com.factual/clojure`, some edition of dunaj or worse Jaunt something or other.

phillord:
Clearly, injecting Clojure is a concern, and the approach that I have taken is very blunt, here. If I can find out the concerns, then I can improve the patch as is. There is, clearly, already an option to disable the injection entirely. This could also be extended to provide an "disable-clojure-injection" option. This would put the onus on people not using Clojure to switch this behaviour off.

At the moment, the only accurate error messages I am seeing is in *nrepl-server* -- I never look at this unless the REPL fails to pop up. The REPL does provide this warning:

WARNING: CIDER's version (0.12.0) does not match cider-nrepl's version (nil). Things will break!

which is correct, but not helpful, since it does not say *why* they are out-of-sync (this message is bad anyway, now that we have dependency injection, but this is a separate issue).

So, the current situation is not nice. Happy to code  up any alternatives that people can think off.

Malabarba:
Thanks for this @phillord.
Personally, I think it wouldn't be very nice of cider to inject  a Clojure version. If I were working on a project that uses. 1.7, I would want to develop it directly on 1.7 to avoid accidentally using functions or features from 1.8.

I also agree that we need to be more informative when the user has an unsupported Clojure version. Fortunately, there's already something in the works for that.

Malabarba:
And, of course, nothing prevents you from injecting Clojure as a dependency on your project. So your use case should be covered too. :-)

gtrak:
I prefer a decent error message to injection.  Many things can break across
a project between clojure versions, like AOT.  Dev-time deps should be as
close to production as possible.

llasram:
I will strongly argue that this injection in particular is not an option.  1.6->1.7 is where CLJ-1738 was introduced and marked WONTFIX.  About half of my company's codebase breaks on 1.7.

phillord:
@Malabarba Sure, this is true, but now anyone who is using Cider on a project which they wish to maintain compat <1.7 has to make sure that they do this. Now, in my case, I am using Clojure 1.6 features, so I have to use Clojure 1.6, but I do not need 1.7 features. Agreed, of course, I would prefer that cider just runs in with Clojure 1.6. But it doesn't. So, I cannot do what both you and @gtrak suggests. Cider cannot do close to 1.6 development.

It's not a major disaster, though; I can develop with cider in 1.7, then test 1.6 in continuous integration using lein profiles (which is how I test 1.7 and 1.8 at the moment).

So, I would suggest that this should be the default -- there is already a mechanism for disabling the injection. Of course, it would be equally possible to leave the default (i.e. cider breaks on Clojure<1.7) projects, fix the error message, and document how to inject 1.7. I can add that, if you would prefer, although I think it is not optimal.

@llasram I have sympathy for your position, but I think it's not relevant. Cider already has a mechanism for disabling injection, and it already requires the use of 1.7. If CLJ-1738 has been turned down, you are just stuffed. You're going to have to change your code base sooner or later, regardless of cider.

Malabarba:
> @Malabarba Sure, this is true, but now anyone who is using Cider on a project which they wish to maintain compat <1.7 has to make sure that they do this. Now, in my case, I am using Clojure 1.6 features, so I have to use Clojure 1.6, but I do not need 1.7 features. [...]

I think there was a bit of confusion, or maybe I misunderstand your point.

I think we should not inject Clojure by default. Instead we should present an informative message. Users who depend on 1.6 can always change the dependency in their `:repl` profile, or they can inject 1.7 by adding this to their init file:

    (cider-add-to-alist 'cider-jack-in-dependencies "org.clojure/clojure" "1.7.0")

Here's my logic of why that is better.

## If we inject Clojure 1.7

1. Users of 1.6 are forced to use 1.7 during development, possibly without realizing it.
2. Users of 1.7 are OK.
3. Users of 1.8 are forced to disable injection.

## If we don't inject, but provide a message

1. Users of 1.6 are forced to use 1.7 during development (via a `:repl` profile or via injection) .
2. Users of 1.7 are OK.
3. Users of 1.8 are OK.

## Conclusion

It seems to me that injection harms case `3.`, without really benefiting cases `1.` or `2.`.

> I can add that, if you would prefer, although I think it is not optimal.

Don't worry, there's already someone implementing a better error message.

arrdem:
Artur beat me to the punch. The question is what constitutes sufficiently better error messages. I feel somewhat chastised by the above comment that the cider-nrepl version warnings go to `*nrepl-server* ...`, rather than `*messages*`. It would seem that reporting either an old Clojure version or no Clojure version to the user and aborting the REPL so the user notices would be the thing to do.

Malabarba:
We'll have a bold red message right on the repl. Since the repl pops up after connecting, where basically guaranteed the user will see it. :-)

bbatsov:
> We'll have a bold red message right on the repl. Since the repl pops up after connecting, where basically guaranteed the user will see it. :-)

👍

phillord:
@Malabarba Again, my patch is rough and ready, and needs improvement.

First, I would argue over your notion of "users of 1.6". In my project, I am quite happy to use 1.6, 1.7 or 1.8. I state in the project file because this is the minimal version that the project needs, not the one I expect it to run. "1.6" to my understanding means "1.6 or later".

# With Injection

 * User opens project which happens to support old clojure versions -- it works by using 1.7
 * User opens project with maximal clojure range of 1.6, cider breaks (correctly!)
 * User opens project with minimal version of 1.8, injection needs to be disabled.

The first of these is correct, I think, but could offer a warning. The last of these is wrong, and I agree, I need to think of a way around this. 

# Without Injection

 * User opens project which supports old clojure versions -- it breaks.

An error message is an advance, of course. But, we are back to add dependencies and configuring things that ought to have been gone. 

Finally, the user lacks the expressivity they need -- what I really want to say is "ensure a minimal version of clojure 1.7, but use 1.8 if the project requires it". So, there is still functionality missing.

I have a background feeling that, maybe, we are missing a solution. Am I correct in thinking that the problem happens because we are using the provided profile for cider-nrepl. If this had a hard 1.7 dependency, then for a 1.6 (or earlier) project, using cider-nrepl would just force use of 1.7 anyway?

gtrak:
I think cider controlling the clojure version of the project it's in is
extremely invasive and would alienate users.

All clojure projects I've seen or worked on did not include cider in the
final artifacts, because others on the team may be using a different
version of cider or not cider at all.  That would be up to their user
profile.

If cider does not work with that version of clojure, then that project
should be pinned to an earlier version of cider until the developers make
the decision to upgrade themselves.

gtrak:
I personally have made efforts to stick to lowest-common denominator functionality if at all possible, I'm not sure what's motivating this effort, but I would wonder if it can't be done in a way that keeps compatibility across versions.

I don't see the problem that this is fixing.

phillord:
@gtrak The motivation for the dependency injection is to make it all "just work" without modifying things. It should just work. But it does not. I agree, cider should not control the version of Clojure for the project -- I do not want to change the 1.6 dependency of my project to 1.7 because of cider.  But cider *has* got a requirement for a minimal version of Clojure. Like or it not, a cider REPL must be 1.7 or greater, or you lose functionality.

At the moment, the perverse thing is that cider fails for people who maintain backward compatibility with older versions of their code, but works just fine for those who push their clojure requirements up to the latest whether they need it or not.

phillord:
@Malabarba Being daft -- the simpler solution is to inject Clojure 1.8 (or Clojure latest). Now dependency injection will not interfere with projects running Clojure 1.8 and will also work with projects running 1.7.

What about an "cider-always-inject-latest-clojure" option?

gtrak:
>  I do not want to change the 1.6 dependency of my project to 1.7 because of cider.

I'm happy to make this change manually if I know that's a requirement, lose editor functionality if it's not often used, or use an older version of cider.  If I know I am upgrading the project I can expect a level of effort ahead of time due to fixing up tests and that sort of thing.  If cider or another tool injected it and tests/compilation/running fails later when run from CLI, that creates about 2-3 hours of headscratching which will ultimately end up in me cursing cider's implicit behavior.  This is why all cider deps have been shaded with mranderson.  Clojure is the most invasive and risky dep there is in a clojure project.

phillord:
@gtrak If you are happy with an older version of cider, so that you can work with clojure < 1.7, then why would you worry about a later version of cider that you are not using, injecting it's minimal dependency somewhere along the line.

If you must have 1.6 or earlier, then you cannot use cider latest, so you should use an older one. If cider 0.12 or master must have clojure 1.7 in it's REPL, then it should make sure it gets *at least* 1.7, probably offering a warning if the project states "at least 1.6".

To me, this is ideal functionality. My patch does not achieve this, of course, nor am I 100% clear how to achieve that. But I would like to understand what the best functionality is.

Malabarba:
> What about an "cider-always-inject-latest-clojure" option?

If it defaults to nil, I think that sounds ok, though I'd prefer `cider-inject-latest-clojure`.
@bbatsov?

bbatsov:
> cider-inject-latest-clojure

Or `cider-auto-inject-latest-clojure`. In general I'm fine with the idea, as long as this is not on by default. And I still believe we need a warning message.

Malabarba:
> And I still believe we need a warning message.

Yes. There's something in the works on #1698

phillord:
@Malabarba @bbatsov It falls short of my ideal functionality, but it's very easy to implement while my ideal functionality is pretty hard!

I'll patch this up tomorrow.

gtrak:
> @gtrak If you are happy with an older version of cider, so that you can work with clojure < 1.7, then why would you worry about a later version of cider that you are not using, injecting it's minimal dependency somewhere along the line.

@phillord I would add a profile for the dep in that project to work with the new cider, or I would set up a different emacs config for the old specifically for that project, I'm happy using the latest cider and latest deps, what I'm worried about is the situation where someone updates cider elisp and breaks their clojure build.

dpsutton:
image


You can see this over in the issues. I have the code working and would love some polish before pushing. This issue popped up because the seattle version dropped support for clojure 1.6.0 which left cryptic warnings about nrepl versions rather than the clojure version mismatch.

I'm not sure what the release schedule is but I can make sure that its ready tonight. I see that there are 3 different build styles supported: lein, boot and gradle. Is the syntax for clojure support the same across all three?

I have a single boot version to test against and then a few project.clj's from lein projects but just need to make sure that the parsing is working correctly.

As it stands, right now, I parse the project files for the version info. I figured that this was correct enough since if there was middleware problems, I might not be able to get a response. However, after talking with @Malabarba he's given me some insight into just asking the repl which version it is running.

phillord:
@gtrak To be honest, I do not understand your point. You are worried about updating cider and breaking your clojure *build*? But injecting dependencies has no affect on your build, testing or anything. It just changes the REPL you get.

I am worried about updating my cider and finding that it no longer works, when it should work.

Regardless, we have a solution, which I can implement. I think it is not ideal, as it requires people to manually configure a clojure version, but at least it's better than at current, when it is not at all obvious how a user should do this.

phillord:
@dpsutton If you make a PR, then I'll be happy to test this out. I agree that parsing the project file is not a good idea, unless you are using lein pprint or some such. The clojure dependency could be specified in too many places.

phillord:
@Malabarba @bbatsov How about this? It provides a minimal, max or what ever the user wants version.

phillord:
@dpsutton Assuming this PR is accepted, why not add to your error message a click button that sets this value in .dir-locals.el. 

So, we'd have "Cider supports minimal version of Clojure 1.7. [Set] the minimal version for this project".

dpsutton:
@phillord what would it mean for you to set a minimum version other than Clojure 1.7.0? Cider itself does not support less than that. Whatever text string you might set somewhere is meaningless without some code changes.

phillord:
@dpsutton Well, "Set the version for this project", then. It would mean either 1.7, 1.8 or anything else you felt you wanted to add.

dpsutton:
@phillord I'm confused. the version for the project is set in project.clj or build.boot. Seems really bad to me to use a different version of clojure for development and production. And changing that would have to be super super obvious that you made a strange change. 

I also don't know how that would work. Would you load whatever specified version of Clojure, and then allow overriding it if it was below's cider's abilities? Is it easy to unload a Clojure? What if the requested version isn't local? I'm not sure how the fetching works and if you can stall while it fetches a new version. 

One idea that i played around with was parsing the project.clj/boot.build file. But that idea sounds worse and worse as its a brittle and simple second parser that could disagree with the real implementation. Much better to ask java what version of Clojure is requested. but how does java itself read this since its a clojure data structure which specifies which version of Clojure to use. 

Right now the most simple thing to do is to just change the Clojure version string in the build file without commiting those changes while these other ideas are thought about or abandoned.

phillord:
@dpsutton "Further thoughts: could you specify some cider specific minimum version to use somewhere? Seems risky to say you can use 1.7 on all of your projects. Then allow overrides in local directories? So now there is a specific cider props file somewhere."

Yes, that's what I was suggesting. Once this PR is merged, then putting this:

     ((clojure-mode
        (cider-jack-in-auto-inject-clojure . :latest)))

into .dir-locals.el does the trick. I was suggesting adding a button to your error message to achieve this. There is even a handy function in core Emacs for adding to this file correctly.

Malabarba:
The problem is that the error message also comes up in other situations. Like connecting to a server without the middleware, or jacking-in with a command where we don't know how to inject the middleware. So we can just say "click this button and we'll fix that", we need to point people to the manual so they can understand what's going on.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * Fringe indicators highlight which sexps have been loaded. Disable it with `cider-use-fringe-indicators`.
 * New command: `cider-inspect-last-result`.
 * `cider-cljs-lein-repl` now also supports figwheel.
+* Option `cider-jack-in-auto-inject-clojure` enables the user to specific a
+  version of Clojure for CIDER, independent of the project.
```
Malabarba: > Option `cider-jack-in-auto-inject-clojure` allows the user to specify a version of Clojure to be injected. This can be used to override the version used in a project.

phillord:
@Malabarba okay, unfortunate, but not much that can be done.

cider.el
```diff
@@ -229,6 +229,19 @@ Sub-match 1 must be the project path.")
 (cider-add-to-alist 'cider-jack-in-dependencies
                     "org.clojure/tools.nrepl" "0.2.12")
 
+(defcustom cider-jack-in-auto-inject-clojure nil
+  "Version of clojure to auto-inject into REPL.
+
+If nil, do not inject clojure into the REPL.  If :latest, inject
+`cider-latest-clojure-version', which should approximate to the
+most recent version of clojure. If :minimal, inject
```
Malabarba: `. I` -> `.  I`

test/cider-tests.el
```diff
@@ -584,6 +584,28 @@
                      "-d org.clojure/tools.nrepl\\:0.2.12 -d cider/cider-nrepl\\:0.10.2 repl -m cider.nrepl/cider-middleware -s wait"))
     (should (string= (cider-inject-jack-in-dependencies "--no-daemon clojureRepl" "gradle") "--no-daemon clojureRepl"))))
 
+(ert-deftest cider-inject-clojure-dependency ()
+  (should
+   (equal
+    '(("org.clojure/clojure" "1.7.0"))
```
Malabarba: This will break if we ever change minimal
.

test/cider-tests.el
```diff
@@ -584,6 +584,28 @@
...
+    (let ((cider-jack-in-auto-inject-clojure :minimal))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (equal
+    '(("org.clojure/clojure" "bob"))
+    (let ((cider-jack-in-auto-inject-clojure "bob"))
+      (cider-add-clojure-dependencies-maybe nil))))
+  (should
+   (equal
+    '(("org.clojure/clojure" "1.8.0"))
```
Malabarba: this will break if we ever change latest

test/cider-tests.el
phillord: Yes. And then it can be changed. I can do 

    (equal (list (list "org.clojure/clojure" cider-minimum-clojure-version))
              ...)

But then it will not pick up the issue when cider-minimum-clojure-version is set to "1,8.0" or something by mistake.

Happy to fix, but think it's a less good test.

Malabarba: Yes, I would prefer to change it.

> I can do 

>    (equal (list (list "org.clojure/clojure" cider-minimum-clojure-version))
>              ...)

Yes. Or ``` `(("org.clojure/clojure" ,cider-minimum-clojure-version))```

>But then it will not pick up the issue when cider-minimum-clojure-version is set to "1,8.0" or something by mistake.

Think of it this way. In its current form it's not checking whether the variable is set to a valid version string, it's just checking whether the variable's value is exactly `"1.7.0"`. Would you write a test that simply checks if `(equal cider-minimum-clojure-version "1.7.0")`? :-) (I wouldn't)

phillord: No worries, will do.

Agree, a valid version string test would be better. Would I write `(equal cider-minimum-clojure-version "1.7.0")` -- yeah, probably, if it was critical. It would require me to screw up in two places then, before it went wrong. But, you are right, the duplication is a pain.

cider-interaction.el
```diff
@@ -223,6 +223,9 @@ namespace-qualified function of zero arity."
 (defvar cider-minimum-clojure-version "1.7.0"
   "Minimum supported version of Clojure.")
 
+(defvar cider-latest-clojure-version "1.8.0"
```
bbatsov: This should be `defconst` I guess.

cider.el
```diff
@@ -229,6 +229,19 @@ Sub-match 1 must be the project path.")
...
+`cider-minimum-clojure-version', which will be the lowest
+version cider supports. If a string, use this as the version number."
+  :type '(choice (const :tag "None" nil)
```
bbatsov: I'd use symbols instead of keywords here, as this is the traditional Elisp way. I definitely don't expect to see keywords as  config values anywhere.

test/cider-tests.el
bbatsov: Guess we should simply let-bind those in the tests.

phillord:
All done, I hope.

I've added something also -- see the commit message!

bbatsov:
So, is everyone happy with the current state of affairs?

expez:
> So, is everyone happy with the current state of affairs?

Looks good to me! :+1:

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Fringe indicators highlight which sexps have been loaded. Disable it with `cider-use-fringe-indicators`.
 * New command: `cider-inspect-last-result`.
 * `cider-cljs-lein-repl` now also supports figwheel.
+* Option `cider-jack-in-auto-inject-clojure` enables the user to specify a
+  version of Clojure for CIDER. This allows the user to override the version
+  used in a project, particular if it lower than minimum required for CIDER.
```
expez: missing 'is' in 'if it lower than'

cider.el
```diff
@@ -288,6 +304,24 @@ string is quoted for passing as argument to an inferior shell."
    " -- "
    params))
 
+(defun cider-add-clojure-dependencies-maybe (dependencies)
+  "Return DEPENDENCIES with an added Clojure dependency is requested.
```
expez: should probably be 'as requested'

bbatsov:
Two more things:

* One test is failing
* Drop the `.` from the commit message's title

bbatsov:
This functionality should also be mentioned in the manual, otherwise I doubt someone would find it.

phillord:
All done, and one thing that I had missed also.

bbatsov:
👍

PR #1700, opened 2016-04-21
check clojure version from running jvm

dpsutton:

    - difficult since i'm calling out to nrepl. Willing if there's a straightforward way to mock these calls and check what repl message was generated.
    - 107 of 107
    - 4 preexisting issues
    - doesn't feel applicable but willing to add
    - not applicable

Add a version check of clojure during the pre-flight checks of setting up the repl. This has become an issue because of the recent minimum supported version of clojure went to 1.7.0 with the seattle release. However, it prints only a cryptic message about nrepl mismatch rather than the real cause of the error.

New version:
image

versus previous behavior:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* Ensure Clojure version meets minimum supported by CIDER (1.7.0).
 * Fringe indicators highlight which sexps have been loaded. Disable it with `cider-use-fringe-indicators`.
 * New command: `cider-inspect-last-result`.
 
--- a/cider.el
+++ b/cider.el
@@ -621,6 +621,17 @@ In case `default-directory' is non-local we assume the command is available."
                                "Can't determine nREPL's version.\nPlease, update nREPL to %s."
                                cider-required-nrepl-version)))
 
+(defun cider--check-clojure-version-supported ()
+  "Ensure that we are meeting the minimum supported version of Clojure."
+  (if-let ((clojure-version (cider--clojure-version)))
+      (when (version< clojure-version cider-minimum-clojure-version)
+        (cider-repl-manual-warning "installation/#prerequisites"
+                                   "Clojure version (%s) is not supported (minimum %s). CIDER will not work."
+                                   clojure-version cider-minimum-clojure-version))
+    (cider-repl-manual-warning "installation/#prerequisites"
+                               "Clojure version information could not be determined. Requires a minimum version %s."
+                               cider-minimum-clojure-version)))
+
 (defun cider--check-middleware-compatibility ()
   "CIDER frontend/backend compatibility check.
 Retrieve the underlying connection's CIDER-nREPL version and checks if the
@@ -646,6 +657,7 @@ buffer."
   (cider-make-connection-default (current-buffer))
   (cider-repl-init (current-buffer))
   (cider--check-required-nrepl-version)
+  (cider--check-clojure-version-supported)
   (cider--check-middleware-compatibility)
   (cider--debug-init-connection)
   (cider--subscribe-repl-to-server-out)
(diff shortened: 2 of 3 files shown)


dpsutton:
seems like an erroneous failure:

> 
In toplevel form:
cider-inspector.el:114:9:Error: `cider-inspect-read-and-inspect' is an obsolete function (as of 0.13.0); use `cider-inspect-expr' instead.
make: *** [cider-inspector.elc-test] Error 1

I haven't been anywhere near there that I know of and its complaining about an obsolete function. And just realized that its obsolete as of 0.13.0 so i'm not sure what's going on there at all.

However, that is some supremely slick software that i was able to see the error output formatted so nice. Well done whoever.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New Features
 
+* Ensure clojure version meets minimum supported by cider (1.7.0).
+
```
Malabarba: Remove this blank line

cider.el
```diff
@@ -646,13 +649,26 @@ buffer."
...
+(defun cider--check-clojure-version-supported ()
+  "Ensure that we are meeting the minimum supported version of clojure."
+  (let* ((response (cider-nrepl-sync-request:eval "(clojure-version)"))
+         (clojure-version (nrepl-dict-get response "value")))
+    (setq clojure-version (replace-regexp-in-string "\"" ""
+                                                    (or clojure-version
+                                                        "0.0.0")))
+    (when (version< clojure-version cider-minimum-clojure-version)
+      (cider-repl-manual-warning "installation/#prerequisites"
+                                 "Clojure version (%s) is not supported (minimum %s). Things will break!"
```
Malabarba: Change "things will break" to "CIDER will not work".

Malabarba:
Also capitalise the commit message

dpsutton:
I think your last message confirms it for me but this would be an amend and force push then i presume.

It's also super nice that you take such a detailed look at everything so the codebase remains nice.

Malabarba:
Thanks. 
It looks good to me. I'll let @bbatsov have a final glance at this before merging. :+1:

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* Ensure clojure version meets minimum supported by cider (1.7.0).
```
bbatsov: Clojure
CIDER

cider.el
```diff
@@ -93,6 +93,9 @@ project inference will take place."
 Normally it won't be used, unless `pkg-info' fails to extract the
 version from the CIDER package or library.")
 
+(defconst cider-minimum-clojure-version "1.7.0"
```
bbatsov: Place this near the required nREPL version.

cider.el
```diff
@@ -646,13 +649,26 @@ buffer."
...
+  (cider--check-clojure-version-supported)
   (cider--check-middleware-compatibility)
   (cider--debug-init-connection)
   (cider--subscribe-repl-to-server-out)
   (when cider-auto-mode
     (cider-enable-on-existing-clojure-buffers))
   (run-hooks 'cider-connected-hook))
 
+(defun cider--check-clojure-version-supported ()
+  "Ensure that we are meeting the minimum supported version of clojure."
```
bbatsov: Clojure

cider.el
```diff
@@ -646,13 +649,26 @@ buffer."
...
+  (let* ((response (cider-nrepl-sync-request:eval "(clojure-version)"))
```
bbatsov: All this is not needed. We have a function `cider--clojure-version`.

bbatsov:
In general follow the structure of `cider--check-required-nrepl-version` as what you're trying to do is exactly the same.

dpsutton:
@bbatsov updated with your requests. Don't know why I didn't look into that and rather tried to reinvent everything. thanks for the pointers and you keep a clean code base.

PR #1703, opened 2016-04-21
Remove a reference to cider-inspect-read-and-inspect.

cskksc:
It fixes the bytecomp travis builds.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -112,7 +112,7 @@ With a second prefix argument it prompts for an expression to eval and inspect."
   (pcase arg
     (1 (cider-inspect-last-sexp))
     (4 (cider-inspect-defun-at-point))
-    (16 (cider-inspect-read-and-inspect))))
+    (16 (call-interactively #'cider-inspect-expr))))
 
 ;; Operations
 (defun cider-inspector--value-handler (_buffer value)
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -112,7 +112,7 @@ With a second prefix argument it prompts for an expression to eval and inspect."
   (pcase arg
     (1 (cider-inspect-last-sexp))
     (4 (cider-inspect-defun-at-point))
-    (16 (cider-inspect-read-and-inspect))))
+    (16 (call-interactively #'cider-inspect-expr))))
```
bbatsov: This change seems unrelated to the primary purpose of the commit. It's also unclear to me what exactly do we gain from this change.

bbatsov:
Commit message titles should not end with `.`.

cider-inspector.el
Malabarba: This is necessary because I've marked that function as obsolete.

cskksc: This fixes the travis bytecomp builds, https://travis-ci.org/clojure-emacs/cider/jobs/124647689#L969 and https://travis-ci.org/clojure-emacs/cider/jobs/124622642#L1065

cskksc: Sorry about the commit message thing. I seem to be making this mistake repeatedly. Won't happen again. :-)

PR #1710, opened 2016-04-24
Integrate ns-vars-with-meta op

cskksc:

Thanks!

The current ns-browser relies on the `cider-repl-ns-cache` to decide font-locks.
This gives us incorrect font-locks for namespaces which are present on the classpath,
but not loaded in the REPL.
We change this to get the data from the nREPL middleware op `ns-vars-with-meta`,
in case of a cache miss; clojure-emacs/cider-nrepl#346.

### DIFF
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -74,25 +74,25 @@
   (setq-local truncate-lines t)
   (setq-local cider-browse-ns-current-ns nil))
 
-(defun cider-browse-ns--text-face (namespace text)
-  "Return font-lock-face for TEXT.
-The var info is fetched from the running repl and a font-lock face is decided.
+(defun cider-browse-ns--text-face (var-meta)
+  "Return font-lock-face for a var.
+VAR-META contains the metadata information used to decide a face.
 Presence of \"arglists-str\" and \"macro\" indicates a macro form.
 Only \"arglists-str\" indicates a function. Otherwise, its a variable.
 If the NAMESPACE is not loaded in the REPL, assume TEXT is a fn."
-  (let ((var-info (cider-resolve-var namespace text)))
-    (cond
-     ((not var-info) 'font-lock-function-name-face)
-     ((and (nrepl-dict-contains var-info "arglists")
-           (string= (nrepl-dict-get var-info "macro") "true"))
-      'font-lock-keyword-face)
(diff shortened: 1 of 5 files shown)


cider-browse-ns.el
```diff
@@ -74,25 +74,23 @@
   (setq-local truncate-lines t)
   (setq-local cider-browse-ns-current-ns nil))
 
-(defun cider-browse-ns--text-face (namespace text)
-  "Return font-lock-face for TEXT.
-The var info is fetched from the running repl and a font-lock face is decided.
+(defun cider-browse-ns--text-face (var-meta)
+  "Return font-lock-face for a var. VAR-INFO is its metadata information.
```
bbatsov: The docstring will produce `checkdoc` warnings.

bbatsov:
Overall the changes look good to me.

cskksc:
Done :-)

cider-browse-ns.el
```diff
@@ -74,25 +74,24 @@
   (setq-local truncate-lines t)
   (setq-local cider-browse-ns-current-ns nil))
 
-(defun cider-browse-ns--text-face (namespace text)
-  "Return font-lock-face for TEXT.
-The var info is fetched from the running repl and a font-lock face is decided.
+(defun cider-browse-ns--text-face (var-meta)
+  "Return font-lock-face for a var.  VAR-META is its metadata information.
```
bbatsov: The first line of a docstring should be just one sentence.

cskksc: Done :-)

cider-client.el
```diff
@@ -937,6 +945,19 @@ CONTEXT represents a completion context for compliment."
     (nrepl-dict-get response "formatted-edn")))
 
 
+;;; Cache helpers
+
+(declare-function cider-resolve-ns-symbols "cider-resolve")
+
+(defun cider-ns-vars-with-meta (ns)
```
Malabarba: I don't think we need this anymore. Now that we have an op that returns exactly what we need, there's no need to use the cache. It complicates the code and has no advantage (now that the real op is fast enough).

cskksc: This only has an advantage when fetching huge namespaces, like the core ns'. But yeah, this fn could be avoided. I'll submit a PR today.

PR #1713, opened 2016-04-25
Change priority for guessing proper type of conn

aJchemist:
Existing code have a problem that `clojurec-mode` is always having `"clj"` type of connection.

now it can have a either `"cljs"` or `"clj"` type of connection by `(setq cider-repl-type ...)` in target buffer

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -215,6 +215,8 @@ such a link cannot be established automatically."
   "Return the matching connection type (clj or cljs) for the current buffer."
   (cond
    ((derived-mode-p 'clojurescript-mode) "cljs")
+   ((derived-mode-p 'clojurec-mode) cider-repl-type)
+   ((derived-mode-p 'clojurex-mode) cider-repl-type)
    ((derived-mode-p 'clojure-mode) "clj")
    (cider-repl-type)
    (t "clj")))


Malabarba:
I don't want to rely on the fact that the default value of cider-repl-type is clj. The only reason I didn't set the  default value as nil was to ensure that the users still had a working REPL in case we messed up up logic somewhere.  
Instead, you should add a clause for clojurec-mode (and clojurex-mode) above the Clojure clause. 

However, there's a very recent  feature that we've been building gradually that might dismiss the need for this altogether. Currently, if you load a buffer in clojurec-mode, it gets loaded on **both*" connections. 
We haven't applied the same logic to regular evaluation, but it's something that might make sense. Maybe we should try it out, now that the 0.13 snapshots have just begun, to see if it bothers anyone. @expez @bbatsov

aJchemist:
@Malabarba As you say that something more integration is going on, I can close this PR and use it just for me until next release

aJchemist:
Anyway in case of merging, I've revised the commit as you guided.

bbatsov:
> However, there's a very recent feature that we've been building gradually that might dismiss the need for this altogether. Currently, if you load a buffer in clojurec-mode, it gets loaded on *both" connections. 
> We haven't applied the same logic to regular evaluation, but it's something that might make sense. Maybe we should try it out, now that the 0.13 snapshots have just begun, to see if it bothers anyone. @expez @bbatsov

Yeah, I guess we should move in this direction, as the current approach is rather brittle, as demonstrated by tickets like this one.

bbatsov:
@aJchemist your commit message seems outdated. Overall I'm fine with merging this until we come up with a better solution.

aJchemist:
@bbatsov don't merge this. This commit has another issue that when I change buffer-local `cider-repl-type` in a cljc buffer, then `*cider-repl CLJS*` buffer's `cider-repl-type` will changed to "clj" in somewhat case.

bbatsov:
Guess I'll close this then.

PR #1714, opened 2016-04-25
Fix arity of `cider-change-buffers-designation'

danielsz:

Thanks!

* Allow user to set the argument of `cider-change-buffers-designation'
  in a non-interactive call (programmatically).

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -1066,14 +1066,13 @@ If CONN is not provided the user will be prompted to select a connection."
                   repl-buffer-name)
     (or (match-string 1 repl-buffer-name) "<no designation>")))
 
-(defun cider-change-buffers-designation ()
-  "Change the designation in cider buffer names.
+(defun cider-change-buffers-designation (designation)
+  "Change the DESIGNATION in cider buffer names.
 Buffer names changed are cider-repl and nrepl-server."
-  (interactive)
+  (interactive (list (read-string (format "Change CIDER buffer designation from '%s': "
+					  (cider-extract-designation-from-current-repl-buffer)))))
   (cider-ensure-connected)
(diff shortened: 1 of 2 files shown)


bbatsov:
Seems to me the current commit message is a bit misleading, your description illustrates the actual gist of the change "Allow user to set the argument of `cider-change-buffers-designation' in a non-interactive call (programmatically)".

danielsz:
Hi Bozhidar, 

I have no intentions to mislead, only to communicate. 
If I failed to do so clearly, please help me out with a question of what you think should be restated. 
If I had to guess, I would say you don't understand the reason for this PR, right? You want a use case. 

So the use case for this PR is that I want to call  `cider-change-buffers-designation` in an Emacs Lisp script. It is as simple as that. Before the PR, `cider-change-buffers-designation` only works interactively. After the PR, `cider-change-buffers-designation` works both interactively, and as a regular function that takes an argument.

Furthermore, it is just better Emacs Lisp and closer to the spirit of functional programming. Why does that function hide user input in local state? Why doesn't it leverage the standard `interactive` facility?  You can see the benefits immediately, as they simplify the tests. When the function accepts an argument, you don't need no fancy `noflet` form to advise around the user input gathering. 

The takeaway is that when local state in a function comes from user input, it should always be an argument of that function. The function will be more versatile, will better the chances to be free of side-effects, will be easier to read and test. Emacs Lisp provides a structured approach to apply this principle via the `interactive` form.

Thank you.

bbatsov:
I never doubted the benefit of the PR itself, I just disliked the commit message. Anyways, I changed it myself. Thanks!

danielsz:
:+1: Thank you!

PR #1715, opened 2016-04-25
Always fetch ns-vars-with-meta from nREPL middleware

cskksc:
As discussed in https://github.com/clojure-emacs/cider/pull/1710

/cc @Malabarba


Malabarba:
LGTM

bbatsov:
👍

PR #1729, opened 2016-05-04
[Fix #1725] Display class names in eldoc for interop forms

### LINKED ISSUE #1725: Improve the eldoc for interop forms
Currently the eldoc for interop forms is not very clear. We simply display the name of the method and all possible arglists from all matching classes. I'm reasonably sure not all users are aware of this.

We can improve this a bit in the following manner:

* display the class name when the method/field in question is present in just one class
* display the number of matching classes (or their names) before the method name in cases of ambiguities

Some the end results should be something like "SomeClass/someMethod: ..." and "(n classes)/someMethod: ...".

cskksc:

Thanks!

screen shot 2016-05-04 at 12 41 38 pm

### DIFF
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -35,6 +35,7 @@
 (require 'cider-common) ; for cider-symbol-at-point
 (require 'cider-compat)
 (require 'cider-util)
+(require 'nrepl-dict)
 
 (require 'seq)
 
@@ -61,22 +62,68 @@ For convenience, some functions are already provided for this purpose:
   :group 'cider
   :package-version '(cider . "0.13.0"))
 
+(defcustom cider-eldoc-max-class-names-to-display 3
+  "The maximum number of classes to display in an eldoc string.
+An eldoc string for Java interop forms can have a number of classes prefixed to
+it, when the form belongs to more than 1 class.  When, not nil we only display
+the names of first `cider-eldoc-max-class-names-to-display' classes and add
+a \"& x more\" suffix. Otherwise, all the classes are displayed."
+  :type 'integer
+  :safe #'integerp
+  :group 'cider
+  :package-version '(cider . "0.13.0"))
+
+(defun cider--eldoc-format-class-names (class-names)
+  "Return a formatted CLASS-NAMES prefix string.
+CLASS-NAMES is a list of classes to which a Java interop form belongs.
+Only keep the first `cider-eldoc-max-class-names-to-display' names, and
+add a \"& x more\" suffix.  Return nil if the CLASS-NAMES list is empty or
+mapping `cider-eldoc-ns-function' on it returns an empty list."
+  (when-let ((eldoc-class-names (seq-remove #'null (mapcar (apply-partially cider-eldoc-ns-function) class-names)))
+             (eldoc-class-names-length (length eldoc-class-names)))
+    (cond
+     ;; truncate class-names list and then format it
+     ((and cider-eldoc-max-class-names-to-display
+           (> eldoc-class-names-length cider-eldoc-max-class-names-to-display))
+      (format "(%s & %s more)"
+              (thread-first eldoc-class-names
+                (seq-take cider-eldoc-max-class-names-to-display)
+                (cider-string-join " ")
+                (cider-propertize 'ns))
+              (- eldoc-class-names-length cider-eldoc-max-class-names-to-display)))
+
+     ;; format the whole list but add surrounding parentheses
+     ((> eldoc-class-names-length 1)
+      (format "(%s)"
+              (thread-first eldoc-class-names
+                (cider-string-join " ")
+                (cider-propertize 'ns))))
+
+     ;; don't add the parentheses
+     (t (format "%s" (car eldoc-class-names))))))
+
 (defun cider-eldoc-format-thing (ns symbol thing)
   "Format the eldoc subject defined by NS, SYMBOL and THING.
 THING represents the thing at point which triggered eldoc.  Normally NS and
 SYMBOL are used (they are derived from THING), but when empty we fallback to
 THING (e.g. for Java methods)."
-  (if (and ns (not (string= ns ""))
-           symbol (not (string= symbol "")))
-      (let ((ns (funcall cider-eldoc-ns-function ns)))
-        (if (and ns (not (string= ns "")))
-            (format "%s/%s"
-                    (cider-propertize ns 'ns)
-                    (cider-propertize symbol 'var))
-          ;; in case `cider-eldoc-ns-function' returned nil or empty string
-          (cider-propertize symbol 'var)))
-    ;; we're probably dealing with some interop form
-    (cider-propertize thing 'var)))
+  (if-let ((method-name (if (and symbol (not (string= symbol "")))
+                            symbol
+                          thing))
+           (ns-or-class (if (and ns (stringp ns))
+                            (funcall cider-eldoc-ns-function ns)
+                          (cider--eldoc-format-class-names ns))))
+      (format "%s/%s"
+              ;; we set font-lock properties of classes in `cider--eldoc-format-class-names'
+              ;; to avoid font locking the parentheses and "& x more"
+              ;; so we only propertize ns-or-class if not already done
+              (if (get-text-property 1 'face ns-or-class)
+                  ;; it is already propertized
+                  ns-or-class
+                (cider-propertize ns-or-class 'ns))
+              (cider-propertize method-name 'var))
+    ;; in case ns-or-class is nil
+    (cider-propertize method-name 'var)))
 
 (defun cider-highlight-args (arglist pos)
   "Format the the function ARGLIST for eldoc.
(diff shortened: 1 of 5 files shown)


cider-eldoc.el
```diff
@@ -201,25 +202,29 @@ This includes the arglist and ns and symbol name (if available)."
...
   (when (and (cider-connected-p)
              ;; don't clobber an error message in the minibuffer
              (not (member last-command '(next-error previous-error))))
     (let* ((sexp-info (cider-eldoc-info-in-current-sexp))
-           (thing (car sexp-info))
+           (thing (nrepl-dict-get (car sexp-info) "thing"))
            (pos (cadr sexp-info))
            (eldoc-info (cider-eldoc-info thing))
-           (ns (nth 0 eldoc-info))
+           (ns (nrepl-dict-get (car sexp-info) "ns" (nth 0 eldoc-info)))
```
cskksc: This is added for cases like, `java.lang.String/.length`. It wouldn't be right to just show the whole list here.
One thing to consider; what if the class name entered by user is not a valid one ? Should we keep a check to see if the entered class is a member of the classes list returned by middleware ? What should we display then ?

cider-eldoc.el
```diff
@@ -66,17 +66,19 @@ For convenience, some functions are already provided for this purpose:
...
+                (if (and (stringp ns) (not (string= ns "")))
                     (cider-propertize ns 'ns)
-                    (cider-propertize symbol 'var))
-          ;; in case `cider-eldoc-ns-function' returned nil or empty string
-          (cider-propertize symbol 'var)))
-    ;; we're probably dealing with some interop form
-    (cider-propertize thing 'var)))
+                  ;; it would be a list in case of a Java interop form
+                  ;; which is present in multiple classes. Eg: .length
+                  (cider-propertize (mapconcat #'identity ns ",") 'ns))
```
cskksc: Should the `,` be a defcustom ?

cskksc:
The tests have failed in the `make elpa` step. I'm sure a re-trigger would fix them.

cider-eldoc.el
bbatsov: We should add some config value to limit the number of classes displayed - e.g. 3 by default. If we have more than 3 we'll show something like "A B C & 3 more".

cider-util.el
```diff
@@ -123,6 +124,16 @@ find a symbol if there isn't one at point."
               (forward-sexp -1)))
           (cider-symbol-at-point)))))
 
+(defun cider-ns-thing-at-point ()
+  "Return destructured `cider-symbol-at-point'
+If the symbol at point is of the form ns-name/thing-name, returns a dict
+(\"ns\" \"ns-name\" \"thing\" \"thing-name\")"
```
bbatsov: You forgot the final `.` here.

test/cider-eldoc-tests.el
```diff
@@ -44,16 +44,21 @@
...
     (it "returns eldoc formatted thing"
-      (expect (cider-eldoc-format-thing "" "" ".toString")
-              :to-equal ".toString"))))
+      (expect (cider-eldoc-format-thing "java.lang.String" "" ".startsWith")
+              :to-equal "java.lang.String/.startsWith")))
 
+  (describe "when the given Java interop form belongs to multiple classes"
+    (it "joins the class list into a string"
+      (expect (cider-eldoc-format-thing '("java.lang.String" "java.lang.StringBuffer" "java.lang.CharSequence" "java.lang.StringBuilder") "" ".length")
+              :to-equal "java.lang.String, java.lang.StringBuffer, java.lang.CharSequence, java.lang.StringBuilder/.length"))))
```
bbatsov: Drop the `,`. Having just a space between class names is good enough.

cider-eldoc.el
bbatsov: What do you mean by "entered by the user"?

cskksc: So for example, when the cursor is on the form `(java.lang.X/.length)`, the eldoc would be shown as is ; `(java.lang.X/.length): ([this])`, because we always prioritise the ns entered by user.
But from the information returned by the middleware, we know that `.length` is not a member of class `X`. Should we display some warning in such case ? But if that class is not loaded, or we don't have its eldoc information, our warning would be totally wrong. So right now we assume that class entered by the user is always correct.

The `& 3 more` thing sounds nice.

cskksc: Done. I realised that `& more` is not just nice. It becomes mandatory to have for things like `toString`.
screen shot 2016-05-04 at 2 56 59 pm

bbatsov: > So for example, when the cursor is on the form (java.lang.X/.length), the eldoc would be shown as is ; (java.lang.X/.length): ([this]), because we always prioritise the ns entered by user.

Got it. Guess we should handle those scenarios more gracefullly, although I don't think they are very common. I've rarely seem something use the full class name with a method invocation. File a separate ticket for this and we'll address it after we're done with the current ticket.

cider-eldoc.el
```diff
@@ -201,25 +237,29 @@ This includes the arglist and ns and symbol name (if available)."
...
               (cdr cider-eldoc-last-symbol)
             (when-let ((eldoc-info (cider-sync-request:eldoc thing)))
-              (let ((arglist (nrepl-dict-get eldoc-info "eldoc"))
-                    (ns (nrepl-dict-get eldoc-info "ns"))
-                    (symbol (nrepl-dict-get eldoc-info "name")))
+              (let* ((arglist (nrepl-dict-get eldoc-info "eldoc"))
+                     (ns (nrepl-dict-get eldoc-info "ns"))
+                     (class (nrepl-dict-get eldoc-info "class"))
+                     (symbol (nrepl-dict-get eldoc-info "name"))
+                     (container (if (and ns (not (string= ns "")))
```
bbatsov: container? :-)

bbatsov: I'd just name this `ns-or-class` or something like this.

cider-eldoc.el
bbatsov: Btw, probably there should be some option to compact packages like we do for clojure symbols (as some people will complain about the long names). E.g. we could display `java.lang` as `j.l.`, just `lang` or skip it completely. Not super important, but it'd be nice to have something like this.

cskksc: >  File a separate ticket for this and we'll address it after we're done with the current ticket.

Yeah, will file a ticket once we close this.

> Btw, probably there should be some option to compact packages 

Done. Used the existing `cider-eldoc-ns-function`. Also renamed the `container` var.

bbatsov: > Done. Used the existing cider-eldoc-ns-function.

Hopefully you've tested what happens when this is set to a function returning `nil`. :-)

cider-eldoc.el
```diff
@@ -61,22 +62,62 @@ For convenience, some functions are already provided for this purpose:
...
+   ((and cider-eldoc-max-class-names-to-display
+         (> (length class-names) cider-eldoc-max-class-names-to-display))
+    ;; truncate the list
+    (format "(%s & %s more)"
+            (cider-propertize
+             (mapconcat cider-eldoc-ns-function (seq-take class-names cider-eldoc-max-class-names-to-display) " ")
+             'ns)
+            (- (length class-names) cider-eldoc-max-class-names-to-display)))
+
+   ;; format everything and add brackets
```
bbatsov: brackets -> braces/parentheses

cider-eldoc.el
```diff
@@ -61,22 +62,62 @@ For convenience, some functions are already provided for this purpose:
...
+   ((> (length class-names) 1)
+    (format "(%s)"
+            (cider-propertize (mapconcat cider-eldoc-ns-function class-names " ") 'ns)))
+
+   ;; dont add the surrounding brackets
```
bbatsov: dont -> don't

cider-eldoc.el
```diff
@@ -61,22 +62,62 @@ For convenience, some functions are already provided for this purpose:
...
+CLASS-NAMES is a list of classes to which a Java interop form belongs.
+Only keep the first `cider-eldoc-max-class-names-to-display' names, and
+add a \"& x more\" suffix."
+  (cond
+   ((and cider-eldoc-max-class-names-to-display
+         (> (length class-names) cider-eldoc-max-class-names-to-display))
+    ;; truncate the list
+    (format "(%s & %s more)"
+            (cider-propertize
+             (mapconcat cider-eldoc-ns-function (seq-take class-names cider-eldoc-max-class-names-to-display) " ")
```
bbatsov: There's also a `cider-string-join` function.

cider-eldoc.el
```diff
@@ -61,22 +62,62 @@ For convenience, some functions are already provided for this purpose:
...
+             'ns)
+            (- (length class-names) cider-eldoc-max-class-names-to-display)))
+
+   ;; format everything and add brackets
+   ((> (length class-names) 1)
+    (format "(%s)"
+            (cider-propertize (mapconcat cider-eldoc-ns-function class-names " ") 'ns)))
```
bbatsov: One more place you can using string-join.

cider-eldoc.el
cskksc: > Hopefully you've tested what happens when this is set to a function returning nil. :-)

😞 I forgot to do that. There was a check only when `ns-or-class` was a clojure ns. I've corrected it now and added a corresponding test.

cider-eldoc.el
cskksc: Done. `cider-string-join` also works with `thread-first`, which makes this a bit simple :-)

PR #1732, opened 2016-05-05
Document case when exec-path does not include non-standard paths

arbscht:
Noting a not-unusual Emacs configuration issue I ran into.

When `boot`/`lein` is not found in `exec-path`, then `cider-jack-in` fails with a cryptic error message. Emacs's stacktrace is enlightening if a user knows or cares to investigate, but this may be worth documenting as a troubleshooting tip anyway.

--

Thanks!

### DIFF
--- a/doc/troubleshooting.md
+++ b/doc/troubleshooting.md
@@ -114,6 +114,13 @@ that the order here matters.
 Emacs doesn't load the new files, it only installs them on disk.  To see the
 effect of changes you have to restart Emacs.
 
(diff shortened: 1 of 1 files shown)


Malabarba:
Thanks @arbscht. In this case, I think it would be much better (and not too difficult) to catch this error in Emacs and present a more informative message.

arbscht:
Fair enough @Malabarba. I see there is already code intended to handle this case, but there are bugs. I'll raise a separate issue.

PR #1733, opened 2016-05-05
Ensure cider-jack-in fails gracefully when commands are not found

arbscht:
**Problem:** Running `cider-jack-in` with a Boot project when `boot` or `boot.sh` is not found in `exec-path` yields an error like this:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  locate-file-internal(nil ("/usr/local/bin" "/usr/bin" "/bin" "/usr/local/games" "/usr/games" "/usr/lib/emacs/24.5/x86_64-linux-gnu") ("") 1)
  locate-file(nil ("/usr/local/bin" "/usr/bin" "/bin" "/usr/local/games" "/usr/games" "/usr/lib/emacs/24.5/x86_64-linux-gnu") ("") 1)
  executable-find(nil)
  cider--boot-present-p()
  cider-jack-in(nil)
  call-interactively(cider-jack-in record nil)
  command-execute(cider-jack-in record)
  execute-extended-command(nil "cider-jack-in")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

**Expected:** a more user-friendly message like this:
```The boot executable (specified by `cider-lein-command' or `cider-boot-command') isn't on your `exec-path'```

The source of the bug seems to be that `cider-boot-command` can be nil since it has a non-literal value, whereas `cider--boot-present-p` assumes it is non-nil. I've added a fallback literal value `"boot"` to guard against this causing problems.

Affects cider-20160501.1921 (melpa) and git master.

---

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -215,14 +215,16 @@ Sub-match 1 must be the project path.")
     ("lein" cider-lein-command)
     ("boot" cider-boot-command)
     ("gradle" cider-gradle-command)
-    (_ (error "Unsupported project type `%s'" project-type))))
+    (_ (user-error "Unsupported project type `%s'" project-type))))
 
 (defun cider-jack-in-resolve-command (project-type)
+  "Determine the resolved file path to `cider-jack-in-command' if it can be
+found for the PROJECT-TYPE"
   (pcase project-type
     ("lein" (cider--lein-resolve-command))
     ("boot" (cider--boot-resolve-command))
     ("gradle" (cider--gradle-resolve-command))
-    (_ (error "Unsupported project type `%s'" project-type))))
+    (_ (user-error "Unsupported project type `%s'" project-type))))
     
 (defun cider-jack-in-params (project-type)
   "Determine the commands params for `cider-jack-in' for the PROJECT-TYPE."
@@ -230,8 +232,7 @@ Sub-match 1 must be the project path.")
     ("lein" cider-lein-parameters)
     ("boot" cider-boot-parameters)
     ("gradle" cider-gradle-parameters)
-    (_ (error "Unsupported project type `%s'" project-type))))
-
+    (_ (user-error "Unsupported project type `%s'" project-type))))
 
 
 ;;; Jack-in dependencies injection
@@ -662,7 +663,7 @@ choose."
 
 ;; TODO: Implement a check for `cider-lein-command' over tramp
 (defun cider--lein-resolve-command ()
-  "Check if `cider-lein-command' is on the `exec-path'.
+  "Find `cider-lein-command' on `exec-path' if possible, or return `nil'.
 
 In case `default-directory' is non-local we assume the command is available."
   (or (file-remote-p default-directory)
@@ -670,7 +671,7 @@ In case `default-directory' is non-local we assume the command is available."
       (executable-find (concat cider-lein-command ".bat"))))
 
 (defun cider--boot-resolve-command ()
-  "Check if `cider-boot-command' is on the `exec-path'.
+  "Find `cider-boot-command' on `exec-path' if possible, or return `nil'.
 
 In case `default-directory' is non-local we assume the command is available."
   (or (file-remote-p default-directory)
@@ -678,7 +679,7 @@ In case `default-directory' is non-local we assume the command is available."
       (executable-find (concat cider-boot-command ".exe"))))
 
 (defun cider--gradle-resolve-command ()
-  "Check if `cider-gradle-command' is on the `exec-path'.
+  "Find `cider-gradle-command' on `exec-path' if possible, or return `nil'.
 
 In case `default-directory' is non-local we assume the command is available."
   (or (file-remote-p default-directory)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Malabarba** [comment]
I like this solution. IIRC, there are situations in which the exec-path might get set after our defcustom is defined, so it's a good idea to have a sane fallback. 
@arbscht Please also document this as a bug fix in Changelog.md

**Comment by bbatsov** [comment]
Is there the same problem with `lein` as well?

**Comment by bbatsov** [comment]
Btw, I guess you've seen the code here, right - https://github.com/clojure-emacs/cider/blob/master/cider.el#L497

**Comment by arbscht** [comment]
@bbatsov I don't think lein is affected. lein and gradle are defined as literal values, whereas boot is subject to change due to invoking `executable-find`. Compare:

```
(defcustom cider-lein-command
  "lein"
...)

(defcustom cider-boot-command
  (or (executable-find "boot")
      (executable-find "boot.sh")) ;; <-- this can be nil
...)

```

The nil value of `cider-boot-command` trickles down to `cider-present-p` and raises an error before the branching in [L476](https://github.com/clojure-emacs/cider/blob/master/cider.el#L476) can proceed.

This fix adds a default fallback literal value so that if `executable-find` returns nil, the `cider-present-p` check will still evaluate properly and return nil, so that the existing error handling path can happen.

**Comment by bbatsov** [comment]
I get what the fix does. I simply dislike we're doing different things for `lein`, `boot` and `gradle`. Maybe we should just remove `executable-find`. If someone uses `boot.sh` they'll just update the defcustom. @Malabarba thoughts?

File: cider.el [hunk 1]
```diff
@@ -110,7 +110,8 @@ version from the CIDER package or library.")
 
 (defcustom cider-boot-command
   (or (executable-find "boot")
-      (executable-find "boot.sh"))
+      (executable-find "boot.sh")
+      "boot")
```
bbatsov: If we do something like this the first check would be redundant. As I mentioned earlier - I'd rather simplify this to be just a constant to keep in aligned with other similar values.

Malabarba:
Ok. I'm with changing it to just boot.

bbatsov:
One more thing - the real place to fix the bug is the place where a nil can be passed to `executable-find` in the code I linked to. This should definitely be addressed as well.

arbscht:
I agree with changing the value to just "boot", and let "boot.sh" users simply customize the variable, instead of magically nesting `executable-find` in multiple layers.

I'm not sure if the main code path needs to protect itself against such nils. It seems like a reasonable assumption that a custom variable is non-nil, given that it is typed `'string` and Emacs enforces or interprets the type at `set-variable` and `customize-variable`. This is only unreliable in case of initializing the value to nil (as currently the case), but the problem really is that nil was used for a string type custom variable (no fallback).

cider.el
arbscht: If not searching for "boot.sh", then there's no need to search for "boot" at this point either. Simply setting "boot" should be enough as `executable-find` is called later with this value.

bbatsov:
Fine. Let's just change the default and wrap this.

On Thursday, 5 May 2016, Abhishek Reddy <notifications@github.com> wrote:

> I agree with changing the value to just "boot", and let "boot.sh" users
> simply customize the variable, instead of magically nesting
> executable-find in multiple layers.
>
> I'm not sure if the main code path needs to protect itself against such
> nils. It seems like a reasonable assumption that a custom variable is
> non-nil, given that it is typed 'string and Emacs enforces or interprets
> the type at set-variable and customize-variable. This is only unreliable
> in case of initializing the value to nil (as currently the case), but the
> problem really is that nil was used for a string type custom variable (no
> fallback).
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/clojure-emacs/cider/pull/1733#issuecomment-217309807>
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

arbscht:
Here's a somewhat bigger change to normalize behaviour between the present-p check and the command search code paths. This allows custom variables for `cider-{lein,boot,gradle}-command` to be simple string command names; the resolved location can safely be nil; the error message if a command is not found shows the configured name of the missing command.

cider.el
```diff
@@ -226,6 +217,13 @@ Sub-match 1 must be the project path.")
     ("gradle" cider-gradle-command)
     (_ (error "Unsupported project type `%s'" project-type))))
 
+(defun cider-jack-in-resolve-command (project-type)
+  (pcase project-type
+    ("lein" (cider--lein-resolve-command))
+    ("boot" (cider--boot-resolve-command))
+    ("gradle" (cider--gradle-resolve-command))
+    (_ (error "Unsupported project type `%s'" project-type))))
```
Malabarba: `user-error`

cider.el
```diff
@@ -235,6 +232,7 @@ Sub-match 1 must be the project path.")
     ("gradle" cider-gradle-parameters)
     (_ (error "Unsupported project type `%s'" project-type))))
 
+
```
bbatsov: Kill this blank line.

PR #1734, opened 2016-05-06
Fix eldoc errors for empty parentheses

cskksc:
I introduced this bug yesterday. When the cursor is on empty parentheses, an error is displayed in the minibuffer.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -128,11 +128,11 @@ find a symbol if there isn't one at point."
   "Return destructured `cider-symbol-at-point'.
 If the symbol at point is of the form ns-name/thing-name, returns a dict
 \(\"ns\" \"ns-name\" \"thing\" \"thing-name\")."
-  (let* ((sym (cider-symbol-at-point))
-         (ns-thing (split-string sym "/")))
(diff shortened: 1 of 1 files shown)


cider-util.el
```diff
@@ -128,11 +128,11 @@ find a symbol if there isn't one at point."
   "Return destructured `cider-symbol-at-point'.
 If the symbol at point is of the form ns-name/thing-name, returns a dict
 \(\"ns\" \"ns-name\" \"thing\" \"thing-name\")."
-  (let* ((sym (cider-symbol-at-point))
-         (ns-thing (split-string sym "/")))
-    (if (< (length ns-thing) 2)
-        (nrepl-dict "thing" (car ns-thing))
-      (nrepl-dict "ns" (car ns-thing) "thing" (cadr ns-thing)))))
+  (if-let ((sym (cider-symbol-at-point))
+           (ns-thing (split-string sym "/")))
```
cskksc: `sym` can be nil sometimes and this is not handled.

cskksc:
The tests have failed at `make melpa`.

PR #1738, opened 2016-05-07
[Fix #1735] Display eldoc only for valid forms

### LINKED ISSUE #1735: Improve class names in eldoc for interop forms
## Actual behavior
screen shot 2016-05-06 at 8 19 13 am

In a clojure buffer with active nREPL connection, enter the form ` (java.lang.X/.length)`
When the cursor is on the form `(java.lang.X/.length)`, the eldoc is shown as is ; `(java.lang.X/.length): ([this])`, because the ns entered by user is always prioritised.

## Suggested behaviour

From the information returned by the middleware, we know that `.length` is not a member of class X. Should we display some warning in such case ? But if that class is not loaded, or we don't have its eldoc information, our warning would be totally wrong.

This is just a "good to have" sort of thing, as the class name is rarely used with an interop form.
Relevant PR, #1729 .

cskksc:

Thanks!

If the user has entered invalid class/ns name, like `(X/.length)` or `(X/defn)`, the eldoc is not displayed.

### DIFF
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -210,7 +210,7 @@ if the maximum number of sexps to skip is exceeded."
       ;; Don't do anything if current word is inside a string, vector,
       ;; hash or set literal.
       (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
-        (list (cider-ns-thing-at-point) argument-index)))))
+        (list (cider-symbol-at-point) argument-index)))))
 
 (defun cider-eldoc--convert-ns-keywords (thing)
   "Convert THING values that match ns macro keywords to function names."
@@ -250,15 +250,25 @@ This includes the arglist and ns and symbol name (if available)."
               (let* ((arglist (nrepl-dict-get eldoc-info "eldoc"))
                      (ns (nrepl-dict-get eldoc-info "ns"))
                      (class (nrepl-dict-get eldoc-info "class"))
-                     (symbol (nrepl-dict-get eldoc-info "name"))
+                     (name (nrepl-dict-get eldoc-info "name"))
+                     (member (nrepl-dict-get eldoc-info "member"))
                      (ns-or-class (if (and ns (not (string= ns "")))
                                       ns
-                                    class)))
+                                    class))
+                     (name-or-member (if (and name (not (string= name "")))
+                                         name
+                                       (format ".%s" member))))
                 ;; middleware eldoc lookups are expensive, so we
                 ;; cache the last lookup.  This eliminates the need
                 ;; for extra middleware requests within the same sexp.
-                (setq cider-eldoc-last-symbol (list thing ns-or-class symbol arglist))
-                (list ns-or-class symbol arglist)))))))))
+                (setq cider-eldoc-last-symbol
+                      (list thing ns-or-class name-or-member arglist))
+                (list ns-or-class name-or-member arglist)))))))))
+
+(defun cider--eldoc-remove-dot-sym (sym)
+  "Remove the preceding \".\" from a namespace qualified SYM and return sym.
+Only useful for interop forms.  Clojure forms would be returned unchanged."
+  (when sym (replace-regexp-in-string "/\\." "/" sym)))
 
 (defun cider-eldoc ()
   "Backend function for eldoc to show argument list in the echo area."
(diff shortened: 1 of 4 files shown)


cider-eldoc.el
```diff
@@ -260,6 +260,21 @@ This includes the arglist and ns and symbol name (if available)."
                 (setq cider-eldoc-last-symbol (list thing ns-or-class symbol arglist))
                 (list ns-or-class symbol arglist)))))))))
 
+(defun cider--eldoc-ns (sexp-info eldoc-info)
+  "Extract and return the correct namespace from SEXP-INFO and ELDOC-INFO.
+Return the namespace entered by user, given in SEXP-INFO, if its valid.
```
bbatsov: its -> it's

cider-eldoc.el
```diff
@@ -260,6 +260,21 @@ This includes the arglist and ns and symbol name (if available)."
...
+Otherwise return the namespace given by the middleware, from ELDOC-INFO,
+if any.  Else return nil."
+  (if-let ((middleware-ns (nth 0 eldoc-info))
+           (user-ns (nrepl-dict-get (car sexp-info) "ns" middleware-ns)))
+      ;; return user-ns only if its valid
```
bbatsov: it's

cider-eldoc.el
```diff
@@ -260,6 +260,21 @@ This includes the arglist and ns and symbol name (if available)."
                 (setq cider-eldoc-last-symbol (list thing ns-or-class symbol arglist))
                 (list ns-or-class symbol arglist)))))))))
 
+(defun cider--eldoc-ns (sexp-info eldoc-info)
```
bbatsov: A few unit tests would be useful.

cider-eldoc.el
cskksc: I also forgot to add tests. On it.

cskksc:
Done :-)

test/cider-eldoc-tests.el
```diff
@@ -146,3 +146,34 @@
       (let ((cider-eldoc-max-num-sexps-to-skip 2))
         (expect (cider-eldoc-beginning-of-sexp) :to-equal nil)
         (expect (point) :to-equal 4)))))
+
+(describe "cider--eldoc-ns"
+  (describe "when the user has entered a valid ns"
+    (it "returns that ns"
+      (expect (cider--eldoc-ns `(,(nrepl-dict "ns" "ns-a" "thing" "thing-a") 0)
```
bbatsov: Why is the quasi-quoting needed here?

bbatsov: Ah, nevermind. I misread the code.

bbatsov:
Just one question - does this code handle properly namespace aliases?

cskksc:
Nice catch. It doesn't work with aliases. I'll fix this right now. Sorry about this.

cskksc:
@bbatsov Yes. I am working on the PR. It'll be done in a bit.

cskksc:
@bbatsov,

What went wrong:

1. When we introduced eldoc for classes, the method `ns-thing-at-point` was used to get the `thing` and `ns` and `thing` was sent to the middleware. This was done because the previous way getting eldoc involved sending, `java.lang.String/.length`, and it returned nil response for interop forms.

2. But the middleware is not able to find eldoc for a symbol from a required ns if we just pass a symbol. It needs the complete namespace prefixed form. This caused the breakage.

3. The root cause of the nil response for interop forms was not  just the namespace prefix. The "." causes problems too. So,`java.lang.String/length` is a valid input. Or just `.length` also works.

This PR;

So `ns-thing-at-point` method has been removed, and we use the old `symbol-at-point` to get eldoc from the middleware. We just remove the `.` from a namespace qualified form returned by `symbol-at-point`, before sending it to the middleware.

Also, this means that the thing var could be a `ns/thing` or just `thing`. So we destructure it before displaying it in the minibuffer. This automatically fixes the problem of having a wrong ns prefix before the symbol. The eldoc would be nil for such forms and we won't display them.

I've tested it with these symbols from the `cider.nrepl.middleware.info` ns.
```
(require '[clojure.string :as str])
(require '[cider.nrepl.middleware.util.java :as java])
(require '[cider.nrepl.middleware.util.namespace :as ns])

;; clj
(defn)
(clojure.core/defn)
(clojure.string/blank?)
(str/blank?)
(cider.nrepl.middleware.ns/ns-vars-clj)
(ns/ensure-namespace)
(java/member-info (u/as-sym "java.lang.String") (u/as-sym "length"))
;; invalid. should be empty
(clojure.X/defn)

;; interop
(.length)
(java.lang.String/.length)
;; invalid. should be empty
(java.lang.X/.length)
```

Have I missed anything ?

cskksc:
Also please review cider-nrepl#352 first. It's another mistake I made. It will eliminate the need for `cider-destructure-sym`.

bbatsov:
> I've tested it with these symbols from the cider.nrepl.middleware.info ns.

Seems your testing was exhaustive. It's bad that we can write an automated test for this in Emacs...

bbatsov:
> Also please review clojure-emacs/cider-nrepl#352 first. It's another mistake I made. It will eliminate the need for cider-destructure-sym.

Merged.

cskksc:
> It's bad that we can't write an automated test for this in Emacs...

Yeah.. That would've been of great help.

Issue #1638, opened 2016-03-29
Alt doc for higher-order functions


achesnais:
## Expected behavior

When using a higher-order function, CIDER could display the function signature of whichever function is being manipulated in the mini-buffer, rather than that of the higher-order function itself (or both if there's a clever way to do that).
For example, when writing `(partial some-function ...)`, the minibuffer would display the signature of `some-function`, not that of `partial`.
Similar behaviour could be expected of `map`, `filter`, etc.

I find I'm often going back to the signature of functions I'm manipulating with higher-order ones, as the order of arguments is always crucial when using `partial`, `map`, etc.
## Actual behavior

Currently CIDER displays the doc of whichever function comes first in the s-exp.

Malabarba:
I can see the benefit in that, but I can also see it might be surprising. So I'm on the fence about it. :-)

bbatsov:
Btw, something like this is on my personal radar. I'll probably implement it when I get do adding eldoc for variables (as the implementation is related - you have to consider the symbol at point in addition to the symbol at the beginning of the list).

bbatsov:
This is the related ticket, btw https://github.com/clojure-emacs/cider/issues/1572

Issue #1572, opened 2016-02-16
Add support for variables in eldoc


bbatsov:
In Emacs Lisp even variables have eldoc:

![image](https://cloud.githubusercontent.com/assets/103882/13078024/43cd9582-d4c5-11e5-9828-5d8042d92369.png)

Adding this to cider should be easy.

bbatsov:
Btw, @cskksc as you've been doing some eldoc-related work lately you might eventually take a look at this as well. Basically we just need to make eldoc consider the symbol-at-point in addition to the first symbol of the list at point (which we currently do). So in a form like `(map inc ...)` if your cursor is over `inc` you'd see its docstring and when you move forward you'd see once again the docstring for `map`.

cskksc:
Yeah. Will do. :-)

PR #1749, opened 2016-05-15
Alt doc for higher-order functions

cskksc:

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [#1720](https://github.com/clojure-emacs/cider/issues/1720): Add a command `cider-eval-sexp-at-point` to evaluate the form around point (bound to `C-c C-v v`).
 * [#1564](https://github.com/clojure-emacs/cider/issues/1564): CIDER's internal namespaces and vars are filtered from the ns-browser and apropos functions.
 * [#1725](https://github.com/clojure-emacs/cider/issues/1725): Display class names in eldoc for interop forms.
+* [#1572](https://github.com/clojure-emacs/cider/issues/1572): Add support for variables in eldoc.
 
 ### Changes
 
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -111,7 +111,7 @@ mapping `cider-eldoc-ns-function' on it returns an empty list."
      ;; don't add the parentheses
      (t (format "%s" (car eldoc-class-names))))))
 
-(defun cider-eldoc-format-thing (ns symbol thing function-p)
+(defun cider-eldoc-format-thing (ns symbol thing type)
   "Format the eldoc subject defined by NS, SYMBOL and THING.
 THING represents the thing at point which triggered eldoc.  Normally NS and
 SYMBOL are used (they are derived from THING), but when empty we fallback to
@@ -120,6 +120,9 @@ is non-nil.  Else format it as a variable."
   (if-let ((method-name (if (and symbol (not (string= symbol "")))
                             symbol
                           thing))
+           (propertized-method-name (if (equal type 'function)
+                                        (cider-propertize method-name 'var)
+                                      (cider-propertize method-name 'font-lock-variable-name-face)))
            (ns-or-class (if (and ns (stringp ns))
                             (funcall cider-eldoc-ns-function ns)
                           (cider--eldoc-format-class-names ns))))
@@ -131,31 +134,30 @@ is non-nil.  Else format it as a variable."
                   ;; it is already propertized
                   ns-or-class
                 (cider-propertize ns-or-class 'ns))
-              (if function-p
-                  (cider-propertize method-name 'var)
-                (cider-propertize method-name 'font-lock-variable-name-face)))
+              propertized-method-name)
     ;; in case ns-or-class is nil
-    (if function-p
-        (cider-propertize method-name 'var)
-      (cider-propertize method-name 'font-lock-variable-name-face))))
-
-(defun cider-eldoc-function-p (eldoc-info)
-  "Return non-nil if ELDOC-INFO contains a value of \"doc\"."
-  (lax-plist-get eldoc-info "arglists"))
+    propertized-method-name))
 
 (defun cider-eldoc-format-variable (thing pos eldoc-info)
-  (let ((ns (lax-plist-get eldoc-info "ns-or-class"))
-        (symbol (lax-plist-get eldoc-info "name-or-member"))
-        (doc (lax-plist-get eldoc-info "doc")))
-    (when doc
-      (format "%s: %s" (cider-eldoc-format-thing ns symbol thing nil) doc))))
+  "Return the formatted eldoc string for a variable.
+THING is the variable name.  POS will always be 0 here.
+ELDOC-INFO is a p-list containing the eldoc information."
+  (let ((ns (lax-plist-get eldoc-info "ns"))
+        (symbol (lax-plist-get eldoc-info "symbol"))
+        (docstring (lax-plist-get eldoc-info "docstring")))
+    (when docstring
+      (format "%s: %s" (cider-eldoc-format-thing ns symbol thing 'var)
+              docstring))))
 
 (defun cider-eldoc-format-function (thing pos eldoc-info)
-  (let ((ns (lax-plist-get eldoc-info "ns-or-class"))
-        (symbol (lax-plist-get eldoc-info "name-or-member"))
+  "Return the formatted eldoc string for a function.
+THING is the function name.  POS is the argument-index of the functions
+arglists.  ELDOC-INFO is a p-list containing the eldoc information."
+  (let ((ns (lax-plist-get eldoc-info "ns"))
+        (symbol (lax-plist-get eldoc-info "symbol"))
         (arglists (lax-plist-get eldoc-info "arglists")))
     (format "%s: %s"
-            (cider-eldoc-format-thing ns symbol thing t)
+            (cider-eldoc-format-thing ns symbol thing 'function)
             (cider-eldoc-format-arglist arglists pos))))
 
 (defun cider-highlight-args (arglist pos)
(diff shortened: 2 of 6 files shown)


cskksc:
I have checked this with these forms:

```
;; clojure
(defn)
(clojure.string/blank?)
(clojure.core/defn)
(str/blank?)
(cider.nrepl.middleware.ns/ns-vars-clj)
(ns/ensure-namespace)
(java/member-info (u/as-sym "java.lang.String") (u/as-sym "length"))
;; invalid. should be empty
(clojure.X/defn)

;; interop
(.length)
(java.lang.String/.length)
;; invalid. should be empty
(java.lang.X/.length)


;; combinations

(map inc [1 2 3 4 5])

(reduce + [1 2 3 4 5])

((comp inc (partial + 3)) 3)

(map :a [{:a 1 :b 2} {:a 3 :b 4}])
```

Please try out once :-)

cider-eldoc.el
```diff
@@ -202,15 +202,21 @@ if the maximum number of sexps to skip is exceeded."
...
-               (argument-index (1- beginning-of-sexp)))
-      ;; If we are at the beginning of function name, this will be -1.
-      (when (< argument-index 0)
-        (setq argument-index 0))
-      ;; Don't do anything if current word is inside a string, vector,
-      ;; hash or set literal.
-      (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
-        (list (cider-symbol-at-point) argument-index)))))
+    (let ((thing-at-point (cider-symbol-at-point))
+          (pos-thing-at-point (point)))
```
cskksc: I'm not sure if this is right. But even without this var, eldoc seems to work correctly.

cskksc: I think I should remove this var. It highlights the argument, but it is subtle so I din't notice the difference. But in a sexp like `(map inc ...)`, inc wouldn't have any arguments to highlight.
And even if the mapping function is something like `+` which can take 2 args, the cursor position would be such that it would never need to highlight the arguments.
Does this make sense ?

cider-eldoc.el
```diff
@@ -202,15 +202,19 @@ if the maximum number of sexps to skip is exceeded."
...
+    (let ((thing-at-point (cider-symbol-at-point)))
+      (when-let ((beginning-of-sexp (cider-eldoc-beginning-of-sexp))
+                 (argument-index (1- beginning-of-sexp)))
+        ;; If we are at the beginning of function name, this will be -1.
+        (when (< argument-index 0)
+          (setq argument-index 0))
+        ;; Don't do anything if current word is inside a string, vector,
+        ;; hash or set literal.
+        (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
+          (nrepl-dict
```
bbatsov: Don't use an `nrepl-dict` for this. I'd go with something simpler like a plist or an alist.

cider-eldoc.el
```diff
@@ -276,9 +280,20 @@ Only useful for interop forms.  Clojure forms would be returned unchanged."
...
-           (pos (cadr sexp-info))
-           (eldoc-info (cider-eldoc-info (cider--eldoc-remove-dot-sym thing)))
+           (first-thing (nrepl-dict-get sexp-info "first-thing"))
+           (thing-at-point (nrepl-dict-get sexp-info "thing-at-point"))
+           (pos (nrepl-dict-get sexp-info "pos"))
+           ;; If the thing-at-point has eldoc info available, display that.
+           ;; Else display the eldoc for first-thing-at-point.
+           ;; So in (map inc ...), if the cursor is over `inc`, it's eldoc would
+           ;; be displayed. Otherwise display eldoc of `map`.
+           (first-thing-eldoc-info (cider-eldoc-info (cider--eldoc-remove-dot first-thing)))
```
bbatsov: The `let*`

cider-eldoc.el
```diff
@@ -202,15 +202,19 @@ if the maximum number of sexps to skip is exceeded."
...
+      (when-let ((beginning-of-sexp (cider-eldoc-beginning-of-sexp))
+                 (argument-index (1- beginning-of-sexp)))
+        ;; If we are at the beginning of function name, this will be -1.
+        (when (< argument-index 0)
+          (setq argument-index 0))
+        ;; Don't do anything if current word is inside a string, vector,
+        ;; hash or set literal.
+        (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
+          (nrepl-dict
+           "first-thing" (cider-symbol-at-point)
```
bbatsov: rename first-thing to first-sexp-sym or sym-at-sexp-beginning. I was this to have a more descriptive name.

bbatsov:
There should also be some `defcustom` to restore the old behaviour, as some people might prefer it.

cskksc:
Ill make all the changes and update this PR.

cskksc:
Done :-) All the `let`s still don't look pretty enough, but better than before I guess.

cider-eldoc.el
```diff
@@ -202,15 +211,30 @@ if the maximum number of sexps to skip is exceeded."
...
-               (argument-index (1- beginning-of-sexp)))
-      ;; If we are at the beginning of function name, this will be -1.
-      (when (< argument-index 0)
-        (setq argument-index 0))
-      ;; Don't do anything if current word is inside a string, vector,
-      ;; hash or set literal.
-      (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
-        (list (cider-symbol-at-point) argument-index)))))
+    (let ((sym-at-point (cider-symbol-at-point)))
+      ;; cider-eldoc-beginning-of-sexp is nil for variables
```
Malabarba: This comment makes it sound like nothing will be done if we're not at the beginning of sexp. Could you elaborate to make it clearer?

cider-eldoc.el
```diff
@@ -202,15 +211,30 @@ if the maximum number of sexps to skip is exceeded."
...
+      (when-let ((beginning-of-sexp (cider-eldoc-beginning-of-sexp))
+                 (argument-index (1- beginning-of-sexp)))
+        (let* (;; Don't do anything if current word is inside a string, vector,
+               ;; hash or set literal.
+               (sym-at-sexp-beginning (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
```
Malabarba: `(char-after (1- (point)))` => `(char-before (point))`

cider-eldoc.el
```diff
@@ -202,15 +211,30 @@ if the maximum number of sexps to skip is exceeded."
...
-        (setq argument-index 0))
-      ;; Don't do anything if current word is inside a string, vector,
-      ;; hash or set literal.
-      (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
-        (list (cider-symbol-at-point) argument-index)))))
+    (let ((sym-at-point (cider-symbol-at-point)))
+      ;; cider-eldoc-beginning-of-sexp is nil for variables
+      (when-let ((beginning-of-sexp (cider-eldoc-beginning-of-sexp))
+                 (argument-index (1- beginning-of-sexp)))
+        (let* (;; Don't do anything if current word is inside a string, vector,
```
Malabarba: Word = symbol, right?

cider-eldoc.el
Malabarba: Will this actually work consistently? For instance, what if the symbol at point is the second symbol inside a hashmap? Would you catch that here?

bbatsov:
The highlighting of the current argument doesn't make sense for the symbol at point's docstring, but makes a lot of sense for everything else. And as far as higher order functions go I believe the idea on the ticket for them was to hardcode a list of known such functions and to have some special treatment for them. That's not something important, I'm just pointing it out. 

Also now that you can get an eldoc for the variable name at point - we should font-lock the type of the eldoc and for variables we should probably show the first line of their docstring as they don't have arguments. This is how eldoc works for variables in Elisp.

cider-eldoc.el
bbatsov: The original code checked only lists (something surrounded with `()`). Eldoc will never be triggered inside a vector or some other literal.

cskksc: Yep. This works because we move the cursor to the start of the sexp with `(cider-eldoc-beginning-of-sexp)`.

cskksc:
This actually doesn't get the eldoc for a variable. It only considers the `(map inc ...)` situation, #1638.
I'll raise another PR with those changes, for #1572.

cider-eldoc.el
```diff
@@ -202,15 +211,30 @@ if the maximum number of sexps to skip is exceeded."
...
+               ;; hash or set literal.
+               (sym-at-sexp-beginning (unless (member (or (char-after (1- (point))) 0) '(?\" ?\{ ?\[))
+                                        (cider-symbol-at-point)))
+               (sym-at-point-eldoc-info (cider-eldoc-info (cider--eldoc-remove-dot sym-at-point)))
+               (sym-at-sexp-beginning-eldoc-info (cider-eldoc-info (cider--eldoc-remove-dot sym-at-sexp-beginning))))
```
Malabarba: Inline this function call so we'll don't do the request when we don't have to.

bbatsov:
> This actually doesn't get the eldoc for a variable. It only considers the (map inc ...) situation, #1638.
> I'll raise another PR with those changes, for #1572.

Meaning that you'll get eldoc for the second symbol after your cursor goes pass it? Or you just get eldoc at point for whatever symbol now? I still don't quite get what exactly does the PR solve.

cskksc:
@bbatsov This PR just handles what this comment describes [here](https://github.com/clojure-emacs/cider/issues/1572#issuecomment-217732657):

> Basically we just need to make eldoc consider the symbol-at-point in addition to the first symbol of the list at point (which we currently do). So in a form like (map inc ...) if your cursor is over inc you'd see its docstring and when you move forward you'd see once again the docstring for map.

For the eldoc of any form at point, we require a small nrepl change [here](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/info.clj#L312). I've was a bit busy yesterday so I couldn't follow up on this. But I'm gonna finish both of these tickets today. I've already made the nrepl change, just have to change the elisp interface. I'll submit the nrepl PR, and the "eldoc for any form at point including variables" and all the changes @Malabarba suggested here, by tonight. :-)

bbatsov:
OK, thanks for the update.

cskksc:
I've made all the changes and tested everything. This PR now includes both, `eldoc for variables` and `eldoc for higher order functions`.

I've tried to make the functions small enough to avoid comments all together. There's probably some scope for more refactoring here, but haven't made any of those changes here.

Depends on clojure-emacs/cider-nrepl#360.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [#1720](https://github.com/clojure-emacs/cider/issues/1720): Add a command `cider-eval-sexp-at-point` to evaluate the form around point (bound to `C-c C-v v`).
 * [#1564](https://github.com/clojure-emacs/cider/issues/1564): CIDER's internal namespaces and vars are filtered from the ns-browser and apropos functions.
 * [#1725](https://github.com/clojure-emacs/cider/issues/1725): Display class names in eldoc for interop forms.
+* [#1638](https://github.com/clojure-emacs/cider/issues/1638): Alternate documentation for higher-order functions.
```
bbatsov: Just to clarify - the issue about alternative eldoc for higher order functions was different, so I wouldn't call this HOF-related. The idea there was in a situation like this:

```
(map inc [1 2 3|])
```

to show the eldoc for `inc`, not `map` (| denotes the cursor). This means this cannot be generic, as changing the eldoc like this doesn't make sense for every higher order function - it should be applied just for a list of know functions.

cider-eldoc.el
```diff
@@ -73,6 +73,15 @@ a \"& x more\" suffix. Otherwise, all the classes are displayed."
   :group 'cider
   :package-version '(cider . "0.13.0"))
 
+(defcustom cider-eldoc-sexp-at-point t
```
bbatsov: Rename this to `cider-eldoc-show-for-symbol-at-point`.

cider-eldoc.el
```diff
@@ -73,6 +73,15 @@ a \"& x more\" suffix. Otherwise, all the classes are displayed."
...
+  "When non-nil, display eldoc for symbol at point if available.
+So in `(map inc ...)` when the cursor is over `inc` its eldoc would be
```
bbatsov: We don't use backquotes in Elisp docstrings.

doc/configuration.md
```diff
@@ -12,6 +12,14 @@ experience.
 
 ![Eldoc](images/eldoc.png)
 
+CIDER also supports alternate eldoc for higher order functions. So in `(map inc ...)`
```
bbatsov: also would show the eldoc for the symbol at point

cider-eldoc.el
```diff
@@ -111,11 +111,12 @@ mapping `cider-eldoc-ns-function' on it returns an empty list."
      ;; don't add the parentheses
      (t (format "%s" (car eldoc-class-names))))))
 
-(defun cider-eldoc-format-thing (ns symbol thing)
+(defun cider-eldoc-format-thing (ns symbol thing function-p)
```
bbatsov: Why don't you just call the final argument `type` so it would be 'function, 'method, 'var, etc? I hate boolean flags.

test/cider-eldoc-tests.el
```diff
@@ -76,47 +76,47 @@
 
   (describe "when ns is given and it exists"
     (it "returns formatted eldoc strings of form ns/symbol"
-      (expect (cider-eldoc-format-thing "clojure.core" "map" "map")
+      (expect (cider-eldoc-format-thing "clojure.core" "map" "map" nil)
```
bbatsov: And this is why I hate them - they result in passing around a lot of nil/t (which don't carry much meaning).

CHANGELOG.md
cskksc: Oh. That makes sense. The first thing to do is change the commit message probably so the issue isn't incorrectly closed. And the commit should still be merged right ?

bbatsov: Yep. The references you were using were the primary source of my confusion - as what was suggested in the ticket was different and I said we can simplify it if we extend my idea about variable eldoc to all symbols at point. :-)

test/cider-eldoc-tests.el
cskksc: Yep. I'll change them.

cskksc:
I'm done with all the changes. Before uploading, wanted to change something else. Is [this](https://github.com/clojure-emacs/cider/blob/master/cider-util.el#L603) a little confusing ? How about we change it to `var -> font-lock-variable-name-face` and `fn -> font-lock-function-name-face` ?

CHANGELOG.md
cskksc: Yeah.. Sorry about that. Now I understand correctly 💡 . I'll change that in another PR though. This one is huge even now :-)

cskksc:
So most of the things are handled in this PR. Found a issue with this.
The `cider-eldoc-info-at-point`, returns eldoc even if when the cursor is on a line which is commented out.
I want to avoid doing, `go back to start of the line, check char at point and then decide`. Have we implemented something similar anywhere ?

Malabarba:
This checks if you're in a comment or a string:

    (let ((s (syntax-ppss)))
      (or (nth 4 s)
          (nth 3 s)))

bbatsov:
Just wrap the code @Malabarba shared with you in some helper function and you're good to go.

expez:
We already have `cider-in-string-p` and `cider-in-comment-p`.

bbatsov:
Guess we forgot about them. Good catch!

bbatsov:
P.S. I'm a bit surprised those aren't in `clojure-mode` itself.

cskksc:
Thanks! Ready to merge now I think.

expez:
> P.S. I'm a bit surprised those aren't in clojure-mode itself.

haha ;) I argued for that, and initially opened the PR against that repo, but you felt they belonged in CIDER.

bbatsov:
> haha ;) I argued for that, and initially opened the PR against that repo, but you felt they belonged in CIDER.

Now I remembered - my concern was that we weren't using them anywhere in `clojure-mode` itself.

Issue #1746, opened 2016-05-12
Show "See Also" links for all functions


cskksc:
Right now (#1736), the middleware has `See Also` information only for `clojure.core`.
But it would be good to be able to show the links for all namespaces. Parsing the docstring and generating links from there is one way to do it. So, the docstring  would look something like,

```
Returns a lazy sequence of the items in coll for which
(pred item) returns true. pred must be free of side-effects.
Returns a transducer when no collection is provided.

See also: `clojure.core/remove', `clojure.core/keep', `user-ns/user-fn'
```

So we just need to collect all the `` `' `` pairs and replace them with buttons which would lookup the corresponding documentation.

Relevant discussion [here](https://github.com/clojure-emacs/cider/issues/1736#issuecomment-217754434)

Malabarba:
Two things:

1. We should use ``` `clojure.core/remove` ```, not ``` `clojure.core/remove'```.
2. We should try to linkify any occurrence of ``` `function-in-quotes` ``` in the docstring, not just the "see also:" links.

cskksc:
Fixed by #1750.

PR #1750, opened 2016-05-15
[#1746] Parse docstrings to create help xrefs

cskksc:
The existing doc buffers can only create "See Also" hyperlinks to functions returned by the middleware.
This would allow users to specify clojure vars in docstrings and have them hyperlinked in the same way.


cider-doc.el
```diff
@@ -451,19 +451,16 @@ Tables are marked to be ignored by line wrap."
...
                   (let* ((ns-sym-split (split-string ns-sym "/"))
                          (see-also-ns (car ns-sym-split))
                          (see-also-sym (cadr ns-sym-split))
                          ;; if the fn belongs to the same ns,
                          ;; don't display the namespace prefixed name
                          (symbol (if (equal ns see-also-ns) see-also-sym ns-sym)))
-                    (insert-button symbol
-                                   'type 'help-xref
-                                   'help-function (apply-partially #'cider-doc-lookup symbol)))
+                    (insert "`" symbol "`"))
```
cskksc: Should the `` ` `` be configurable ?

Just put this up here for review. This is not ready for merge.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,22 @@ Keyboard shortcut               | Description
 <kbd>^</kbd>                    | Browse all namespaces.
 <kbd>n</kbd>                    | Go to next line.
 <kbd>p</kbd>                    | Go to previous line.
+
+## Documentation buffers include "See Also" links
+
+You can add links to other functions by including its name in `` `...` `` in the docstring.
```
bbatsov: Technically this will work for all vars + java symbols.

cider-doc.el
bbatsov: I guess hardcoding this is ok for now. Most people use backticks. The only slight problem is that some people use backticks for locals as well, so we might create some links that will be broken. Guess ideally we should check that thing is the backticks doesn't appear in the arglists, but that's not a big deal.

If only there was some good convention about structuring docstrings in Clojure (like in Elisp)...

Malabarba: Yes. I prefer it hard coded too. We could ask people

Malabarba: We could ask people on the Clojure list what style they tend to use on docstrings.

cskksc: > If only there was some good convention about structuring docstrings in Clojure (like in Elisp)...
😄  yeah..

@Malabarba Yep. Would you please put a message there ? And on slack probably. I think it would be more effective than me asking.

cskksc: @Malabarba I'm posting a message on the thread. You repost it or something, if people don't post replies.

cskksc: Done. https://groups.google.com/forum/#!topic/clojure/b9NOo0hOT9g

bbatsov: I think we can skip the configuration for now - if someone starts complaining, we'll revisit this.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,22 @@ Keyboard shortcut               | Description
...
+If the function is in another namespace, then you'll have to include the full
+namespace qualified name in the docstring.
+
+![Cider Documentation](images/cider_see_also.gif)
```
bbatsov: This should say [CIDER See Also] I guess.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,22 @@ Keyboard shortcut               | Description
...
+
+Example function with a docstring containing links:
+
+```
+(defn test-fn
+  "Test function.
+  Also see: `clojure.core/map`, `clojure.core/reduce`, `defn`,
+  `var-meta-whitelist`, `cider.nrepl.middleware.info/resolve-var`"
```
bbatsov: Add the final `.` to this docstring.

doc/miscellaneous_features.md
bbatsov: its name -> their names

bbatsov:
Guess we should move forward to complete this. Address the small remarks and rebase the PR.

cskksc:
Yep. Ill finish this now. Lets go ahead with `` ` ``.

> I think we can skip the configuration for now - if someone starts complaining, we'll revisit this.

Guessing from the replies the post got, this is going to be very soon 😄

bbatsov:
Making this configurable via a function might be best then.

On Thursday, 19 May 2016, Chaitanya Koparkar <notifications@github.com>
wrote:

> Yep. Ill finish this now. Lets go ahead with `.
>
> I think we can skip the configuration for now - if someone starts
> complaining, we'll revisit this.
>
> Guessing from the replies the post got, this is going to be very soon 😄
>
> —
> You are receiving this because you commented.
> <https://github.com/clojure-emacs/cider/pull/1750#issuecomment-220382061>
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

cskksc:
Sorry for the delay. I've started working on this now and Ill update this PR in sometime. Another thing which I realised when doing this; should we remove the start and end separators (`` ` ``, or `[[ ]]`) from the docstring ? Emacs, clojure-mode or cider may be, highlights the text between `` ` ``, so it looks good. But I tried out the separators which James was using, `[[ ... ]]` and those brackets don't look so good in the doc buffer.

screen shot 2016-05-21 at 4 05 30 pm

cider-util.el
```diff
@@ -454,6 +454,100 @@ restore it properly when going back."
...
+  "Parse and return the first clojure symbol in BUFFER.
+Uses `cider-doc-xref-regexp' to search the symbols.
+Return nil if there are no more matches in the buffer."
+  (with-current-buffer (or buffer (current-buffer))
+    (when (re-search-forward cider-doc-xref-regexp nil t)
+      (list "symbol" (match-string 1)
+            "beg" (match-beginning 1)
+            "end" (match-end 1)))))
+
+(defcustom cider-parse-symbol-docstring-fn #'cider--parse-symbol-docstring
```
cskksc: The function is a `defcustom` but most cases will be handled by the variables. Not sure if this is required.  We require the variables to insert the around symbols returned by the middleware.

cider-util.el
```diff
@@ -454,6 +454,100 @@ restore it properly when going back."
...
+that.  A similar reference `forward' is made to traverse in the opposite
+direction.  Variable `help-forward-label' specifies the text for that."
+  (interactive "b")
+
+  ;; parse the docstring and create xrefs for symbols
+  (save-excursion
+    (goto-char (point-min))
+    (if-let (current-sym (funcall cider-parse-symbol-docstring-fn))
+        (while current-sym
+          (when-let ((symbol (lax-plist-get current-sym "symbol"))
```
cskksc: Is there a better way to do it ? Something like `while-let` ?

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+
+Example function with a docstring containing links:
+
+```
+(defn test-fn
+  "Test function.
+  Also see: `clojure.core/map`, `clojure.core/reduce`, `defn`,
+  `var-meta-whitelist`, `cider.nrepl.middleware.info/resolve-var`.
+  Also creates references to java interop forms like `java.lang.String/.length`"
+  []
```
cskksc: I forgot the `.` here. I'll add it with the other changes after you review this.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+  (+ 1 1))
+```
+
+#### Configuration:
```
bbatsov: Drop the `:` here.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+
+If you want to change the `` ` `` used to identify clojure symbols,
```
bbatsov: Guess you can just say "change the delimiters (backticks by default)" here.

cider-util.el
```diff
@@ -454,6 +454,100 @@ restore it properly when going back."
         (if tail (setcdr tail nil))))
     (setq help-xref-stack-item item)))
 
+(defcustom cider-doc-xref-start "`"
+  "The string indicating the start of a clojure symbol.
```
bbatsov: Capitalize usages of Clojure - it's a name after all.

cider-util.el
```diff
@@ -454,6 +454,100 @@ restore it properly when going back."
...
+
+(defcustom cider-doc-xref-end "`"
+  "The string indicating the end of a clojure symbol.
+This is used to create and parse xrefs in the doc buffers."
+  :type 'string
+  :safe #'stringp
+  :group 'cider
+  :package-version '(cider . "0.13.0"))
+
+(defcustom cider-doc-xref-regexp "`\\(.*?\\)`"
```
bbatsov: Can't this be inferred since we know the start and end delimiter? It seems to me we can have just this defcustom, or just the other two.

cider-util.el
bbatsov: The current code looks ok to me.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
 <kbd>^</kbd>                    | Browse all namespaces.
 <kbd>n</kbd>                    | Go to next line.
 <kbd>p</kbd>                    | Go to previous line.
+
+## Documentation buffers include "See Also" links
+
+You can add links to other functions by including their names in `` ` `` in the docstring.
```
bbatsov: Are you sure those backticks will be rendered properly?

links -> references
functions -> vars

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+If the function is in another namespace, then you'll have to include the full
```
bbatsov: Same here.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+namespace qualified name in the docstring.
+
+![CIDER See Also](images/cider_see_also.gif)
```
bbatsov: This image can be resized a bit - right now it wastes a lot of vertical space. I'd just just resizing your Emacs to be just as big as needed before doing the gif capture.

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+
+Example function with a docstring containing links:
```
bbatsov: links -> references

doc/miscellaneous_features.md
```diff
@@ -185,3 +185,49 @@ Keyboard shortcut               | Description
...
+
+For most cases, the variables specified above provide sufficient flexibility to configure
+the docstring parser. But for some reason, if you want to change the function used to
+parse the docstring, you can do so by:
+
+```
+(setq cider-parse-symbol-docstring-fn #'your-fn)
+```
+
+Please refer to its documentation for the more details.
```
bbatsov: The "please" here is redundant.

cider-util.el
Malabarba: The if-let needs an extra pair of parentheses.

cider-util.el
cskksc: I had done the same thing before. But since it is being used in a regexp, for symbols like `[[`, we would have to set it to `\\[\\[`. And for some reason, this doesn't work well in the `insert` function used [here](https://github.com/clojure-emacs/cider/pull/1750/files#diff-67cc413fb696b5d1b980f1571be07f7bR465). I guess we could do a string replace to remove the backquotes before inserting. I'll try that.

Malabarba: I don't think we need these many variables. The way we insert see-also links doesn't have to be customizable, it's only ever going to be used inside cider. 

Just make xref-regexp customizable. The creation of links has to support the different link notation other people use, so it makes sense to be configurable.

cskksc: You're right. But right now we're not deleting the delimiters from the doc buffers. So the cider generated "See Also" links wont be visually consistent with the the user generated ones.

Malabarba: I don't mind that, and I think that's easy to fix (by deleting the delimiters).

Malabarba: Not that we need to fix it immediately m

cskksc: Oh, alright. But it would be difficult to delete them if we only accept the regexp. It may be not be straightforward to infer the delimiters from the regexp. How about we only keep the start and end delimiters customisable and infer the regexp from there ?

I dint do this before because while inserting the symbols with a delimiter like `\\[\\[`, I din't do a string replace for the `\` which resulted in broken links. It was pretty stupid.

So we'll have only 2 defcustom's; `cider-doc-xref-end` and `cider-doc-xref-end`. We could also remove the function defcustom. I don't think it'll be that useful. Is that ok ?

Malabarba: > Oh, alright. But it would be difficult to delete them if we only accept the regexp. 

Not at all, we just say that the regexp needs to have a group numbered 1 which contains the var name. The we can remove the delimiters by replacing the entire match with `(match-string 1)`. I'm happy to do that myself after we merge this PR.

cskksc: 👍 Lets only keep the regexp then.

cskksc:
I've addressed all the comments. This can be merged I think.
@Malabarba The diff looks much better with just 1 defcustom :-).

doc/miscellaneous_features.md
cskksc: Yep. Checked with local mkdocs.

cider-util.el
```diff
@@ -454,6 +454,72 @@ restore it properly when going back."
         (if tail (setcdr tail nil))))
     (setq help-xref-stack-item item)))
 
+(defcustom cider-doc-xref-regexp "`\\(.*?\\)`"
+  "The regexp used to search Clojure vars in doc buffers."
+  :type 'string
```
Malabarba: :type regexp

Malabarba:
I'll probably make a couple more improvements to this, but it LGTM for merging. I'll let @bbatsov do the honors.

cskksc:
Alright :-)

bbatsov:
👍

cskksc:
I've been trying this out for the last two days and have not found any bugs. What about you guys ? If not, then we can post a message on the mailing list, linking to the docs, for completeness.

@Malabarba It looks even better with your changes :-) `replace-match` FTW

PR #1753, opened 2016-05-19
Add font-lock-variable-name-face in cider-propertize

cskksc:
Add an option `'var` to `cider-propertize` which would propertize the text as a variable. This is not compulsory, as the method also accepts any `face`. But it currently maps `'var` to `font-lock-function-name-face` which could be confusing. So the mapping is changed to `var -> font-lock-variable-name-face` and `fn -> font-lock-function-name-face`

### DIFF
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -121,8 +121,8 @@ is non-nil.  Else format it as a variable."
                             symbol
                           thing))
            (propertized-method-name (if (equal type 'function)
-                                        (cider-propertize method-name 'var)
-                                      (cider-propertize method-name 'font-lock-variable-name-face)))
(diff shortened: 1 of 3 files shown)


cider-eldoc.el
```diff
@@ -121,8 +121,8 @@ is non-nil.  Else format it as a variable."
                             symbol
                           thing))
            (propertized-method-name (if (equal type 'function)
-                                        (cider-propertize method-name 'var)
-                                      (cider-propertize method-name 'font-lock-variable-name-face)))
```
cskksc: An example where the current mapping could be confusing.

bbatsov:
👍

PR #1757, opened 2016-05-22
[Fix #1577] Show first line of docstring in ns browser

### LINKED ISSUE #1577: Show first-line of docstring on the ns browser

mallt:
Show the first line of the docstring in the ns browser after the var name.
-----------------

### DIFF
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -114,6 +114,31 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
                             'cider-browse-ns-current-ns ns)))
       (goto-char (point-min)))))
 
+(defun cider-browse-ns--first-doc-line (doc)
+  "Return the first line of the given DOC string.
+If the first line of the DOC string contains multiple sentences, only
+the first sentence is returned.  If the DOC string is nil, a Not documented
+string is returned."
+  (if doc
+      (let* ((split-newline (split-string (read doc) "\n"))
+             (first-line (car split-newline)))
+        (cond
+         ((string-match "\\. " first-line) (substring first-line 0 (match-end 0)))
+         ((= 1 (length split-newline)) first-line)
+         (t (concat first-line "..."))))
+    "Not documented."))
+
+(defun cider-browse-ns--items (namespace)
+  "Return the items to show in the namespace browser of the given NAMESPACE.
+Each item consists of a ns-var and the first line of its docstring."
+  (let* ((ns-vars-with-meta (cider-sync-request:ns-vars-with-meta namespace))
+         (propertized-ns-vars (nrepl-dict-map #'cider-browse-ns--properties ns-vars-with-meta)))
+    (mapcar (lambda (ns-var)
+              (let* ((doc (nrepl-dict-get-in ns-vars-with-meta (list ns-var "doc")))
+                     (first-doc-line (cider-browse-ns--first-doc-line doc)))
+                (concat ns-var " " (propertize first-doc-line 'font-lock-face 'font-lock-doc-face))))
+            propertized-ns-vars)))
+
 ;; Interactive Functions
 
 ;;;###autoload
@@ -123,8 +148,7 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
   (with-current-buffer (cider-popup-buffer cider-browse-ns-buffer t)
     (cider-browse-ns--list (current-buffer)
                            namespace
-                           (nrepl-dict-map #'cider-browse-ns--properties
-                                           (cider-sync-request:ns-vars-with-meta namespace)))
+                           (cider-browse-ns--items namespace))
     (setq-local cider-browse-ns-current-ns namespace)))
 
 ;;;###autoload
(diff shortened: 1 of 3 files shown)


cider-browse-ns.el
```diff
@@ -123,9 +123,14 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
   (with-current-buffer (cider-popup-buffer cider-browse-ns-buffer t)
     (cider-browse-ns--list (current-buffer)
                            namespace
-                           (nrepl-dict-map #'cider-browse-ns--properties
-                                           (cider-sync-request:ns-vars-with-meta namespace)))
-    (setq-local cider-browse-ns-current-ns namespace)))
+                           (let* ((ns-vars (nrepl-dict-map #'cider-browse-ns--properties
+                                                           (cider-sync-request:ns-vars-with-meta namespace))))
+                             (mapcar (lambda (ns-var)
+                                       (let* ((doc (nrepl-dict-get (cider-var-info (concat namespace "/" ns-var)) "doc"))
```
Malabarba: I think that the return value of cider-sync-request:ns-vars-with-meta already contains the docstrings. We should use that instead of making an additional nrepl request for each var.

mallt: Ok thanks, then I missed that. I will check it and update the PR.

mallt: @Malabarba It seems the ns-vars-with-meta only contains the arglists, not the docstring. Should a new sync request be defined for this?

Malabarba: No need for a new request. We can just add docstrings to `ns-vars-with-meta`.

mallt: @Malabarba I've added the docstring to ns-vars-with-meta in cider-nrepl, but the strings are fully quoted, f.ex. "\"Converts first character of the string to upper-case, all other\\\n  characters to lower-case.\"". What would be the best way to receive the strings without the extra quotes and with proper newlines? Thanks!

Malabarba: I think it's safe to just call `read` on it. It will return the original string.

mallt: Works perfectly, thanks!

bbatsov:
Btw, can you also post a screenshot showing the change here?

mallt:
Here's a screenshot of clojure.string:
screenshot-cider-browse-ns

bbatsov:
I think there should be some `...` when the first line of a docstring doesn't end with a `.`.  If there's more than one sentence in the first line of the docstring we should just use the first sentence. Might be a good idea to indicate explicitly something is not documented, although I'm on the edge about this.

bbatsov:
Also - use the docstring face to font-lock the docstrings.

mallt:
Thanks for the suggestions, I will include them in a new commit. The "Not documented." string is the default for missing doc strings in other screens, so I think it's better indeed to add this here as well.

expez:
> I think there should be some ... when the first line of a docstring doesn't end with a .. If there's more than one sentence in the first line of the docstring we should just use the first sentence. 

I think a lot of docstrings (not everyone is as particular as you are!) won't contain a period when there's only a single line in the docstring. These lines would get a confusing `...` appended if the above suggestion is implemented.

 There's already some truncating going on here, form the looks of the screenshot, so how about this:

* When the line ends with a period, and is short enough, display that line.
* When the line is short enough, but doesn't end with a period, display that line.
* When the first line goes on forever and ever, truncate and append `...`

Malabarba:
FWIW, I don't think we need an indicator of absent docstrings. It's just going to add to the clutter in a buffer that's already pretty dense.

bbatsov:
Btw, we can also add some config option to control whether those are  displayed or not. And maybe some major mode shortcut.

bbatsov:
> I think a lot of docstrings (not everyone is as particular as you are!) won't contain a period when there's only a single line in the docstring. These lines would get a confusing ... appended if the above suggestion is implemented.

Well, this would encourage some people to start writing better docstrings. :-) On a more serious note - if a line doesn't feature a newline we won't be adding `...` to it anyways. I assumed this obvious, so I didn't list it as a special case. 

As for truncating long lines - for this to work well it should be based on the current window size. It's something nice to do, but I don't think it's important for the first version of this feature.

mallt:
Here's a new screenshot containing examples of every case mentioned in the suggestions. I think this looks pretty clean now, but feel free to comment. :) 
screenshot-cider-browse-ns1

bbatsov:
@mallt Looks great!

Malabarba:
@mallt I've deployed the other change to clojars.

cider-browse-ns.el
```diff
@@ -114,6 +114,24 @@ contents of the buffer are not reset before inserting TITLE and ITEMS."
...
       (goto-char (point-min)))))
 
+(defun cider-browse-ns--items (namespace)
+  "Return the items to show in the namespace browser of the given NAMESPACE.
+Each item consists of a ns-var and the first line of its docstring."
+  (let* ((ns-vars-with-meta (cider-sync-request:ns-vars-with-meta namespace))
+         (propertized-ns-vars (nrepl-dict-map #'cider-browse-ns--properties ns-vars-with-meta)))
+    (mapcar (lambda (ns-var)
+              (let* ((doc (nrepl-dict-get-in ns-vars-with-meta (list ns-var "doc")))
+                     (first-doc-line (if doc
```
bbatsov: Extract this logic into a helper function, so you could write some unit tests for it.

bbatsov:
Apart from my remark, everything looks good. Ping me when you update the code and we'll merge the PR.

mallt:
@bbatsov I've added a helper function and a few unit tests.

PR #1773, opened 2016-06-02
[Fix #1772] Autocomplete fails when nREPL not connected

### LINKED ISSUE #1772: Autocomplete fails when nREPL not connected
## Expected behavior

Can auto complete typing as well even if nREPL not connected.

## Actual behavior

When nREPL not connected, it will show below error message when typing:

```
Error running timer `ac-update-greedy': (wrong-type-argument stringp nil)
Error running timer `ac-show-menu': (wrong-type-argument stringp nil)
```

1. Launch Emacs
2. Switch to a temporary buffer by `C-x b tmp-buffer`
3. `M-x clojure-mode`
4. `M-x cider-mode`
5. Type something, like `hello world`, to trigger `auto-complete`

## Related code

cider-client.el:cider-current-session

```
(defun cider-current-session ()
  "The REPL session to use for this buffer."
  (with-current-buffer (cider-current-connection)
    nrepl-session))
```

`(cider-current-connection)` will return nil if nREPL not connected.

### CIDER version information

```
;; CIDER 0.12.0 (Seattle), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_45
```

### Emacs version

GNU Emacs 24.5.1 (x86_64-apple-darwin14.1.1, NS apple-appkit-1345.38)

### Operating system

OS X 10.11 "El Capitan"

redraiment:

Thanks!

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -859,13 +859,14 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
 (defun cider-sync-request:complete (str context)
   "Return a list of completions for STR using nREPL's \"complete\" op.
 CONTEXT represents a completion context for compliment."
-  (when-let ((dict (thread-first (list "op" "complete"
-                                       "session" (cider-current-session)
-                                       "ns" (cider-current-ns)
-                                       "symbol" str
-                                       "context" context)
-                     (cider-nrepl-send-sync-request nil 'abort-on-input))))
-    (nrepl-dict-get dict "completions")))
+  (if (cider-current-connection)
+    (when-let ((dict (thread-first (list "op" "complete"
+                                         "session" (cider-current-session)
+                                         "ns" (cider-current-ns)
+                                         "symbol" str
+                                         "context" context)
+                       (cider-nrepl-send-sync-request nil 'abort-on-input))))
+      (nrepl-dict-get dict "completions"))))
 
 (defun cider-sync-request:info (symbol &optional class member)
   "Send \"info\" op with parameters SYMBOL or CLASS and MEMBER."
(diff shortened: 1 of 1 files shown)


cider-client.el
```diff
@@ -859,13 +859,14 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
...
   "Return a list of completions for STR using nREPL's \"complete\" op.
 CONTEXT represents a completion context for compliment."
-  (when-let ((dict (thread-first (list "op" "complete"
-                                       "session" (cider-current-session)
-                                       "ns" (cider-current-ns)
-                                       "symbol" str
-                                       "context" context)
-                     (cider-nrepl-send-sync-request nil 'abort-on-input))))
-    (nrepl-dict-get dict "completions")))
+  (if (cider-current-connection)
```
bbatsov: `(when (cider-connected-p)` would be better.

redraiment: Changed. Thanks for suggestion.

Malabarba:
Thanks for the effort. This seems like the wrong way to fix this.
None of our `-request:*` functions check for connection. 

I think the proper fix would be to avoid calling this function in the first place when there's no connection.

bbatsov:
> I think the proper fix would be to avoid calling this function in the first place when there's no connection.
>  Show all checks

I was about to mention exactly this in the morning, but got distracted and forgot about it. You're completely right.

bbatsov:
@redraiment Please, move the check higher up the call stack, as suggested by @Malabarba.

redraiment:
Sure. Thanks! I'm going to do it.

bbatsov:
OK, let me know when you do this.

redraiment:
After research, it would be better to fixed in ac-cider project rather than cider.

PR #1774, opened 2016-06-02
[Fix #1767] Read and eval defun at point

### LINKED ISSUE #1767: Feature request: call function at point and prompt arguments
When writing a clojure function I often find it helpful to repeatedly call the function with the same parameters and get the result as an overlay in the buffer.
I use a simple function for this at the moment that applies the hard coded arguments I want to test to the function at point:

`(defun call-clj-function-at-point ()
  (interactive)
  (save-buffer) 
  (cider-interactive-eval (concat "(apply "
                                  (cider-current-ns) "/"
                                  (second (split-string (cider-defun-at-point))) 
                                  " [4 6])") nil (list (point) (point)) nil))`

Applied to a simple sum function this gives the following result:

`(defn sum
  [a b]
  (+ a b)) => 10`

Instead of hard coding the arguments, I would like to be able to supply them the first time I call this function (f.ex. in the minibuffer). When I call the function again, I would like
the previous arguments to be applied by default, but get the option to supply different arguments to test other cases.

Does this sound like a feasible idea? Or are there already alternatives available to accomplish the same result?

Thanks!

mallt:
Insert the defun at point in the minibuffer and place the point next to the function name to allow passing arguments. Output the result of the function call to the minibuffer or as overlay in the buffer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [#1725](https://github.com/clojure-emacs/cider/issues/1725): Display class names in eldoc for interop forms.
 * [#1572](https://github.com/clojure-emacs/cider/issues/1572): Add support for variables in eldoc.
 * [#1736](https://github.com/clojure-emacs/cider/issues/1736): Show "See Also" links for functions/variables in documentation buffers.
+* [#1767](https://github.com/clojure-emacs/cider/issues/1767): Add a command `cider-read-and-eval-defun-at-point` to insert the defun at point into the minibuffer for evaluation (bound to `C-c C-v .`).
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1223,10 +1223,11 @@ otherwise it's evaluated interactively."
             (cider-nrepl-sync-request:eval (cider-defun-at-point)))
         (cider-eval-defun-at-point)))))
 
-(defun cider-read-and-eval ()
-  "Read a sexp from the minibuffer and output its result to the echo area."
+(defun cider-read-and-eval (&optional value)
+  "Read a sexp from the minibuffer and output its result to the echo area.
+If VALUE is non-nil, it is inserted into the minibuffer as initial input."
   (interactive)
-  (let* ((form (cider-read-from-minibuffer "Clojure Eval: "))
+  (let* ((form (cider-read-from-minibuffer "Clojure Eval: " value))
          (override cider-interactive-eval-override)
          (ns-form (if (cider-ns-form-p form) "" (format "(ns %s)" (cider-current-ns)))))
     (with-current-buffer (get-buffer-create cider-read-eval-buffer)
@@ -1237,6 +1238,15 @@ otherwise it's evaluated interactively."
       (insert form)
       (let ((cider-interactive-eval-override override))
         (cider-interactive-eval form)))))
+
+(defun cider-read-and-eval-defun-at-point ()
+  "Insert the toplevel form at point in the minibuffer and output its result.
+The point is placed next to the function name in the minibuffer to allow
+passing arguments."
+  (interactive)
+  (let* ((fn-name (cadr (split-string (cider-defun-at-point))))
+         (form (concat "(" fn-name ")")))
+    (cider-read-and-eval (cons form (length form)))))
 
 
 ;; Connection and REPL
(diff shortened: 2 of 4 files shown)


cider-mode.el
```diff
@@ -297,6 +297,7 @@ Configure `cider-cljs-lein-repl' to change the ClojureScript REPL to use."]
     (define-key map (kbd "C-c C-p") #'cider-pprint-eval-last-sexp)
     (define-key map (kbd "C-c C-f") #'cider-pprint-eval-defun-at-point)
     (define-key map (kbd "C-c M-:") #'cider-read-and-eval)
+    (define-key map (kbd "C-c . M-:") #'cider-read-and-eval-defun-at-point)
```
Malabarba: We have a keymap for evaluations now. Let's use that.

bbatsov: Yep, I'd go with `C-c v .` or something similar.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * [#1725](https://github.com/clojure-emacs/cider/issues/1725): Display class names in eldoc for interop forms.
 * [#1572](https://github.com/clojure-emacs/cider/issues/1572): Add support for variables in eldoc.
 * [#1736](https://github.com/clojure-emacs/cider/issues/1736): Show "See Also" links for functions/variables in documentation buffers.
+* [#1767](https://github.com/clojure-emacs/cider/issues/1767): Read and eval defun at point.
```
bbatsov: Change the wording a bit here. "Add a command that does this and that. It's bound to ..."

cider-mode.el
```diff
@@ -291,6 +291,7 @@ Configure `cider-cljs-lein-repl' to change the ClojureScript REPL to use."]
     (define-key map (kbd "C-c C-v r") #'cider-eval-region)
     (define-key map (kbd "C-c C-v n") #'cider-eval-ns-form)
     (define-key map (kbd "C-c C-v v") #'cider-eval-sexp-at-point)
+    (define-key map (kbd "C-c C-v .") #'cider-read-and-eval-defun-at-point)
```
bbatsov: You also have to update the manual and mention the new command in command listings there.

cider-interaction.el
```diff
@@ -1238,6 +1238,19 @@ otherwise it's evaluated interactively."
       (let ((cider-interactive-eval-override override))
         (cider-interactive-eval form)))))
 
+(defun cider-read-and-eval-defun-at-point ()
+  "Insert the toplevel form at point in the minibuffer and output its result.
+The point is placed next to the function name in the minibuffer to allow
+passing arguments."
+  (interactive)
+  (let* ((fn-at-point (concat "("
```
bbatsov: I think you've conflated a bit here extracting the function name and building the string, that you're going to insert in the minibuffer. Adding a few more let bindings should make this code cleaner.

cider-interaction.el
```diff
@@ -1238,6 +1238,19 @@ otherwise it's evaluated interactively."
...
+                              (cider-current-ns) "/"
+                              (cadr (split-string (cider-defun-at-point)))
+                              ")"))
+         (form (cider-read-from-minibuffer "Clojure Eval: "
```
bbatsov: Why aren't you using `read-and-eval` directly? I'm guessing this was done to get the results displayed in an overlay, but I'm not sure it's really needed in such situations.

mallt: Since you want to call the defun at point I guess it makes sense to show the result as an overlay? With cider-use-overlays you can still tweak the behavior to fit your needs, so I think this is more flexible than being restricted to the minibuffer output.

bbatsov: But you're not really evaluating the defun at point - this would always result in a var. You're just using its name to build the expression to evaluate. That's why I don't like this - sends out the wrong message IMO.

mallt: Ok I see how this might be confusing. I'll stick to the minibuffer output.

artemyarulin:
Hi @mallt, I'm really happy to have `cider-read-and-eval-defun-at-point` in Cider, but it doesn't work for CLJS (saying that `exit-minibuffer` needs a Clojure REPL).  What do you think - would it be hard to support it? Asking it here first before creating an issue, thanks.

Malabarba:
That might be a genuine bug. Are you able to evaluate things with `C-c M-:`? What about `C-c C-c`?

artemyarulin:
Oh it works now, weird, thank you very much

Issue #1782, opened 2016-06-13
Disable symbol documentation in minibuffer?


vspinu:
I find display of the documentation in the minibuffer a non-feature. It's almost never useful, interferes with eldoc, and it constantly resizes/recenteres the display. How do I disable that? Thanks.

cskksc:
@vspinu Are you referring to the docs which are displayed for `def`'d vars ? Right now that is tied together with `eldoc` for functions, so there's no way to disable it independently. You'll have to disable entire `eldoc` for this. Although it would be easy to add a `defcustom` to control this. @bbatsov @Malabarba ?

bbatsov:
> I find display of the documentation in the minibuffer a non-feature. It's almost never useful, interferes with eldoc, and it constantly resizes/recenteres the display. How do I disable that? Thanks.

This is exactly how Elisp's eldoc works, although it shouldn't resize or recenter the display. Can you share some repro steps for this?

> Although it would be easy to add a defcustom to control this. @bbatsov @Malabarba ?

Obviously we can do this, but I don't want to add a config option for absolutely anything someone might have an issue with. If more people find this behaviour problematic then a defcustom would make more sense. Right now I'd rather focus on what exactly is problematic for @vspinu and try to address it.

vspinu:
> Are you referring to the docs which are displayed for def'd vars ? 

Not only. I am seeing blinking doc for functions as well. For instance when I navigate to `(count ...)` I see a blinking full size docstring and immediately after the eldoc argument string. 

This blinking is not comming from eldoc. It shows as a message in the `*Messages*` buffer. Eldoc inhibits pollution of `*Messages*` buffer so it's either something else or eldoc breaks somewhere.  

> This is exactly how Elisp's eldoc works,

Not quite. Elisp eldoc never shows the full docstring, only the first line. This doesn't resize minibufer or recenter the display.  BTW, eldoc function should respect `eldoc-echo-area-use-multiline-p` which cider currently doesn't.

> Can you share some repro steps for this?

For example
```clojure
(def blabla 
  "Some doc.

  Another line.
  Another line"
  nil)
```

Go to last visible line in the script window and insert `(count blabla)`. When cursor is moved over `count` I see the doc of `count` for a fraction of a second and window is recentered. Same happens when you go over  `blabla`. First I see blabla doc for a fraction of a second, then `count` argumetns. The behavior is somewhat non-systematic, due to caching I guess.

[...]

PR #1785, opened 2016-06-14
Feature/multiline eldoc

cskksc:
For #1782. We now consider the value of `eldoc-echo-area-use-multiline-p` while formatting the docstring for a var. It's semantics are;

| eldoc-echo-area-use-multiline-p | Behaviour |
| ------------- | ------------- |
| `t`  | never attempt to truncate messages  |
| `nil`  | messages are always truncated to fit in a single line  |
| `truncate-sym-name-if-fit` or anything else | symbol name may be truncated if it will enable the function arglist or documentation string to fit on a single line |

This PR does not implement the last case. Any ideas on how to ? We could use `cider-abbreviate-ns`, or `cider-last-ns-segment`. But I'm not sure if truncating the name this way is right. The name may be more important than eldoc.

Thanks!

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -4,22 +4,6 @@ experience.
 
 ## Basic configuration
 
-* Enable `eldoc` in Clojure buffers:
-
-```el
-(add-hook 'cider-mode-hook #'eldoc-mode)
-```
-
-![Eldoc](images/eldoc.png)
-
-CIDER also would show the eldoc for the symbol at point. So in (map inc ...)
-when the cursor is over inc its eldoc would be displayed. You can turn off this
-behaviour by:
-
-```el
-(setq cider-eldoc-display-for-symbol-at-point nil)
-```
-
 * Suppress auto-enabling of `cider-mode` in `clojure-mode` buffers, when starting
   CIDER:
 
@@ -139,6 +123,34 @@ More details can be found [here](https://github.com/clojure-emacs/cider/issues/9
 ```el
 (setq cider-filter-regexps '(".*nrepl"))
 ```
+
+## Configuring eldoc
+
+* Enable `eldoc` in Clojure buffers:
+
+```el
+(add-hook 'cider-mode-hook #'eldoc-mode)
+```
+
+![Eldoc](images/eldoc.png)
+
+* CIDER also would show the eldoc for the symbol at point. So in (map inc ...)
+when the cursor is over inc its eldoc would be displayed. You can turn off this
+behaviour by:
+
+```el
+(setq cider-eldoc-display-for-symbol-at-point nil)
+```
+
+* CIDER respects the value of `eldoc-echo-area-use-multiline-p` when
+displaying documentation in the minibuffer. You can customize this variable to change
+its behaviour.
+
+| eldoc-echo-area-use-multiline-p | Behaviour |
+| ------------- | ------------- |
+| `t`  | Never attempt to truncate messages. Complete symbol name and function arglist or variable documentation will be displayed even if echo area must be resized to fit.|
+| `nil`  | Messages are always truncated to fit in a single line of display in the echo area.  |
+| `truncate-sym-name-if-fit` or anything non-nil | Symbol name may be truncated if it will enable the function arglist or documentation string to fit on a single line. Otherwise, behavior is just like `t` case. |
 
 ## Overlays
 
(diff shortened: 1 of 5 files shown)


Malabarba:
We could hide the namespace from vars of the current namespace, and vars that are already referenced by their full name (i.e., only display the namespace when the var is referenced by an `:alias` or by a `:refer`).

bbatsov:
Can I see some screenshots of this?

bbatsov:
@Malabarba's idea is sound. We can also compress all other namespaces by using one letter per namespace segment (we've got config options to do this in the REPL and for eldoc already).

cskksc:
@bbatsov Sorry I've not updated anything here in a while. I've been travelling for the last 3 days. I'll post some screenshots and the implementation by tomorrow. 

> We can also compress all other namespaces by using one letter per namespace segment (we've got config options to do this in the REPL and for eldoc already).

I guess we could do both these things. Truncate by cider-abbreviate-ns/cider-last-ns-segment first, and if it still doesn't fit, try what @Malabarba said. Otherwise, truncate with `substring`.

cskksc:
1. when the docstring fits in the minibuffer



2. `(setq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)` and truncate using `cider-abbreviate-ns`


3. `(setq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)` and truncate using `cider-last-ns-segment`


4. `(setq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)` and exclude namespace from the var because we're in the same namespace


5.  `(setq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)` and truncating the symbol won't prevent it from resizing. So we show the whole thing


6. `(setq eldoc-echo-area-use-multiline-p nil)`. so we just leave out the entire var and truncate the docstring if required

cskksc:
I still have to write tests for this. This will be ready for code review tomorrow :-)

bbatsov:
> (setq eldoc-echo-area-use-multiline-p nil). so we just leave out the entire var and truncate the docstring if required

I'm confused about this. Shouldn't we just get the default behaviour then? And by this I mean the current behaviour.

cskksc:
@bbatsov When `eldoc-echo-area-use-multiline-p` is nil, we have to fit the eldoc string in the available minibuffer space. That's why we truncate it. To get the current behaviour, we have to set `eldoc-echo-area-use-multiline-p` to `t`.

cskksc:
@bbatsov I'm sorry if I've created a confusion by saying
> Otherwise, return the full thing.

in the previous comment.

That was the case when `eldoc-echo-area-use-multiline-p` is t. This specification has a lot of special cases. We're gonna need a lot of tests :-) Planning to work on those today.

bbatsov:
OK, I was simply confused by the screenshot, as it didn't feature the symbol at all.

cskksc:
I've added the tests. Its ready for review now.
Inspired (copied 😛) from [Emacs](https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/eldoc.el#L371).

cskksc:
Also, the commit `Remove read from cider-browse-ns--first-doc-line` is not required anymore, since we are not using the function here. But I assume we may want to use that fn later.

bbatsov:
I'd also mention something about this in the manual.

cskksc:
I've updated the manual :-)

doc/configuration.md
```diff
@@ -140,6 +124,40 @@ More details can be found [here](https://github.com/clojure-emacs/cider/issues/9
...
+displaying documentation in the minibuffer. You can customize this variable to change
+its behaviour.
+
+| eldoc-echo-area-use-multiline-p | Behaviour |
+| ------------- | ------------- |
+| `t`  | Never attempt to truncate messages. Complete symbol name and function arglist or variable documentation will be displayed even if echo area must be resized to fit.|
+| `nil`  | Messages are always truncated to fit in a single line of display in the echo area.  |
+| `truncate-sym-name-if-fit` or anything non-nil | Symbol name may be truncated if it will enable the function arglist or documentation string to fit on a single line. Otherwise, behavior is just like `t` case. |
+
+CIDER's current behaviour can be achieved by;
```
bbatsov: I don't understand this line.

cskksc: I've removed it completely. Even I thought that its meaning is unclear. Lets leave it out. Other information given here should be sufficient I guess.

PR #1788, opened 2016-06-20
Fix orphaned nrepl-messages buffer after cider-quit

brandm:
Hi all

My observation after cider-quit and answering both questions with yes
is that in the function cider--close-connection-buffer: First the
buffer *nrepl-messages [...]* is killed and then it is recreated
during killing the buffer *cider-repl [...]*.

The attached patch solves the problem for me by killing nrepl-messages
after cider-repl on master and 0.12.0. "make test" succeeds.

Michael


Malabarba:
LGTM

bbatsov:
👍 Thanks!

PR #1791, opened 2016-06-23
[Fix #1646] Add option to control default apropos act on symbol

### LINKED ISSUE #1646: Enable a default selection for cider-apropos-act-on-symbol
`cider-apropos-act-on-symbol` presents a selection on what action to take.

It would speed-up / improve my flow if a default action could be configured (in my case `find-def`).

mallt:
Add an option `cider-apropos-default-act-on-symbol` to control the default action to be applied on the symbol found by an apropos search.

-----------------

### DIFF
--- a/cider-apropos.el
+++ b/cider-apropos.el
@@ -41,6 +41,17 @@
 
 (push cider-apropos-buffer cider-ancillary-buffers)
 
+(defcustom cider-apropos-actions '(("display-doc" . cider-doc-lookup)
+                                   ("find-def" . cider--find-var)
+                                   ("lookup-on-grimoire" . cider-grimoire-lookup))
+  "Controls the actions to be applied on the symbol found by an apropos search.
+The first action key in the list will be selected as default.  If the list
+contains only one action key, the associated action function will be
+applied automatically.  An action function can be any function that receives
+the symbol found by the apropos search as argument."
+  :type '(alist :key-type string :value-type function)
+  :group 'cider)
+
 (defun cider-apropos-doc (button)
   "Display documentation for the symbol represented at BUTTON."
   (cider-doc-lookup (button-get button 'apropos-symbol)))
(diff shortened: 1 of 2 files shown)


bbatsov:
I think a much better approach would be to make the list of actions configurable as well, because right now it's not really obvious for users where those actions are coming from (as they are hardcoded).

mallt:
Ok good point, thanks. When the list of actions itself is configurable I guess setting the default action is not useful anymore because the user can then specify a single action in the action list that will be applied by default. I'll replace the defcustom `cider-apropos-default-act-on-symbol` by `cider-apropos-actions`.

Malabarba:
> When the list of actions itself is configurable I guess setting the default action is not useful anymore because the user can then specify a single action in the action list that will be applied by default.

Can we set the default to be the first action on the list? Then the user can have a list of options **and** still choose their preferred default by just changing the order.

mallt:
I think always having a default might be disturbing for users that frequently switch the action. But maybe they can put nil as first element in the list of actions so they can always choose the action manually?

Malabarba:
Ah, I see what you mean. But that's only because you're using the wrong argument (`INITIAL-INPUT`) instead of `DEF`.
Just replace that last line with `nil nil nil nil cider-apropos-default-act-on-symbol)` and you'll see what I mean. The default doesn't pre-populate the input, but it still gets used if you immediately hit RET. This where there's nothing to lose by having a default.

mallt:
I see, that's perfect, thanks for the tip!

bbatsov:
Guess this is settled then. :-)

mallt:
What is the preferred defcustom type for this action list? Using a set gives an easy config in the customization menu, but changing the order (to set the default action) can only be done by editing the lisp expression by hand. Is this ok?
cider-apropos-actions

bbatsov:
Sounds ok to me.

Malabarba:
You can also make the custom type be something like `(repeat (choice (const...) (const...) (const...)))`. You won't get the cute checkboxes, but the effect will be very similar and you'll be able to change the order. 

Also, do we need that `quit`? Isn't that the same as hitting C-g?

bbatsov:
> Also, do we need that quit? Isn't that the same as hitting C-g?

It is. I added it for users who are new to Emacs, obviously it's not really needed.

cider-apropos.el
```diff
@@ -139,16 +148,14 @@ optionally search doc strings (based on DOCS-P), include private vars
...
-                                 '("display-doc"
-                                   "find-def"
-                                   "lookup-on-grimoire"
-                                   "quit"))))
+  (let ((action (if (= 1 (length cider-apropos-actions))
+                    (car cider-apropos-actions)
+                  (completing-read (format "Choose action to apply to `%s`: " symbol)
+                                   cider-apropos-actions nil nil nil nil (car cider-apropos-actions)))))
     (pcase action
       ("display-doc" (cider-doc-lookup symbol))
```
bbatsov: Btw, I meant that I'd move this mapping itself to some configuration, so everyone can easily add more actions for the apropos commands.

mallt: Ok I see, I guess it would be easiest then if the defcustom is just a list of functions?

bbatsov: Rather some alist of names to functions mappings, as you'll still want to display some meaningful action names in the minibuffer.

cider-apropos.el
```diff
@@ -139,17 +150,15 @@ optionally search doc strings (based on DOCS-P), include private vars
...
-    (pcase action
-      ("display-doc" (cider-doc-lookup symbol))
-      ("find-def" (cider--find-var symbol))
-      ("lookup-on-grimoire" (cider-grimoire-lookup symbol))
-      ("quit" nil)
-      (_ (user-error "Unknown action `%s`" action)))))
+  (let* ((first-action-key (car (car cider-apropos-actions)))
+         (action-key (if (= 1 (length cider-apropos-actions))
+                     first-action-key
+                   (completing-read (format "Choose action to apply to `%s`: " symbol)
```
Malabarba: This format should probably be:
`(completing-read (format "Choose action to apply to `%s` (default %s): " symbol first-action-key)`
To match the usual emacs notation for completion with default.

PR #1799, opened 2016-07-09
cider-repl-closing-return completes { and [

pyridine:
This pull request edits `cider-repl-closing-return` (`C-<return>`) so that it inserts `]` and `}` where necessary, instead of completing every braced expression with a `)`.

Observe: say you've typed this out at the REPL:
```clojure
user> (def foo {:bar ['baz 'qux
```

If you hit `C-<return>`, you'd expect it to complete your expression and eval it:

````clojure
user> (def foo {:bar [baz qux)))
RuntimeException Unmatched delimiter: )  clojure.lang.Util.runtimeException (Util.java:221)
````

With this edit, `cider-repl-closing-return` works as expected, inserting `}` and `]` to match their respective braces. Because emacs is treating all three of these brace types as expression delimiters, the auto-complete should work with each of them.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
 * Rebind `cider-eval-region` to `C-c C-v r`.
 * Rebind `cider-eval-ns-form` to `C-c C-v n`.
 * [#1577](https://github.com/clojure-emacs/cider/issues/1577): Show first line of docstring in ns browser.
-* cider-repl-closing-return (`C-<Return>`) now also completes brackets (`[]`) and curly braces (`{}`) in an expression.
+* `cider-repl-closing-return` (`C-<Return>`) now also completes brackets (`[]`) and curly braces (`{}`) in an expression.
 
 ### Bugs fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 * Rebind `cider-eval-region` to `C-c C-v r`.
 * Rebind `cider-eval-ns-form` to `C-c C-v n`.
 * [#1577](https://github.com/clojure-emacs/cider/issues/1577): Show first line of docstring in ns browser.
+* cider-repl-closing-return (`C-<Return>`) now also completes brackets (`[]`) and curly braces (`{}`) in an expression.
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cider-repl.el [hunk 1]
```diff
@@ -765,8 +765,11 @@ text property `cider-old-input'."
   (goto-char (point-max))
   (save-restriction
     (narrow-to-region cider-repl-input-start-mark (point))
-    (while (ignore-errors (save-excursion (backward-up-list 1)) t)
-      (insert ")")))
+    (let ((matched-delimiter nil)))
+      (while (ignore-errors (save-excursion
+	   		       (backward-up-list 1)
+			       (setq matched-delimiter (string (following-char)))) t)
+      (insert (cdr (assoc matched-delimiter '(("(" . ")") ("[" . "]") ("{" . "}")))))))
```
Malabarba: I think the same can be achieved by just calling `syntax-ppss`. It returns a list where one of the items is the delimiter that would close the current sexp.

cider-repl.el
```diff
@@ -765,8 +765,11 @@ text property `cider-old-input'."
   (goto-char (point-max))
   (save-restriction
     (narrow-to-region cider-repl-input-start-mark (point))
-    (while (ignore-errors (save-excursion (backward-up-list 1)) t)
-      (insert ")")))
+    (let ((matching-delimiter nil))
```
bbatsov: You should also update the docstring of this command.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 * Rebind `cider-eval-region` to `C-c C-v r`.
 * Rebind `cider-eval-ns-form` to `C-c C-v n`.
 * [#1577](https://github.com/clojure-emacs/cider/issues/1577): Show first line of docstring in ns browser.
+* cider-repl-closing-return (`C-<Return>`) now also completes brackets (`[]`) and curly braces (`{}`) in an expression.
```
bbatsov: Wrap the name of the command in backticks as well.

cider-repl.el
```diff
@@ -765,8 +765,11 @@ text property `cider-old-input'."
...
+      (while (ignore-errors (save-excursion
+			      (backward-up-list 1)
```
bbatsov: Maybe you should extract the bulk of the logic to a helper function, so it can reused and unit tested.

bbatsov:
I'll merge this in its current state. Feel free to add some tests down the road if you feel like it.

PR #1801, opened 2016-07-11
Make mouse-over tooltips less intrusive

### LINKED ISSUE #1782: Disable symbol documentation in minibuffer?

cskksc:
We disable the tooltips when `help-at-pt-display-when-idle` is non-nil.

The Emacs customization interface also allows us to set this to `In certain situations`; which sets it to a list `(keymap local-map button kbd-help help-echo)` by default.

> If the value is a list, the help only gets printed if there is a text or overlay property at point that is included in this list.

I have not added any special consideration for this, as both `help-echo` and `kbd-help` are in this list by default. I don't think any special handling is required, and we'll just skip the messages unless the value is nil. Is this ok ? 


bbatsov:
I think that's fine for now.

Issue #1804, opened 2016-07-15
Feature request: remember location when going up level in cider inspector


mallt:
When navigating large data structures, it would be useful if the inspector would remember the location in the structure when moving back up with `cider-inspector-pop`.

bbatsov:
Guess so. This should be relatively easy to implement.

mallt:
Thanks a lot guys!

PR #1807, opened 2016-07-19
[#1804] Remember location after cider-inspector-pop

cskksc:
We push our current location in the buffer i.e`(point)`, onto a stack, before every `cider-inspector-push` operation. Then when we're navigation back using `cider-inspector-pop`,
we just `goto-char` to that location.

I've not used this feature of CIDER much. But does it make sense to do something like this for `next-page` and `prev-page` too ?

out

/cc @mallt 

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -114,11 +114,41 @@ With a second prefix argument it prompts for an expression to eval and inspect."
     (4 (cider-inspect-defun-at-point))
     (16 (call-interactively #'cider-inspect-expr))))
 
+(defvar cider-inspector-location-stack nil
+  "A stack used to save point locations in inspector buffers.
+These locations are used to emulate save-excursion between
+`cider-inspector-push' and `cider-inspector-pop' operations.")
+
+(defvar cider-inspector-page-location-stack nil
+  "A stack used to save point locations in inspector buffers.
+These locations are used to emulate save-excursion between
+`cider-inspector-next-page' and `cider-inspector-prev-page' operations.")
+
+(defvar cider-inspector-last-command nil
+  "Contains the value of the most recently used `cider-inspector-*' command.
+This is used as an alternative to the built-in `last-command'. Whenever we
+invoke any command through M-x and its variants, the value of `last-command'
+is not set to the command it invokes.")
+
 ;; Operations
 (defun cider-inspector--value-handler (_buffer value)
   (cider-make-popup-buffer cider-inspector-buffer 'cider-inspector-mode)
   (cider-inspector-render cider-inspector-buffer value)
-  (cider-popup-buffer-display cider-inspector-buffer t))
+  (cider-popup-buffer-display cider-inspector-buffer t)
+  (with-current-buffer cider-inspector-buffer
+    (when (eq cider-inspector-last-command 'cider-inspector-pop)
+      (setq cider-inspector-last-command nil)
+      ;; Prevents error message being displayed when we try to pop
+      ;; from the top-level of a data struture
+      (when cider-inspector-location-stack
+        (goto-char (pop cider-inspector-location-stack))))
+
+    (when (eq cider-inspector-last-command 'cider-inspector-prev-page)
+      (setq cider-inspector-last-command nil)
+      ;; Prevents error message being displayed when we try to
+      ;; go to a prev-page from the first page
+      (when cider-inspector-page-location-stack
+        (goto-char (pop cider-inspector-page-location-stack))))))
 
 (defun cider-inspector--out-handler (_buffer value)
   (cider-emit-interactive-eval-output value))
@@ -157,12 +187,14 @@ current buffer's namespace."
 
 (defun cider-inspector-pop ()
   (interactive)
+  (setq cider-inspector-last-command 'cider-inspector-pop)
   (cider-nrepl-send-request
    (list "op" "inspect-pop"
          "session" (cider-current-session))
    (cider-inspector-response-handler (current-buffer))))
 
 (defun cider-inspector-push (idx)
+  (push (point) cider-inspector-location-stack)
   (cider-nrepl-send-request
    (list "op" "inspect-push"
          "idx" idx
@@ -181,6 +213,7 @@ current buffer's namespace."
 
 Does nothing if already on the last page."
   (interactive)
+  (push (point) cider-inspector-page-location-stack)
   (cider-nrepl-send-request
    (list "op" "inspect-next-page"
          "session" (cider-current-session))
@@ -191,6 +224,7 @@ Does nothing if already on the last page."
 
 Does nothing if already on the first page."
   (interactive)
+  (setq cider-inspector-last-command 'cider-inspector-prev-page)
   (cider-nrepl-send-request
    (list "op" "inspect-prev-page"
          "session" (cider-current-session))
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -157,12 +166,16 @@ current buffer's namespace."
 
 (defun cider-inspector-pop ()
   (interactive)
+  ;; We cannot pop anything if we're at the top level
+  (when cider-inspector-location-stack
+    (push 'pop cider-inspector-location-stack))
```
Malabarba: I don't understand why this is necessary. Can't we just `(goto-char (pop cider-inspector-location-stack))` in the `cider-inspector--value-handler` function?

cskksc: We only have to `goto-char` if this handler is called by the `pop` operation. I just pushed a patch where we check this through the `last-command`. This was certainly not the best way to do this :-)

bbatsov:
> I've not used this feature of CIDER much. But does it make sense to do something like this for next-page and prev-page too ?

I'm not quite certain what you mean by this.

Malabarba:
I think it might make sense to have it, but it would probably complicate the feature a lot (as the two trackings would interfere with each other), so maybe it's best to leave it like this.

cskksc:
@bbatsov This PR, in its current state would only preserve cursor location across `cider-inspector-push/pop` operations in the inspector buffer. These are the ones used to navigate down/up in a data structure, so I think these would be most frequently used. But we also have `cider-inspector-next-page/prev-page` to handle paginated responses. So I was wondering if preserving cursor location across page navigation would be useful too :-)

@Malabarba You're right. It would complicate tracking method calls in the handler. 

But lets keep this PR open for another day. I'll try this today, and if I'm able to get it to work and if we're happy with the added complexity vs usefulness of the feature, then lets merge it. Otherwise lets stick to only preserving cursor location across `cider-inspector-push/pop` :-)

bbatsov:
So, what's the verdict here - are we extending this or merging it as it is?

cider-inspector.el
```diff
@@ -114,11 +114,18 @@ With a second prefix argument it prompts for an expression to eval and inspect."
     (4 (cider-inspect-defun-at-point))
     (16 (call-interactively #'cider-inspect-expr))))
 
+(defvar cider-inspector-location-stack nil
```
bbatsov: I'd say this should be a buffer-local variable.

cider-inspector.el
```diff
@@ -114,11 +114,18 @@ With a second prefix argument it prompts for an expression to eval and inspect."
...
+These locations are used to emulate save-excursion after a pop operation.")
+
 ;; Operations
 (defun cider-inspector--value-handler (_buffer value)
   (cider-make-popup-buffer cider-inspector-buffer 'cider-inspector-mode)
   (cider-inspector-render cider-inspector-buffer value)
-  (cider-popup-buffer-display cider-inspector-buffer t))
+  (cider-popup-buffer-display cider-inspector-buffer t)
+  (when (eq last-command 'cider-inspector-pop)
+    (with-current-buffer cider-inspector-buffer
```
bbatsov: Isn't the inspector buffer already the current buffer?

cskksc:
I'm sorry I forgot to post an update :-) I have gotten it to work. We can just use 2 stacks to track the position. But I ran into something else. When I invoke any `cider-*` command through ido, tracking the method calls via `last-command` doesn't work. It sets the last command to something like `ido-last-command`. Need to look into this further.

I am AFK for the last two days. I will definitely  finish this tomorrow :-) 


Sent from my iPhone

Malabarba:
>  It sets the last command to something like `ido-last-command`. Need to look into this further.

Yes, that's just how last-command works. The same happens if you invoke the command with regular M-x

bbatsov:
Probably we shouldn't rely on `last-command` at all, but rather bake this into `cider-inspector-pop`. Using `last-command` is generally a recipe for trouble...

cskksc:
> Yes, that's just how last-command works. The same happens if you invoke the command with regular M-x

It is surprising I guess. I expected it to get the value of the command selected through M-x.

> Probably we shouldn't rely on last-command at all, but rather bake this into cider-inspector-pop. Using last-command is generally a recipe for trouble...

Yep. It seems so. We're gonna have to track this via our own var I guess. Or through the stack like in the previous patch. I'll try and see we could get it done any other way.


On a side note, I have a doubt about this. Why have we used callback's here ? All the actions seem synchronous with the middleware response. What I guessed is that its because we have a good abstraction of the form `nrepl-make-response-handler`. Should we use the form `cider-sync-request:*` here ? Not necessarily in this feature.

bbatsov:
> On a side note, I have a doubt about this. Why have we used callback's here ? All the actions seem synchronous with the middleware response. What I guessed is that its because we have a good abstraction of the form nrepl-make-response-handler. Should we use the form cider-sync-request:* here ? Not necessarily in this feature.

We don't really need it. In the early days of the project we were trying to use callbacks everywhere as nREPL doesn't natively support sync evaluations and our emulation of those was poor. This code hasn't be updated in quite a while. Async handlers have a ton of room for improvement, this has been discussed in some older open tickets - unfortunately no one has had to the time to tackle them in years...

bbatsov:
The discussions about callbacks is here https://github.com/clojure-emacs/cider/issues/1099

cskksc:
@bbatsov Ah ok. It makes sense now. I'll read those tickets and we can start working on it next. I have some free time now :-)

For now, we can track this through our own global `cider-inspector-last-command`. It is not the cleanest solution, but it works. I just have to make all those vars buffer local. But please check if this approach is ok.

bbatsov:
> I just have to make all those vars buffer local. But please check if this approach is ok.

The solution looks good. Just use `defvar-local` and we should be good to go.

cider-inspector.el
```diff
@@ -136,6 +166,7 @@ displayed.
 
 Used for all inspector nREPL ops."
   (nrepl-make-response-handler buffer
+
```
bbatsov: I think you've added this blank line by accident.

bbatsov:
> Thanks! I'll read those tickets and we can start working on it next.

That'd be great. This refactoring is way way overdue... Many of the older tickets are some essential changes that got stale mostly due to lack of time to work on them (comint is another such example).

cider-inspector.el
```diff
@@ -114,11 +114,41 @@ With a second prefix argument it prompts for an expression to eval and inspect."
...
+(defvar cider-inspector-last-command nil
+  "Contains the value of the most recently used `cider-inspector-*' command.
+This is used as an alternative to the built-in `last-command'. Whenever we
+invoke any command through M-x and its variants, the value of `last-command'
+is not set to the command it invokes.")
+
 ;; Operations
 (defun cider-inspector--value-handler (_buffer value)
   (cider-make-popup-buffer cider-inspector-buffer 'cider-inspector-mode)
   (cider-inspector-render cider-inspector-buffer value)
```
cskksc: This is preventing us from using a buffer-local var. This method clears out all the stored vars. Let's keep them as normal vars for now ? I guess our popup method's could use an optional param, which controls this behavior. But it would changes in a lot of places. I'll create a separate PR for this. Is that ok ?

cider-inspector.el
```diff
@@ -114,11 +114,41 @@ With a second prefix argument it prompts for an expression to eval and inspect."
...
+
+(defvar cider-inspector-last-command nil
+  "Contains the value of the most recently used `cider-inspector-*' command.
+This is used as an alternative to the built-in `last-command'. Whenever we
+invoke any command through M-x and its variants, the value of `last-command'
+is not set to the command it invokes.")
+
 ;; Operations
 (defun cider-inspector--value-handler (_buffer value)
   (cider-make-popup-buffer cider-inspector-buffer 'cider-inspector-mode)
```
cskksc: This is preventing us from creating a buffer-local var. It clears out all the state when creating a new buffer. I guess our this method could use an optional param which controls this behavior. But it would involve changes in a lot of places. Let's handle this in a separate PR ?

cskksc:
Ill start reading the code and let's break the refactoring tasks down into smaller tickets we could handle. I'll start working on this from tomorrow :hammer: :-)

cider-inspector.el
bbatsov: Ah, I forgot about this. Yeah, guess we can revisit this later. I'll merge the PR in its current state.

Issue #1794, opened 2016-06-28
printing chars breaks cider repl printing


ysmiraak:
## Expected behavior

`(apply pr (map char (range 128)))`
Should print all the ASCII chars (and it does for lein repl in the terminal).

## Actual behavior

In cider repl, it prints ASCII from 64 `@` to 90 `Z`, and then the printing functionality breaks down: The repl still returns results but won't print anything anymore.

## Steps to reproduce the problem

Do `(apply pr (map char (range 128)))`,
And then print something else.
The repl won't print but only returns nil.

### CIDER version information

```
;; CIDER 0.13.0snapshot (package: 20160623.12), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_72
```

### Lein/Boot version

Lein 2.6.1

### Emacs version

"GNU Emacs 24.5.1 (x86_64-apple-darwin14.5.0, NS apple-appkit-1348.17)"

### Operating system

OS X 10.11 "El Capitan"

dpsutton:
I can reliably reproduce this issue with `(pr (char 27))`. I'm gonna investigate if this is getting caught in the decoder of the nrepl messages.

dpsutton:
ah, and its not printing @ .. z as stated above, but its printing 0-27 (and only half of the char 27 key) which it displays as 

    ;; ======================================================================
    user> (apply pr (map char (range 128)))
    \ \ \ \ \ \ \ \ \backspace \tab \newline \ \formfeed \return \ \ \ \ \ \ \ \ \ \ \ \ \ \
    nil
    user> 


![image](https://cloud.githubusercontent.com/assets/6377293/16892649/cd87c83e-4adf-11e6-98bf-a5246ef395e7.png)


The 27 key (not sure of the terminology here) is escape, represented as `\^[`

dpsutton:
Neat. This is actually an emacs bug and dies in list/ansi-color.el in the function ansi-color-apply

dpsutton:
(defvar-local ansi-color-context nil
      "Context saved between two calls to `ansi-color-apply'.
    This is a list of the form (CODES FRAGMENT) or nil.  CODES
    represents the state the last call to `ansi-color-apply' ended
    with, currently a list of ansi codes, and FRAGMENT is a string
    starting with an escape sequence, possibly the start of a new
    escape sequence.")

dpsutton:
its this bit of code in ansi-apply-color:

     (let (fragment)
[...]

PR #1813, opened 2016-07-24
Flush ansi color context after printing

dpsutton:

References https://github.com/clojure-emacs/cider/issues/1794

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -526,6 +526,14 @@ Return the position of the prompt beginning."
         (set-marker cider-repl-prompt-start-mark prompt-start)
         prompt-start))))
 
+(defun cider-repl--flush-ansi-color-context ()
+  "Flush ansi color context after printing.  
+When there is a possible unfinished ansi control sequence,
+ `ansi-color-context` maintains this list."
+  (when (and ansi-color-context (stringp (cadr ansi-color-context)))
+    (insert-before-markers (cadr ansi-color-context))
+    (setq ansi-color-context nil)))
+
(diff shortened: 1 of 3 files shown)


Malabarba:
Answering the question on the issue: I like the way it's implemented here (a function rather than inline code), and I think it shouldn't cause much problem (at least, not more problem than the current code).

However, it looks like you broke: ./test/cider-util-tests.el:141:

cider-repl.el
```diff
@@ -526,6 +526,14 @@ Return the position of the prompt beginning."
         (set-marker cider-repl-prompt-start-mark prompt-start)
         prompt-start))))
 
+(defun cider-repl--flush-ansi-color-context ()
+  "Flush ansi color context after printing.  When there is a possible
```
bbatsov: Add a line break after the first sentence. Standard Emacs documentation convention.

test/cider-repl-tests.el
```diff
@@ -68,3 +68,20 @@
 ;;  Javadoc: (javadoc java-object-or-class)
 ;;     Exit: <C-c C-q>
 ;;  Results: Stored in vars *1, *2, *3, an exception in *e;"))))
+
+(defun create-cider-string (s)
```
bbatsov: This could use some docstring as it's not immediately obvious what it does.

bbatsov: Maybe this should be named `cider-create-repl-output-string`?

dpsutton:
Ok, fixed the logic but need to move my test. I've got to head out for a bit though and I'll do it when i get back.

thanks for the heads up.

Issue #1677, opened 2016-04-14
REPL doesn't interpret carriage return


thomas-morgan:
I entered `(println "foo\rbar")` in the Cider REPL, hoping to see

```
bar
nil
```

Instead I saw

```
foobar
nil
```

This breaks progress bars among other things, as reported here: <http://stackoverflow.com/questions/34649151/progress-indicator-bar-auto-changed-line-in-cider-nrepl-buffer/34651898#34651898>.

For comparison, `(display "foo\rbar\n")` displays `bar` in the REPL started by `run-guile`.

### CIDER version information

```
;; CIDER 0.11.0 (Bulgaria), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_60
```

### Lein/Boot version

Leiningen 2.6.1 on Java 1.8.0_60 OpenJDK 64-Bit Server VM

### Emacs version

GNU Emacs 24.5.2 (x86_64-unknown-linux-gnu, GTK+ Version 2.24.28) of 2015-09-25 on machine

### Operating system

NixOS 15.09

bbatsov:
Seems you're right - this is a bug. I'm not sure what's causing it, though. The message we send to the nREPL server seems correct to me:

```
(--> 
  ns  "cider-demo.core"
  op  "eval"
  session  "4bab8d15-cbc4-4ec7-a6ad-087c51c37845"
  code  "(println \"foo\\rbar\")\n"
  file  "*cider-repl cider-demo*"
  line  58
  column  17
  id  "55"
)
```

bbatsov:
@dpsutton Maybe you'd like to investigate another REPL bug? :-)

dpsutton:
This appears to be a bug in the `nrepl--bdecode-1` function, 
```

;; normalise any platform-specific newlines
        (let* ((original (buffer-substring-no-properties beg end))
               (result (replace-regexp-in-string "\r" "" original)))
          (cons nil (nrepl--push result stack))))))
```

in nrepl-client.el.  I'm gonna go learn about platform specific newlines as I've never really understood the difference between \r and \r\n, etc. If i remove that line, i'm getting `^M` in the buffer, which is the perennial problem with git on multiple platforms. So I just need to learn the actual issue at hand instead of a hand-wavy idea about newlines.

dpsutton:
I guess I don't know if a `\r` by itself is valid. I've always seen the windows style as `\r\n`, which is why the code was only removing the `\r` by itself. 
```
[...]

PR #1814, opened 2016-07-26
Handle lone carriage returns as newline

dpsutton:

Substitute newline for single carriage returns

https://github.com/clojure-emacs/cider/issues/1677

this doesn't totally solve the issue actually. The issue poster was hoping that "foo\rbar" would right foo, return to the beginning of the same line, and then write bar. However, we were stripping carriage returns with no action. We now act as lein repl does, replacing this lone carriage returns with newlines.

The fundamental issue was that text-based status bars didn't work, and it sounds like we don't have plans to fix this, or at least not at the moment. There are lots of things to consider like markers, how we parse up the bencoded string to know when to move to next line, when to zap over current line, etc, and that would be a much larger change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## master (unreleased)
 
+### New Features
+### Changes
+### Bugs Fixed
+* [#1677](https://github.com/clojure-emacs/cider/issues/1677): Interpret `\r` as a newline
+
+
 ## 0.13.0 (2016-07-25)
 
 ### New Features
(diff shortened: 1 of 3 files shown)


bbatsov:
The changes look good. Mention them in the changelog and we're good to go.

> The fundamental issue was that text-based status bars didn't work, and it sounds like we don't have plans to fix this, or at least not at the moment.

Well, I wouldn't mind fixing this, but as you said - it's probably going to require a bit of work.

dpsutton:
Added to changelog along with categories for bugs/features/fixes on master (unreleased)

PR #1821, opened 2016-08-10
Use sync requests in inspector-*

cskksc:
As discussed in https://github.com/clojure-emacs/cider/pull/1807#issuecomment-234760146, the inspector-* fn's use callbacks to interact with the middleware. This PR changes those fn's to use the `cider-sync-request:*`.

Thanks!

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -131,7 +131,114 @@ invoke any command through M-x and its variants, the value of `last-command'
 is not set to the command it invokes.")
 
 ;; Operations
-(defun cider-inspector--value-handler (_buffer value)
+;;;###autoload
+(defun cider-inspect-expr (expr ns)
+  "Evaluate EXPR in NS and inspect its value.
+Interactively, EXPR is read from the minibuffer, and NS the
+current buffer's namespace."
+  (interactive (list (cider-read-from-minibuffer "Inspect expression: " (cider-sexp-at-point))
+                     (cider-current-ns)))
+  (when-let (value (cider-sync-request:inspect-expr expr ns (or cider-inspector-page-size 32)))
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -131,117 +132,78 @@ invoke any command through M-x and its variants, the value of `last-command'
...
-   (list "op" "inspect-set-page-size"
-         "session" (cider-current-session)
-         "page-size" page-size)
-   (cider-inspector-response-handler (current-buffer))))
+  (interactive "nPage size: ")
+  (when-let (value (cider-sync-request:inspect-set-page-size page-size))
+    (cider-inspector--value-handler value)))
 
 ;; Render Inspector from Structured Values
+(defun cider-inspector--value-handler (value)
```
cskksc: Probably we should change this name. But I couldn't think of anything better.

bbatsov: How about `cider-inspector--render-value`?

cider-client.el
```diff
@@ -988,6 +988,59 @@ CONTEXT represents a completion context for compliment."
       (error (car (split-string err "\n"))))
     (nrepl-dict-get response "formatted-edn")))
 
+(defun cider-sync-request:inspect-pop ()
```
bbatsov: As all those commands are unlikely to used outside the inspector perhaps they should be in its own source file.

cskksc: Done.

cider-inspector.el
cskksc: Yep. I hadn't used that one because we already have a couple of `render` methods here :-) But that does seem to be the best one. Changed it now.

PR #1826, opened 2016-08-20
[Fix #1825] Show generated test input

### LINKED ISSUE #1825: Show test.check's generated input for failing tests
CIDER nicely displays the expected and actual outcome for tests in the clojure.test framework. However, when the tests are based on test.check the output lacks information about the generated inputs.

For example, consider the following property:
```clojure
(defspec whatever
  (prop/for-all [x gen/int]
    (= 0 x)))
```

Checking it in CIDER yields this output:
```
Fail in whatever
expected: result
  actual: false
```

It would be nice to also learn which `x` caused the test to fail. To this end, I would suggest to capture that very piece of information from the test.check results and display it in the result buffer whenever applicable.

Technically, the information about the failing input has to be propagated to `cider-nrepl.middleware.test/report`, where it could be stored and passed to the Emacs plugin for display. Unfortunately, `clojure.test.check.clojure-test.defspec` doesn't do so yet, [throwing away](https://github.com/clojure/test.check/blob/test.check-0.9.0/src/main/clojure/clojure/test/check/clojure_test.cljc#L16) everything except the success status, but the extension should be relatively simple. The utility library test.chuck already demonstrates a way to go: it [calls](https://github.com/gfredericks/test.chuck/blob/17ce0f1d86dc5f3bcef2a8f75df58a625dd3a101/src/com/gfredericks/test/chuck/clojure_test.cljc#L28) the report function, in our case `cider-nrepl.middleware.test/report`, with an event of `:type ::shrunk` when it has worked out the smallest failing inputs. As a first step, one might extend `cider-nrepl.middleware.test/report` to handle events of that type and have CIDER display the information in the result buffer. Then the same concept could be applied to test.check's `defspec`. I'll play around with it...

sebastianpoeplau:
Display the test input generated by `test.check`. The respective change to `cider-nrepl` is [pending](https://github.com/clojure-emacs/cider-nrepl/pull/371).

-----------------

Thanks!


bbatsov:
LGTM. Thanks!

PR #1833, opened 2016-08-26
Display fn spec in the doc buffers

cskksc:
We get the formatted spec strings through the middleware `:info` op, and we just display those in the doc buffers. Depends on clojure-emacs/cider-nrepl#374

Thanks!


cider-doc.el
```diff
@@ -455,6 +456,10 @@ Tables are marked to be ignored by line wrap."
                                         (browse-url (button-get x 'url))))
           (insert ".\n"))
         (insert "\n")
+        (when spec
+          (emit "Spec: " 'font-lock-function-name-face)
+          (mapc (lambda (s) (insert s "\n")) spec))
+        (insert "\n")
```
Malabarba: I think this line should be inside the `(when ...)

Malabarba:
:+1: besides the comment

bbatsov:
Looks good to me as well. 

On an unrelated note - probably it would make sense to add some command to display just the spec in some fancier manner (e.g. inline). Enriching eldoc with spec info would also be pretty cool.

cider-doc.el
cskksc: Done :-)

cskksc:
> Enriching eldoc with spec info would also be pretty cool.

Yep, it would be. But I think for most users, `eldoc-echo-area-use-multiline-p` would be set to `truncate-sym-name-if-fit`, the default value. So the multiline strings won't be displayed. Should we still add it in ? It would be displayed when `eldoc-echo-area-use-multiline-p` is t. 

> probably it would make sense to add some command to display just the spec in some fancier manner (e.g. inline)

I'm not sure what `inline` is. Do we have some other feature that uses this ? I could figure it out from there.

bbatsov:
> Yep, it would be. But I think for most users, eldoc-echo-area-use-multiline-p would be set to truncate-sym-name-if-fit, the default value. So the multiline strings won't be displayed. Should we still add it in ? It would be displayed when eldoc-echo-area-use-multiline-p is t.

Well, that's why we have a manual. Curious users will figure this out. :-)

> I'm not sure what inline is. Do we have some other feature that uses this ? I could figure it out from there :-)

I call the use of overlays "inline". Not a big deal, just something that crossed my mind.

cskksc:
@bbatsov Oh, ok. Got it. We can do something like this. I'll post a separate PR for that. The eldoc integration is pretty straightforward. Lets include that here.

PR #1844, opened 2016-09-05
[Fix #1824] Cider-jack-in dependency exclusions

### LINKED ISSUE #1824: Can no longer jack-in to an inherited clojure version
## Expected behavior

Under 0.12.0 (and earlier), I could specify the clojure version in one library and then have other libraries/apps that depend on that one so I don't have to keep all the common library versions updated in various projects that use them.

This still seems to work fine under `lein repl`.

## Actual behavior

cider-jack-in results in a stack trace that basically starts with an error about nrepl-server-sentinel and "cider.nrepl.middleware.version: Too many arguments to def."

There are a bunch of "Unable to resolve symbol: some-> in this context" java.lang.Exceptions.

And then it ends with a "java.lang.Exception: Unable to resolve symbol: some? in this context."
The full stack trace is available at https://github.com/jimrthy/clojure-introduction/blob/cider/transitive-clojure-version/cider-transitive-version/README.md

- git clone https://github.com/jimrthy/clojure-introduction
- git checkout cider/transitive-clojure-version
- cd cider-transitive-version
- lein install
- cd ../usage
- lein deps :tree
- Verify that it inherits org.clojure.clojure 1.9.0-alpha10 from cider-transitive-version
- Verify that lein repl starts correctly (I seem to be able to cider-connect to this, so I think I have a work-around)
- Try to cider-jack-in to the usage project, see the stack trace
- Uncomment the clojure dependency in usage/project.clj
- Verify that cider-jack-in works correctly

### CIDER version information

CIDER 0.13.0 (California), nREPL 0.2.12
Clojure 1.9.0-alpha10, Java 1.7.0_79

### Lein version

2.6.1

### Emacs version

GNU Emacs 24.3.1 (x86_64-px-linux-gnu, GTK+ Version 3.10.7) of 2014-03-07 on lamiak, modified by Debian

### Operating system

Ubuntu 14.04.3 LTS

benedekfazekas:

Add option to define exclusions for injected dependecies. Add
`org.cloure/clojure` exclusion to `org.clojure/tools.nrepl` to mitigate
problem with tools.nrepl clojure dependency.

Note about boot: the latest stable release does not support defining
depedency exclusions on the command line. However, this feature is
available in the upcoming 2.7.x release. This feature will be added
for boot when 2.7.x is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 * [#1677](https://github.com/clojure-emacs/cider/issues/1677): Interpret `\r` as a newline.
 * [#1819](https://github.com/clojure-emacs/cider/issues/1819): Handle properly missing commands on `cider-jack-in`.
+* Add option to define exclusions for injected dependecies. Fixes [#1824](https://github.com/clojure-emacs/cider/issues/1824): Can no longer jack-in to an inherited clojure version.
 
 ## 0.13.0 (2016-07-25)
 
--- a/cider.el
+++ b/cider.el
@@ -245,6 +245,14 @@ found for the PROJECT-TYPE"
 (cider-add-to-alist 'cider-jack-in-dependencies
                     "org.clojure/tools.nrepl" "0.2.12")
 
+(defvar cider-jack-in-dependencies-exclusions nil
+  "List of exclusions for jack in dependencies.
+
+Elements of the list are artifact name and list of exclusions to apply for the artifact.")
+(put 'cider-jack-in-dependencies-exclusions 'risky-local-variable t)
+(cider-add-to-alist 'cider-jack-in-dependencies-exclusions
+                    "org.clojure/tools.nrepl" '("org.clojure/clojure"))
+
 (defcustom cider-jack-in-auto-inject-clojure nil
   "Version of clojure to auto-inject into REPL.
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * [#1677](https://github.com/clojure-emacs/cider/issues/1677): Interpret `\r` as a newline.
 * [#1819](https://github.com/clojure-emacs/cider/issues/1819): Handle properly missing commands on `cider-jack-in`.
+* [#1824](https://github.com/clojure-emacs/cider/issues/1824): Can no longer jack-in to an inherited clojure version
```
bbatsov: I believe this can be phrased a bit better (and the sentence should end with a `.`).

cider.el
```diff
@@ -245,6 +245,12 @@ found for the PROJECT-TYPE"
 (cider-add-to-alist 'cider-jack-in-dependencies
                     "org.clojure/tools.nrepl" "0.2.12")
 
+(defvar cider-jack-in-dependencies-exclusions nil
```
bbatsov: This should be a defcustom IMO.

cider.el
```diff
@@ -245,6 +245,12 @@ found for the PROJECT-TYPE"
...
+  "List of exclusions for jack in dependencies where elements are lists of artifact name and list of exclusions to apply for the artifact.")
```
bbatsov: This docstring doesn't match Emacs's docstring convetions. Break it down and reformat it properly.

cider.el
```diff
@@ -245,6 +245,12 @@ found for the PROJECT-TYPE"
...
+(put 'cider-jack-in-dependencies-exclusions 'risky-local-variable t)
```
bbatsov: I'm not sure what's the idea behind this bit of code and one that follows it.

cider.el
```diff
@@ -300,18 +306,26 @@ string is quoted for passing as argument to an inferior shell."
   (concat (cider-boot-command-prefix (append dependencies plugins))
           (cider-boot-repl-task-params params middlewares)))
 
-(defun cider--list-as-lein-artifact (list)
+(defun cider--lein-artifact-exclusions (exclusions)
```
bbatsov: Where's the matching function for Boot 2.7?

bbatsov:
Maybe this should also be mentioned somewhere in the manual?

cider.el
benedekfazekas: Boot 2.7 is not yet released afaik. Will add support for boot as well when there is a stable release adding the feature enabling this.

cider.el
benedekfazekas: I think it is more like an internal thing for cider and "cider plugins" like clj-refactor just like `cider-jack-in-dependencies` which is a `defvar` as well

CHANGELOG.md
benedekfazekas: will do

cider.el
benedekfazekas: will do

cider.el
benedekfazekas: I basically followed the convention worked out with @Malabarba for `cider-jack-in-dependencies` for the exclusions too.

benedekfazekas:
will add an explanation to the manual too

cider.el
Malabarba: I agree with Benedek.

cider.el
Malabarba: I figured that the  jack-in dependencies are risky because they determine code that gets run. I'm not sure about the exclusions though.

cider.el
bbatsov: Seems to me we can add this even now if the command line params are already known. Not a bit deal, though.

cider.el
benedekfazekas: well, strictly speaking exclusions modify what is pulled in via the dependencies so it has an effect on what gets run too I guess...

benedekfazekas:
I looked at installation.md where auto injecting cider dependencies is describe. I kinda feel now that this is a quite internal bugfix which should not interest users rather as I said above cider "plugin" authors. Happy to reconsider if you think adding something is necessary.

PR #1852, opened 2016-09-25
Fix usage of RET in several keymaps

madand:
Replace occurrences of `[return]` in keymap definitions with
`(kbd "RET")`.

### DIFF
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -54,7 +54,7 @@
     (set-keymap-parent map cider-popup-buffer-mode-map)
     (define-key map "d" #'cider-browse-ns-doc-at-point)
     (define-key map "s" #'cider-browse-ns-find-at-point)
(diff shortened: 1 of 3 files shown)


Malabarba:
Just out of curiosity, is there a difference? :)

madand:
Yes, [there is](http://ergoemacs.org/emacs/emacs_key_notation_return_vs_RET.html). 

I found this issue because I use `C-M` instead of the `Return` key, and it did not worked in `cider-browse-ns-mode`.

Malabarba:
Isee. In that case, I think we can remove the C-m binding from [this line](https://github.com/clojure-emacs/cider/pull/1852/files#diff-fee1d15f6ebd550179a96d1cee6fa512R60), right?

madand:
Ah, nice catch! I have rewritten my commit with that line deleted.

bbatsov:
👍

Issue #1831, opened 2016-08-25
unicode apostrophes in help banner


johnv02139:
## Expected behavior

Words will appear as words, with no funny numbers in them.  :)

## Actual behavior

The apostrophes in the help banner show up as u\2019

## Steps to reproduce the problem

Admittedly I have an unusual setup.  I'm running emacs (24.5.1) on OS X inside of a Putty on a Windows box.

Perhaps emitting a -*- line at the top would work.  Or we could simply replace the Unicode apostrophes with ASCII.  Or, if everyone else is happy with it, you can just close this.  But it doesn't give a nice first impression, if you happen to have an environment like mine.

### CIDER version information

;; CIDER 0.13.0 (California), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_45

### Lein/Boot version

2.7.0

### Emacs version

24.5.1

### Operating system

Running on OS X, but connected to OS X via ssh from Putty and displaying emacs -nw within Putty

Malabarba:
What's the value of `buffer-file-coding-system` in you repl buffer?

johnv02139:
nil

Malabarba:
Could you try doing `M-x set-buffer-file-coding-system` and choosing utf8?

johnv02139:
Unfortunately that definitely didn't change the apostrophes.  As far as I could tell, it didn't refresh the buffer at all, but maybe it did it so quickly I just didn't notice.

I have the same problem viewing the cider-repl.el file.  There, I tried revert-buffer-with-coding-system, and tried utf-8 and a few other things, and actually saw the display change, but nothing I tried actually made it right.

Note: I use Putty to ssh to the OS X machine, then run "emacs -nw" from the command line.  I tried viewing the .el file with "less" in the terminal, and it didn't display the characters correctly there, either.  So it's not just emacs.

Putty has settings for "translation", and my settings appear to be correct, but maybe there's some subtle thing somewhere that I haven't checked.

Malabarba:
If you type `C-x 8 RET 2019`, on some other buffer (unrelated to cider), does it display correctly?

johnv02139:
Nope.

[...]

PR #1857, opened 2016-10-06
Replace \u2019 with ASCII apostrophe

johnv02139:
The welcome banner contained a few instances of Unicode character u2019, a RIGHT SINGLE QUOTATION MARK.  This doesn't display well in ASCII-only terminals.  Replace it with a simple apostrophe character: '

https://github.com/clojure-emacs/cider/issues/1831


bbatsov:
Thanks! Seems I really used some odd apostrophes somehow - no idea how this happened...

Issue #1758, opened 2016-05-23
Cider locks up Emacs regularly after cider-connect


mrrodriguez:
## Expected behavior

When using cider-connect to connect to a pre-existing nREPL server, I expect it to connect and be responsive once the prompt is available.  As I continue to interact and work in the REPL I do not expect frequent lock-ups that lock up Emacs itself, not only the REPL prompt.
## Actual behavior

I use cider-connect to connect to a pre-existing nREPL server.  I get a prompt fairly quickly with something like

```
user>
```

However, if I try to type any character, Emacs/Cider becomes fully locked up.  This lasts for anywhere from 5-10 seconds.  Once it becomes responsive again, some characters I typed during the lock-up typically show up.  Then I can start requiring my namespaces etc.  This all happens at normal speeds at this point.  However, frequently as I am doing more work in the editor with this connection present, I get frequent lock-ups again that lock the whole editor.  I do not know what it is doing during this time.
## Steps to reproduce the problem

My steps above are all you need to do to see this problem.
### CIDER version information

```
;; CIDER 0.12.0  nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_91
```

Note: I see the same behavior in Java 6 as well.
### Lein/Boot version

```
Lein 2.6.1
```
### Emacs version

```
GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2015-04-10 on builder10-9.porkrind.org
```
### Operating system

```
OS X 10.11.4 "El Capitan"
```

Malabarba:
1. Do `M-x toggle-debug-on-quit`.
2. Reproduce the hang (or wait for it to happen)
3. Hit `C-g` while it's hanged.

You'll get a backtrace you can paste here.

mrrodriguez:
This is from when it was locking up immediately after the Cider REPL prompt buffer prompt showed up after `cider-connect`.

```
Debugger entered--Lisp error: (quit)
  redisplay_internal\ \(C\ function\)()
  read-event(nil t 2)
  sit-for(2)
  execute-extended-command(nil "cider-connect")
[...]

Issue #1568, opened 2016-02-14
Clicking on folded nrepl dicts in the messages buffer results in an error


bbatsov:
We had this problem in the past, but @Malabarba fixed it. Guess we've got a regression.

bbatsov:
Just verified the old fix is here https://github.com/clojure-emacs/cider/blob/master/nrepl-client.el#L1287

I'm at a loss...

bbatsov:
Just consulted the docs:

```
Perform the action specified by a button at location pos. pos may be either a buffer position or a mouse-event. If use-mouse-action is non-nil, or pos is a mouse-event (see Mouse Events), try to invoke the button’s mouse-action property instead of action; if the button has no mouse-action property, use action as normal. pos defaults to point, except when push-button is invoked interactively as the result of a mouse-event, in which case, the mouse event’s position is used. If there’s no button at pos, do nothing and return nil, otherwise return t.
```

Seems our code is correct, but for some reason the `mouse-action` is not registered at all.

Malabarba:
It might be an emacs bug. I'll try to reproduce on 25.1 tomorrow.

bbatsov:
Yeah, maybe. That'd be fun. :-)

Malabarba:
Yeah, I was working on it. Can't believe I spent half an hour on this. This is really one of Emacs dark corners. :P

PR #1861, opened 2016-10-10
[Fix #1859] In message log insert large objects on request only

### LINKED ISSUE #1859: Optimize the performance of the nREPL message log
As discussed in https://github.com/clojure-emacs/cider/issues/1758 the nREPL message logging is pretty slow (especially when big data structures are involved) and enabling logging causes constant freezes.

To alleviate the problem logging was disabled by default, but we still need to analyze and optimize the problematic code (`nrepl--pp` seems to be the main culprit here). 

//cc @vspinu

vspinu:
Instead of removing the pp logic I have moved it to a different place. The pp print of big dicts, lists and strings now happens on requests only, i.e. when the user clicks the expansion button. This completely removes logger overhead. In my tests I don't see `nrepl-message-log` in the profiler tree anymore. 

Not closing the #1758 as there seem to be much more going on there.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1099,8 +1099,8 @@ The message is logged to a buffer described by
         (re-search-forward "^(" nil t)
         (delete-region (point-min) (- (point) 1)))
       (goto-char (point-max))
-      (nrepl--pp (nrepl-decorate-msg msg type)
-                 (nrepl--message-color (lax-plist-get (cdr msg) "id"))
+      (nrepl-log-pp-object (nrepl-decorate-msg msg type)
+                 (nrepl-log--message-color (lax-plist-get (cdr msg) "id"))
                  t)
       (when-let ((win (get-buffer-window)))
         (set-window-point win (point-max)))
@@ -1135,7 +1135,7 @@ BUTTON defaults the button at point."
           (goto-char start)
           (delete-overlay button)
           (delete-region start end)
-          (nrepl--pp obj)
+          (nrepl-log-pp-object obj)
           (delete-char -1)))
     (error "No button at point")))
 
@@ -1152,7 +1152,7 @@ BUTTON defaults the button at point."
           (nrepl-log-expand-button button)
           (setq button (next-button pos)))))))
 
-(defun nrepl--expand-button-mouse (event)
+(defun nrepl-log--expand-button-mouse (event)
   "Expand the text hidden under overlay button.
 EVENT gives the button position on window."
   (interactive "e")
@@ -1169,10 +1169,10 @@ EVENT gives the button position on window."
                  'face 'link
                  'help-echo "RET: Expand object."
                  ;; Workaround for bug#1568.
-                 'local-map '(keymap (mouse-1 . nrepl--expand-button-mouse)))
+                 'local-map '(keymap (mouse-1 . nrepl-log--expand-button-mouse)))
   (insert "\n"))
 
-(defun nrepl--message-color (id)
+(defun nrepl-log--message-color (id)
   "Return the color to use when pretty-printing the nREPL message with ID.
 If ID is nil, return nil."
   (when id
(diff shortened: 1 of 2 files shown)


nrepl-client.el
```diff
@@ -1145,49 +1155,67 @@ EVENT gives the button position on window."
...
-
-(defun nrepl--pp (object &optional foreground)
-  "Pretty print nREPL OBJECT, delimited using FOREGROUND."
-  (if (not (and (listp object)
-                (memq (car object) '(<-- --> dict))))
-      (progn (when (stringp object)
-               (setq object (substring-no-properties object)))
-             (pp object (current-buffer))
-             (unless (listp object) (insert "\n")))
+(defun nrepl--insert-button (label object)
```
bbatsov: Guess a better name would be `nrepl--log-insert-button` or something like this.

nrepl-client.el
```diff
@@ -1145,49 +1155,67 @@ EVENT gives the button position on window."
...
+      ;; listlike objects
+      (cond
+       ((memq head '(<-- -->))
+        (nrepl--pp-listlike object foreground button))
+       ((eq head 'dict)
+        (if (and button (> (length object) 1))
+            (nrepl--insert-button "(dict ...)" object)
+          (nrepl--pp-listlike object foreground button)))
+       (t
+        (if (and button (> (length object) 10))
```
vspinu: It looks rather useless option to me. Even though it's arbitrary I don't see why would anyone care to change that.

If we add config here, we should probably add a config for the minimal size of the dict that is buttonized. Currently I am buttonizing all non-empty dicts. One option would be to have same constant for both dicts and lists, but that would mean printing potentially deeply nested objects if at each level there is a small list.

Also strings are truncated if they are longer than 80chars. Do you want an option for that as well? :)

vspinu: BTW, there is a now a defunct `nrepl-dict-max-message-size` which I forgot to remove. IMO, unfolding dicts is an unnecessary overhead. People can always click when needed and it saves quite some space.

One slight drawback of this design that you cannot copy-paste or save the log directly because there is no text in those buttons in the first place. We can have a small utility `nrepl-log-unfold-all-buttons` to handle this.

bbatsov: > If we add config here, we should probably add a config for the minimal size of the dict that is buttonized. Currently I am buttonizing all non-empty dicts. One option would be to have same constant for both dicts and lists, but that would mean printing potentially deeply nested objects if at each level there is a small list.

Oh, I've missed this. I guess top-level dicts shouldn't be folded, though, as you'll get all your context about a request/response. 

I generally don't like magic numbers flowing around. I'm fine with not using defcustoms, but even consts make it easier to read the code. You can spruce up some comments here and there in the folding logic, as some decision might not be obvious to everyone.

nrepl-client.el
bbatsov: > One slight drawback of this design that you cannot copy-paste or save the log directly because there is no text in those buttons in the first place. We can have a small utility nrepl-log-unfold-all-buttons to handle this.

Sounds reasonable to me.

nrepl-client.el
vspinu: >  I guess top-level dicts shouldn't be folded,

The top dicts are not folded, only 1st level dicts are (aka those which display with `(dict ...)`. It looks like this:

```
(--> 
  op  "describe"
  session  "67aafc4f-8ba7-4b41-9a20-8f418ca0da8c"
  id  "8"
)
(<-- 
  aux  (dict ...)
  id  "8"
  ops  (dict ...)
  session  "67aafc4f-8ba7-4b41-9a20-8f418ca0da8c"
  status  ("done")
  versions  (dict ...)
)
```

> You can spruce up some comments here and there in the folding logic, as some decision might not be obvious to everyone.

Ok. I will get all these into a local let. You can extract them into variables if it will ever be a need for it.

nrepl-client.el
vspinu: > Guess a better name would be nrepl--log-insert-button or something like this.

Then all others should be renamed for consistency: nrepl--expand-button, nrepl--message-color, nrepl--expand-button-mouse, nrepl--pp. There is no consistency with respect to `nrepl-log-..` and `nrepl-message-..` prefix either.

bbatsov: Yeah, ideally we should renamed all the related stuff in some consistent manner, but that's not super important. If you'd like to tackle it - that'd be great, otherwise I can do it down the road.

nrepl-client.el
bbatsov: Sounds good to me.

vspinu:
Ok. I have parameterized the constants and renamed the main logging functions with `nrepl-log-` prefix. 

I also fixed the keymap problems on buttons on this ocassion. Fix in #1568, 814692dd7c5e5a9 used `local-map` property, which completely overwrites the major-mode map. So things like TAB, n,p and now e/E wouldn't work on buttons.

bbatsov:
Thanks!

PR #1863, opened 2016-10-10
[Fix #1707] Disable line truncating in cider-test-report-mode

### LINKED ISSUE #1707: Break the text in the error field of cider-test-error-report
Here is an example of how the error of a failing test:

image

The problem I'm having is that this is quite unreadable, as it's just one long line and I have to scroll to read it all.  

If I hit <kbd>RET</kbd> on the error to get the stracktrace buffer to popup, I can read the message with proper text breaks.  This is now my current workflow.

It looks this is just dumping the `ExceptionInfo` on one line.  Perhaps we can split it into `msg` and `cause` and wrap the message text so it's readable without scrolling?

bkruczyk:
This removes setting buffer local variable `truncate-lines` for `cider-test-report` buffer which caused #1707.

Although I don't know if line wrapping in whole report buffer is wanted behaviour?

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -198,7 +198,6 @@
 
 \\{cider-test-report-mode-map}"
   (setq buffer-read-only t)
-  (setq-local truncate-lines t)
   (setq-local electric-indent-chars nil))
 
 ;; Report navigation


bbatsov:
@jeffvalk mentioned that this causes some things to break, so I'm not sure it's a good idea.

Did you test the result of the change extensively?

bkruczyk:
Content of `cider-test-report` start with summary in which the longest line is about 40 characters long (eg. `Ran 31 assertions, in 7 test functions`), so I think it won't be affected by word-wrapping. Also, if buffer width is lesser than 50 and window widht is lesser than frame width line truncating is enabled automatically (value of variable `truncate-partial-width-windows`).

Then we have specific non passing tests and we can have there arbitrarily long sexp-s which will be wrapped with this change. On the one hand wrapped sexp-s can look ugly in narrow frames and often unmodifiable buffers does line truncating by default, but on the other we don't have line truncating enabled by default in `clojure-mode` and report mode shows portions of Clojure code, so maybe it would be consistent to not truncate it.

bbatsov:
One more thing - I've noticed we're enabling line truncating in every similar CIDER mode, so disabling it here would introduce an inconsistency. Maybe this was just copy-pasting, maybe we did it for a reason. Frankly, I don't remember many of the reasons behind some decisions we've made so far. :-) So we have to be extra careful...

bkruczyk:
Sure, maybe this is really non-issue. Too bad that text-wrapping can't be enabled on region or be specified as text property.

bbatsov:
It might be an option to have some global config for CIDER whether to use line-wrapping or line-truncation for popup-buffers. This seems like the Emacs way to solve such an issue. Everyone is happy and no risks are taken. :-)

bkruczyk:
Ok, I'll see what I can do. For now, I think we can close this pr.

PR #1865, opened 2016-10-12
[Fix #943] Make cider-repl-result-prefix work for multiline strings

### LINKED ISSUE #943: repl pretty printing and repl result prefix does not go together
perhaps that is intentional or a logical side effect so this is more like a question than a bugreport really.

```
; CIDER 0.8.2 (Java 1.7.0_17, Clojure 1.6.0, nREPL 0.2.6)
```

using `example-config` with 24.3 emacs

bkruczyk:
This commit fixes issue where only first line of multilne string
returned by repl was prefixed.

As it changes repl output string, font locking prefix and repl output
separately would not work and was removed.

--
This means that `cider-repl-result-prefix` won't be font locked as comment and rest of output will by font loced with `cider-font-lock-as-clojure`. This looks nice if prefix is someting like `;; => ` and not so bad with other values :)

If old behaviour is something necessary it can be done with some more work, let me know if this is good enough.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -613,12 +613,14 @@ If BOL is non-nil insert at the beginning of the line."
           (goto-char cider-repl-input-start-mark)
           (when (and bol (not (bolp)))
             (insert-before-markers "\n"))
-          (insert-before-markers (propertize cider-repl-result-prefix 'font-lock-face 'font-lock-comment-face))
-          (if cider-repl-use-clojure-font-lock
-              (insert-before-markers (cider-font-lock-as-clojure string))
-            (cider-propertize-region
-                '(font-lock-face cider-repl-result-face rear-nonsticky (font-lock-face))
-              (insert-before-markers string))))))
+          (let ((prefixed-string (mapconcat (lambda (s) (concat cider-repl-result-prefix s))
+                                            (butlast (split-string string "\n") 1)
+                                            "\n")))
+            (if cider-repl-use-clojure-font-lock
+                (insert-before-markers (cider-font-lock-as-clojure prefixed-string))
+              (cider-propertize-region
+                  '(font-lock-face cider-repl-result-face rear-nonsticky (font-lock-face))
+                (insert-before-markers prefixed-string)))))))
     (cider-repl--show-maximum-output)))
 
 (defun cider-repl-newline-and-indent ()


Malabarba:
Ok, first of all, thanks for the patch!

Second, this doesn't seem to fix #943. IIUC, that issue was due to pretty-printing using a different nrepl-dict key, not due to multi-line.

Third, I think the effect this will have on font-locking won't be ideal. Instead, maybe we should use `cider-font-lock-as-clojure` on the string first, and *then* do the splitting+prefixing.

bbatsov:
I'll have to take a look at the issue again, but before the problem was that when you were using pretty-printing the result returned was always `nil`. I recall originally I thought it'd be easy to fix this, but it turned out it wasn't and then I seem to have forgotten about the ticket.

bkruczyk:
@Malabarba Well I don't know if that was the original issue here, but for me above patch works as expected. 

I've noticed that aside from pretty-printing multiline strings are not prefixed, e.g. for prefix `;; ` evaluating
`user> "foo\n\bar\nbaz"` would return
```
;; foo
bar
baz
```

and not 

```
;; foo
;; bar
;; baz
```

Maybe it's accidental but patch works with pretty printing as the pretty-printed result is passed as `string` in function `cider-repl-emit-result`.

If you have any sugestions how to improve the solution or give me a hint what the root issue may be I'll be glad to work on this.

bbatsov:
The original issue was that when you enable both a result prefix and pretty-printing you wouldn't get the prefix at all - just the result. I'm not sure what when you have a multiline result it has to have the eval prefix on every line - that looks kind of odd to me. I'm interested to hear what other CIDER users/developers think about this matter.

Malabarba:
Like I said, I like the patch, and I'm fine with merging as soon as we fix the third point:

> Third, I think the effect this will have on font-locking won't be ideal. Instead, maybe we should use cider-font-lock-as-clojure on the string first, and then do the splitting+prefixing.

bbatsov:
> Like I said, I like the patch, and I'm fine with merging as soon as we fix the third point:

Even if we assume this is the desirable behaviour it can't be changed only here - similar result prefix logic exists for the minibuffer...

bkruczyk:
> The original issue was that when you enable both a result prefix and pretty-printing you wouldn't get the prefix at all - just the result. I'm not sure what when you have a multiline result it has to have the eval prefix on every line - that looks kind of odd to me. I'm interested to hear what other CIDER users/developers think about this matter.

Alright, it looks like current behaviour is different from described in issue. For me it works like this:
```
(setq cider-repl-use-pretty-printing)

user> (for [x (range 10)] (range x))

(()
 (0)
 (0 1)
 (0 1 2)
 (0 1 2 3)
 (0 1 2 3 4)
 (0 1 2 3 4 5)
 (0 1 2 3 4 5 6)
 (0 1 2 3 4 5 6 7)
 (0 1 2 3 4 5 6 7 8))
```
```
(setq cider-repl-result-prefix ";; => ")

user> (for [x (range 10)] (range x))

;; => (()
 (0)
 (0 1)
 (0 1 2)
 (0 1 2 3)
 (0 1 2 3 4)
 (0 1 2 3 4 5)
 (0 1 2 3 4 5 6)
 (0 1 2 3 4 5 6 7)
 (0 1 2 3 4 5 6 7 8))
```
> Even if we assume this is the desirable behaviour it can't be changed only here - similar result prefix logic exists for the minibuffer...

I'll check what can I do with this and with font locking.

bbatsov:
Guess someone fixed the original issue (maybe accidentally). As for the prefix on every line - after doing some thinking it might be good to configure to either show the result prefix on every line or just use whitespace padding.

bkruczyk:
Thank you for your sugestions, I think they are worth of separate pull request.

PR #1868, opened 2016-10-15
[Fix #1707] Add customization for line truncating in special buffers

### LINKED ISSUE #1707: Break the text in the error field of cider-test-error-report

bkruczyk:
As proposed in #1863 this commit allows to customize line truncating in CIDER's special buffers such as `cider-test-report` and `cider-doc`.

As most of those buffers (except `cidet-test-report-mode` and `cider-connections-buffer-mode`) use modes deriving `special-mode` I named it `cider-special-mode-truncate-lines`, if you can think of better name please let me know.

Side note: shouldn't `cider-test-report-mode` derive `special-mode` instead of `fundamental-mode`?
Also setting `buffer-read-only t` is redundant when deriving `special-mode`.

-----------------

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -42,6 +42,13 @@ prompt if that throws an error."
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-special-mode-truncate-lines t
+  "If non-nil, contents of CIDER's special buffers will be line-truncated.
+Should be set before loading CIDER."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
+
 (defun cider--should-prompt-for-symbol (&optional invert)
   "Return the value of the variable `cider-prompt-for-symbol'.
 Optionally invert the value, if INVERT is truthy."
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -125,6 +125,17 @@ More details can be found [here](https://github.com/clojure-emacs/cider/issues/9
 (setq cider-filter-regexps '(".*nrepl"))
 ```
 
+* By default contents of CIDER's special buffers such as `*cider-test-report*`
+  or `*cider-doc*` are line truncated. You can set
+  `cider-special-mode-truncate-lines` to `nil` to make those buffers use word
+  wrapping instead of line truncating.
+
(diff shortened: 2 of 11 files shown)


bbatsov:
There's just one small caveat - this defcustom needs to be set prior to loading anything from CIDER - once the mode definitions have been loaded its value won't matter.

bbatsov:
Probably this should be explained in its docstring and maybe in the manual.

bkruczyk:
I've fixed docstring and added documentation.

cider-common.el
```diff
@@ -42,6 +42,13 @@ prompt if that throws an error."
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-special-mode-truncate-lines t
+  "If non-nil, contents of CIDER's special buffers will be line-truncated.
+Should be set before starting CIDER."
```
bbatsov: starting -> loading

If you require CIDER before setting this it won't have any effect (even if you haven't started CIDER). At least for the files CIDER would load by default.

doc/configuration.md
```diff
@@ -125,6 +125,17 @@ More details can be found [here](https://github.com/clojure-emacs/cider/issues/9
 (setq cider-filter-regexps '(".*nrepl"))
 ```
 
+* By default contents of CIDER's special buffers such as `*cider-test-report*`
+  or `*cider-doc*` are line truncated. You can set
+  `cider-special-mode-truncate-lines` to `nil` to make those buffers use word
+  wrapping instead of line truncating.
+
+  This variable should be set before starting CIDER.
```
bbatsov: Same here.

cider-common.el
bkruczyk: Right.
Done :)

PR #1877, opened 2016-11-02
Add option to control whether the cider inspector window fills its frame

mallt:
Added an option to control whether the cider inspector window fills its frame. When non-nil, the cider inspector window will be shown as the only window instead of as a popup.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -53,6 +53,12 @@ The page size can be also changed interactively within the inspector."
   :group 'cider-inspector
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-inspector-fill-frame nil
+  "Controls whether the cider inspector window fills its frame."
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -53,6 +53,11 @@ The page size can be also changed interactively within the inspector."
   :group 'cider-inspector
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-inspector-window-fill-frame nil
```
bbatsov: I don't like this name this much. @Malabarba any better ideas?

mallt: Hi guys, I'd gladly change the name of the defcustom if you have any suggestions. Thanks!

Malabarba: I think just `cider-inspector-fill-frame` is descriptive enough.

Issue #1876, opened 2016-10-30
pprinting expressions _always_ uses window width as *print-right-margin*


cemerick:
I have my splits around 120 columns each, so this is really not what I want: 120 columns pretty-printed isn't very…pretty. At least IMO.

It'd be nice to be able to disable the binding of `*print-right-margin*`, or set it at some fixed width in cider. Absent that, here's what I'm using:

``` emacs
; don't pprint wider than 80 columns, regardless of the width of the current split
(let ((-cider--pprint--orig (symbol-function 'cider--nrepl-pprint-request-plist)))
  (defun cider--nrepl-pprint-request-plist (right-margin &optional pprint-fn)
    (funcall -cider--pprint--orig 80 pprint-fn)))
```

(Go ahead, make fun of my elisp, I care not.)

bbatsov:
It will be pretty easy to add some defcustom that would override the current default.

dpsutton:
Should we obey the `fill-column` variable in emacs and just ignore this completely?

cemerick:
Makes sense to me, especially given high DPI screens, etc.

dpsutton:
I think this can be closed now.

PR #1880, opened 2016-11-18
Use fill column rather than window width in pprint

dpsutton:

-----------------

Thanks!

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -994,7 +994,7 @@ CONTEXT represents a completion context for compliment."
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "formatted-code")))
 
-(defun cider-sync-request:format-edn (edn &optional right-margin)
+(defun cider-sync-request:format-edn (edn right-margin)
   "Perform \"format-edn\" op with EDN and RIGHT-MARGIN."
   (let* ((response (thread-first (list "op" "format-edn"
                                        "session" (cider-current-session)
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -112,6 +112,15 @@ change the setting's value."
   :type 'boolean
   :group 'cider-repl)
 
+(defcustom cider-repl-pretty-print-width nil
+  "Control the width of pretty printing on the REPL.
+This sets the wrap point for pretty printing on the repl.  If nil, it
+defaults to the variable `fill-column'."
+  :type '(restricted-sexp  :match-alternatives
+                          (integerp 'nil))
+  :group 'cider-repl
+  :package-version '(cider . "0.15.0"))
+
 (defcustom cider-repl-use-clojure-font-lock t
   "Non-nil means to use Clojure mode font-locking for input and result.
 Nil means that `cider-repl-input-face' and `cider-repl-result-face'
@@ -718,7 +727,7 @@ If NEWLINE is true then add a newline at the end of the input."
      (cider-column-number-at-pos input-start)
      (unless (or (not cider-repl-use-pretty-printing)
                  (string-match-p "\\`[ \t\r\n]*\\'" input))
-       (cider--nrepl-pprint-request-plist (1- (window-width)))))))
+       (cider--nrepl-pprint-request-plist (cider--pretty-print-width))))))
 
 (defun cider-repl-return (&optional end-of-input)
   "Evaluate the current input string, or insert a newline.
@@ -787,6 +796,12 @@ text property `cider-old-input'."
   (setq cider-repl-use-pretty-printing (not cider-repl-use-pretty-printing))
   (message "Pretty printing in REPL %s."
            (if cider-repl-use-pretty-printing "enabled" "disabled")))
+
+(defun cider--pretty-print-width ()
+  "Returns the width to use for pretty-printing."
+  (or cider-repl-pretty-print-width
+      fill-column
+      80))
 
 (defun cider-repl-switch-to-other ()
   "Switch between the Clojure and ClojureScript REPLs for the current project."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 * [#1707](https://github.com/clojure-emacs/cider/issues/1707): Allow to customize line truncating in CIDER's special buffers.
+* [#1876](https://github.com/clojure-emacs/cider/issues/1876): Set pretty-printing width with `cider-repl-pretty-print-width`. If this variable is not set, fall back to `fill-column`.
 
 ## 0.14.0 (2016-10-13)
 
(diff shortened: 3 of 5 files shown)


dpsutton:
Should address [1876](https://github.com/clojure-emacs/cider/issues/1876)

bbatsov:
Are we certain that everyone prefers the use of `fill-column` for this? I thought we were going to make this configurable, but obviously I don't mind a simpler solution. This change also requires a changelog entry.

bbatsov:
Also - maybe it's worth mentioning somewhere in the manual that `fill-column` controls the pretty-printing behaviour.

dpsutton:
> Are we certain that everyone prefers the use of fill-column for this? 

We are looking for a solution to print width deviating from window width. This is exactly the purpose of `fill-column`. In my mind, this is solved everywhere by emacs rather than in a particular mode. Or at least that was my thinking :)

Also, as to the reason I didn't put it into the changelog, this "reverts" cider back to the standard behavior of emacs, and my thinking was that therefore there was no need to announce to the user how to use the fill-column, in the same manner that we dn't tell them how to make new windows. But I agree with you that this probably merits a changelog.

I'll also investigate the failed tests, hopefully they are just being weird.

dpsutton:
Also, test failed to fluke:
> Dependency clojure-mode failed to install: Package `emacs-24.4' is unavailable

bbatsov:
> Also, as to the reason I didn't put it into the changelog, this "reverts" cider back to the standard behavior of emacs, and my thinking was that therefore there was no need to announce to the user how to use the fill-column, in the same manner that we dn't tell them how to make new windows. But I agree with you that this probably merits a changelog.

Well, the behaviour changes, so this has to be noted somewhere.

cider-client.el
```diff
@@ -999,7 +999,8 @@ CONTEXT represents a completion context for compliment."
   (let* ((response (thread-first (list "op" "format-edn"
                                        "session" (cider-current-session)
                                        "edn" edn)
-                     (append (cider--nrepl-pprint-request-plist right-margin))
+                     (append (cider--nrepl-pprint-request-plist (or right-margin
```
bbatsov: OK, let's leave it as is.

dpsutton:
Readme has been updated. I didn't mean to challenge that it should. I was just putting my thinking at the time out there. But then I realized I should have just updated the README. Sorry :)

dpsutton:
Thinking about it further, would you prefer a defcustom for pretty-print-width and fall back to fill-column? Just thinking about it more and perhaps your initial reaction was correct?

bbatsov:
> Thinking about it further, would you prefer a defcustom for pretty-print-width and fall back to fill-column? Just thinking about it more and perhaps your initial reaction was correct?

Yes, I would.

dpsutton:
Got the defcustom. Still split into multiple commits for easy review. Let me know if I need to change anything and I'll squash for ya.

dpsutton:
Tests pass locally for me. The tests are failing here for emacs <25 as they miss when-let and if-let, and the emacs 25 binary doesn't seem to work correctly. It's reporting its version as 23.x or something.

I've got the defcustom here; multiple commits for review. Let me know if you like and I can squash.

cider-repl.el
```diff
@@ -112,6 +112,13 @@ change the setting's value."
   :type 'boolean
   :group 'cider-repl)
 
+(defcustom cider-repl-pretty-print-width -1
```
bbatsov: I'd rather default to `nil`. Hate this notion of using negative numbers...

cider-repl.el
```diff
@@ -112,6 +112,13 @@ change the setting's value."
...
+  "Control the width of pretty printing on the repl.
```
bbatsov: repl -> REPL

cider-repl.el
```diff
@@ -112,6 +112,13 @@ change the setting's value."
...
+This sets the wrap point for pretty printing on the repl. If set to a negative
```
bbatsov: There should be 2 spaces after `.`.

cider-repl.el
```diff
@@ -112,6 +112,13 @@ change the setting's value."
...
+number, it defaults to the variable `fill-column`."
+  :type 'integer
+  :group 'cider-repl)
```
bbatsov: Add `:package-version` here.

bbatsov:
Apart from my small remarks, the code looks good. Apply the changes suggested and squash the commits into one.

cider-repl.el
dpsutton: I agree. But when i "typed" it with `:type 'integer` it was saying that there was a variable mismatch. I'm assuming this meant that `nil` wasn't acceptable as an integer. I'd much rather it be nil.

cider-repl.el
```diff
@@ -788,6 +795,12 @@ text property `cider-old-input'."
   (message "Pretty printing in REPL %s."
            (if cider-repl-use-pretty-printing "enabled" "disabled")))
 
+(defun cider--pretty-print-width ()
```
dpsutton: (let ((cider-repl-pretty-print-width 42)
          (fill-column 80))
      (cider--pretty-print-width))

Oddly enough, this shadows the `defcustom`. However, when in a test, it doesn't change its value away from -1.

    (describe "cider--pretty-print-width"
              (let ((cider-repl-pretty-print-width 42)
                     (fill-column 80))
                (it "prefers cider-repl-pretty-print-width"
                    (expect (cider--pretty-print-width)
                            :to-equal cider-repl-pretty-print-width)))
              (let ((cider-repl-pretty-print-width -1)
                     (fill-column 42))
                (it "falls back to fill-column otherwise"
                    (expect (cider--pretty-print-width)
                            :to-equal fill-column))))

The above test fails for some reason in thinking that `cider-repl-pretty-print-width` is -1 still rather than the shadowed binding.

dpsutton: This seems to get around that but i still don't understand why.

    (describe "cider--pretty-print-width"
      (it "falls back to fill-column"
        (let ((old-width cider-repl-pretty-print-width))
          (setq cider-repl-pretty-print-width -1)
          (expect (cider--pretty-print-width)
                  :to-equal fill-column)
          (setq cider-repl-pretty-print-width old-width)))

      (it "prefers cider-repl-pretty-print-width"
        (let ((old-width cider-repl-pretty-print-width))
          (setq cider-repl-pretty-print-width 42)
          (expect (cider--pretty-print-width)
                  :to-equal cider-repl-pretty-print-width)
          (setq cider-repl-pretty-print-width old-width))))

cider-repl.el
bbatsov: I think you need to use a composite type here - see https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html#Composite-Types and `(choice alternative-types…)`.

cider-repl.el
bbatsov: The first test looks completely correct to me. Can't imagine how the let binding wouldn't work. I guess you can run the test using the debugger and see what exactly happens there.

dpsutton:
Uses nil as sentinel, test updated. Should be good to go. Thanks for all the pointers along the way!

bbatsov:
👍

Issue #1882, opened 2016-11-22
Cider removed `cider-compat.el` which appears to have removed support for emacs 24.x not 24.3


dpsutton:
## Expected behavior
CIDER should be able to jack in

## Actual behavior

`cider-add-to-alist: Invalid function: (cons (assoc car alist))` is the error message.

This can be seen from the build logs

https://api.travis-ci.org/jobs/177236922/log.txt?deansi=true

```
cider.el:812:1:Warning: the following functions are not known to be defined:
    when-let, if-let, exact-buff
Wrote /home/travis/build/clojure-emacs/cider/cider.elc
Compiling /home/travis/build/clojure-emacs/cider/cider-util.el...


In cider-add-to-alist:
cider-util.el:338:39:Warning: `(cons (assoc car alist))' is a malformed
    function
cider-util.el:348:15:Warning: reference to free variable `cons'
```

I believe that `cider-compat.el` was removed and `subr-x.el` was included thinking this included `if-let`. However, while `subr-x.el` exists on emacs 24.4+, it does not include the macros `if-let`. I believe that this commit mistakenly moved support for emacs to 25+.

## Steps to reproduce the problem

Install CIDER from elpa for git in emacs 24.4.
The build logs from https://travis-ci.org/clojure-emacs/cider/builds/177236920 support this as well

tommy-mor:
getting this problem as well, on v 24.5

bbatsov:
Ops, guess I forgot those macros were added later. I guess by Emacs 26 your actually mean 25. :-)

tommy-mor:
I installed emacs from github using the --HEAD flag in homrbrew, and when I ckeck the version it says v 26. I guess that makes me somewhere in between

PR #1883, opened 2016-11-23
Revert "Drop support for Emacs 24.3"

dpsutton:

This reverts commit 70971d0a918053d86e76e4ff8248b25a8646940e.

Should be a fix for [1882](https://github.com/clojure-emacs/cider/issues/1882)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: true
 language: emacs-lisp
 env:
+  - EMACS_BINARY=emacs-24.3-bin PATH=$HOME/.evm/bin:$PATH MAKE_TEST=test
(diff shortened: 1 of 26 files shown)


benedekfazekas:
I guess you still need to keep this line: https://github.com/dpsutton/cider/blob/master/cider.el#L15 and the `.travis.yml` changes to stop supporting 24.3 only

bbatsov:
I don't really want to revert this. It won't work with Emacs 24.3 anymore because of `clojure-mode`, so I'd rather just add back `if-let` and `when-let`.

dpsutton:
Sounds good. I know you're busy so in case you wanted to just revert I wanted you to have a pull request so that you just had to approve and merge.

PR #1888, opened 2016-12-11
Allow auto-display of REPL without cursor focusing

### LINKED ISSUE #1872: Option for opening REPL but not moving cursor
We have the option `cider-repl-pop-to-buffer-on-connect`, which affects what happens when CIDER connects to the REPL. When it's on, the REPL buffer is opened and my cursor is moved to it. When it's off, the REPL buffer is not opened.

I would like the REPL buffer to be opened but my cursor to _not_ be moved to it.

This is my current solution:

```
(setq cider-repl-pop-to-buffer-on-connect nil)
(defadvice cider-repl-init (after display-repl-buffer)
  (display-buffer buffer))
```

It's not ideal because it means that when I do `C-c M-J`, only one of the REPLs is opened, instead of both, which is what happens when `cider-repl-pop-to-buffer-on-connect` is enabled (https://github.com/raxod502/radian/issues/105).

raxod502:
Resolves #1872. I updated all of the places where
`cider-repl-pop-to-buffer-on-connect` was used. This should be perfectly
reverse-compatible, since the only new behavior happens if you specifically set
the variable to `display-only`.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -84,8 +84,13 @@
 (defcustom cider-repl-pop-to-buffer-on-connect t
   "Controls whether to pop to the REPL buffer on connect.
 
-When set to nil the buffer will only be created."
-  :type 'boolean
+When set to nil the buffer will only be created, and not displayed.  When
+set to `display-only' the buffer will be displayed, but it will not become
+focused.  Otherwise the buffer is displayed and focused."
+  :type '(choice (const :tag "Create the buffer, but don't display it" nil)
+                 (const :tag "Create and display the buffer, but don't focus it"
+                   display-only)
+                 (const :tag "Create, display, and focus the buffer" t))
   :group 'cider-repl)
 
 (defcustom cider-repl-display-in-current-window nil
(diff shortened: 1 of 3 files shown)


cider-repl.el
```diff
@@ -290,8 +293,11 @@ client process connection.  Unless NO-BANNER is non-nil, insert a banner."
     (cider-repl--insert-banner-and-prompt buffer))
   (when cider-repl-display-in-current-window
     (add-to-list 'same-window-buffer-names (buffer-name buffer)))
-  (when cider-repl-pop-to-buffer-on-connect
-    (pop-to-buffer buffer))
+  (cond
```
bbatsov: Using `pcase` is better for this type of matching.

cider-repl.el
```diff
@@ -84,8 +84,11 @@
 (defcustom cider-repl-pop-to-buffer-on-connect t
   "Controls whether to pop to the REPL buffer on connect.
 
-When set to nil the buffer will only be created."
-  :type 'boolean
+When set to nil the buffer will only be created. When set to `display-only'
+the buffer will be displayed, but the cursor will not be focused in it."
+  :type '(choice (const :tag "Only create the buffer" nil)
```
bbatsov: Create the buffer, but don't display it.

doc/using_the_repl.md
```diff
@@ -70,6 +70,13 @@ Normally, the REPL buffer is auto-displayed in a separate window after
 (setq cider-repl-pop-to-buffer-on-connect nil)
 ```
 
+If you want the REPL buffer to be auto-displayed, but don't want the cursor to
```
bbatsov: but don't want it to be focused.

(the cursor doesn't become focused - a window becomes focused)

bbatsov:
A few minor things:

* prefix the commit message with `[Fix #1872]`
* add a changelog entry
* fix my inline remarks

raxod502:
How does this look?

bbatsov:
Looks great! 👍

Issue #1875, opened 2016-10-28
cljc buffers not evaluated in cljs repl by cider-load-buffer


leppert:
## Expected behavior

After running `cider-jack-in-clojurescript`, `cider-load-buffer` loads the cljc buffer in both the Clojure and ClojureScript REPLs per the [`cider-load-buffer` docstring](https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L1526-L1527).
## Actual behavior

The buffer is evaluated in the Clojure REPL but not the ClojureScript REPL.
## Steps to reproduce the problem
1. Check out this sample project: https://github.com/leppert/cljc-bug
2. Open [src/cljc_bug/core.cljc](https://github.com/leppert/cljc-bug/blob/master/src/cljc_bug/core.cljc)
3. Run `cider-jack-in-clojurescript`
4. `C-c M-n` to load the namespace. Note: this works correctly in both REPLs
5. `C-c C-k` to eval the buffer
6. Calling `(foo "bar")` in the Clojure REPL will now correctly print `bar Hello, World!`
7. Calling `(foo "bar")` in the ClojureScript REPL will throw the error `WARNING: Use of undeclared Var /foo at line 1 <cljs repl>`

![image](https://cloud.githubusercontent.com/assets/26643/19821543/e9745d74-9d2c-11e6-8b68-1a1c3a65b107.png)
### CIDER version information

```
;; CIDER 0.15.0snapshot (package: 20161016.424), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_40
```
### Lein/Boot version

`Leiningen 2.7.1 on Java 1.8.0_40 Java HotSpot(TM) 64-Bit Server VM`
### Emacs version

`GNU Emacs 25.1.1 (x86_64-apple-darwin16.0.0, NS appkit-1504.00 Version 10.12 (Build 16A320)) of 2016-09-21`
### Operating system

macOS 10.12 (16A320)

bbatsov:
Sounds like some regression to me. Probably fixing this would be easy, but I won't be able to look into this for a while.

leppert:
@bbatsov I'll give it a stab.

leppert:
To narrow this down further, it does appear that:

- [`cider--cljc-or-cljx-buffer-p`](https://github.com/clojure-emacs/cider/blob/5a9b3ced54f52abfa3632a9e8d382be49cd4faff/cider-util.el#L85-L90) returns correctly
[...]

PR #1892, opened 2016-12-19
Allow specifying connection for load buffer

dpsutton:
When running a cljc buffer, you can load the code into more than just a
single connection. This is managed by cider-map-connections. However,
the load buffer command was reaching out to the current connection for
the session id, leading the entire mechanism working correctly except
that it would talk to the OTHER repl session inside of the repl.

-----------------

Thanks!

Fix for [Issue 1875](https://github.com/clojure-emacs/cider/issues/1875)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 * [#1707](https://github.com/clojure-emacs/cider/issues/1707): Allow to customize line truncating in CIDER's special buffers.
 * [#1876](https://github.com/clojure-emacs/cider/issues/1876): Set pretty-printing width with `cider-repl-pretty-print-width`. If this variable is not set, fall back to `fill-column`.
+* [#1875](https://github.com/clojure-emacs/cider/issues/1875): Ensure cljc buffers can load buffer into both clj and cljs repls.
 
 ## 0.14.0 (2016-10-13)
 
--- a/cider-client.el
+++ b/cider-client.el
@@ -766,7 +766,10 @@ Return the REPL buffer given by `cider-current-connection'.")
 
 (defun cider-current-session ()
   "Return the eval nREPL session id of the current connection."
-  (with-current-buffer (cider-current-connection)
+  (cider-session-for-connection (cider-current-connection)))
+
+(defun cider-session-for-connection (connection)
+  (with-current-buffer connection
     nrepl-session))
 
 (defun cider-current-messages-buffer ()
@@ -852,7 +855,9 @@ loaded.
 If CONNECTION is nil, use `cider-current-connection'.
 If CALLBACK is nil, use `cider-load-file-handler'."
   (cider-nrepl-send-request (list "op" "load-file"
-                                  "session" (cider-current-session)
+                                  "session" (if connection
+                                                (cider-session-for-connection connection)
+                                              (cider-current-session))
                                   "file" file-contents
                                   "file-path" file-path
                                   "file-name" file-name)


dpsutton:
So this commit should fix this instance, but basically any function that can be used with map connections is a potential source for the same bug. We'll need to add optional conncetion parameters, or honestly, make every function require a connection

bbatsov:
I'm a bit confused - I thought that with `cljc` buffers our strategy was to dispatch evals to both connections. I might not be understanding what exactly are you trying to solve, though. I definitely find it odd that `current-connection` should accept a param forcing it...

dpsutton:
We do. We `cider-map-connections` and we pass a function that takes a connection. However, a connection is just a buffer and we need two bits of information from the buffer. 

1) Where to write the output and 
2) what nrepl session id to use.

The problem is best exemplified by the following nrepl messages from the cljs buffer. 
```

(<--
  changed-namespaces  (dict)
  id  "12"
  repl-type  "cljs"
  session  "a94bea76-0c4b-47bb-b5d7-4daa9e589b3c"  ;; <- this is the session for the cljs repl
  status  ("state")
)
(-->
  op  "load-file"
  session  "32aa402c-2d52-43d3-92f9-b8c7a635f8a8" ;; <- this is the session id for the clj repl
  file  "(ns cljc-bug.core)

(defn foo
  "I don't do a whole lot."
  ..."
  file-path  "/home/dan/projects/clojure/cljc-bug/src/cljc_bug/core.cljc"
  file-name  "core.cljc"
  id  "13"
)
```

The problem is that we are explicitly using another connection for the cljs buffer. But the `cider-request:load-file` function cannot respect the other connection. It uses that buffer, but when it asks for which session id to use, it is incapable of resolving *which* connection to use. It can only call give me current connection, which goes to whichever connection buffer CIDER wants and returns the nrepl-session id buffer local var. So in essence, we are telling the loader which connection to use but we have to rely on a global variable for it to resolve which session id to use.

Once this happens, its using the cljs repl buffer to talk to the clj repl because the session id is ultimately the most important specifier. This is what causes the apparent session id change in the above nrepl messages.

Also, the function could be better named. All is does it grab the sessoin id from the buffer local var. `Get-session-id-from-connection` makes more sense.

dpsutton:
Concise answer: we are mapping across both connections for cljc buffers. However, the load buffer reaches out to get the session id to use from a function that returns session id for the current connection. Therefore the clj and cljs buffers both use the same session id. So the code is loaded into clj repl twice and cljs repl 0 times.

bbatsov:
Got it. So the fix is definitely wrong. :-) You should simply get the session id from the connection that was passed to the load command instead of messing with `cider-current-session`.

dpsutton:
I agree with you but probably in the opposite way. In the assumption that there's always only a single connection that makes sense, its okay for evaluation handlers to reach out to a global current session. I would make getting an nrepl session id *require* a connection buffer.

I think this is the first of many bugs. Any random handler could reach out and use the wrong session id. I would make connection required rather than optional, but this would take quite a bit of checking.

My argument against your thinking is that to get a nrepl session, you open up a buffer and return the buffer local var, which is exactly what this function does. It doesn't make sense to me to have another function that does the exact same thing. Further, since there can be cljc buffers, the notion of current-session kinda doesn't have meaning: _which_ current connection.

bbatsov:
> My argument against your thinking is that to get a nrepl session, you open up a buffer and return the buffer local var, which is exactly what this function does. It doesn't make sense to me to have another function that does the exact same thing. Further, since there can be cljc buffers, the notion of current-session kinda doesn't have meaning: which current connection.

When I wrote all of this code there was no `cljc`, no clojurescript support, no nothing. I don't use them and I didn't really think how to support them properly. This was just bolted on the current code by suggestion @expez if memory serves. Properly supporting the notion of multiple connections for every source buffer will require more changes for sure. At any rate - you can't really push something name `current-session` to act on the non-current connection - it simply makes no sense from a naming standpoint. Those names made a lot of sense when the code was authored - of that I'm certain.

bbatsov:
One more thing - the reason that cljc and ClojureScript are so poorly supported is that none of the main maintainers actually uses them (much) - I've never used them at all. With few people contributing fixes for cljc and cljs no one can expect codebase that's actually designed to support them... CIDER right now is very much a Clojure-only tool that has some crude support for something else...

dpsutton:
I hope I didn't give you the impression that I was being critical or harsh. I really enjoy CIDER and I've made a commitment to stay with supporting it. I understand how fundamental assumptions like "there is always a connection you can ask for" can change and how wide ranging the refactor is to handle going from 1 to 2. 

On a question of style, do you prefer me to leave this function alone and write a new one that is `get-nrepl-session-from-connection` or something similar and we start using that one? My hesitation is that it would have essentially the exact same implementation, but maybe this one could throw a meaningful error when no connection was provided.

PS, I have been working on CIDER since I was lucky enough to have dinner with you at the seattle clojure west. Ever since then I've been trying to work on bugs and help people with issues. I am by no means critical of CIDER.

bbatsov:
> I hope I didn't give you the impression that I was being critical or harsh. I really enjoy CIDER and I've made a commitment to stay with supporting it. I understand how fundamental assumptions like "there is always a connection you can ask for" can change and how wide ranging the refactor is to handle going from 1 to 2.

I just wanted to make it clear that the root of most such problems is that we never did the work to properly support those workflows. It's just a collection of workarounds at the moment... Eventually we'll have to rethink some of the fundamentals, but this will require a lot of work.

> On a question of style, do you prefer me to leave this function alone and write a new one that is get-nrepl-session-from-connection or something similar and we start using that one? My hesitation is that it would have essentially the exact same implementation, but maybe this one could throw a meaningful error when no connection was provided.

Add some function `cider-connection-session` (and maybe `cider-connection-tooling-session`). You can implement the old functions in terms of the new ones. 

> PS, I have been working on CIDER since I was lucky enough to have dinner with you at the seattle clojure west. Ever since then I've been trying to work on bugs and help people with issues. I am by no means critical of CIDER.

:-) Happy to hear this! I'm fine with people being critical, btw. There a ton of things that are horrible and need to be improved. My main frustration is that I don't have the time to work on fixing them... I'm always happy to see more people contributing on a steady base.

bbatsov:
You'll also have to do rebase.

bbatsov:
This is a bugfix, so it should be mentioned in the changelog.

dpsutton:
Changelog updated.

bbatsov:
👍

PR #1894, opened 2016-12-20
Bugfix: (pcase) accepts only backquote in Emacs 24.5.1

dottedmag:
Introduced in ce349eb24f2, caused error like:

    error in process filter: pcase--u1: Unknown upattern `(quote display-only)'


bbatsov:
Thanks for spotting and fixing this!

PR #1898, opened 2016-12-23
Bind terminal TAB in stacktrace-mode

### LINKED ISSUE #1897: TAB doesn't work in stacktrace-mode
tl;dr `<tab>` bound but not `TAB` in stacktrace buffers.

Here are the keybindings in `stacktrace-mode` according to `C-h m`:

```
Stacktrace mode defined in ‘cider-stacktrace.el’:
Major mode for filtering and navigating CIDER stacktraces.

key             binding
---             -------

ESC             Prefix Command
SPC             scroll-up-command
-               negative-argument
0               cider-stacktrace-cycle-all-causes
1               cider-stacktrace-cycle-cause-1
2               cider-stacktrace-cycle-cause-2
3               cider-stacktrace-cycle-cause-3
4               cider-stacktrace-cycle-cause-4
5               cider-stacktrace-cycle-cause-5
6 .. 9          digit-argument
<               beginning-of-buffer
>               end-of-buffer
?               describe-mode
a               cider-stacktrace-toggle-all
c               cider-stacktrace-toggle-clj
d               cider-stacktrace-toggle-duplicates
g               revert-buffer
h               describe-mode
j               cider-stacktrace-toggle-java
r               cider-stacktrace-toggle-repl
t               cider-stacktrace-toggle-tooling
DEL             scroll-down-command
S-SPC           scroll-down-command
<backtab>       cider-stacktrace-cycle-all-causes
<remap>         Prefix Command
<tab>           cider-stacktrace-cycle-current-cause

M-.             cider-stacktrace-jump
M-n             cider-stacktrace-next-cause
M-p             cider-stacktrace-previous-cause
```

As you can see, `<tab>` is bound but `TAB` is not. This has the consequence that `cider-stacktrace-cycle-current-cause` can't be triggered in the terminal, since `TAB` is sent instead of `<tab>`.

This should be a pretty simple fix (just bind `TAB` in addition to `<tab>`).

### CIDER version information

```
;; CIDER 0.15.0snapshot (package: 20161219.713), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_112
```

### Lein/Boot version

Leiningen 2.7.1

### Emacs version

25.1.1

### Operating system

El Capitan 10.11.6

raxod502:
This should fix #1897.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -195,7 +195,7 @@ The error types are represented as strings."
     (define-key map "4" #'cider-stacktrace-cycle-cause-4)
     (define-key map "5" #'cider-stacktrace-cycle-cause-5)
     (define-key map "0" #'cider-stacktrace-cycle-all-causes)
-    (define-key map [tab] #'cider-stacktrace-cycle-current-cause)
+    (define-key map (kbd "TAB") #'cider-stacktrace-cycle-current-cause)
     (define-key map [backtab] #'cider-stacktrace-cycle-all-causes)
     (easy-menu-define cider-stacktrace-mode-menu map
       "Menu for CIDER's stacktrace mode"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#1707](https://github.com/clojure-emacs/cider/issues/1707): Allow to customize line truncating in CIDER's special buffers.
 * [#1876](https://github.com/clojure-emacs/cider/issues/1876): Set pretty-printing width with `cider-repl-pretty-print-width`. If this variable is not set, fall back to `fill-column`.
 * [#1875](https://github.com/clojure-emacs/cider/issues/1875): Ensure cljc buffers can load buffer into both clj and cljs repls.
+* [#1897](https://github.com/clojure-emacs/cider/issues/1897): Bind TAB in stacktrace buffers in the terminal.
 
 ## 0.14.0 (2016-10-13)
 


cider-stacktrace.el
```diff
@@ -196,6 +196,7 @@ The error types are represented as strings."
     (define-key map "5" #'cider-stacktrace-cycle-cause-5)
     (define-key map "0" #'cider-stacktrace-cycle-all-causes)
     (define-key map [tab] #'cider-stacktrace-cycle-current-cause)
+    (define-key map (kbd "TAB") #'cider-stacktrace-cycle-current-cause)
```
bbatsov: Do you still need to bind `[tab]`? What about `backtab` (Shift + TAB)? Does this work properly or not?

raxod502: You're right, of course. I've unbound `[tab]`; it's not necessary because `TAB` will work for both terminal and windowed Emacs. (`<tab>` is automatically translated to `TAB` but not the other way around.)

For whatever reason, `<backtab>` seems to work fine in both windowed and terminal Emacs. I have no idea why there are two versions of `<tab>` but only one version of `<backtab>`.

xiongtx: > I have no idea why there are two versions of `<tab>` but only one version of `<backtab>`.

Emacs has a few [named ASCII control characters](https://www.gnu.org/software/emacs/manual/html_node/emacs/Named-ASCII-Chars.html), such as `TAB` and `RET`. There's no ASCII char for "backtab".

PR #1905, opened 2017-01-03
Reconnect to the same host:port after cider-restart if the connection was established with `cider-connect`

### LINKED ISSUE #1895: Restarting a repl does not connect to same port
## Expected behavior

`cider-restart` should connect to the same `host:port` as when I used `cider-connect`.

1. `cider-connect` to connect to a repl running on `localhost:6005`.
2. Then call `cider-restart`.
3. The new repl should connect to `localhost:6005` as before.

## Actual behavior

`cider-restart` seems to use `cider-jack-in` instead, and does not connect to the repl running on `localhost:6005`.

cskksc:
It's been a while since I've worked on anything in CIDER. Feels good to be back :-)

This PR fixes #1895.

@bbatsov, the approach is the same as what you outlined in the comment.

1. Create a new buffer-local var `cider-connection-created-with` to track how the connection was created. It defaults to `'jack-in`, and we set it to `'connect` in `cider-connect`.
2. In `cider--restart-connection`, we do the right thing based on the type of  `conn`.

Thanks!

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -245,6 +245,10 @@ namespace-qualified function of zero arity."
     map)
   "Minibuffer keymap used for reading Clojure expressions.")
 
+(defvar-local cider-connection-created-with 'jack-in
+  "Save how the connection was created.
+Its value can be either 'jack-in or 'connect.")
+
 (defun cider-read-from-minibuffer (prompt &optional value)
   "Read a string from the minibuffer, prompting with PROMPT.
 If VALUE is non-nil, it is inserted into the minibuffer as initial-input.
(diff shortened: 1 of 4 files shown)


cider-interaction.el
```diff
@@ -1742,17 +1746,22 @@ and all ancillary CIDER buffers."
 (defun cider--restart-connection (conn)
   "Restart the connection CONN."
   (let ((project-dir (with-current-buffer conn nrepl-project-dir))
-        (buf-name (buffer-name conn)))
+        (buf-name (buffer-name conn))
+        ;; save these variables before we kill the connection
+        (conn-type (with-current-buffer conn cider-connection-created-with))
```
bbatsov: `conn-type` is a bit misleading here. Maybe `conn-creation-method`?

cider-interaction.el
```diff
@@ -245,6 +245,10 @@ namespace-qualified function of zero arity."
     map)
   "Minibuffer keymap used for reading Clojure expressions.")
 
+(defvar-local cider-connection-created-with 'jack-in
+  "Save how the connection was created. Its value can be either 'jack-in or
```
bbatsov: The second sentence should be on a separate line.

cider-interaction.el
```diff
@@ -1742,17 +1746,22 @@ and all ancillary CIDER buffers."
...
     ;; Workaround for a nasty race condition https://github.com/clojure-emacs/cider/issues/439
     ;; TODO: Find a better way to ensure `cider-quit' has finished
     (message "Waiting for CIDER connection %s to quit..."
              (cider-propertize buf-name 'bold))
     (sleep-for 2)
-    (if project-dir
-        (let ((default-directory project-dir))
-          (cider-jack-in))
-      (error "Can't restart CIDER connection for unknown project"))))
+    (if (eq conn-type 'connect)
```
bbatsov: I guess you can make this a `pcase` and add some error handling in case the `conn-type` is unexpectedly something different than the two possible values.

bbatsov:
It's good to have you back! I've added a few small remarks.

cskksc:
Done :-)

PR #1906, opened 2017-01-03
[Fix #1776] Add new customization variable `cider-test-defining-names`.

### LINKED ISSUE #1776: Forms which macroexpand into deftest aren't recognised by cider-test-run-test
## Expected behavior

`cider-test-run-test` should run the test at point, even if it is not directly declared in the Emacs buffer with `deftest`.

You might use a convenience macro which macroexpands into a `deftest` form as a means of following [Stuart Sierra's advice](https://stuartsierra.com/2016/05/19/fixtures-as-caches) to avoid `clojure.test`'s fixtures mechanism, for example.
## Actual behavior

"No test at point"

``` clojure
(ns fancy-test
  (:require [clojure.test :refer :all]))

(defn- do-setup []
  (println "Expensive resource setup"))

(defmacro deffancytest [sym & body]
  `(deftest ~sym
     (do-setup)
     ~@body))

(deffancytest fancytest
  (is (= 1 1)))
```

Invoke `cider-test-run-test` with the point in `fancytest` and you'll get "No test at point" in the minibuffer.
### CIDER version information

```
;; CIDER 0.12.0 (Seattle), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_40
```
### Lein/Boot version

boot 2.5.5
### Emacs version

24.4.1
### Operating system

OS X 10.11

ncalexan:
This fix is strictly simpler than that suggested in
https://github.com/clojure-emacs/cider/issues/1776#issuecomment-223758069.  It
has the advantage of being much more robust.

I partially implemented the suggested fix, and witnessed the following two
issues.  First, full "macroexpand" does not leave a recognizable `deftest` form;
it generally leaves a `(def test-... (fn [] ...))` sexp.  That implies that a
recursive macroexpansion would be required, with each step of the recursion
checking for a recognized form.  Second, even recognizing such a form is tricky,
because the expansion may refer to deftest in an aliased namespace (e.g.,
`t/deftest` or `clojure.test/deftest` rather than bare `deftest`).  One could
then try to identify possible namespaces using the current environment, but this
is getting complicated.

Therefore, I think it better to have the user configure their environment to
help them solve this problem.  (And, of course, future work can implement the
full macroexpansion approach.)

-----------------

- [n/a] You've added tests (if possible) to cover your change(s)
- [n/a] You've updated the readme (if adding/changing user-visible functionality)
- [n/a] You've updated the refcard (if you made changes to the commands listed there)

Thanks!

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -59,6 +59,15 @@
   :group 'cider-test
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-test-defining-forms '("deftest" "defspec")
+  "Forms that define individual tests.
+CIDER considers the top-level form around point to define a test if the
+form starts with one of these forms.
+Add to this list to have CIDER recognize additional test defining macros."
+  :type '(repeat string)
+  :group 'cider-test
+  :package-version '(cider . "0.15.0"))
+
 (defvar cider-test-last-summary nil
   "The summary of the last run test.")
 
@@ -680,7 +689,7 @@ is searched."
           (cider-test-execute ns (list var)))
       (let ((ns  (clojure-find-ns))
             (def (clojure-find-def)))
-        (if (and ns (member (car def) '("deftest" "defspec")))
+        (if (and ns (member (car def) cider-test-defining-forms))
             (progn
               (cider-test-update-last-test ns (cdr def))
               (cider-test-execute ns (cdr def)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add new customization variable `cider-special-mode-truncate-lines`.
 * Add an option `cider-inspector-fill-frame` to control whether the cider inspector window fills its frame.
 * [#1893](https://github.com/clojure-emacs/cider/issues/1893)Add negative prefix argument to `cider-refresh` to inhibit invoking of cider-refresh-functions
+* [#1776](https://github.com/clojure-emacs/cider/issues/1776)Add new customization variable `cider-test-defining-forms` allowing new test defining forms to be recognized.
 
 ### Changes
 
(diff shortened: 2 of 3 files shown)


cider-test.el
```diff
@@ -59,6 +59,12 @@
   :group 'cider-test
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-test-defining-names '("deftest" "defspec")
+  "Macro names that define individual tests."
```
bbatsov: I think you should expand the description here - if I didn't know what this was about I'd never figure it out. :-)

cider-test.el
```diff
@@ -59,6 +59,12 @@
...
+  :type 'list
```
bbatsov: The actual type should be `'(repeat string)`.

cider-test.el
```diff
@@ -59,6 +59,12 @@
   :group 'cider-test
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-test-defining-names '("deftest" "defspec")
```
bbatsov: I'd prefer `cider-test-defining-forms/macros`. Names is a bit vague I think.

bbatsov:
I'm obviously fine with simple solutions. :-)

This should definitely be added to the manual as well - with more extensive explanations, otherwise few people are going to figure out the possibility of customization even exists.

bbatsov:
Sorry about closing the PR - wrong button. :-)

ncalexan:
@bbatsov thanks for such a fast and helpful review.  I force pushed a new patch since this rewrote the entire original.

Also, +1 to CIDER's great contributing checklist -- one of the best I've used on github.  So far, a great contribution experience \o/

bbatsov:
> Also, +1 to CIDER's great contributing checklist -- one of the best I've used on github. So far, a great contribution experience \o/

Happy to hear this! :-) Thanks for your contribution!

Issue #1896, opened 2016-12-22
java class output doesn't show in cider-repl buffer


sound2gd:
## Expected behavior
print the `something` and `nil`
## Actual behavior
print the `nil`,`something` is gone
## Steps to reproduce the problem

```clojure
(.println (java.lang.System/out) "something")
```

### CIDER version information
```
;; CIDER 0.15.0snapshot (package: 20161201.757), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_101
```

### Lein/Boot version

Leiningen 2.7.1 on Java 1.8.0_101 Java HotSpot(TM) 64-Bit Server VM 

### Emacs version

25.1.1

### Operating system

Mac OS X Sierra

bhagany:
This is pretty much the same issue as #1588.  I have a proof-of-concept fix here: https://github.com/bhagany/cider-nrepl/tree/repl-err.  Hopefully soon to be a PR.

bhagany:
Also, I'm currently stuck on the fact that the printed output gets printed before the prompt that causes it - that is, the prompt moves down and the output prints above it. If anyone has a solution to that, please let me know.

bhagany:
This is fixed with #387

Issue #1598, opened 2016-03-03
stderr shows up in nrepl-messages, with no id


tommy-mor:
I am using clojure quil to make drawings, but runtime errors from quil are being sent to nrepl-messages.
For example of runtime errors from quil in nrepl-messages:
http://pastebin.com/hu1eN5RM
(I think these are stdout because they have \n's in them, and when using different environments, they show up in stdout)
To me (although I am no expert) it seems like this is the correct behavior, as normal clojure errors show up in nrepl-messages aswell.
The problem is that clojure **quil errors do not have an id attribute** , so when they show up in the nrepl-messages buffer, I get
`error in process filter: nREPL: No response handler with id nil found`
This is super annoying, because everytime that message shows up, my editor lags for a few hundered milliseconds, which is incredibly annoying because when multiple errors show up it can lag for ~3 seconds.
This happens when running any type of quil project.
I do not know if the problem is because of CIDER's handling of stderr. It also might be leinigen with it's handling of stderr. I do not know. (sorry if I put this bug report on the wrong project)
## Environment & Version information

`CIDER 0.11.0snapshot`
using cider-nrepl 0.11.0snapshot aswell
using nREPL 0.2.10
this is when using a standalone repl

I also get it using the cider-jack-in command, using

```
;; CIDER 0.11.0snapshot (package: 20160229.922), nREPL 0.2.12
;; Clojure 1.7.0, Java 1.7.0_79
```

I just want the editor lags to go away, which means redirecting or stopping these idless repl messages

bbatsov:
Is this still an issue? Usually such problems are caused by some version mismatch.

Malabarba:
If it's really the case that nrepl can send messages without an ID, then we definitely shouldn't complain about it in cider. But I think it's strange that nrepl would do that.

bbatsov:
I'm pretty sure something like this should never happen, but this might be some nREPL bug.

Tommy-tf2:
Still is an issue, but I will change around versions of packages. It also may be worth noting that it is stderr.
Any ideas on how to fix this within nrepl? Is there any config settings I could change?

Tommy-tf2:
I also have already tried difserent packages,  I changed from nrepl 2.12 and 2.10

bbatsov:
Btw, can you post here the messages you see in `*nrepl-messages ...*`?

Tommy-tf2:
http://pastebin.com/hu1eN5RM
[...]

Issue #1404, opened 2015-11-09
Silent fail: no visible stacktrace upon an exception


timur87:
Hi, 

The exception occurs in a seperate thread created by a (future). Clojure version is 1.7.0. :pre :post conditions are also now shown.

No exception is shown on CIDER REPL or on nREPL server when execution fails with an exception: 

``` clojure
(defn- add-state
  [m]
  (debug "Test")
  (throw (Exception.)))
```

> 15-Nov-09 18:01:27 dyn125 DEBUG [test-prj] - Test

Some project details:
; CIDER 0.10.0snapshot (package: 20151106.2226) (Java 1.8.0_45, Clojure 1.7.0, nREPL 0.2.10 (12))

Any ideas how to fix this problem?

bbatsov:
Can you share with us your `*nrepl-messages*` log?

timur87:
Here is a simplified version of the last lines, hopefully it's so okay. There are huge maps in place of huge maps in places where I wrote huge map and some string in place of some string. After model addition the exception has been thrown with (throw (Exception.))

<addr>
(<- 
  id  "8"
  out  "15-Nov-09 22:29:44 linux-dv1t DEBUG [execution-environment-integrator] - Error sequence will be executed!! (huge map)"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
(<- 
  id  "8"
  out  "\n"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
(<- 
  id  "8"
  out  "15-Nov-09 22:29:44 linux-dv1t ERROR [execution-environment-integrator] - Error sequence will be executed!! (str)"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
(<- 
  id  "8"
  out  "\n"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
(<- 
  id  "8"
  out  "15-Nov-09 22:29:44 linux-dv1t DEBUG [execution-environment-integrator] - Adding final model state (huge map)"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
(<- 
  id  "8"
  out  "\n"
  session  "a2411d18-f78b-4f62-8e36-350df1c07768"
)
</addr>

bbatsov:
Yeah, I can reproduce this problem as well. Seems to that this is some nREPL bug/limitation. @cemerick any idea what's going on here?

cemerick:
[...]

PR #1907, opened 2017-01-05
Route stdout/stderr to different buffers based on regexps

cskksc:
Check out the very short [demo](https://www.youtube.com/watch?v=nhej5XE2THE&feature=youtu.be) (15 sec), with 

`(setq cider-redirect-output-regexps '(("INFO: .*" . "*cider-log*")))`.

If there are multiple matches, we use the first match but display a warning.

-----------------

Thanks!

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -89,7 +89,7 @@ set to `display-only' the buffer will be displayed, but it will not become
 focused.  Otherwise the buffer is displayed and focused."
   :type '(choice (const :tag "Create the buffer, but don't display it" nil)
                  (const :tag "Create and display the buffer, but don't focus it"
-                   display-only)
+                        display-only)
                  (const :tag "Create, display, and focus the buffer" t))
   :group 'cider-repl)
 
@@ -123,7 +123,7 @@ change the setting's value."
 This sets the wrap point for pretty printing on the repl.  If nil, it
 defaults to the variable `fill-column'."
   :type '(restricted-sexp  :match-alternatives
-                          (integerp 'nil))
+                           (integerp 'nil))
   :group 'cider-repl
   :package-version '(cider . "0.15.0"))
 
@@ -154,6 +154,16 @@ you'd like to use the default Emacs behavior use
   :type 'boolean
   :group 'cider-repl
   :package-version '(cider . "0.11.0"))
+
+(defcustom cider-redirect-output-regexps
+  '()
+  "A list to define mappings from regexps to buffer names.
+Output matching the regexps would be redirected to the corresponding buffers.
+For example, if the value is ((\"INFO: .*\" . \"*cider-log*\")), all output
+strings containing INFO, would be redirected to the *cider-log* buffer."
+  :type 'list
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
 
 
 ;;;; REPL buffer local variables
@@ -571,12 +581,29 @@ If BOL is non-nil insert at the beginning of line."
             (set-marker cider-repl-output-end (1- (point)))))))
     (cider-repl--show-maximum-output)))
 
+(defun cider--interactive-out-buffer (string default)
+  "Return a buffer to emit output to.
+If STRING matches with any of the regexps from `cider-redirect-output-regexps',
+return the corresponding buffer.  Otherwise, return DEFAULT."
+  (let ((buffer-candidates (seq-reduce (lambda (acc p)
+                                         (if (string-match-p (car p) string)
+                                             (cons (cdr p) acc)
+                                           acc))
+                                       cider-redirect-output-regexps '())))
+    (when (> (length buffer-candidates) 1)
+      (warn "Multiple regexps matched in cider-interactive-output-map, %s" string))
+    (or (car buffer-candidates) default)))
+
 (defun cider-repl--emit-interactive-output (string face)
   "Emit STRING as interactive output using FACE."
-  (with-current-buffer (cider-current-repl-buffer)
-    (let ((pos (cider-repl--end-of-line-before-input-start))
-          (string (replace-regexp-in-string "\n\\'" "" string)))
-      (cider-repl--emit-output-at-pos (current-buffer) string face pos t))))
+  (let ((buffer (cider--interactive-out-buffer string (cider-current-repl-buffer))))
+    (if (eq buffer (cider-current-repl-buffer))
+        (let ((pos (cider-repl--end-of-line-before-input-start))
+              (string (replace-regexp-in-string "\n\\'" "" string)))
+          (cider-repl--emit-output-at-pos (current-buffer) string face pos t))
+      (let ((output-buffer (or (get-buffer buffer) (cider-popup-buffer buffer t))))
+        (cider-emit-into-popup-buffer output-buffer string)
+        (pop-to-buffer buffer)))))
 
 (defun cider-repl-emit-interactive-stdout (string)
   "Emit STRING as interactive output."
@@ -599,8 +626,12 @@ FORMAT is a format string to compile with ARGS and display on the REPL."
 (defun cider-repl--emit-output (buffer string face &optional bol)
   "Using BUFFER, emit STRING font-locked with FACE.
 If BOL is non-nil, emit at the beginning of the line."
-  (with-current-buffer buffer
-    (cider-repl--emit-output-at-pos buffer string face cider-repl-input-start-mark bol)))
+  (let ((buf (cider--interactive-out-buffer string buffer)))
+    (if (eq buf buffer)
+        (cider-repl--emit-output-at-pos buf string face cider-repl-input-start-mark bol)
+      (let ((output-buffer (or (get-buffer buf) (cider-popup-buffer buf t))))
+        (cider-emit-into-popup-buffer output-buffer string)
+        (pop-to-buffer buf)))))
 
 (defun cider-repl-emit-stdout (buffer string)
   "Using BUFFER, emit STRING as standard output."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 * Add an option `cider-inspector-fill-frame` to control whether the cider inspector window fills its frame.
 * [#1893](https://github.com/clojure-emacs/cider/issues/1893): Add negative prefix argument to `cider-refresh` to inhibit invoking of cider-refresh-functions
 * [#1776](https://github.com/clojure-emacs/cider/issues/1776): Add new customization variable `cider-test-defining-forms` allowing new test defining forms to be recognized.
+* [#1525](https://github.com/clojure-emacs/cider/issues/1525): Add the ability to route stdout and stderr to different buffers based on regular expressions
+
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -631,14 +631,16 @@ The handler simply inserts the result value in BUFFER."
 The OUTPUT can be sent to either a dedicated output buffer or the current
 REPL buffer.  This is controlled by `cider-interactive-eval-output-destination'.
 REPL-EMIT-FUNCTION emits the OUTPUT."
-  (pcase cider-interactive-eval-output-destination
-    (`output-buffer (let ((output-buffer (or (get-buffer cider-output-buffer)
-                                             (cider-popup-buffer cider-output-buffer t))))
-                      (cider-emit-into-popup-buffer output-buffer output)
-                      (pop-to-buffer output-buffer)))
-    (`repl-buffer (funcall repl-emit-function output))
-    (_ (error "Unsupported value %s for `cider-interactive-eval-output-destination'"
-              cider-interactive-eval-output-destination))))
+  (let ((buffer (cider--interactive-out-buffer output cider-interactive-eval-output-destination)))
+    (pcase buffer
+      (`output-buffer (let ((output-buffer (or (get-buffer cider-output-buffer)
+                                               (cider-popup-buffer cider-output-buffer t))))
+                        (cider-emit-into-popup-buffer output-buffer output)
+                        (pop-to-buffer output-buffer)))
+      (`repl-buffer (funcall repl-emit-function output))
+      (_ (let ((output-buffer (or (get-buffer buffer) (cider-popup-buffer buffer t))))
+           (cider-emit-into-popup-buffer output-buffer output)
+           (pop-to-buffer buffer))))))
 
 (defun cider-emit-interactive-eval-output (output)
   "Emit OUTPUT resulting from interactive code evaluation.
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -137,6 +137,22 @@ More details can be found [here](https://github.com/clojure-emacs/cider/issues/9
 (setq cider-special-mode-truncate-lines nil)
 ```
 
+* You can redirect certain type of output to buffers instead of the REPL.
+  This is configured by adding mappings between regular expressions and
+  buffers in `cider-redirect-output-regexps`. Any output that matches any
+  regular expression will be redirected to the corresponding buffer.
+  Note that if there are multiple matches, the first match is used,
+  and a warning is displayed.
+
+  Consider this example;
+
+```el
+(setq cider-redirect-output-regexps '(("INFO: .*" . "*cider-log*")))
+```
+
+  This would cause all output strings containg "INFO" (logging output)
+  to be redirected to the `*cider-log*` buffer.
+
 ## Configuring eldoc
 
 * Enable `eldoc` in Clojure buffers:


cider-repl.el
```diff
@@ -571,12 +579,29 @@ If BOL is non-nil insert at the beginning of line."
             (set-marker cider-repl-output-end (1- (point)))))))
     (cider-repl--show-maximum-output)))
 
+(defun cider--interactive-out-buffer (string default)
```
cskksc: This is not the best name for this. Suggestions please.

cider-repl.el
```diff
@@ -155,6 +155,14 @@ you'd like to use the default Emacs behavior use
   :group 'cider-repl
   :package-version '(cider . "0.11.0"))
 
+(defcustom cider-redirect-output-regexps
+  '()
+  "A list to define mappings from regexps to buffer names.
```
bbatsov: I guess it'd be nice to give an example of the mappings in the docstring.

dpsutton:
Are we building a new logging system? I know this issue popped up on someone's project, but is this a poorly implemented logging system? Should the "fix" be to use tools.logging or timbre?

dpsutton:
Good point. I'm just worried someone will pull these projects up in another IDE/repl/whatever and just be deluged by stuff because we are hiding java.lang.System/out

cskksc:
@bbatsov Done.

@dpsutton We'll only redirect the output for the regexps users have configured. But it's true that this may be confusing and unexpected at times. Any ideas ?

bbatsov:
Ideally we should ask some users why are they logging to the console before making a decision. I really wonder why they prefer such an unconventional form of logging...

expez:
A while back I noticed that when I call `start` at the REPL I got logs in the REPL.  But when I relied on `cider-refresh` to call `start` for me I got the project's logs in `nrepl-server`.  I'm not sure why this happened, but whenever I tripped up I got an unusable REPL until I restarted the app.

I use timbre for logging and this probably happened because I had the `println` appender as part of the config (I've since removed this appender altogether).  I think that appender is part of the default configuration.  In any event, using the println appender might be something people do because you can then just start your app with `java -jar ... >> log_file 2>&1`.  I expect this pattern comes more naturally to many, instead of say using a file-appender.

In any event, I'll be glad to get a separate buffer for all my logs :)

agzam:
I feel bad. I think it was me who started complaining about this. When I switched from lein to boot, setting `:verbose` option for cljs-compiler as it turned out works differently. When working with lein everything (verbose log) would appear right in the repl buffer, with boot only warnings and errors (and the rest would go to nrepl-server buffer). It looked strange to me at that time. Now I see that I was wrong. It seems it's better that way.

bbatsov:
I'll leave this PR for a while to collect more feedback.

@cskksc Sorry for asking you to work on this when it might not get merged in the end. After @dpsutton's remark I really started doubting whether this was a good idea or not.

bbatsov:
//cc @cch1

cskksc:
@bbatsov No problem. I agree with @dpsutton that this may cause a lot of confusion.

bhagany:
fwiw, I like the redirect buffers.  While I don't quite understand logging to stdout and not wanting to see the output at the repl, I do understand getting used to that behavior once you discover it.  Anyway, `cider-redirect-output-regexps` is the kind of thing that people won't se unless they look for it, and they'll be happy when they find it.  I think they only time it will cause confusion is if they `setq` and forget, and that's true of everything in emacs.

@agzam you shouldn't feel bad.  People have been complaining about output unexpectedly going to the server buffer for at least a year, and I'd be surprised if it weren't longer (this is pretty much a longstanding nrepl bug that CIDER now works around).  Just in the boot world, this bug caused _all_ output from pods to go to the server buffer, which is how I encountered it.  And anyway, I only found your bug report on boot-cljs after #387 was already merged :).

bbatsov:
My main issue with adding this feature is that it adds a certain overhead to every output operation and it makes the pretty convoluted output handlers even more messy. If there's no compelling reason to support redirecting output then it's probably not worth doing it. On the other hand if the messy nREPL handlers were reworked in some saner manner this might be more doable. 

One other thing to consider is do you really want to redirect output from code that was evaluated in the REPL itself?

bhagany:
Good points.  I don't really see myself using this feature if it's merged - I don't want the mental overhead of remembering which output goes where.  I do have some sympathy for people who were relying on the old behavior, but if tailing a file in eshell or whatever is deemed a good enough workaround, I'm fine with that.

bbatsov:
There are also `auto-revert-tail-mode` and `itail` - I'd say they are more convenient for this purpose than using a shell. :-)

cch1:
Thanks @bbatsov for bringing me into this conversation.

I am using `tools.logging` with slf4j+log4j.  Using a console logger sends output to the STDOUT defined before Clojure starts.  I could log to a file instead of the console and `auto-revert-tail-mode`** but then behavior for non-cider uses (e.g. `lein X` from the command line) would change.  I could manage two loggers (or two profiles) -but now this is getting a convoluted.

I'm pretty happy now using `cider-jack-in` and opening a buffer on `*nrepl-server*`.  At one point I was under the impression that buffer would become unavailable on a future release -that would be painful.  The ability to redirect output to multiple buffers is not a huge deal for me as very little non-log4j-console output ends up on STDOUT.  The one regular exception, ironcially, being nrepl startup ("nREPL server started on port 64458 on host 127.0.0.1 - nrepl://127.0.0.1:64458").  The ability to segregate STDOUT and STDERR (`*nrepl-server-stdout*` and `*nrepl-server-stderr*`) would be nice...

The ability to add targeted hooks around the startup of the `*nrepl-server*` buffer would be cool.  I would hook in `log4j-mode` or `logview-mode` and `read-only-mode`, for example.  As an aside, why is the `*nrepl-server*` not read-only by default?

**in my experience logging to a file and using `auto-revert-tail-mode` is unsatisfactory.  Having a new process reopen the same file for output (e.g. restarting the REPL) causes the mode to stop working.  It's only when you realize that obvious log messages are missing that you realize you need to revisit the file manually.  PITA.

bbatsov:
> I'm pretty happy now using cider-jack-in and opening a buffer on *nrepl-server*. At one point I was under the impression that buffer would become unavailable on a future release -that would be painful.

That just happened. The old behaviour was simply the result of a bug, that was finally fixed on `master`. :-)

cch1:
Uh oh.  So what is the proposed work-around?  If this PR is the workaround, then I'm all for it.  Losing the visibility of STDOUT of the nrepl process will be painful.

bhagany:
@cch1 you won't lose visibility - the output that used to go to the server buffer will now go to your repl buffer, except for the nrepl startup messages you mentioned.

dpsutton:
@cch1 in the intermediate time while we discuss this, it is easy to restore the old behavior by running CIDER from melpa stable which will reference a version of cider-nrepl from before this behavior change.

Also, I want to summarize what your needs are to make sure I understand them, allowing you to correct any misunderstandings.

You don't particularly care where output happens, as long as there is a buffer that has all output. You are OK with stdout happening in the repl.

If the above is correct, I think you should be fine. This change doesn't prevent (to my knowledge) the output to the nrepl server messages buffer. What it does is route output additionally to all interested buffers, ie, repls.

bhagany:
> This change doesn't prevent (to my knowledge) the output to the nrepl server messages buffer. What it does is route output additionally to all interested buffers, ie, repls.

I incorrectly implied that stdout/stderr output would no longer go to the server buffer.  What @dpsutton said here is correct.

bbatsov:
> I incorrectly implied that stdout/stderr output would no longer go to the server buffer. What @dpsutton said here is correct.

Hmm, why would it still go there? I thought your patch redirected all related IO streams. In general I don't think anything should go to the nrepl-server buffer - it exists just because this is how the Emacs process spawning facilities work. Unless some output came straight from leiningen (or whatever was used to start the nrepl-server) it doesn't belong there.

bhagany:
To be honest, I don't really know why output still goes to the server buffer.  When I wrote it, I thought I was completely redirecting the java streams, and it surprised me that it works the way it does.  I didn't investigate much, but I can if it's important to you.

bbatsov:
> I didn't investigate much, but I can if it's important to you.

It's not super important, I just really don't want people to rely on something that's an implementation glitch. As this bug has been around since day one many people thought it was some feature. :-)

bhagany:
Alright, I already have a few theories.  I should have some time this weekend.

bhagany:
Well, it turned out to be the simplest thing: https://github.com/clojure-emacs/cider-nrepl/pull/388

dpsutton:
From @cch1 

>  If this PR is the workaround, then I'm all for it. Losing the visibility of STDOUT of the nrepl process will be painful. 

Do we maybe want to think about that pull request before merging it? I think we may be removing their relied upon behavior for a bugfix that no one is really in a rush for. I agree that its the correct thing to do, but in time and not immediately. thoughts?

bhagany:
I'm fine either way

bbatsov:
> Do we maybe want to think about that pull request before merging it? I think we may be removing their relied upon behavior for a bugfix that no one is really in a rush for. I agree that its the correct thing to do, but in time and not immediately. thoughts?

I'll think about it, but this is definitely going to get merged sooner or later. Postponing this without thinking of some alternative log handling won't serve anyone.

cch1:
I wish this were not viewed as just a hack for log handling.  What other early-loaded Java code binds to STDOUT/STDERR?  They too will be affected.

bbatsov:
> I wish this were not viewed as just a hack for log handling. What other early-loaded Java code binds to STDOUT/STDERR? They too will be affected.

Well, this output is not going to be visible anywhere unless you're using `jack-in`. What happens when using `cider-connect` and there's no process buffer then? That output normally just gets lost. Seems reasonable to just treat all the output in a consistent manner.

cch1:
So the principle is going to be that `cider-jack-in` swallows STDOUT?  Today if I run `cider-jack-in` or `cider-connect`, I am able to see STDOUT (one in emacs, the other from the shell).  Now `cider-jack-in` will be unusable.  :(

bhagany:
> Well, this output is not going to be visible anywhere unless you're using jack-in. What happens when using cider-connect and there's no process buffer then? That output normally just gets lost.

stdout and stderr used to go to whatever process you're `cider-connect`ing to.  This could be a terminal window somewhere, and I think it would only be lost if the process is backgrounded.

cch1:
> stdout and stderr used to go to whatever process you're cider-connecting to. This could be a terminal window somewhere, and I think it would only be lost if the process is backgrounded.

Exactly.

bhagany:
> Now cider-jack-in will be unusable. :(

I don't understand why?

cch1:
@bhagany IF STDOUT is swallowed by the changes recently made on master (per @bbatsov 's comment above) then `cider-jack-in` will be unusable for me.  Assuming your PR can direct STDOUT from the jacked-in nrepl process to a buffer based on a regexp, it would restore the usability

bbatsov:
> stdout and stderr used to go to whatever process you're cider-connecting to. This could be a terminal window somewhere, and I think it would only be lost if the process is backgrounded.

I meant it's not available in Emacs and if you don't redirect it there you won't see it.

bhagany:
> IF STDOUT is swallowed by the changes recently made on master (per @bbatsov 's comment above) then cider-jack-in will be unusable for me.

I'm not sure what comment you're referring to here, but in any case, all of the output goes to the repl buffer.  It's not swallowed.

bhagany:
> I meant it's not available in Emacs and if you don't redirect it there you won't see it.

Ah, okay, got it.

expez:
It seems we're still debating whether this is a good idea or not, and I'd like to raise the point that for certain apps it's a huge problem when you're doing dev work at e.g. `trace` or `debug` presently.  When everything is dumped into the REPL you lose your prompt (due to scrolling or a scrolling related bug) and am unable to interact with the repl.  

Furthermore, any interactions, while your app is trace logging in the background, become potentially unreadable, and you certainly can't interact with the result (e.g. to copy paste) because it will scroll of the screen too fast.

The last time I got into this unfortunate situation I had a hard time recovering my REPL even after stopping the app.  The prompt had disappeared altogether and I was left looking for a point in the buffer where I could hit <kbd>RET</kbd> to restore it.  Is it the last line?  At the end of the last line?  The second to last line?  I can't remember the answer anymore, but I did get my prompt back, eventually.

bbatsov:
It's certain that this feature has certain usefulness. As I said before - my main issue with it is the intrusive implementation right now (mandated by the poor design of the nREPL response handling code). We spoke with @cskksc in private and decide it's best to focus on improving the design of the nREPL handlers before trying to solve the output redirection problem.

jcf:
> It's certain that this feature has certain usefulness. As I said before - my main issue with it is the intrusive implementation right now (mandated by the poor design of the nREPL response handling code). We spoke with @cskksc in private and decide it's best to focus on improving the design of the nREPL handlers before trying to solve the output redirection problem.

@bbatsov can you revert the change in master to bring back the old behaviour of logging to the nREPL server buffer?

The "fix" has broken Emacs for me when log output gets large (I've got a 5.8M REPL buffer that causes Emacs to hang every time I interact with it), and I have to manually clear my buffer losing interesting information every time.

I have log rotation via logback, and can send things to both stdout and files just fine, but with what's in master Emacs is now performing pretty badly.

When nREPL design is improved maybe change the way logging works in Cider?

bbatsov:
> @bbatsov can you revert the change in master to bring back the old behaviour of logging to the nREPL server buffer?

That's not going to happen. I think that the benefits of properly routing the output outweigh the downsides - after all logging to stdout can always be disabled. Fixing the REPLs performance when the text there gets a lot is something else that we should eventually address. 

@cskksc is now working on improvements that should make the stdout redirect idea feasible, so your inconvenience should be temporary. 

> When nREPL design is improved maybe change the way logging works in Cider?

It's unlikely that this is going to happen any time soon. nREPL has seen pretty much no changes in quite a lot of time and I don't see a lot of volunteers to work on improving it.

bbatsov:
Btw, applying this patch locally or going back to CIDER 0.14 is also an option for the time being.

cch1:
> the benefits of properly routing the output outweigh the downsides

What are the benefits?

bhagany:
> the old behaviour of logging to the nREPL server buffer

I'd like to address this, because there seems to be some misapprehension about it.  The issue that's fixed in master isn't about logging.  CIDER doesn't know anything about logging in particular. The fix is about output streams.  To illustrate,

```clojure 
user> (.println java.lang.System/out "foo")
```

If you expect that to print something at the repl, you'd be disappointed if you're using a stable CIDER release.  This is what the fix is aimed at.  CIDER did this with `System/err` as well, like with exceptions,

```clojure
user> (try
  (throw (Exception. "hi"))
  (catch Exception e
    (.printStackTrace e)))
```

The upshot here is that Java stuff couldn't print to your repl before, and that caused lots of problems that people can't fix themselves.  In contrast, redirecting logs is a matter of configuration.

cch1:
@bhagany , when you say "print to the repl", I've never wanted/asked for/seen that behavior. In my experience, `(.println java.lang.System/out "foo")` sends output to my `*nrepl-server*` buffer, not to my *cider-repl* buffer.  

This is what I want and what I fear will be taken away be the changes @bbatsov referenced above.

Maybe I am confused as to what is actually being changed... I hope so.  Without the ability to have java's `java.lang.System/out` in a buffer somewhere, I'm blind.

dpsutton:
@cch1

bhagany:
> when you say "print to the repl", I've never wanted/asked for/seen that behavior

Plenty of others do want it and have asked for it\*, and they are justified in doing so.  You have workarounds for your problem, whereas the people on the other side of this are stuck if it doesn't change.  As for seeing it, try it in `lein repl`, and I would guess, many other repls that people use regularly.

I don't understand why `(.println java.lang.System/out "foo")` having a different behavior from `(println "foo")` should be the default behavior.  It's certainly not intuitive that it should be this way.

> Without the ability to have java's java.lang.System/out in a buffer somewhere, I'm blind.

It seems to me that `*cider-repl*` counts as "a buffer somewhere", but in any case, this PR is being designed to accommodate your use case.  It's being worked on, and you'll be able to put your log messages in the buffer of your choosing, assuming there's a regex that matches them.

\* https://github.com/clojure-emacs/cider/issues/1896
https://github.com/clojure-emacs/cider/issues/1598
https://github.com/clojure-emacs/cider/issues/1404
https://github.com/clojure-emacs/cider-nrepl/issues/381
https://github.com/boot-clj/boot-cljs/issues/138
https://github.com/boot-clj/boot/issues/544

cch1:
> This output goes to the nrepl-server buffer. It now (as of a recent patch) also goes to the cider buffer.

Thanks @dpsutton , for clarifying the issues involved here.  I honestly did not completely understand them before.

The patch behavior seems to introduce redundancy.  What's the point of even having the *nrepl-server* buffer if its content is duplicated in *cider-nrepl*?

When using Cider to connect to programs emitting to STDOUT as a general practice (CLI tools, etc) it seems like the nrepl dialog is going to be conflated with STDOUT.

If the fear is that people aren't looking at `*nrepl-server*` and thus need to be made aware of the their own program's STDOUT, it seems like an easy answer: open the `*nrepl-server*` buffer.   Is that unreasonable?

(Now just reading @bhagany 's comment)  Before I respond, please know that I appreciate everyone's efforts here.  And the prospect of an eventual fix to route based on regex seems like it might address my use case nicely.  Now responding to your particular points:

I will have an unpleasant workaround to my particular problem.  But don't others have an even easier workaround: open the `*nrepl-server*` buffer.  Am I missing something?

When I run `lein repl` in a terminal, I know I only have one output stream (unless I redirect STDOUT to a file).  I expect more from a windowing/multibuffer environment like emacs.

My position: STDOUT is not the same as the nrepl dialog unless I'm in a constrained environment.

bhagany:
> The patch behavior seems to introduce redundancy.

The redundancy is resolved here: https://github.com/clojure-emacs/cider-nrepl/pull/388.

> When using Cider to connect to programs emitting to STDOUT as a general practice (CLI tools, etc) it seems like the nrepl dialog is going to be conflated with STDOUT.

nrepl dialog _is_ stdout.

> Is that unreasonable?

In my opinion, yes.  In the boot ecosystem in particular, the output that is split between buffers is extremely confusing, because parts of it go through a Java intermediary.

> I will have an unpleasant workaround to my particular problem.

Why is it unpleasant?  It's a regex that, as far as I can tell, meets your wants completely.  That's less unpleasant than not working without any options to fix it, which is what you're advocating I deal with for my use case.

dpsutton:
> When I run lein repl in a terminal, I know I only have one output stream (unless I redirect STDOUT to a file). I expect more from a windowing/multibuffer environment like emacs.

That's one way to look at it. Another is to ensure that repl's meet a standard. I would hate to work on a project with someone who had a suppressing repl and I was subjected to lots of messages. This makes me not want to make CIDER's repl suppress things that may bother the rest of the ecosystem. Also why I'm not super sold on the notion of filtering output from the repl buffer, this patch.

bbatsov:
I've said this numerous times and I'm tired of repeating it - the old behavior was always considered a bug by me. Some output got bound early and ended up in a buffer **dedicated exclusively** for the process that spawned the nREPL server - e.g. `lein`, `boot`, whatever. People were warned not to rely on this and yet they did. We're considering the implementation of the redirect feature just for the sake of such users, although in principle I agree with @dpsutton that re-routing output in this manner is probably not a good idea in general.

Btw, we already have a feature to route the output of interactive evaluations (those done in source buffers) to an output buffer instead of to the REPL - probably people unhappy about the recent changes should find it useful. I don't recall whether we had something similar for the REPL itself.

bbatsov:
I guess we can close this for now.

cch1:
> Btw, we already have a feature to route the output of interactive evaluations (those done in source buffers) to an output buffer instead of to the REPL - probably people unhappy about the recent changes should find it useful. I don't recall whether we had something similar for the REPL itself.

I remain frustrated by console logging being intermingled with REPL output.  I realize this ship has already sailed... but can you summarize the current options _other than switching to logging to a file_? 
```
;; CIDER 0.21.0 (New York), nREPL 0.6.0
;; Clojure 1.10.0, Java 1.8.0_162
```

bbatsov:
> I remain frustrated by console logging being intermingled with REPL output. I realize this ship has already sailed... but can you summarize the current options other than switching to logging to a file?

Well, I wouldn't say that ship has sailed - it's just nobody wanted to tackle the flexible routing idea. I still believe that's the best course of action and I hope at some point we'll revisit this.

The old behaviour (where some things would end up in the server's buffer is still available) it's just not the default anymore. Just set `cider-redirect-server-output-to-repl` to nil and you're done.

cch1:
> The old behaviour (where some things would end up in the server's buffer is still available) it's just not the default anymore. Just set `cider-redirect-server-output-to-repl` to nil and you're done.

That worked perfectly.  Thanks!

Issue #1909, opened 2017-01-05
Fuzzy completion woes


alexander-yakushev:
So, I've been using company together with company-quickhelp for a while now. For the most part, I'm pretty content with how everything works, except for fuzzy completion. I've tried the `company-flx` route as described in the docs, but it doesn't solve the problem:

- `company-flx` works by polling the backend with empty prefix (`""`) and then performing the filtration on the client-side. This does a poor job with Compliment since the latter doesn't spit out _absolutely all_ candidates when queried with `""`. E.g. a regular project has tens of thousands of classes loaded, and we can't just return all of them. Thus, Compliment actually expects at least something in the query to start looking into heavyweight sources like classes and methods.
- `company-flx` is *too fuzzy*, using very wide regexp to filter candidates. Even despite the intelligent ranking algorithm, I am quite confused by some of the suggestions it offers. This may be subjective, of course.

I could let some of the missing fuzziness go, if not for [simple classname completion](http://bytopia.org/2015/03/06/smart-classname-completion-for/). I do plenty of interop these days, and it is absolutely painful having to Google that to import `ProgressEvent` I actually have to write `com.amazonaws.services.s3.model.ProgressEvent`.

Long story short, I've made a blunt hack _for myself_.

```elisp
(defun cider-company-capf-advice (old-fun &rest args)
  (let ((completion-styles (list 'cider)))
    (apply old-fun args)))

(defun cider-company-completions (string table predicate point)
  (cider-complete string))

(advice-add 'company-capf :around #'cider-company-capf-advice)
(add-to-list 'completion-styles-alist
             '(cider
               cider-company-completions
               cider-company-completions
               "CIDER backend-driven completion style."))
```

Basically, I add an adivce to `company-capf` that forces it to skip all its machinery and call straight to `cider-complete`. This works for me because I don't use company for anything except CIDER — but I can imagine a solution like this breaks all other company backends.

I create this issue with the intent to discuss the following points:

- Is fuzzy completion of any value? I vote yes, especially for the classname example above. That's certainly Cursive's strong suite.
- Should we continue promoting `company-flx`? I don't feel like it offers good UX for CIDER particularly.
- Can the advice hack be improved and embedded into CIDER by default? From what I see we don't need more than UI from company/standard Emacs completion mechanisms, so should we limit our functionality because of aribitrary restrictions in their "logic" part?

expez:
[...]

PR #1910, opened 2017-01-06
Add company-cider backend

alexander-yakushev:
#1909 follow-up.

This patch adds `company-cider` backend that for the most part calls directly to the already existing CIDER functions. An update to docs was also made to describe how to enable the backend.

I've made a function that sets `company-cider` a sole completion backend. Is it OK? Maybe it is enough to drop `company-capf` from the list?

Regarding the docs, I also request comments. I took the liberty to drop the mentions of company-flx, but I feel there is a better way to explain the new setup.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -600,6 +600,22 @@ in the buffer."
                                         (cider-eldoc-thing-type eldoc-info))
               (cider-eldoc-format-arglist arglists 0)))))
 
+;; Fuzzy completion for company-mode
+
+(defun cider-company-unfiltered-candidates (string table predicate point)
+  "Return CIDER completion candidates as is, without filtering them by prefix."
+  (cider-complete string))
+
+(add-to-list 'completion-styles-alist
+             '(cider
+               cider-company-unfiltered-candidates
+               cider-company-unfiltered-candidates
+               "CIDER backend-driven completion style."))
+
+(defun cider-company-enable-fuzzy-completion ()
+  "Enable backend-driven fuzzy completion in the current buffer."
+  (setq-local completion-styles '(cider)))
+
 (defun cider-stdin-handler (&optional buffer)
   "Make a stdin response handler for BUFFER."
   (nrepl-make-response-handler (or buffer (current-buffer))
--- a/doc/code_completion.md
+++ b/doc/code_completion.md
@@ -55,13 +55,19 @@ the time of this writing it's still in development).
 By default `company-mode` will provide completion candidates with the assumption
 that whatever you've typed so far (e.g. `map-`) is a completion prefix (meaning
 you'd get only candidates that have `map-` in the beginnings of their names).
+You can get enhanced fuzzy completion with the CIDER-specific completion style
+by adding:
 
-You can get enhanced fuzzy completion with `company-mode` if you install the
-additional package [company-flx](https://github.com/PythonNut/company-flx).
-This is powered internally by [flx](https://github.com/lewang/flx).
+```el
+(add-hook 'cider-repl-mode-hook #'cider-company-enable-fuzzy-completion)
+(add-hook 'cider-mode-hook #'cider-company-enable-fuzzy-completion)
+```
 
-Basically with `company-flx` enabled typing something like `mp` will show you
-`map-indexed` as one of the possible completion candidates.
+Now `company-mode` will accept certain fuzziness when matching candidates
+against the prefix. For example, typing `mp` will show you `map-indexed` as one
+of the possible completion candidates, `cji` will complete to `clojure.java.io`,
+etc. Different completion examples are
+listed [here](https://github.com/alexander-yakushev/compliment/wiki/Examples).
 
 ### Completion annotations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add new customization variable `cider-font-lock-reader-conditionals` which toggles syntax highlighting of reader conditional expressions based on the buffer connection.
 * Add new face `cider-reader-conditional-face` which is used to mark unused reader conditional expressions.
 * [#1544](https://github.com/clojure-emacs/cider/issues/1544): Add a new defcustom `nrepl-use-ssh-fallback-for-remote-hosts` to control the behavior of `nrepl-connect` (and in turn that of `cider-connect`) for remote hosts.
+* [#1910](https://github.com/clojure-emacs/cider/issues/1910): Add custom company-mode completion style to show fuzzy completions from Compliment.
 
 ### Changes
 


bbatsov:
> I've made a function that sets company-cider a sole completion backend. Is it OK? Maybe it is enough to drop company-capf from the list?

I think that'd be better. Some people might be using a mixture of backends and this would affect them.

doc/code_completion.md
```diff
@@ -30,8 +30,15 @@ After installation, company can be turned on  globally, like so --
 (add-hook 'cider-mode-hook #'company-mode)
 ```
 
+To enable CIDER-specific company backend, add the following hooks:
```
bbatsov: a CIDER-specific

use the following code (you're actually adding hook functions to the hooks)

doc/code_completion.md
```diff
@@ -30,8 +30,15 @@ After installation, company can be turned on  globally, like so --
...
+
+```el
+(add-hook 'cider-repl-mode-hook #'company-cider-make-exclusive)
+(add-hook 'cider-mode-hook #'company-cider-make-exclusive)
+```
+
 When `company-mode` is thus enabled, it will receive completion information
```
bbatsov: I'd keep the original paragraph before this code and add some explanations about the different behaviour afterwards - with a link to "Fuzzy candidate matching" (or you can just move that section here).

cider-interaction.el
```diff
@@ -600,6 +600,22 @@ in the buffer."
                                         (cider-eldoc-thing-type eldoc-info))
               (cider-eldoc-format-arglist arglists 0)))))
 
+(defun company-cider (command &optional arg &rest ignored)
+  "`company-mode' completion backend for CIDER."
+  (interactive (list 'interactive))
+  (cl-case command
```
bbatsov: Why not use `pcase` here?

bbatsov:
Looks good to me overall. Apart from my small remarks you should mention this in the changelog.

cider-interaction.el
alexander-yakushev: Mindlessly copied from some other backend. Point taken.

alexander-yakushev:
I would like anyone to try this out to verify it doesn't only work on my setup:).

Will rebase when ready.

bbatsov:
OK, I'll try to source you some testers from Twitter as I'm short on time to testing myself right now.

xiongtx:
If `company-backends` is turned into a local variable, wouldn't user modifications of it apply to only a single buffer? If the user tries to change `company-backends` in a CIDER buffer, it won't apply to other buffers, even other CIDER buffers.

Might a [*mode*-local variable](https://www.emacswiki.org/emacs/ModeLocal) be a better idea? `mode-local.el` is a part of GNU Emacs.

alexander-yakushev:
To be honest, I've stolen the idea of a buffer-local variable from company-eclim. If you say that it is better to make the variable mode-local, so be it.

alexander-yakushev:
@xiongtx Is this what you have in mind?

dgutov:
> mode-local.el is a part of GNU Emacs

It's part of CEDET, and not used too widely.

On the other hand, changing the buffer-local value of `company-backends` is a perfectly common practice.

xiongtx:
I'll defer to the author of `company` on how `company-backends` should be overridden 🙂 .

alexander-yakushev:
Thanks for chiming in, Dmitry!

dgutov:
Not a problem. But also see the last comment in the original issue. If what I'm suggesting is right, it should lead to a smaller, less disruptive change.

bbatsov:
@alexander-yakushev Seems no one volunteered to test this, but we can merge this non-the-less. Squash and rebase and ping me when this is good to go.

doc/code_completion.md
```diff
@@ -54,14 +54,20 @@ the time of this writing it's still in development).
...
-you'd get only candidates that have `map-` in the beginnings of their names).
+you'd get only candidates that have `map-` in the beginnings of their
+names). You can get enhanced fuzzy completion by enabling the CIDER-specific
+company backend in your config:
 
-You can get enhanced fuzzy completion with `company-mode` if you install the
-additional package [company-flx](https://github.com/PythonNut/company-flx).
-This is powered internally by [flx](https://github.com/lewang/flx).
+```el
+(add-hook 'cider-repl-mode-hook #'company-cider-enable-backend)
```
julienfantin: That's should be `company-cider-enable-backend`

julienfantin:
This is not working for me.

I tried the exclusive config to make sure nothing else was interfering in my setup but I did not get any completion results nor errors?

stardiviner:
I have tested the code, it works. But will break the behavior of mixed backends users. Like grouped backends with `company-yasnippet`. So I suggest a better solution. Or just don't override the `company-backends` by default, let user decide. Like me, I will mix backend `company-yasnippet` with function:
```elisp
(defun my-company-add-backend-locally (backend)
    "Add a backend in my custom way.

\(my-company-add-backend-locally 'company-robe\)
"
    (if (local-variable-if-set-p 'company-backends)
        (add-to-list 'company-backends `(,backend :with company-yasnippet))
      (add-to-list (make-local-variable 'company-backends)
                   `(,backend :with company-yasnippet))
      ))
```

alexander-yakushev:
@stardiviner Apparently, @dgutov was correct after all (to nobody's surprise) suggesting to use a custom completion style rather than a custom backend. I have rewritten the PR. Could you please test if this solution doesn't break your setup?

@julienfantin Thanks for pointing out a mistake! Could you please try if the current version works for you?

dgutov:
Sorry, right about what? I never said a custom backend won't work with fuzzy-matching. There are such third-party backends for other languages already.

I suggested the completion-styles approach because it takes less code, and because it might work with `C-M-i` as well. But nobody has tested that hupothesis, so far.

alexander-yakushev:
Correct about it being a "better" way, if it actually will help stardiviner's usecase.

`C-M-i` seems to work fuzzily for me now.

dgutov:
> C-M-i seems to work fuzzily for me now.

Cool. This UI has a few peculiar behaviors, such as moving point to the position where disambiguation is required, which probably won't work with it. But still, having the basics work is a good result.

cider-interaction.el
```diff
@@ -600,6 +600,22 @@ in the buffer."
                                         (cider-eldoc-thing-type eldoc-info))
               (cider-eldoc-format-arglist arglists 0)))))
 
+;; Fuzzy completion for company-mode
+
+(defun cider-company-compliment-candidates (string table predicate point)
+  "Return Compliment candidates as is, without filtering them."
```
bbatsov: Just one thing about the use of compliment in the names - this calls into the middleware which uses compliment for clj connections and cljs-tooling for cljs connections. Don't think cljs-tooling supports the fuzzy matching, so some mentions of this might be useful.

bbatsov:
Looks OK to me. I wonder if we shouldn't move all the completion stuff to a dedicated file as there's already quite a lot of completions-related code.

cider-interaction.el
```diff
@@ -600,6 +600,22 @@ in the buffer."
...
+  (cider-complete string))
+
+(add-to-list 'completion-styles-alist
+             '(cider
+               cider-company-compliment-candidates
+               cider-company-compliment-candidates
+               "CIDER backend-driven completion style."))
+
+(defun cider-company-enable-fuzzy-completion ()
```
bbatsov: There should be a matching disable command I guess.

stardiviner:
@alexander-yakushev It works with the `completion-styles`.

cider-interaction.el
alexander-yakushev: This is supposed to be put on a hook and enables itself per-buffer. Does it still warrant a disable?

bbatsov: Well, it's not a big deal but I can imagine some peoople will use this without a hook and it's generally a good idea to have a way to undo the effect of enabling this.

bbatsov:
Guess this is in a good enough shape now. Thanks, @alexander-yakushev!

PR #1912, opened 2017-01-09
Reader conditionals font lock

alvinfrancis:
This PR adds syntax highlighting to non-matching platform reader conditionals (with respect to the buffer connection type) in the same vein as SLIME does with Common Lisp feature expressions.  By default, this feature checks the connection type of the buffer and grays out the code that will be skipped for that connection.

This adds the following user facing customisations:
* `cider-font-lock-reader-conditionals` which toggles this feature.  Defaults to true.
* `cider-reader-conditional-face` which is the face used to mark unused reader conditional expressions.  Defaults to inheriting from `font-lock-comment-face`.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * [#1893](https://github.com/clojure-emacs/cider/issues/1893): Add negative prefix argument to `cider-refresh` to inhibit invoking of cider-refresh-functions
 * [#1776](https://github.com/clojure-emacs/cider/issues/1776): Add new customization variable `cider-test-defining-forms` allowing new test defining forms to be recognized.
 * [#1860](https://github.com/clojure-emacs/cider/issues/1860): Add `cider-repl-history` to browse the REPL input history and insert elements from it into the REPL buffer.
+* Add new customization variable `cider-font-lock-reader-conditionals` which toggles syntax highlighting of reader conditional expressions based on the buffer connection.
+* Add new face `cider-reader-conditional-face` which is used to mark unused reader conditional expressions.
 
 ### Changes
 
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -406,6 +406,12 @@ The value can also be t, which means to font-lock as much as possible."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-font-lock-reader-conditionals t
+  "Apply font-locking to unused reader conditional expressions depending on the buffer CIDER connection type."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
+
 (defface cider-deprecated-face
   '((((background light)) :background "light goldenrod")
     (((background dark)) :background "#432"))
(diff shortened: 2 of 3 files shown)


dpsutton:
It looks like the tests fail on emacs 24.x because `font-lock-ensure` doesn't exist until emacs 25.x

cider-mode.el
```diff
@@ -426,6 +432,98 @@ The value can also be t, which means to font-lock as much as possible."
...
+
+(defun cider--anchored-search-suppressed-forms-internal (limit)
+  "Helper function for `cider--anchored-search-suppressed-forms`.
+
+LIMIT is the same as the LIMIT in `cider--anchored-search-suppressed-forms`"
+  (let ((expr (read (current-buffer)))
+        (start (save-excursion (backward-sexp) (point))))
+    (when (<= (point) limit)
+      (forward-sexp)
+      (if (not (string-equal (symbol-name expr) (concat ":" (cider-connection-type-for-buffer))))
```
dpsutton: I see you are `(read (current-buffer))` and then calling `(symbol-name expr)` on it. This is reading clojure code into the emacs lisp reader and then calling symbol name yes? Is this a bit dangerous?

dpsutton: I accidentally hit the review button. Did not mean to make this a "review" I was just looking over your code sorry.

test/cider-font-lock-tests.el
```diff
@@ -0,0 +1,104 @@
...
+
+(defmacro test-with-temp-buffer (content &rest body)
+  "Evaluate BODY in a temporary buffer with CONTENT."
+  (declare (debug t)
+           (indent 1))
+  `(with-temp-buffer
+     (insert ,content)
+     (clojure-mode)
+     (cider-mode)
+     (font-lock-ensure)
```
dpsutton: Unfortunately this function `font-lock-ensure` is emacs 25+

dpsutton:
I cranked up a cljc buffer and jacked in clojurescript so that I had a clojurescript and clojure repl attached to the buffer. You rely on `cider-connection-type-for-buffer` but this reported "clj" for the cljc buffer. Is there a way we can modify this so that it knows that both reader conditionals should have highlights? Or what would be your thoughts in this situation?

cider-mode.el
```diff
@@ -426,6 +432,98 @@ The value can also be t, which means to font-lock as much as possible."
...
+  (when (and cider-font-lock-reader-conditionals
+             (cider-connected-p))
+    (when (search-forward-regexp cider-reader-conditionals-regexp limit t)
+      (let ((start (match-beginning 0))
+            (state (syntax-ppss)))
+        (if (or (nth 3 state) (nth 4 state)) ; inside string or comment?
+            (cider--search-reader-conditionals limit)
+          (when (<= (point) limit)
+            (ignore-errors
+              (cl-assert (<= (point) limit))
```
dpsutton: What's the point of this assert inside of the `when` form with the same check?

dpsutton:
PS, I really like the idea.

Just a question for you: why are the match data results put into global vars rather than returned from the functions?

cider-mode.el
alvinfrancis: Yes, that's what it is doing. This should be safe since `read` doesn't `eval` the code.  Same goes for `symbol-name`.

test/cider-font-lock-tests.el
alvinfrancis: Yeah, didn't test for versions under 25. In lieu of that though, it should be enough to use `font-lock-fontify-buffer` instead. I'll make the changes.

cider-mode.el
alvinfrancis: Oops. That shouldn't be there anymore. It's an artefact of the original SLIME code when I was porting it. I'll remove it.

alvinfrancis:
First off, thanks @dpsutton for the review.  It's much appreciated.

The `match-data` is just a minor performance thing. It uses the global vars just to prevent constructing new lists.

As for the `cljc` thing, it works if you change modes in between `clojure-mode` and `clojurescript-mode` as this updates the connection appropriately.  Ideally, using `cider-toggle-buffer-connection` should also work but I've been having problems with that command.  It seems to be bugged right now.  I think it would be best if that was fixed and `clojurec-mode` relies on that instead.

cider-mode.el
dpsutton: I ran read on some clojure code and it was usually fine with the exception of the clj/cljs reader macros. Can we guarantee that we won't read any of these reader macros?

dpsutton:
> Ideally, using cider-toggle-buffer-connection should also work but I've been having problems with that command. It seems to be bugged right now. It actually seems to drop known connections. I think it would be best if that was fixed and clojurec-mode relies on that instead.

If you want to file an issue on this I can start working on it.

dpsutton:
Oh, PS, your tests look super nice.

alvinfrancis:
> If you want to file an issue on this I can start working on it.

Ok. I'll gather what information I can on it and open an issue.

Thanks again!

cider-mode.el
alvinfrancis: It can read those and it will throw an `invalid-read-syntax` but that error is expected and handled by a `condition-case` in the main anchored matcher.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Add an option `cider-inspector-fill-frame` to control whether the cider inspector window fills its frame.
 * [#1893](https://github.com/clojure-emacs/cider/issues/1893): Add negative prefix argument to `cider-refresh` to inhibit invoking of cider-refresh-functions
 * [#1776](https://github.com/clojure-emacs/cider/issues/1776): Add new customization variable `cider-test-defining-forms` allowing new test defining forms to be recognized.
+* [#1893](https://github.com/clojure-emacs/cider/issues/1893)Add negative prefix argument to `cider-refresh` to inhibit invoking of cider-refresh-functions
```
bbatsov: This looks unrelated.

cider-mode.el
```diff
@@ -426,6 +432,97 @@ The value can also be t, which means to font-lock as much as possible."
...
+  '((t (:inherit font-lock-comment-face)))
+  "Face used to mark unused reader conditional expressions."
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
+
+(defconst cider-reader-conditionals-regexp "\\(?:#\\?@?[[:space:]\n]*(\\)")
+
+(defvar cider--search-reader-conditionals-match-data (list nil nil))
+
+(defvar cider--anchored-search-suppressed-forms-match-data (list nil nil))
```
bbatsov: The constant and the vars should ideally have some docstrings as well.

cider-mode.el
bbatsov: Unfortunately to my knowledge there's no easy way to replace the EL reader with something else. There's `edn.el`, but it brings along dependencies that I'd rather not have around. Parsing properly Clojure expressions in Emacs is a no go in the moment...

CHANGELOG.md
alvinfrancis: My bad. Must have had a bad merge or rebase at some point. Will remove that.

cider-mode.el
alvinfrancis: Adding docstrings.  And looking at it again makes me realise I could/should condense the match-data var to just one.

cider-mode.el
alvinfrancis: In this case, the EL reader is enough since the only desired functionality is parsing a keyword.  It's enough to catch `invalid-read-syntax` errors. That would mean that the thing read doesn't match (is not a keyword) and the matcher should just skip them.

bbatsov:
> As for the cljc thing, it works if you change modes in between clojure-mode and clojurescript-mode as this updates the connection appropriately. Ideally, using cider-toggle-buffer-connection should also work but I've been having problems with that command. It seems to be bugged right now. It actually seems to drop known connections. I think it would be best if that was fixed and clojurec-mode relies on that instead.

Some bug report would be useful indeed.

Btw, the new face should be added to popular themes like https://github.com/bbatsov/zenburn-emacs and https://github.com/bbatsov/solarized-emacs

Can you post a screenshot of the default appearance of the unused reader conditionals? (don't have time to test this manually myself right now)

bbatsov:
P.S. You'll also have to rebase this branch on top of the current `master`.

alvinfrancis:
> Some bug report would be useful indeed.

I've raised the issue here: #1913 

> Btw, the new face should be added to popular themes like https://github.com/bbatsov/zenburn-emacs and https://github.com/bbatsov/solarized-emacs

The new face, by default, inherits from `font-lock-comment-face`.  I figured that would be sufficient.  Do you figure the new face should still be added to existing themes?

> Can you post a screenshot of the default appearance of the unused reader conditionals? (don't have time to test this manually myself right now)

Here are some screenshots.

screen shot 2017-01-12 at 2 13 50 pm
screen shot 2017-01-12 at 2 12 27 pm

screen shot 2017-01-12 at 2 22 37 pm
screen shot 2017-01-12 at 2 22 08 pm

> P.S. You'll also have to rebase this branch on top of the current master.

Okay. I'll take the opportunity to squash the commits.

bbatsov:
> The new face, by default, inherits from font-lock-comment-face. I figured that would be sufficient. Do you figure the new face should still be added to existing themes?

In your theme this is ok and it's a decent default, but I'd definitely pick some shade of gray for both mentioned themes.

bbatsov:
Anyways, this does not prevent us to merge this, so more people can play with it! Thanks for the great contribution!

PR #1922, opened 2017-01-16
Require cider-classpath

johnv02139:
Require `cider-classpath` in cider-repl.el.

This file accesses or references the cider-classpath functionality, but it's not required by this file, or by any file that requires this one.  I found that with a vanilla setup, doing `,` and then `classpath` did not work, without explicitly the explicit `require`.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -40,6 +40,7 @@
 (require 'cider-compat)
 (require 'cider-util)
 (require 'cider-resolve)
+(require 'cider-classpath)
 
 (require 'clojure-mode)
 (require 'easymenu)


bbatsov:
The proper solution would be to use `declare-function` for the problematic function. This was done on purpose to avoid loading the classpath functionality at CIDER startup time.

bbatsov:
I see the `declare-function` is there and that the common in question is autoloaded. Can't imagine why this is not working for you...

johnv02139:
OK.  I figured it might have been intentional, but this seemed like the fastest way to ask.  I'll try to figure out why it doesn't work for me, and open up a bug if I can't get it.

PR #1923, opened 2017-01-17
Add subsecond-resolution timestamps to *nrepl-messages*

dottedmag:
This helps in debugging unexpected slowness and pinpointing the
offender.

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1099,6 +1099,7 @@ described by `nrepl-message-buffer-name-template'."
         (re-search-forward "^(" nil t)
         (delete-region (point-min) (- (point) 1)))
       (goto-char (point-max))
+      (insert (format-time-string "%Y-%m-%0d %H:%M:%S.%N\n"))
       (nrepl-log-pp-object (nrepl-decorate-msg msg type)
                            (nrepl-log--message-color (lax-plist-get (cdr msg) "id"))
                            t)


nrepl-client.el
```diff
@@ -1099,6 +1099,7 @@ described by `nrepl-message-buffer-name-template'."
         (re-search-forward "^(" nil t)
         (delete-region (point-min) (- (point) 1)))
       (goto-char (point-max))
+      (insert (format-time-string "%Y-%m-%0d %H:%M:%S.%N\n"))
```
bbatsov: How does this look right now? I find it odd that the timestamps are ending with a newline - that's not exactly common for a log type of message. Due to the multi-line nature of most messages I kind of dislike just dumping some timestamp before each of them. You can easily tell right away whether the preceding or following timestamp is the one you're looking for this way.

dottedmag: I have inserted a newline because every message is already multiline one, so would
not move `(<--`/`(-->` to the right. Without a newline it was harder to see whether
the message is a request or a reply.

Looks pretty usable for me:
screen shot 2017-01-18 at 09 01 45

bbatsov: Yeah, I imagined this was your idea and this was the end result. I don't like it, as it breaks the visual flow of the buffer and it doesn't clearly associate the timestamp with the dictionary. Why not simply insert a timestamp key to the request/response objects?

bbatsov:
I'll close this, but I'll keep #1952 around so we could implement some timestamps eventually.

PR #1925, opened 2017-01-21
Remove session from nrepl interaction

dpsutton:
The most fundamental notion of CIDER interaction is the connection, which is the repl (or process connected to it). These each hold two sessions, the tooling and standard sessions. The important part is that you always know which session to use if you have a connection: you want the standard one mostly and sometimes the tooling connection. 

However, we keep registering the two independently. The importance of this is that we don't take care to make sure they go in lockstep. The end result is that we'll say something like do this request (with session number) in the clojure connection, and then do it in the clojurescript connection. This causes the _same session_ to be used in two different connections. This can cause the same code to happen in the two different repls, but handled by the same clojure backend, rather than ever hitting the clojurescript one. Worse, when returning results, the tooling reports that it is a clojure repl and we record this. This is what causes cljs repls to be marked as clojure repls.

This is a tangible bug because when we evaluate clojurescript code, we dont ask for a clojurescript repl, we look for buffers named with clojurescript in them! To wit,

    (defun cider--guess-cljs-connection ()
      "Hacky way to find a ClojureScript REPL.
    DO NOT USE THIS FUNCTION.
    It was written only to be used in `cider-map-connections', as a workaround
    to a still-undetermined bug in the state-stracker backend."
      (when-let ((project-connections (cider-find-connection-buffer-for-project-directory
                                       nil :all-connections))
                 (cljs-conn
                  ;; So we have multiple connections. Look for the connection type we
                  ;; want, prioritizing the current project.
                  (or (seq-find (lambda (c) (string-match "\\bCLJS\\b" (buffer-name c)))
                                project-connections)
                      (seq-find (lambda (c) (string-match "\\bCLJS\\b" (buffer-name c)))
                                (cider-connections)))))
        (unless cider--has-warned-about-bad-repl-type
          (setq cider--has-warned-about-bad-repl-type t)
          (read-key
           (concat "The ClojureScript REPL seems to be is misbehaving."
                   (substitute-command-keys
                    "\nWe have applied a workaround, but please also file a bug report with `\\[cider-report-bug]'.")
                   "\nPress any key to continue.")))
        cljs-conn))

Further, this is the first step on the roadmap to making connections first class objects. The plan is to make connections aware of their client buffers and buffers aware of all potential connections and have a preference. Then, when receiving results from the backend, we pick out the session, find the associated connection, and we again have a connection that flows through all code. This allows us to be very flexible in the number and type of connections we handle. It also allows us to update any associated buffers to connections when appropriate (ie, font lock, etc).

This should increase the usability of cljs and cljc cuffers in CIDER, and move to give them first class citizenship alongside clojure buffers.

**Caveats**

I haven't tested much of it. There are many things I don't know about in CIDER, killing sessions, describing sessions, etc. Everything should work, again, because connections uniquely (up to tooling) determine which session to use, and I have allowed us to mark tooling uses.

Also, specifically, the session browser needs some work. We need a new interface, and I'll begin working on this. But in a cljc project, it shows four UUID's. No one could find this useful as no one knows session ids. We need to group them by connection (clj cljs) and then mark them as tooling and standard. Once we do this, then we just describe them and pass along the connection selected, and mark it tooling or not and we are good to go.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -680,7 +680,6 @@ to be appended to the request message."
                             (cider-eval-spinner-handler connection callback)
                           callback)
                         connection
-                        (cider-current-session)
                         ns line column additional-params)
     (cider-spinner-start connection)))
 
@@ -690,7 +689,6 @@ If NS is non-nil, include it in the request."
   (nrepl-sync-request:eval
    input
    (cider-current-connection)
-   (cider-current-session)
    ns))
 
 (defcustom cider-pprint-fn 'pprint
@@ -744,8 +742,8 @@ NS specifies the namespace in which to evaluate the request."
   (nrepl-request:eval input
                       callback
                       (cider-current-connection)
-                      (cider-current-tooling-session)
-                      ns))
+                      ns nil nil nil t  ; tooling
+                      ))
 
 (defalias 'cider-current-repl-buffer #'cider-current-connection
   "The current REPL buffer.
@@ -761,8 +759,7 @@ Return the REPL buffer given by `cider-current-connection'.")
         (nrepl-request:interrupt
          request-id
          (cider-interrupt-handler (current-buffer))
-         (cider-current-connection)
-         (cider-current-session))))))
+         (cider-current-connection))))))
 
 (defun cider-current-session ()
   "Return the eval nREPL session id of the current connection."
@@ -855,9 +852,6 @@ loaded.
 If CONNECTION is nil, use `cider-current-connection'.
 If CALLBACK is nil, use `cider-load-file-handler'."
   (cider-nrepl-send-request (list "op" "load-file"
-                                  "session" (if connection
-                                                (cider-session-for-connection connection)
-                                              (cider-current-session))
                                   "file" file-contents
                                   "file-path" file-path
                                   "file-name" file-name)
@@ -889,7 +883,6 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
   (let* ((query (replace-regexp-in-string "[ \t]+" ".+" query))
          (response (cider-nrepl-send-sync-request
                     `("op" "apropos"
-                      "session" ,(cider-current-session)
                       "ns" ,(cider-current-ns)
                       "query" ,query
                       ,@(when search-ns `("search-ns" ,search-ns))
@@ -904,8 +897,7 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
 (defun cider-sync-request:classpath ()
   "Return a list of classpath entries."
   (cider-ensure-op-supported "classpath")
-  (thread-first (list "op" "classpath"
-                      "session" (cider-current-session))
+  (thread-first (list "op" "classpath")
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "classpath")))
 
@@ -913,7 +905,6 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
   "Return a list of completions for STR using nREPL's \"complete\" op.
 CONTEXT represents a completion context for compliment."
   (when-let ((dict (thread-first (list "op" "complete"
-                                       "session" (cider-current-session)
                                        "ns" (cider-current-ns)
                                        "symbol" str
                                        "context" context)
@@ -923,7 +914,6 @@ CONTEXT represents a completion context for compliment."
 (defun cider-sync-request:info (symbol &optional class member)
   "Send \"info\" op with parameters SYMBOL or CLASS and MEMBER."
   (let ((var-info (thread-first `("op" "info"
-                                  "session" ,(cider-current-session)
                                   "ns" ,(cider-current-ns)
                                   ,@(when symbol (list "symbol" symbol))
                                   ,@(when class (list "class" class))
@@ -936,7 +926,6 @@ CONTEXT represents a completion context for compliment."
 (defun cider-sync-request:eldoc (symbol &optional class member)
   "Send \"eldoc\" op with parameters SYMBOL or CLASS and MEMBER."
   (when-let ((eldoc (thread-first `("op" "eldoc"
-                                    "session" ,(cider-current-session)
                                     "ns" ,(cider-current-ns)
                                     ,@(when symbol (list "symbol" symbol))
                                     ,@(when class (list "class" class))
@@ -949,15 +938,13 @@ CONTEXT represents a completion context for compliment."
 (defun cider-sync-request:ns-list ()
   "Get a list of the available namespaces."
   (thread-first (list "op" "ns-list"
-                      "filter-regexps" cider-filtered-namespaces-regexps
-                      "session" (cider-current-session))
+                      "filter-regexps" cider-filtered-namespaces-regexps)
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "ns-list")))
 
 (defun cider-sync-request:ns-vars (ns)
   "Get a list of the vars in NS."
   (thread-first (list "op" "ns-vars"
-                      "session" (cider-current-session)
                       "ns" ns)
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "ns-vars")))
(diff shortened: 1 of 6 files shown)


cider-client.el
```diff
@@ -855,9 +852,6 @@ loaded.
 If CONNECTION is nil, use `cider-current-connection'.
 If CALLBACK is nil, use `cider-load-file-handler'."
   (cider-nrepl-send-request (list "op" "load-file"
-                                  "session" (if connection
-                                                (cider-session-for-connection connection)
-                                              (cider-current-session))
```
dpsutton: This is a recent fix of one particular instance that this whole branch fixes. We were loading files twice into clojure and never into clojurescript. Bug #1875 . It further marks the clojurescript repl as a clojure type one since its most revent eval response said it was a clj style buffer. (see `cider-repl--state-handler` in cider-repl.el for this. It's a function registered as a callback to every function).

cider-interaction.el
```diff
@@ -1666,7 +1662,7 @@ START and END represent the region's boundaries."
   (cider-ensure-connected)
   (let ((selected-session (completing-read "Describe nREPL session: " (nrepl-sessions (cider-current-connection)))))
     (when (and selected-session (not (equal selected-session "")))
-      (let* ((session-info (nrepl-sync-request:describe (cider-current-connection) selected-session))
+      (let* ((session-info (nrepl-sync-request:describe (cider-current-connection)))
```
dpsutton: Note this is the part that is kinda broken right now by this. Because it always describes the standard repl for the current cider connection. But note it has always been broken, as it is unable to select a session from another connection. (i think :))

cider-interaction.el
```diff
@@ -1685,10 +1681,8 @@ START and END represent the region's boundaries."
   "Close an nREPL session for the current connection."
   (interactive)
   (cider-ensure-connected)
-  (let ((selected-session (completing-read "Close nREPL session: " (nrepl-sessions (cider-current-connection)))))
-    (when selected-session
-      (nrepl-sync-request:close (cider-current-connection) selected-session)
-      (message "Closed nREPL session %s" selected-session))))
+  (nrepl-sync-request:close (cider-current-connection))
+      (message "Closed nREPL session"))
```
dpsutton: I didn't understand why you would ever close a tooling session and not a standard session--or vice versa-- so i figure they should die in lockstep.

cider-interaction.el
```diff
@@ -1700,10 +1694,8 @@ START and END represent the region's boundaries."
           (when (or (not nrepl-server-buffer)
                     ;; Sync request will hang if the server is dead.
                     (process-live-p (get-buffer-process nrepl-server-buffer)))
-            (when nrepl-session
-              (nrepl-sync-request:close buffer nrepl-session))
-            (when nrepl-tooling-session
-              (nrepl-sync-request:close buffer nrepl-tooling-session)))
+            (when nrepl-session or nrepl-tooling-session
+                  (nrepl-sync-request:close buffer)))
```
dpsutton: might need error handling on this in case one or the other has died. but i don't know how this would have happened either way.

nrepl-client.el
```diff
@@ -792,16 +792,18 @@ corresponding type of response."
...
+session already in it. This code will add it as appropriate to prevent
+connection/session drift.
 Return the ID of the sent message."
   (with-current-buffer connection
-    (when (and (not (lax-plist-get request "session"))
-               nrepl-session)
-      (setq request (append request (list "session" nrepl-session))))
+    (when-let ((session (if tooling nrepl-tooling-session nrepl-session)))
+      (setq request (append request
+                            (list "session" session))))
```
dpsutton: note here we were already (in the past) ensuring there was a session included. So why not _only_ do it here so we make sure that we do it right, and do it for what we actually mean. We add the optional `tooling` parameter to specify the tooling session

nrepl-client.el
```diff
@@ -899,45 +900,48 @@ If LINE and COLUMN are non-nil and current buffer is a file buffer, \"line\",
                     "line" line
                     "column" column)))))
 
-(defun nrepl-request:eval (input callback connection &optional session ns line column additional-params)
+(defun nrepl-request:eval (input callback connection &optional ns line column additional-params tooling)
```
dpsutton: this signature is kinda unfortunate. I'd like to add ns, line, column, additional-params into one list and then there would be just two optionals. But this explains any funcalls like `(nrepl-request request connection nil nil nil nil t)` from other areas. I don't like it though.

nrepl-client.el
```diff
@@ -899,45 +900,48 @@ If LINE and COLUMN are non-nil and current buffer is a file buffer, \"line\",
...
-(defun nrepl-sync-request:describe (connection &optional session)
+(defun nrepl-sync-request:describe (connection)
   "Perform :describe request for CONNECTION and SESSION."
-  (if session
-      (nrepl-send-sync-request (list "session" session "op" "describe")
-                               connection)
-    (nrepl-send-sync-request '("op" "describe")
-                             connection)))
+  (nrepl-send-sync-request '("op" "describe")
+                           connection))
```
dpsutton: This will need a tooling optional if we decide that someone would ever want to describe a tooling session. My guess is that an end user need never know this existed. I've been working on CIDER for a few months and I didn't even know :)

nrepl-client.el
dpsutton: ah, i think @Malabarba used a `cl-defun` somewhere that allowed passing optionals with :param, etc. this would be a nice candidate for that.

cider-interaction.el
bbatsov: This functionality exists mostly to kill dead nREPL sessions. Some sessions might hang around of CIDER didn't exit cleanly for instance. Guess I should have documented this better. :-)

nrepl-client.el
bbatsov: Yeah, I guess most similar functions are a good candidate for `cl-defun`.

nrepl-client.el
```diff
@@ -792,16 +792,18 @@ corresponding type of response."
   (with-current-buffer connection
     (number-to-string (cl-incf nrepl-request-counter))))
 
-(defun nrepl-send-request (request callback connection)
+(defun nrepl-send-request (request callback connection &optional tooling)
   "Send REQUEST and register response handler CALLBACK using CONNECTION.
 REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
 \"par1\" ... ). See the code of `nrepl-request:clone',
-`nrepl-request:stdin', etc.
+`nrepl-request:stdin', etc. This expects that the REQUEST does not have a
```
bbatsov: Check your docstrings with either `M-x checkdoc` or with `flycheck`.

nrepl-client.el
```diff
@@ -913,11 +913,12 @@ ADDITIONAL-PARAMS is a plist to be appended to the request message."
                       connection
                       tooling))
 
-(defun nrepl-sync-request:clone (connection)
+(defun nrepl-sync-request:clone (connection &optional tooling)
   "Sent a :clone request to create a new client session.
 The request is dispatched via CONNECTION."
```
bbatsov: You forgot to update this docstring.

bbatsov:
Overall the changes look good. Not sure whether you figured out what `tooling` sessions were, so I'll paste here the answer I gave you on slack:

```
The tooling session is meant for users who want to build commands themselves by simply evaluating some code. It's a remnant of the old days of CIDER before we were using nREPL middleware. If you evaluate some code to get completion candidates or do macroexpansion in the eval session that would interfere with it (e.g. messing up with `*1` and friends). I'm assuming very few people are using this tooling session, so I've been thinking for a while about killing it (and maybe coming up with some different solution to this problem).

CIDER never uses the tooling session directly (all commands that were using it were switch to middleware years go), the functionality was retained here just for the same of people who might prefer to just evaluate something as opposed to writing a middleware. You can see this approach still in projects like `monroe` and `inf-clojure`.
```

I have a feeling nobody uses, so it might be best to simply kill them. 

As an interesting side note - one nrepl session supports only one concurrent evaluation, so one of the ideas I had was to be able to spin sessions on the demand if case the `default` nrepl session is `busy`. That's fairly advanced stuff, but I thought you might find it useful info. 

And two small notes for the commit messages:
* Dont -> Don't
* In the final commit message the title got split in two

nrepl-client.el
dpsutton: if the tooling session is going to go away i suppose this will be a non-issue.

dpsutton:
I think I've cleaned everything up. Let me know any other spots that need some more care, but otherwise I think I'll start working on making sure connections are required parameters for everywhere but the entrance to eval commands

nrepl-client.el
bbatsov: Having a lot of positional params is always an issue. :-)

bbatsov:
Yeah, I guess this is a good enough first step. 

As for the tooling session - I'm not really certain. As I mentioned before using it at least for the tooling middleware makes some sense as this won't block the session dedicated to code evaluation. Most middleware invocations are kind of short, so this users probably don't notice this at all, but the issue exists. On the other hand we might come up with something even smarter (like dynamic session allocation or whatever).

bbatsov:
> I think I'll start working on making sure connections are required parameters for everywhere but the entrance to eval commands

Sounds reasonable.

Issue #1731, opened 2016-05-05
Define our own boot task to make jack-in more robust


Malabarba:
When doing middleware injection in boot projects, we add the cider-nrepl artifact as a dependency by passing the `-d` argument to boot and we add our middleware to tools.nrepl by passing the `-m` argument to the `repl` task.
The `-d` argument is fine, but the `-m` argument is a little finicky. It only works immediately after the `repl` task, so there's nothing we can do if the user instead uses a custom task or something like the very popular `cljs-repl` task.

Instead, cider-nrepl could define its own boot task that properly adds the nrepl middleware. Then all we have to do is inject the dependency with `-d` and add the task before any other task in `cider-boot-params`. This wouldn't require the `-m` argument and would work even if the `repl` task is not being called directly.

bbatsov:
👍

arichiardi:
This I think already work ootb, I am constantly doing `boot cider dev` instead of jacking-in (which sometimes break I think exactly for the reason at the top).

I can take care of this, just need some hint on were to put the task and how/where to require it from.
The task should be this right?

```
deftask cider "CIDER profile"
  []
  (require 'boot.repl)
  (swap! @(resolve 'boot.repl/*default-dependencies*)
         concat '[[org.clojure/tools.nrepl "0.2.12"]
                  [cider/cider-nrepl "0.10.0"]
                  [refactor-nrepl "2.0.0-SNAPSHOT"]])
  (swap! @(resolve 'boot.repl/*default-middleware*)
         concat '[cider.nrepl/cider-middleware
                  refactor-nrepl.middleware/wrap-refactor])
  identity)
```

Malabarba:
1. Almost. It would have to look like this:

        (deftask cider-middleware "CIDER middleware profile"
          []
          (require 'boot.repl)
          (swap! @(resolve 'boot.repl/*default-middleware*)
                 conj 'cider.nrepl/cider-middleware)
          identity)

  - We don't need to add cider-nrepl as a dependency here, because we already add the dependency on the command-line. (And the task wouldn't be available at all if cider-nrepl weren't already a dep)
  - Instead writing `cider.nrepl/cider-middleware`, it would be great if we the middlewares were specified as command-line arguments to this task. Then other middlewares (like refactor-nrepl) could inject themselves too.

2. You can put it in a file called `src/cider/tasks.clj`.
[...]

PR #1926, opened 2017-01-23
Update "Up and Running" part related to boot

arichiardi:
Some moar doc changes for #1731.

Also, @bbatsov if there are other places where I need to bump the minimum `boot` version that you can think of, please send it over :smile:


bbatsov:
👍

PR #1933, opened 2017-01-30
Cljs repl form tests

arichiardi:
On Slack, we had the following report:

> jfntn [7:55 AM] 
>The latest version of cider broke cljs repl type customization for me, no matter the value of `cider-cljs-lein-repl` cider always starts a rhino repl with `cider-jack-in-clojurescript`

> [7:57]  
>I used to just set `(cider-cljs-lein-repl . "(do (load-file \"script/dev.cljc\"))")` in my `.dir-locals.el`, I also tried using setq but I get the same behavior

So I decided to debug and test a little. I am opening this PR for some more testing around the `cider-cljs-repl-form` feature and in order to get some more feedback about the issue.

### DIFF
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -188,6 +188,32 @@
               :and-return-value '())
       (expect (cider-project-type) :to-equal cider-default-repl-command))))
 
+(describe "cider-cljs-repl-form"
+  (describe "when there is no overridden cider-cljs-lein-repl"
+    (it "returns the default form"
+      (expect (cider-cljs-repl-form "lein") :to-equal "(cemerick.piggieback/cljs-repl (cljs.repl.rhino/repl-env))")))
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm not sure what's the value of those tests - you're basically testing that changing the value of a defcustom in Emacs works right now. :-)

I think the actual problem is that you forgot to add a deprecation alias for the old name of the defcustom and people didn't notice the change.

bbatsov:
P.S. More valuable tests would check that the value of the proper defcustom is used depending on the project type.

arichiardi:
Lol yeah that is true :smile: Given that this a side effecting thing, probably I cannot really test it (that is why no tests had been added). I don't know, the reporter said that is running emacs `26` so I guess the problem can be there as well, hopefully we'll hear more from it.

bbatsov:
Seems that #1944 fixed the actual issue.

Issue #1938, opened 2017-02-08
cider-jack-in-clojurescript no longer honors buffer local variable


plexus:
This seems to be a regression in recent versions of CIDER. It's unfortunate, because setting `cider-cljs-lein-repl` in `.dir-locals.el` provided an elegant way to configure the CLJS repl on a per-project basis.

## Expected behavior

Having this as a `.dir-locals.el` in a project:

``` emacs-lisp
((nil . ((cider-cljs-lein-repl    . "(do (reloaded.repl/go) (user/cljs-repl))"))))
```

And calling `M-x cider-jack-in-clojurescript` from a buffer within that project, the ClojureScript REPL should be started using `(do (reloaded.repl/go) (user/cljs-repl))`.

## Actual behavior

The root value of `cider-cljs-lein-repl` is used, ignoring the dir-local/buffer-local value.

## Steps to reproduce the problem

clone [this repo](https://github.com/lambdaisland/redolist/), open `project.clj` in emacs, do `M-x cider-jack-in-clojurescript`. In the `*cider-repl CLJS redolist*` buffer, you should see the line

```
Garden: watching src/clj/css.clj
```

If you don't then this problem is still present.

## Environment & Version information

This worked at least until CIDER 0.13.0. After upgrading to 0.15.0-SNAPSHOT it no longer works.

### CIDER version information

```
;; CIDER 0.15.0snapshot (package: 20170129.1941), nREPL 0.2.12
;; Clojure 1.9.0-alpha14, Java 1.8.0_91
```

### Lein/Boot version

```
Leiningen 2.6.1 on Java 1.8.0_91 Java HotSpot(TM) 64-Bit Server VM
```
### Emacs version

```
26.0.50.1
```

### Operating system

Ubuntu 16.04.1 LTS

julienfantin:
Possibly related to #1933 ?

@plexus does it work as expected if you manually set `cider-cljs-lein-repl` with `setq` for example?

plexus:
That does sound like the same problem, since they are also setting `cider-cljs-lein-repl` through `.dir-locals.el`.

The problem is that the CLJS repl is getting created indirectly in a process handler, at which point the buffer context is no longer there.

I came up with this solution, not sure to call it a fix or a hack though :) should I turn it into a PR?

https://github.com/clojure-emacs/cider/compare/master...plexus:dir_local_cljs_repl_hack

julienfantin:
[...]

PR #1944, opened 2017-02-22
Make cljs repl command configurable again through .dir-locals.el

plexus:
When using `cider-jack-in-clojurescript`, the CLJS repl is started indirectly
from a process handler. The result is that when looking up variables like
`cider-cljs-lein-repl`, it's no longer in the context of a source buffer in the
user's project, and so dir-local (or buffer-local) values no longer apply.

The "fix" is to wrap `cider-create-sibling-cljs-repl` in a lambda which captures
the current buffer.

Would be happy to get feedback on this code.

For more discussion see https://github.com/clojure-emacs/cider/issues/1938 and https://github.com/clojure-emacs/cider/pull/1933

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -498,6 +498,9 @@ based on PROJECT-TYPE."
 The new buffer will correspond to the same project as CLIENT-BUFFER, which
 should be the regular Clojure REPL started by the server process filter."
   (interactive (list (cider-current-connection)))
+  ;; Load variables in .dir-locals.el into the server process buffer, so
+  ;; cider-cljs-*-repl can be set for each project individually.
+  (hack-local-variables)
   (let* ((nrepl-repl-buffer-name-template "*cider-repl CLJS%s*")
          (nrepl-create-client-buffer-function #'cider-repl-create)
          (nrepl-use-this-as-repl-buffer 'new)


cider.el
```diff
@@ -608,7 +608,10 @@ own buffer."
                           (nrepl-use-this-as-repl-buffer repl-buff))
                       (nrepl-start-server-process
                        project-dir cmd
-                       (when cljs-too #'cider-create-sibling-cljs-repl))))))
+                       (when cljs-too (let ((origin-buffer (current-buffer)))
```
bbatsov: My concern about this code right now is that without some comments no one is going to figure out its exact purpose. I do wonder if won't be easier just to set the `default-directory` within `cider-create-sibling-cljs-repl` so it would be self-contained. I'm curious though, what's the default-directory being set to from this process handler. Seems to me that it should still be some directory within the currently visited project.

plexus:
I put some debugging calls in `cider-create-sibling-cljs-repl`, this is what I got

``` emacs-lisp
current-buffer: #<buffer *nrepl-server App*> 
default-directory: ~/LambdaIsland/App/
cider-cljs-lein-repl: (do (require ’figwheel-sidecar.repl-api) (figwheel-sidecar.repl-api/start-figwheel!) (figwheel-sidecar.repl-api/cljs-repl))
```

So the context is the process buffer, the default directory is indeed inside the project, but it's not picking up `.dir-locals.el`. This is my `~/LambdaIsland/App/.dir-locals.el`, you can see `cider-cljs-lein-repl` is different.

``` emacs-lisp
((nil . ((cider-refresh-before-fn . "reloaded.repl/suspend")
         (cider-refresh-after-fn  . "reloaded.repl/resume")
         (cider-cljs-lein-repl    . "(do (reloaded.repl/go) (user/cljs-repl))"))))
```

So `*nrepl-server App*` isn't picking up the directory local variables, which perhaps isn't too surprising since it doesn't visit a file. The emacs manual says

> Directory-local variables also take effect in certain buffers that do not visit a file directly but perform work within a directory, such as Dired buffers (see Dired).

plexus:
Thanks for the feedback, it was worth it to investigate this better. I pushed a different patch that is more straightforward. See what you think.

cider.el
```diff
@@ -498,6 +498,9 @@ based on PROJECT-TYPE."
 The new buffer will correspond to the same project as CLIENT-BUFFER, which
 should be the regular Clojure REPL started by the server process filter."
   (interactive (list (cider-current-connection)))
+  (hack-local-variables)   ;; Load variables in .dir-locals.el into the nREPL
```
bbatsov: I'd just move the comment lines above `(hack-local-variables)`.

cider.el
```diff
@@ -498,6 +498,9 @@ based on PROJECT-TYPE."
...
+                           ;; process buffer, so cider-cljs-*-repl can be set
```
bbatsov: server process buffer
(because we also have connection process buffers here and there)

bbatsov:
The new patch looks much better to me. I had forgotten completely about this function.

Issue #1947, opened 2017-02-27
cider-jack-in fails when cider-enlighten-mode is enabled


plexus:
## Expected behavior

`cider-jack-in` results in a buffer with a REPL prompt

## Actual behavior

The minibuffer pops up with `Lisp expression: `. Entering anything there results in

```
Debugger entered--Lisp error: (error "Not an nREPL dict object: (+ 1 1)")  
signal(error ("Not an nREPL dict object: (+ 1 1)")) 
error("Not an nREPL dict object: %s" (+ 1 1))  
nrepl-dict-get((+ 1 1) "interns")  
cider-resolve-ns-symbols((+ 1 1))  
cider-refresh-dynamic-font-lock((+ 1 1))  
cider-set-buffer-ns((+ 1 1)) 
 cider-repl-set-initial-ns(#<buffer *cider-repl App*>)  
cider-repl-init(#<buffer *cider-repl App*>)  
cider--connected-handler()  run-hooks(nrepl-connected-hook)  
nrepl-start-client-process(nil 42389 #<process nrepl-server>)  
nrepl-server-filter(#<process nrepl-server> "nREPL server started on port 42389 on host 127.0.0.1 - nrepl://127.0.0.1:42389\n")
```

This shows up in nrepl-messages:

```
(--> 
  op  "eval"
  code  "(str *ns*)"
  enlighten  "true"
  session  "b72d41fa-6418-441f-bc52-cc8e7138c986"
  id  "4"
)
(<-- 
  ex  "class java.lang.Exception"
  id  "4"
  root-ex  "class java.lang.Exception"
  session  "b72d41fa-6418-441f-bc52-cc8e7138c986"
  status  ("eval-error")
)
(<-- 
  err  "Exception Debugger not initialized  user/eval27659/fn--27660 (form-init2222324724219143458.clj:1)
"
  id  "4"
  session  "b72d41fa-6418-441f-bc52-cc8e7138c986"
)
```

The user gets an empty repl buffer, but pressing enter does bring up a prompt. Except that it shows `nil` as the namespace.

## Steps to reproduce the problem

Open a Clojure file, `M-x cider-enlighten-mode`, `M-x cider-jack-in`.

### CIDER version information

```
;; CIDER 0.15.0snapshot (package: 20170129.1941), nREPL 0.2.12
;; Clojure 1.9.0-alpha14, Java 1.8.0_91
```

### Emacs version

[...]

PR #1950, opened 2017-03-05
Reorder repl init to occur after debugger init

dpsutton:
When initializing the repl, we ask what namespace we are in by evaling code,
eg (eval "(str *ns*)"). However, enlighten mode hijacks eval messages in the
middleware and uses the debugger on them which causes an error if the debugger
is not yet initialized.

https://github.com/clojure-emacs/cider/issues/1947
-----------------

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -857,16 +857,20 @@ message in the REPL area."
 This function is appended to `nrepl-connected-hook' in the client process
 buffer."
   ;; `nrepl-connected-hook' is run in the connection buffer
-  (cider-make-connection-default (current-buffer))
-  (cider-repl-init (current-buffer))
-  (cider--check-required-nrepl-version)
-  (cider--check-clojure-version-supported)
-  (cider--check-middleware-compatibility)
-  (cider--debug-init-connection)
-  (cider--subscribe-repl-to-server-out)
-  (when cider-auto-mode
-    (cider-enable-on-existing-clojure-buffers))
-  (run-hooks 'cider-connected-hook))
+
+  ;; `cider-enlighten-mode' changes eval to include the debugger, so we inhibit
+  ;; it here as the debugger isn't necessarily initialized yet
+  (let ((cider-enlighten-mode nil))
+    (cider-make-connection-default (current-buffer))
+    (cider-repl-init (current-buffer))
+    (cider--check-required-nrepl-version)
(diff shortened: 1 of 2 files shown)


bbatsov:
LGTM, the relevant issue has to be mentioned in the commit message `[Fix ...` and the changelog.

cider.el
```diff
@@ -858,11 +858,13 @@ This function is appended to `nrepl-connected-hook' in the client process
 buffer."
   ;; `nrepl-connected-hook' is run in the connection buffer
   (cider-make-connection-default (current-buffer))
-  (cider-repl-init (current-buffer))
   (cider--check-required-nrepl-version)
   (cider--check-clojure-version-supported)
   (cider--check-middleware-compatibility)
   (cider--debug-init-connection)
+  ;; cider-repl-init evals code. if enlighten mode is on, eval requires the
```
bbatsov: `cider-repl-init'

if `enlighten-mode' is on

bbatsov:
In the interest of robustness we might also add some check in the enlighten init code to determine whether it can be enabled properly or not. That's not essential for sure, but would be nice to have.

bbatsov:
Don't forget about the changelog entry.

dpsutton:
Ah so sorry. Got that updated.

In regards to your thoughts about checking that enlighten mode can actually run yet, did you want that to be on the elisp side or the clojure side?

bbatsov:
> In regards to your thoughts about checking that enlighten mode can actually run yet, did you want that to be on the elisp side or the clojure side?

Elisp.

dpsutton:
So I put the inhibition in and "undid" the reorder. Read the commit message and see if you agree. If so, I think we are ready to squash and merge.

Also, in the related issue, I thought this was going to be much worse. I was thinking that the startup tasks were spread throughout the codebase, but that was just the two session creation points (tooling and regular). That made me think all the initialization happened at random points. I was happy to be misinformed on that point.

bbatsov:
👍 I love the new approach. No idea why this wasn't my initial suggestion, as it's clearly the natural way to solve this problem.

dpsutton:
Swuashed and commit message cleaned up a little. I think it's ready to merge.

bbatsov:
👍

PR #1951, opened 2017-03-05
Fix expectations in cider-font-lock-tests

dpsutton:
These tests pass in their current form, but they do not pass on a build of emacs
from source tracking master. The tests us cl-every and cl-some and the tests
assert that the results are either `:equal-to t` or `:not :equal-to t`. However,
cl-some does not return t but the truthy value that it finds, and the buttercup
comparison fails with

FAILED: Expected cider-reader-conditional-face to be `eq' to t

Failing to compare a "truthy" value with t. So I am changing all of the
assertions to assert that they are something rather than they are not something,
as we could miss some subtle bugs if the tests thing that `truthy-value :not
:equal-to t`.

Also some line length formatting.

### DIFF
--- a/test/cider-font-lock-tests.el
+++ b/test/cider-font-lock-tests.el
@@ -29,7 +29,7 @@
 
 ;; Utilities
 
-(defmacro test-with-temp-buffer (content &rest body)
+(defmacro cider--test-with-temp-buffer (content &rest body)
   "Evaluate BODY in a temporary buffer with CONTENT."
   (declare (debug t)
            (indent 1))
@@ -40,23 +40,24 @@
      (font-lock-fontify-buffer)
      ,@body))
 
-(defun every-face-is-at-range-p (start end face)
+(defun cider--face-covers-range-p (start end face)
   "Return true if every face from START to END has FACE."
-  (let ((all-faces (cl-loop for i from start to end collect (get-text-property i 'face))))
-    (cl-every (lambda (target-face)
+  (let ((all-faces (mapcar (lambda (i) (get-text-property i 'face)) (number-sequence start end))))
+    (seq-every-p (lambda (target-face)
+                   (or (eq face target-face)
+                       (when (consp target-face)
+                         (seq-contains target-face face))))
+                 all-faces)))
+
+(defun cider--face-exists-in-range-p (start end face)
+  "Return true if FACE exists between START to END."
+  (let ((all-faces (mapcar (lambda (i) (get-text-property i 'face)) (number-sequence start end))))
+    ;; cl-some returns t now but will change to return a truthy value in the future
+    (seq-some (lambda (target-face)
                 (or (eq face target-face)
                     (when (consp target-face)
                       (seq-contains target-face face))))
               all-faces)))
-
-(defun face-exists-at-range-p (start end face)
-  "Return true if FACE exists between START to END."
-  (let ((all-faces (cl-loop for i from start to end collect (get-text-property i 'face))))
-    (cl-some (lambda (target-face)
-               (or (eq face target-face)
-                   (when (consp target-face)
-                     (seq-contains target-face face))))
-             all-faces)))
 
 ;; Tests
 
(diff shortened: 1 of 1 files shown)


bbatsov:
I think you took the wrong approach here - I would fix those predicate methods instead. Something ending with a `-p` should returning a boolean value IMO.

dpsutton:
I thought `-p` functions in emacs lisp returned _truthy_ values, not true or false. Looking at `-p` functions in CIDER, most are built on `and`, `or`, `derived-mode-p` and `string-match-p` all of which return truthy values, never `t`. Honestly, I thought that emacs lisp lacked a true boolean true and used by convention `t` but this does actually look like a special form when I evaluate it.

I can fix the predicates but i think it would look weird:

`(when (cl-some ....) t)` vs `(cl-some ...)`. I think the former looks like non-idiomatic elisp code.

Seeing how these test results are failing on older builds but not newer builds of emacs, it looks like cl-some is changing its contract from returning t to the truthy value it finds. I think there's an easier solution at hand:

`(expect val :to-be t)` -> `(except val)`

and 

`(expect val :not :to-be t)` -> `(expect (not val))`

I guess it doesn't read as well but it seems that with cl-some changing its return value and with the expect form comparing truthy vs true this seems like the most straightforward way to have the tests pass reliably.

Thoughts?

bbatsov:
Elisp and Common Lisp are not known for their brilliantly devised APIs. :-) I do write code like this from time to time, but I generally dislike it. But I understand your point.

As an alternative - you can always simply replace the current checks with `(expect something :to-be-truthy)`. I think it's horrible to expect a predicate method to return some symbol or something like this and to be using this in the expectations.

dpsutton:
ah, `:to-be-truthy` is the check i was looking for but couldn't find any documentation about what they accepted.

I can update the predicate to return t if that is what you prefer. I was explaining why I got to the decision that I made earlier but by no means am I unwilling to code in the style that you prefer for CIDER. I think your "spidey" sense is better than a different contributors styles as when things conform to you the project will feel more cohesive.

bbatsov:
I'm fine with `:to-be-truthy`. One day when we don't support Emacs 24 we'll simply dump all the `cl-` crap and replace it with `seq.el`.

xiongtx:
> we'll simply dump all the `cl-` crap and replace it with `seq.el`

😦 I don't think there's an issue with `cl` here--the tests just need to be written in such a way to accommodate [generalized booleans](http://clhs.lisp.se/Body/26_glo_g.htm#generalized_boolean).

bbatsov:
> You might as well turn every :to-be t into :to-be-truthy. :to-be should be used to test for specific values, not truth. See the Emacs manual on truth and falsehood.

I'm well aware of this possibility and all sorts of Emacs Lisp and Common Lisp peculiarities. But this doesn't mean that I like them. I tolerate `cl-lib`, but I'm definitely not a fan. The only reason why CIDER uses it is the fact that it's bundled with older Emacsen unlike `seq.el`. At any rate - this is something really tiny as far as I'm concerned - I just mentioned it in passing.

bbatsov:
@dpsutton You forgot to update your commit message. Btw, the predicate functions have names that sound a bit off in English (e.g. `face-exists-at-range-p`, `every-face-is-at-range-p`) - this might be an opportunity to rename them.

dpsutton:
@bbatsov commit message and predicate tests renamed. Didn't expect this patch to get so much attention :) It was just a single failing test on emacs from master and something I had been meaning to do.

bbatsov:
Hehe, yeah we might have overdid it a bit. 

I just thought that this file also introduces unprefixed functions and macros. :D Anyways, I doubt this will cause issues in the near future. Guess I simply didn't pay much attention to the PR that added all this code.

dpsutton:
@bbatsov good call on the unprefixed functions. I can fix those up now while i'm "here"

bbatsov:
> @bbatsov good call on the unprefixed functions. I can fix those up now while i'm "here"

Oki-doki. Let's get this over with. :-)

bbatsov:
@dpsutton ping :-)

Btw, those tests should have used `seq.el` instead of `cl-lib`. You can also change this if you're into it - it will take only a minute.

dpsutton:
I've excised all cl.el and prefixed everything. Let me know if you want any other touchups otherwise i think we can squash and put this behind us. :)

bbatsov:
@dpsutton looks great now. Squash and we're good to go!

Issue #1726, opened 2016-04-30
Improve the presentation of nrepl messages


bbatsov:
The presentation of nREPL messages we currently use leaves a ton of room for improvement.

![image](https://cloud.githubusercontent.com/assets/103882/14937348/a20f9202-0eb9-11e6-9a53-e302cb7e4859.png)
- we lack proper font-locking that would make it easier to figure out what is what
- even the standard message keys are displayed in different order in every message
- values are not properly indented
- sometimes several keys end up being displayed on the same line, which gets me every time (I just miss the second key)

bbatsov:
[This article](http://www.wilfred.me.uk/blog/2016/04/28/effortless-major-mode-development/) might be useful for the font-locking code if anyone decides to tackle this.

Malabarba:
Font locking in the sense of font-lock-mode shouldn't be necessary. We can just propertize a specific face to every other sexp.

bbatsov:
True. I mostly want top-level keys to stand out. Bonus points for properly font-locked, strings, nils, etc.

dpsutton:
Did @Malabarba 's commit fully satisfy the request? I find myself reinventing lots of things in cider only to find out that the necessary features already exist, there's a pattern to follow, etc. If yall would put a few pointers in here I'd be happy to run through and work on this.

Malabarba:
My commit only highlights the request's keys, which was the lowest-hanging of the fruits here. I feel that improves the readability significantly.

It would also be nice to have proper string font-locking and proper indentation, but these might be a little more difficult to do efficiently (without making the buffer supper slow) and they're really not high priority for me.

bbatsov:
I'd say that ordering the request/response keys consistently should be relatively easy to do and would be very helpful as well. Tackling the bug with several keys getting displayed on the same line is also important.

bbatsov:
@dpsutton This ticket is still yours if you want it. ;-)

Malabarba:
Given that this only really affects the cider developers, and even we weren't bothered enough to actually do something significant about it, maybe it's not that relevant. People's effort might be better spent elsewhere.

bbatsov:
While this is true, this issue still represents an open thread and requires way less effort than most open tasks we have. I've been trying to do some cleanup lately, as tasks really piled on in the past 5-6 months. Hopefully contributor activity will pick up, as I'm pretty short on time myself.

harold:
After goofing around here for a bit another potential improvement would be:

```
[...]

PR #1956, opened 2017-03-08
[1726] Order keys in nrepl message objects.

harold:
Part of the overall solution described in issue 1726.

In order to have the keys print in a predictable order, we can sort them before printing.

Example before this change:
```
(-->
  ns  "cortex.dataset"
  op  "eval"
  session  "499ca8d8-43c8-40bf-aa19-ff369673ce4c"
  code  "(defn map-sequence->dataset "Turns an infinite training se..."
  file  "/home/harold/src/cortex/src/cortex/dataset.clj"
  line  377
  column  1
  id  "37"
)
```

Example after this change:
```
(-->
  code  "(defn map-sequence->dataset "Turns an infinite training se..."
  column  1
  file  "/home/harold/src/cortex/src/cortex/dataset.clj"
  id  "45"
  line  377
  ns  "cortex.dataset"
  op  "eval"
  session  "499ca8d8-43c8-40bf-aa19-ff369673ce4c"
)
```

Thanks!


harold:
Hello @bbatsov --- inspired by your tweet: https://twitter.com/bbatsov/status/839409503632261121

Some more work on: https://github.com/clojure-emacs/cider/issues/1726

I haven't done _that_ much elisp before, so there may be a better way, but it works for me in simple tests.

bbatsov:
Thanks for tackling this!

`id`, `op` and `session` have special importance as they are part of every request and response. They should always be the 3 first keys that get printed and the rest can be sorted alphabetically.

nrepl-client.el
```diff
@@ -1219,11 +1219,16 @@ FOREGROUND and BUTTON are as in `nrepl-log-pp-object'."
     (let ((head (format "(%s" (car object))))
       (insert (color head))
       (let ((indent (+ 2 (- (current-column) (length head))))
-            (l (point)))
+            (l (point))
+            (sorted-object (if (cdr object)
+                               (apply 'seq-concatenate 'list
+                                      (sort (seq-partition (copy-list (cdr object)) 2)
```
bbatsov: You should use `cl-copy-list`.

harold:
That all makes sense --- I'll do another pass on this tomorrow. I'm out of time for today.

harold:
@bbatsov --- got `id`, `op`, and `session` to print first. I also snuck in the table alignment:

```
(-->
  id      "67"
  op      "eval"
  session "d9131a55-30bf-405c-ac9e-ed4c90be9ad4"
  code    "(defmulti loss
  "Implement a specific loss based on the typ..."
  column  1
  file    "/home/harold/src/cortex/src/cortex/loss.clj"
  line    136
  ns      "cortex.loss"
)```

bbatsov:
Looks great! I was reminded of just one more thing - now we have `time-stamp` which is also special. We should put it either first or after `session` IMO.

harold:
Is that new? I haven't seen it (I'm still running the latest MELPA release).

Adding a string here:
https://github.com/clojure-emacs/cider/pull/1956/files#diff-bc7f0008c444f0fbf63329a565cdf64fR1230

Should be pretty easy, if you want to make any other names special. Notably, the special keys also appear in alphabetical order, so `time-stamp` after `session` would be easiest. ;)

bbatsov:
Pretty new - https://github.com/clojure-emacs/cider/commit/050b2a7065c0a278eaf500f9f7b2a081893989fb

Yeah, I saw adding this would be easy. Just thought now would be the best time to do it.

harold:
@bbatsov --- I'll be out till Monday (tomorrow is my birthday).

Perhaps you'd like to pull this into your repo and add that and merge? Otherwise I can check it out Monday.

Regardless, thanks, this was fun and enlightening. Hacking on Cider is easier than I thought it might be.

bbatsov:
> Perhaps you'd like to pull this into your repo and add that and merge? Otherwise I can check it out Monday.

No worries! I'll wrap this my myself.

> Regardless, thanks, this was fun and enlightening. Hacking on Cider is easier than I thought it might be.

Happy to hear this! Hopefully I'll be seeing more of you here! ;-)

Happy birthday!

PR #1958, opened 2017-03-12
[Fix #1832] Add eldoc info for datomic query input parameters

### LINKED ISSUE #1832: Feature request: eldoc for datomic query input parameters
I guess this request might be too specific, but I think it would be useful if there would be eldoc support for datomic query input parameters.

For example in the following query, the `:in` keyword specifies the list of input parameters:
datomic_query_example
so I guess it should be possible to use them in the eldoc info.

I'm not sure how this should be implemented, but if someone could give directions on how to do it, I would gladly give it a try.

Thanks a lot!

mallt:
This PR adds eldoc info for datomic query input parameters. The parameters specified by the `:in` keyword of the datomic query will be visible in the eldoc info while writing the [`datomic.api/q`](http://docs.datomic.com/clojure/#datomic.api/q) function call:

cider-datomic-eldoc

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -86,6 +86,15 @@ displayed.  When nil, always display eldoc for first symbol of the sexp."
   :group 'cider
   :package-version '(cider . "0.13.0"))
 
+(defcustom cider-eldoc-display-context-dependent-info nil
+  "When non-nil, display context dependent info in the eldoc where possible.
+CIDER will try to add expected function arguments based on the current context,
+for example for the datomic.api/q function where it will show the expected
+inputs of the query at point."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
+
 (defun cider--eldoc-format-class-names (class-names)
   "Return a formatted CLASS-NAMES prefix string.
 CLASS-NAMES is a list of classes to which a Java interop form belongs.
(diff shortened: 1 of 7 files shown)


cider-eldoc.el
```diff
@@ -398,10 +398,18 @@ This includes the arglist and ns and symbol name (if available)."
                                         "arglists" arglists
                                         "docstring" docstring
                                         "type" type)))
-                ;; middleware eldoc lookups are expensive, so we
-                ;; cache the last lookup.  This eliminates the need
-                ;; for extra middleware requests within the same sexp.
-                (setq cider-eldoc-last-symbol (list thing eldoc-plist))
+                ;; check datomic query
```
bbatsov: check if we're dealing with a datomic query

cider-eldoc.el
```diff
@@ -398,10 +398,18 @@ This includes the arglist and ns and symbol name (if available)."
...
+                ;; replace arglists and do not cache
```
bbatsov: You should explain here why this shouldn't be cached.

bbatsov:
Apart from my small remarks this looks ok. I don't know the first thing about the datomic, though, so I have no idea how robust your approach is. :-)

bbatsov:
I'd add a defcustom for this, btw, as it changes the existing consistent behaviour of eldoc. I'm reasonably sure not everyone would appreciate this feature.

mallt:
Thanks a lot for the review!

I'm not too sure about the defcustom, as the code of this PR will only interfere with the current eldoc if it detects a call to the `datomic.api/q` function, so it's pretty isolated already. It shouldn't have an effect on any other function calls, there the regular eldoc behavior will not be altered.

If you ever find the time, you should definitely check out datomic by the way, it's pretty awesome. :)

bbatsov:
> I'm not too sure about the defcustom, as the code of this PR will only interfere with the current eldoc if it detects a call to the datomic.api/q function, so it's pretty isolated already. It shouldn't have an effect on any other function calls, there the regular eldoc behavior will not be altered.

The point is that this would start behaving differently from regular eldoc and this might be considered unwanted by some people.

mallt:
Ok I see, thanks again for the comments! I will update the PR accordingly.

mallt:
@bbatsov Do you think `cider-eldoc-add-context-dependent-args` would be a good name for the defcustom? I thought it might be better to give it a broader name in case this functionality would be extended with other special cases? 

```elisp
(defcustom cider-eldoc-add-context-dependent-args nil
  "When non-nil, cider will add context dependent arguments to the eldoc where possible.
Cider will try to add expected function arguments based on the current context, for
example for the datomic.api/q function where it will show the expected inputs of the
query at point."
  :type 'boolean
  :group 'cider
  :package-version '(cider . "0.15.0"))
```

bbatsov:
Perhaps `cider-eldoc-display-context-dependent-info`?

mallt:
Sounds perfect, thanks!

mallt:
I've addressed all the comments in this new commit.

cider-eldoc.el
```diff
@@ -86,6 +86,15 @@ displayed.  When nil, always display eldoc for first symbol of the sexp."
   :group 'cider
   :package-version '(cider . "0.13.0"))
 
+(defcustom cider-eldoc-display-context-dependent-info nil
```
bbatsov: I'd mention this somewhere in the manual (e.g. in the eldoc section), otherwise I doubt many people will find it.

cider-eldoc.el
```diff
@@ -86,6 +86,15 @@ displayed.  When nil, always display eldoc for first symbol of the sexp."
...
+  "When non-nil, display context dependent info in the eldoc where possible.
+Cider will try to add expected function arguments based on the current context,
```
bbatsov: Cider -> CIDER (for consistency with the rest of the docstrings).

mallt:
I've addressed the last 2 comments, thanks!

mallt:
Ok, done.

bbatsov:
👍

mallt:
Thanks a lot for your comments!

PR #1959, opened 2017-03-15
Add markdown support in the grimmoire buffer

notetiene:
* cider-grimoire.el (cider-create-grimoire-buffer): Activate markdown-mode when
  found.

The `http://conj.io/` gives documentation in the markdown format. I thought it would be nice to get font-lock working. It's a really simple PR.

-----------------

- [x] The commits are consistent with our [contribution guidelines](https://github.com/clojure-emacs/cider/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -79,6 +79,8 @@ opposite of what that option dictates."
     (read-only-mode -1)
     (insert content)
     (read-only-mode +1)
+    (when (fboundp 'markdown-mode)
+      (markdown-mode))
     (goto-char (point-min))
     (current-buffer)))
 


bbatsov:
This was discussed and rejected in the past already. `markdown-mode` is for **editing** markdown and what we really need is something that can **render** markdown in Emacs. Such a mode would super useful.

bbatsov:
If this is made optional, and people are interested in it, I might consider merging this change though, as it's pretty simple.

xiongtx:
Don't think there's a good solution to rendering in Emacs; maybe when WebKit support is built-in?

We *could* [use `eww` to render](https://www.reddit.com/r/emacs/comments/2x4jtt/render_markdown_in_a_buffer/cowyg9e/):

```emacs-lisp
(defun gk-markdown-preview-buffer ()
  (interactive)
  (let* ((buf-this (buffer-name (current-buffer)))
         (buf-html (get-buffer-create
                    (format "*gk-md-html (%s)*" buf-this))))
    (markdown-other-window (buffer-name buf-html))
    (shr-render-buffer buf-html)
    (eww-mode)
    (kill-buffer buf-html)))
```

But the results aren't that different:

Markdown           |  `eww`
:-------------------------:|:-------------------------:
Markdown  |  eww

bbatsov:
@xiongtx I actually like this solution! Let's go with it. I still it still relies on `markdown-mode` so there should probably be some conditional check and its absense we should fallback to the current code.

xiongtx:
I'm not a big fan of this solution. 

If you look at the `eww` rendering, the arities under `Usages` are all joined onto one line. This could be considered a Grimoire problem, but is ugly either way.

bbatsov:
Ah, yeah - it looks pretty different here https://www.conj.io/store/v1/org.clojure/clojure/1.8.0/clj/clojure.core/vector

Guess there's not much we can do then.

xiongtx:
I think the biggest value-add would be font-locking for the Clojure code, not for the MD titles and such. What do you think?

bbatsov:
I agree.

Issue #1962, opened 2017-03-18
cider-eval-region has catastrophic performance


phillord:
I find that cider-eval-region has catastrophic performace, when evaling
largish areas. In worst case it locks up Emacs in a way that Ctrl-G cannot
recover from.

This is true with the current release and with master.

Reproduction. Eval this code (in Emacs!).

#+begin_src elisp
  (defun aaa()
    (interactive)
    (loop for i from 1 to 1000 do
          (insert
           (format "(defn a%s [])\n\n" i))))
#+end_src

Open a new clojure file, start cider, run M-x aaa. Then M-x
mark-whole-buffer, M-x cider-eval-region. The eval messages start up quite
quickly, then get very slow and Emacs becomes unrecoverable.

Interestingly, this is a regression. I tried with this commit:

commit 21b585752358c719ff87ad51246d543c59f60b82
Author: Chaitanya Koparkar <ckoparkar@live.in>
Date:   Mon Apr 25 21:27:31 2016 +0530

    Always fetch ns-vars-with-meta from nREPL middleware (#1715)

And everything worked swimmingly (evaling in a second or so). There is nothing
magic about this commit -- I just noticed that I got different performance on
different computers, and it turned out to be because one Emacs had an old
version of cider at this commit.

### CIDER version information

;; CIDER 0.13.0snapshot (package: 0.14.0), nREPL 0.2.12

(Also cider-latest from MELPA-stable!)

### Lein/Boot version

lein 2.7.1

### Emacs version

Emacs-25 HEAD

### Operating system

Ubuntu 16.04

bbatsov:
@cskksc Can you look into this when you have some time?

phillord:
It's the overlays. The same ones are being created continuous; I think you have creating a quadratic number of them.

Using my reproduction case this is a print out of all the overlays in buffer after evaluating the first ten forms.

    "--make-overlay (127:140:result:(after-string  => #'user/a10 ))"

[...]

PR #1964, opened 2017-03-18
Fix performance in fringe overlay placement

phillord:
Multiple identical fringe overlays were being placed by overlays were being
added to the entire region for every interactive eval in that region.

Add check in cider-interactive-eval-handler, so only first value adds fringe
overlays. Remove all fringe overlays in region before placing new ones.

Addresses #1962

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -693,6 +693,7 @@ REPL buffer.  This is controlled via
                          (current-buffer))
     (save-excursion
       (goto-char beg)
+      (remove-overlays beg end 'cider-fringe-indicator)
       (condition-case nil
           (while (progn (clojure-forward-logical-sexp)
                         (and (<= (point) end)
@@ -709,11 +710,14 @@ or it can be a list with (START END) of the evaluated region."
          (beg (car-safe place))
          (end (or (car-safe (cdr-safe place)) place))
          (beg (when beg (copy-marker beg)))
-         (end (when end (copy-marker end))))
+         (end (when end (copy-marker end)))
+         (fringed nil))
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  (lambda (_buffer value)
                                    (if beg
-                                       (cider--make-fringe-overlays-for-region beg end)
+                                       (unless fringed
+                                         (cider--make-fringe-overlays-for-region beg end)
+                                         (setq fringed t))
                                      (cider--make-fringe-overlay end))
                                    (cider--display-interactive-eval-result value end))
                                  (lambda (_buffer out)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 ### Bugs Fixed
 
+* [#1962](https://github.com/clojure-emacs/cider/issues/1962): Fix performance in fringe overlay placement
 * [#1947](https://github.com/clojure-emacs/cider/issues/1947): Fix error on `cider-jack-in` when `enlighten-mode` is enabled.
 * [#1588](https://github.com/clojure-emacs/cider/issues/1588): Redirect `*err*`, `java.lang.System/out`, and `java.lang.System/err` to REPL buffer on all attached sessions.
 * [#1707](https://github.com/clojure-emacs/cider/issues/1707): Allow to customize line truncating in CIDER's special buffers.


cider-interaction.el
```diff
@@ -709,11 +710,14 @@ or it can be a list with (START END) of the evaluated region."
...
          (end (or (car-safe (cdr-safe place)) place))
          (beg (when beg (copy-marker beg)))
-         (end (when end (copy-marker end))))
+         (end (when end (copy-marker end)))
+         (fringed nil))
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  (lambda (_buffer value)
                                    (if beg
-                                       (cider--make-fringe-overlays-for-region beg end)
+                                       (when (not fringed)
```
bbatsov: `unless`

bbatsov:
This needs a changelog entry.

phillord:
Done

cider-interaction.el
```diff
@@ -709,11 +710,14 @@ or it can be a list with (START END) of the evaluated region."
          (beg (car-safe place))
          (end (or (car-safe (cdr-safe place)) place))
          (beg (when beg (copy-marker beg)))
-         (end (when end (copy-marker end))))
+         (end (when end (copy-marker end)))
+         (fringed nil))
```
bbatsov: I'm just not sure this is the right way to check if there are overlays there or not.

xiongtx: I'd expect functions that create overlays to be idempotent, instead of relying on flags like `fringed`. That is, if `cider--make-fringe-overlays-for-region` is called a second time on the same region, it detects that the overlay already exists and does nothing.

phillord: I think that is a good idea, and the other hunk of the commit tries to address this (belts and braces). But, even here there is a performance issue. `make-fringe-overlays` makes many overlays, not just one. So, at a minimum, you have to move through every top-level sexp and check for an overlay at each point.

phillord: @bbatsov Alternatives?

bbatsov: How about this https://www.gnu.org/software/emacs/manual/html_node/elisp/Finding-Overlays.html ? Seems it would get the job done.

phillord: No, it doesn't help. I have already supplemented `cider--make-fringe-overlays-for-region` with a `delete-overlays`, so that it should be idempotent.

But this does not change the fact that this closure is getting called multiple times. Try adding:

    (princ "call\n" #'external-debugging-output)

as the first form in the lambda. Now evaluating a region with a 100 top level forms in it. The closure gets called 100 times. So, even if we change `cider--make-fringe-overlays` it is still going to get called repeatedly. And yet, for the lifetime of this closure, beg and end never change. Why call a function 100 times with exactly the same values?
Even if that function is works well, it is still going to be slow.

Other options 
  - put the call to `--make-fringe-overlays` outside of the response handler.
  - change the middleware so that it returns the region also
  - analyse "result" so that top-level forms are "fringed" as they are evaled

xiongtx: I may be missing something here, but I'd think that *checking* for an overlay is much less expensive than *applying* an overlay. If the fn is called 100 times and the overlay is applied the first time, but not the remaining 99 times (during which we'd check that the overlay exists in the region and do nothing), is there still a performance issue?

phillord: The main problem with overlays is that the slow insertion/deletion of text anywhere in the buffer. But we are not talking about a single overlay here: we are talking about one for each sexp. Each time the function is called, we check every sexp in the region. If it's called 100 times, this means that the are 100 sexps in the region, so we check  for overlays 10,000 times. So, performance is still quadratic during the evaluation, but it should at least not trash Emacs by leaving many overlays in the buffer.

On a seperate note, probably Emacs needs a soft barrier for the number of overlays going into a buffer.

phillord: Any further thoughts?

bbatsov: Nothing from me. I'd simply prefer a solution that doesn't rely on a closure-captured variable as such code would be easier to comprehend.

phillord: Well, `beg` and `end` are also captured; I just copied what was already there. Regardless, are you happy to accept the PR now?

bbatsov: Yeah, I'll take it as is. 

Just add a changelog entry about the bugfix.

bbatsov: Ah, there's already one.

phillord: Thanks!

PR #1988, opened 2017-05-09
Update various faces when a theme is disabled

### LINKED ISSUE #1987: When disabling a theme, faces are not updated accordingly.
Previously, we were only updating faces when we enable themes.

We need to update the faces whenever a theme is disabled, otherwise after disabling a dark theme like molokai, the background colors are still black in the default emacs theme.

## Expected behavior

When disabling a theme, backgroud colors of error reports etc should be re-calculated.

## Actual behavior

Previous theme's colors are taken into account when disabling a theme.

Enable a theme on emacs startup. (possibly a dark one)
Start cider.
Disable theme.
Run a failing test.
The background color for tests still are like that of the disabled theme.

Before

image

After 

image

### CIDER version information

```
;; CIDER 0.15.0snapshot (package: 20170506.1322), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_45

```

### Lein/Boot version

Leiningen 2.7.1 on Java 1.8.0_45 Java HotSpot(TM) 64-Bit Server VM

### Emacs version

GNU Emacs 25.2.1 (x86_64-apple-darwin16.5.0, Carbon Version 157 AppKit 1504.82) of 2017-04-28

### Operating system

macOS

ustun:
Fixes #1987

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -135,6 +135,11 @@
   "When theme is changed, update `cider-docview-code-background-color'."
   (setq cider-docview-code-background-color (cider-scale-background-color)))
 
+
+(defadvice disable-theme (after cider-docview-adapt-to-theme activate)
+  "When theme is disabled, update `cider-docview-code-background-color'."
+  (setq cider-docview-code-background-color (cider-scale-background-color)))
+
 
 ;; Mode & key bindings
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 
 ### Bugs Fixed
 
+* [#1987](https://github.com/clojure-emacs/cider/issues/1987): Fix: Update faces when disabling a theme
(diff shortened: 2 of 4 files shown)


bbatsov:
👍

cider-doc.el
```diff
@@ -135,6 +135,11 @@
   "When theme is changed, update `cider-docview-code-background-color'."
   (setq cider-docview-code-background-color (cider-scale-background-color)))
 
+
+(defadvice disable-theme (after cider-docview-adapt-to-theme activate)
```
xiongtx: Why not use `advice-add` from `nadvice.el`? It's the [recommended way](https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Named-Functions.html#Advising-Named-Functions) for newer versions of Emacs.

bbatsov: The thought crossed my mind as well, but we'd have to update the existing code for consistency as well. We can always do this down the road.

PR #1994, opened 2017-05-20
[#1990] Adds option to auto-save buffers when calling cider-refresh

### LINKED ISSUE #1990: Save all project files when `cider-refresh` is invoked
When developing on CIDER, I often invoke `cider-refresh`. `cider-refresh` reads the files from the filesystem and sometimes in Emacs, you may not have saved all the files you've touched. As a result, you may see a confusing compilation error message. 

Is it possible to add the ability to autosave open buffers when `cider-refresh` is invoked?

rymndhng:
Fixes #1990. Adds an option to auto-save buffers when calling cider-refresh. There is an option `cider-prompt-save-file-on-cider-refresh` which allows users to customize this behaviour. By default, the user is prompted for each clojure buffer to save.

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -111,6 +111,17 @@ If the symbol `always-save', save the file without confirmation."
   :group 'cider
   :package-version '(cider . "0.6.0"))
 
+(defcustom cider-prompt-save-file-on-cider-refresh t
+  "Controls whether to prompt to save clojure file when calling ‘cider-refresh’.
+If nil, files are not saved.
+If t, the user is prompted to save files if it's been modified.
+If the symbol `always-save', save the files without confirmation."
+  :type '(choice (const t :tag "Prompt to save files if it's been modified")
+                 (const nil :tag "Don't save the files")
+                 (const always-save :tag "Save the files without confirmation"))
+  :group 'cider
+  :package-version '(cider . "0.15.0"))
+
 (defcustom cider-completion-use-context t
   "When true, uses context at point to improve completion suggestions."
   :type 'boolean
@@ -1541,6 +1552,9 @@ refresh functions (defined in `cider-refresh-before-fn' and
   (let ((clear? (member mode '(clear 16)))
         (refresh-all? (member mode '(refresh-all 4)))
         (inhibit-refresh-fns (member mode '(inhibit-fns -1))))
+    (when cider-prompt-save-file-on-cider-refresh
+      (save-some-buffers (eq cider-prompt-save-file-on-cider-refresh 'always-save)
+                         (lambda () (derived-mode-p 'clojure-mode))))
     (cider-map-connections
      (lambda (conn)
        ;; Inside the lambda, so the buffer is not created if we error out.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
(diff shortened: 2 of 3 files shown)


dpsutton:
The functionality probably doesn't need to be in the body of that let, and it would make sense if the first few lines were

```
(cider-ensure-connected)
(cider-ensure-op-supported "refresh")
(cider--offer-save-project-files)
(let ((....
```

This would look nicer as a function so we could sprinkle this around in other locations as well that I'm sure will make sense.

One area of concern: in an IDE we would easily and care-free save all files. In emacs, it's possible we span multiple projects. Should we worry about filtering down to just the current project (using perhaps `nrepl-project-dir` or `(clojure-project-dir)`) so that we do not save buffers that are not involved in the refresh operation or is that an overblown concern?

cider-interaction.el
```diff
@@ -1541,6 +1552,9 @@ refresh functions (defined in `cider-refresh-before-fn' and
   (let ((clear? (member mode '(clear 16)))
         (refresh-all? (member mode '(refresh-all 4)))
         (inhibit-refresh-fns (member mode '(inhibit-fns -1))))
+    (when cider-prompt-save-file-on-cider-refresh
```
bbatsov: That should limited only to project-specific buffers. It would be odd to save all Clojure buffers on refresh IMO.

bbatsov: Not sure if we had some function returning those, implementing something like this should trivial.

bbatsov:
@dpsutton is completely right.

dpsutton:
I haven't felt the pain of this bug as I run prelude and have the autosaving all buffers on by default when switching. Perhaps this is an avenue that we could pursue? Rather than ensuring a particular project's files are saved, allow the user to enable CIDER to always save all buffers. This issue would seemingly go away, and we don't need to worry about project boundaries since the user is the one that gives explicit approval for CIDER to save all buffers all the time?

The functionality is done in prelude through the use of advice. It works well but i've heard function advice is the unwelcome part of emacs lisp, but here's the [functionality](https://github.com/bbatsov/prelude/blob/master/core/prelude-editor.el#L174).

I'm not saying that this is necessarily the correct way forward but it does skirt dealing with comparing file locations and notions of projects while making refreshing easier.

bbatsov:
> I'm not saying that this is necessarily the correct way forward but it does skirt dealing with comparing file locations and notions of projects while making refreshing easier.

I'm not sure what's the Prelude functionality you're referring to. I know I'll be extremely annoyed if I get warning about the unsaved source buffers in unrelated projects. I also know that getting this right is trivial, so I don't see the point in not doing it properly.

dpsutton:
the functionality i was talking about was prelude autosaving files when they lose focus:

```
(advise-commands "auto-save"
                 (switch-to-buffer other-window windmove-up windmove-down windmove-left windmove-right)
                 before
                 (prelude-auto-save-command))
```

I was just offering a different approach that would get us the same answers. I wouldn't have it ask but just continually save all clj(s|c) as you leave buffers.

> I also know that getting this right is trivial, so I don't see the point in not doing it properly.
True. I only wanted to point out a possible alternative.

bbatsov:
> I was just offering a different approach that would get us the same answers. I wouldn't have it ask but just continually save all clj(s|c) as you leave buffers.

I wrote this a long time ago, but I recall it actually saved just the buffer you were into. It's pointless to save all buffers when you lose focus.

bbatsov:
I'll merge this and improve it myself.

rymndhng:
@bbatsov thanks for taking it on. Been a bit busy and haven't had the time to follow up on improving my elisp skills!

Issue #1982, opened 2017-04-28
[Feature request] Integrate CIDER doc displaying with quick-peek or inline-docs


stardiviner:
Hope CIDER can integrate `quick-peek` or `inline-docs` for functions like query docs.

This can be set in an option.

- https://github.com/cpitclaudel/quick-peek
- https://github.com/stardiviner/inline-docs.el

bbatsov:
I'm assuming integrating something like this should be trivial, but unless a lot of people demand such a feature it should probably be implemented as a separate package, so that we don't introduce more dependencies to CIDER.

stardiviner:
I see, I will dive into CIDER source code to check out this. If have any question, will ask for help here. Thanks.

bbatsov:
Sure.

bbatsov:
Btw, I've noticed both libraries you've mentioned are tiny, so I'm not opposed that much to depending on one of them. I'm curious about the invisioned workflow for the inline docs, thought - would this a replacement for eldoc or just a command bound to some keystroke? Would we should just prototypes or docstrings and metadata as well? Guess those are the most important things to consider.

stardiviner:
Those two libraries are just a library, not functional package which can be used directly. There is an existing package which uses `inline-docs.el`. https://github.com/stardiviner/eldoc-overlay-mode. It uses as a replacement of `eldoc` default echo-area display. It can be used for other places too. Just change the way of people viewing their information. Don't need to move eye into bottom echo-area. Just viewing info at inline place.

bbatsov:
I guess we can implement some command in terms of `inline-docs.el` and also mention in the eldoc section of the manual `eldoc-overlay-mode`.

stardiviner:
That's great, this is my first time my Emacs library been used in large project. Happy coding.
Because I don't know `CIDER` much, hope someone can pick this up. and create an branch for this? I will try to dive into the cider source code. Thanks.

bbatsov:
Everything that needs doing is pretty trivial as all the documentation and eldoc code is in `cider-doc.el` and `cider-eldoc.el` and the data simply needs to be presented with the new library. I'll mark this as a long-hanging fruit in case some newcomer is interested in tackling it.

vemv:
PRs welcome for whatever support work is necessary for those libraries to be integrated.

CIDER already has a variety of "APIs" that Elisp hackers can leverage to get something working.

PR #1995, opened 2017-05-21
Added cider-docview-select-buffer custom variable

nahuel:
Added cider-docview-select-buffer custom variable to control if the documentation viewer popup will be autoselected after opening.

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -96,6 +96,11 @@
   :group 'cider-docview-mode
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-docview-select-buffer t
+  "Autoselect the documentation viewer after popup."
+  :type 'boolean
+  :group 'cider-docview-mode
+  :package-version '(cider . "0.7.0"))
 
 
 ;; Faces
@@ -263,7 +268,7 @@ opposite of what that option dictates."
 (defun cider-doc-lookup (symbol)
   "Look up documentation for SYMBOL."
   (if-let ((buffer (cider-create-doc-buffer symbol)))
-      (cider-popup-buffer-display buffer t)
+      (cider-popup-buffer-display buffer cider-docview-select-buffer)
     (user-error "Symbol %s not resolved" symbol)))
 
 (defun cider-doc (&optional arg)


cider-doc.el
```diff
@@ -96,6 +96,11 @@
   :group 'cider-docview-mode
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-docview-select-buffer t
```
bbatsov: That's not really related to `cider-docview-mode`.

cider-doc.el
```diff
@@ -96,6 +96,11 @@
...
+  "Autoselect the documentation viewer after popup."
+  :type 'boolean
+  :group 'cider-docview-mode
+  :package-version '(cider . "0.7.0"))
```
bbatsov: And it wasn't introduced in CIDER 0.7. :-)

cider-doc.el
nahuel: what about changing group to `cider-doc`?

bbatsov:
The name of the variable is misleading as it's not really related to docview mode at all. It's controlling whether the doc buffer would be selected or not. I'm curious why do you want this? You know pressing `q` will dispose of the buffer, right?

nahuel:
So maybe naming it as `cider-doc-select-buffer`?

When browsing code I find useful to map `[f1]` to `cider-doc` and press it over unknown functions to get a quick glance of their doc in an always opened secondary window. I switch quickly between functions docs, so if the window autoselects (the current behaviour) I must do extra steps to return to the code buffer. Pressing `q` will close the doc buffer but leaves the window opened and selected, and is also an extra step.

A better yet behavior for me is to have the doc window always opened and make it autorefresh with the documentation for the function at point, without pressing any key (and without autoselecting the doc window).

bbatsov:
I see. That workflow is a bit odd and doesn't sound like something pleasant to me. I think you actually need https://github.com/clojure-emacs/cider/issues/1982

nahuel:
the libs on #1982 open the documentation inside the code buffer, I like to get it always in the same separated emacs window as cider does with my described workflow.

bbatsov:
> the libs on #1982 open the documentation inside the code buffer, I like to get it always in the same separated emacs window as cider does with my described workflow.

Isn't it preferable not to need a dedicated buffer for this? Seems like a waste of space to me, but I might be missing something...

nahuel:
Is a matter of personal preference, I like to have an "static" layout, so my docs are always in the same place and not popping over the code buffer or opening/closing windows.

bbatsov:
I see. Well, I'd make this consistent with:

```
(defcustom cider-auto-select-error-buffer t
  "Controls whether to auto-select the error popup buffer."
  :type 'boolean
  :group 'cider)
```

so `cider-doc-auto-select-buffer` maybe?

This should be reflected in the changelog as well.

cider-doc.el
```diff
@@ -96,6 +96,11 @@
   :group 'cider-docview-mode
   :package-version '(cider . "0.7.0"))
 
+(defcustom cider-docview-select-buffer t
+  "Autoselect the documentation viewer after popup."
+  :type 'boolean
+  :group 'cider-docview-mode
```
bbatsov: This should be `cider-doc`.

nahuel:
@bbatsov PR was autoclosed, but I fixed the code here in the updated aceab734bc675753484a91b14bdb2aa1b5f120f5

bbatsov:
Feel free to open a new PR.

PR #1997, opened 2017-05-24
fix `find-file-noselect: Wrong type argument: stringp, nil` 

hojinyoo:
- fix a bug that losing `buffer-file-name` after call `(cider--quit-error-window)`
The bug yields an error message `find-file-noselect: Wrong type argument: stringp, nil'  that prevents from loading buffer.
-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1590,12 +1590,10 @@ ClojureScript REPL exists for the project, it is evaluated in both REPLs."
                (or (eq cider-prompt-save-file-on-load 'always-save)
                    (y-or-n-p (format "Save file %s? " buffer-file-name))))
       (save-buffer))
-
     (remove-overlays nil nil 'cider-temporary t)
     (cider--clear-compilation-highlights)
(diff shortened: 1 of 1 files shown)


bbatsov:
I can't really understand how moving the code as you did would fix anything. Some details would be useful.

hojinatmixpanel:
`(cider--quit-error-window)` has a side-effect, at least emacs in Mac OS X. Once the error window is "buried" by the call, the return value of function `buffer-file-name` becomes nil.

bbatsov:
Ah, I get it now - quitting the error buffer probably selects it. Now I understand why this fix works, but I don't think that's the best possible fix as this makes the code pretty brittle - that's going to break again the moment someone rearranges the code. I think that passing `buffer` to `buffer-file-name` is going to fix this problem in a much clear and robust manner.

hojinyoo:
cool. that works as well.

PR #2001, opened 2017-06-02
Add NREPL support for user level messages and warnings

vspinu:
Accompanies clojure-emacs/cider-nrepl#408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#2015](https://github.com/clojure-emacs/cider/pull/2015): Show symbols as special forms *and* macros in `cider-doc`
 * [#2012](https://github.com/clojure-emacs/cider/pull/2012): Support special forms in `cider-apropos` and `cider-grimoire-lookup`.
 * [#2007](https://github.com/clojure-emacs/cider/pull/2007): Fontify code blocks from `cider-grimoire` if possible.
+* Add support for notifications from the NREPL server.
 * [#1990](https://github.com/clojure-emacs/cider/issues/1990): Add new customation variable `cider-save-files-on-cider-refresh` to allow auto-saving buffers when `cider-refresh` is called.
 * Add new function `cider-load-all-files`, along with menu bar update.
 * Add new customization variable `cider-special-mode-truncate-lines`.
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -736,6 +736,25 @@ It is safe to call this function multiple times on the same ID."
     (puthash id handler nrepl-completed-requests)
     (remhash id nrepl-pending-requests)))
 
+(declare-function cider-repl--emit-interactive-output "cider-repl")
+(defun nrepl-notify (msg type)
+  "Handle \"notification\" server request.
+MSG is a string to be displayed.  TYPE is the type of the message.  All
+notifications are currently displayed with `message' function and emitted
+to the REPL."
+  (let* ((face (pcase type
+                 ((or "message" 'nil) 'font-lock-builtin-face)
+                 ("warning" 'warning)
+                 ("error"   'error)))
+         (msg (if face
+                  (propertize msg 'face face)
+                (format "%s: %s" (upcase type) msg))))
+    (cider-repl--emit-interactive-output msg (or face 'font-lock-builtin-face))
+    (message msg)
+    ;; Interactive eval handler covers this message, but it won't be eval
+    ;; middleware using this functionality.
+    (sit-for 2)))
+
 (defvar cider-buffer-ns)
 (defvar cider-special-mode-truncate-lines)
 (declare-function cider-need-input "cider-interaction")
@@ -779,6 +798,9 @@ corresponding type of response."
              (when stderr-handler
                (funcall stderr-handler buffer err)))
             (status
+             (when (member "notification" status)
+               (nrepl-dbind-response response (msg type)
+                 (nrepl-notify msg type)))
              (when (member "interrupted" status)
                (message "Evaluation interrupted."))
              (when (member "eval-error" status)


nrepl-client.el
```diff
@@ -772,6 +785,9 @@ corresponding type of response."
              (when stderr-handler
                (funcall stderr-handler buffer err)))
             (status
+             (when (member "message" status)
```
bbatsov: I wonder if this shouldn't be `notification` or something, to avoid confusion with the term `nREPL message`.

nrepl-client.el
```diff
@@ -729,6 +729,19 @@ It is safe to call this function multiple times on the same ID."
     (puthash id handler nrepl-completed-requests)
     (remhash id nrepl-pending-requests)))
 
+(defun nrepl-display-user-message (msg type)
+  (let* ((head (propertize
+                (if type
+                    (format "NREPL %s: " (upcase type))
```
bbatsov: What's the point of prefixing those with `nREPL`?

nrepl-client.el
```diff
@@ -729,6 +729,19 @@ It is safe to call this function multiple times on the same ID."
     (puthash id handler nrepl-completed-requests)
     (remhash id nrepl-pending-requests)))
 
+(defun nrepl-display-user-message (msg type)
+  (let* ((head (propertize
+                (if type
```
bbatsov: I'd still check what exactly is the type - right now the code looks a bit odd.

nrepl-client.el
```diff
@@ -729,6 +729,19 @@ It is safe to call this function multiple times on the same ID."
...
+                    (format "NREPL %s: " (upcase type))
+                  "NREPL: ")
+                'face 'cider-repl-stderr-face))
+         (msg (format "%s%s" head msg)))
+    (cider-repl-emit-interactive-stderr msg)
```
bbatsov: Do we really want to output this in the REPL as well? Especially font-locked as an error?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* Add support for user level messages from the NREPL server.
```
bbatsov: nREPL :-)

nrepl-client.el
vspinu: I don't see a major conflict here. nREPL message is the whole thing. But `notify` or `user-message` could also be.

nrepl-client.el
vspinu: Because you want to emphasize that message came from the NREPL/clojure and not from the some arbitrary code on emacs side.

nrepl-client.el
vspinu: I am not sure I understand. There is nothing to check at the moment. All messages irrespective of the type are displayed with `(message ...)`.

nrepl-client.el
vspinu: The problem is that on interactive eval the message is immediately covered with the output of the evaluation. So in that case it's easy to miss the message unless it's also in the repl. I have a inline comment on that; not sure how to fix that. 

I think it's rather consistent to pipe all nREPL messages into repl buffer irrespective of their source.  In case of trouble, first and foremost users will check repl buffer.

vspinu: > Especially font-locked as an error?

I don't see more appropriate font. Fontifying it as output is out of option either.

nrepl-client.el
bbatsov: I think this would be somewhat confusing as CIDER typically doesn't prefix message (and most Emacs apps don't do this as well). Users shouldn't really care where a message originated from IMO.

nrepl-client.el
bbatsov: It's not a big deal, I just think we can make it a bit clearer.

nrepl-client.el
vspinu: How about printing to repl, should messages originating form nREPL middleware be qualified as such?

nrepl-client.el
bbatsov: We can easily add something new - e.g. something `green` for informational message. Guess for errors it's ok to use the error face.

nrepl-client.el
bbatsov: Well, technically speaking pretty much everything they get in the REPL originated from nREPL.

nrepl-client.el
xiongtx: What types are there? A docstring describing the parameters would be useful.

bbatsov:
@vspinu Next step - wrapping up this. 😉

bbatsov:
@vspinu Ping :-)

vspinu:
Sorry for stacking this one. I will be back on it by the end of the week.

vspinu:
Done with this one. I think calling it "notification" was a good idea. Will need to update server side for this.

nrepl-client.el
```diff
@@ -736,6 +736,25 @@ It is safe to call this function multiple times on the same ID."
...
+notifications are currently displayed with `message' function and emitted
+to the REPL."
+  (let* ((face (pcase type
+                 ((or "message" 'nil) 'font-lock-builtin-face)
+                 ("warning" 'warning)
+                 ("error"   'error)))
+         (msg (if face
+                  (propertize msg 'face face)
+                (format "%s: %s" (upcase type) msg))))
+    (cider-repl--emit-interactive-output msg (or face 'cider-repl-message-face))
```
bbatsov: Where's this face defined? I don't see in this PR.

nrepl-client.el
```diff
@@ -736,6 +736,25 @@ It is safe to call this function multiple times on the same ID."
     (puthash id handler nrepl-completed-requests)
     (remhash id nrepl-pending-requests)))
 
+(declare-function cider-repl--emit-interactive-output "cider-repl")
+(defun nrepl-notify (msg type)
+  "Handle 'notification' server request.
```
bbatsov: I think there are no single quoted strings in Elisp.

nrepl-client.el
```diff
@@ -736,6 +736,25 @@ It is safe to call this function multiple times on the same ID."
...
+MSG is a string to be displayed. TYPE is the type of the message. All
```
bbatsov: You forgot about the double spaces between sentences.

bbatsov:
Looks good, apart from a few tiny remarks I added.

nrepl-client.el
vspinu: Argh. It's from previous version of the PR. I removed it and replaced with hard-coded faces because this small feature doesn't warant new faces IMO.

nrepl-client.el
vspinu: > You forgot about the double spaces between sentences.

Have you considered either relaxing double space assumption or adding a checkdoc to the tests?  I bet you have much more interesting things to do than hunting dots and spaces in peoples' code ;)

vspinu:
Fixed.

nrepl-client.el
xiongtx: Gotta [sweat the small stuff](https://en.wikipedia.org/wiki/Broken_windows_theory).

bbatsov: I don't want to go against established conventions (even if I don't like them that much). But there should definitely be an automated check for this at some point. As a matter of fact I was under the impression we added one a while ago, but I guess I was mistaken. I don't want to waste my time on pointing out doc issues myself. :-)

PR #2002, opened 2017-06-03
Accompanying changes to  clojure-emacs/cider-nrepl#405

vspinu:
Adjustments needed for clojure-emacs/cider-nrepl#405.

So far there is only one change - simplification of instrumentation metadata received from the server. I cut simplified metadata on the clojure side as it inflates the method side considerably. Also removing the namespace qualification which looks like a spill over of internals from clojure side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 
 ### Bugs Fixed
 
+* [#1699](https://github.com/clojure-emacs/cider/issues/1699): Fix "Method code too large!" error that occurred during instrumentation for debugging.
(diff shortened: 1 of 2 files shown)


bbatsov:
You should add a changelog for the debugger improvements as part of this PR.

bbatsov:
> Also removing the namespace qualification which looks like a spill over of internals from clojure side.

I think they were fully-qualified to avoid potential conflicts, but I guess those are rather unlikely. Even `cider-something` would be enough IMO.

vspinu:
I think `:cider/instrumented` would be the best option.

bbatsov:
> I think :cider/instrumented would be the best option.

No argument from me.

PR #2007, opened 2017-06-10
Fontify Grimoire code blocks if possible

xiongtx:
Follow-up on #1959 and the [discussion](https://github.com/clojure-emacs/cider/pull/1959#issuecomment-291918454) with @bbatsov r.e. fontifying code blocks in the markdown buffer returned by `cider-grimoire`.

Thanks to @jrblevin for addressing jrblevin/markdown-mode#185.

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -35,6 +35,9 @@
 
 (require 'url-vars)
 
+(declare-function markdown-mode "markdown-mode.el")
+(declare-function markdown-toggle-fontify-code-blocks-natively "markdown-mode.el")
+
 (defconst cider-grimoire-url "http://conj.io/")
 
 (defconst cider-grimoire-buffer "*cider-grimoire*")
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good! Just add a changelog entry, refer in the commit message to the ticket this is solving and take care of the byte-compilation warnings.

xiongtx:
I'm not 100% sure if `with-no-warnings` is the right way to address byte-compiler warnings in this situation. I've posted a [question on the Emacs StackExchange](https://emacs.stackexchange.com/questions/33445/prevent-byte-compiler-warnings-after-requiring-features) regarding this.

I'm also not sure why [build 4636.5](https://travis-ci.org/clojure-emacs/cider/jobs/241424471) failed when the others pass.

cider-grimoire.el
```diff
@@ -35,6 +35,8 @@
 
 (require 'url-vars)
 
+(declare-function markdown-mode "markdown-mode.el")
```
bbatsov: You also have to declare `markdown-toggle-fontify-code-blocks-natively`.

xiongtx: :white_check_mark:

bbatsov: One more thing - for this work `markdown-mode` should actually be present. You can just add `markdown-mode` to the dev dependencies in Cask.

xiongtx: 👍

bbatsov:
👍

PR #2012, opened 2017-06-16
Apropos special forms

xiongtx:
These changes are necessary to use clojure-emacs/cider-nrepl#409.

Example of special forms from `cider-apropos`:

image

### DIFF
--- a/cider-grimoire.el
+++ b/cider-grimoire.el
@@ -91,10 +91,13 @@ opposite of what that option dictates."
     (current-buffer)))
 
 (defun cider-grimoire-lookup (symbol)
-  "Look up the grimoire documentation for SYMBOL."
+  "Look up the grimoire documentation for SYMBOL.
+
+If SYMBOL is a special form, the clojure.core ns is used, as is
+Grimoire's convention."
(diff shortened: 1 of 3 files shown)


cider-grimoire.el
```diff
@@ -91,10 +91,13 @@ opposite of what that option dictates."
     (current-buffer)))
 
 (defun cider-grimoire-lookup (symbol)
-  "Look up the grimoire documentation for SYMBOL."
+  "Look up the grimoire documentation for SYMBOL.
+
+If SYMBOL is a special form, the `clojure.core' ns is used, as is
```
bbatsov: Don't quote `clojure.core` here - this is just for Elisp symbols.

bbatsov:
You should also add a changelog entry.

xiongtx:
What's your general rule for releasing new version of `cider-nrepl` and `CIDER`? After some time or some number of commits?

bbatsov:
> What's your general rule for releasing new version of cider-nrepl and CIDER? After some time or some number of commits?

There are no special rules - I release new versions when it seems to me that this is warranted. Usually I aim to cut major releases around major Clojure conferences.

PR #2015, opened 2017-06-18
Indicate both special form and macro in doc

xiongtx:
Some symbols, like `clojure.core/let`, are special forms AND macros. `cider-doc` should indicate this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### New Features
 
-* [#2012](https://github.com/clojure-emacs/cider/pull/2007): Support special forms in `cider-apropos` and `cider-grimoire-lookup`.
+* [#2015](https://github.com/clojure-emacs/cider/pull/2015): Show symbols as special forms *and* macros in `cider-doc`
+* [#2012](https://github.com/clojure-emacs/cider/pull/2012): Support special forms in `cider-apropos` and `cider-grimoire-lookup`.
 * [#2007](https://github.com/clojure-emacs/cider/pull/2007): Fontify code blocks from `cider-grimoire` if possible.
 * [#1990](https://github.com/clojure-emacs/cider/issues/1990): Add new customation variable `cider-save-files-on-cider-refresh` to allow auto-saving buffers when `cider-refresh` is called.
 * Add new function `cider-load-all-files`, along with menu bar update.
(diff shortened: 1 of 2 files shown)


bbatsov:
I'm afraid I don't understand how this can be. Something is either a macro or a special form, can't see how it can be both...

xiongtx:
Consider `clojure.core/let`:

```clojure
user> (meta #'let)
{:added "1.0",
 :ns #namespace[clojure.core],
 :name let,
 :special-form true,
 :file "clojure/core.clj",
 :column 1,
 :line 4325,
 :macro true,
 :arglists ([bindings & body]),
 :doc
 "binding => binding-form init-expr\n\n  Evaluates the exprs in a lexical context in which the symbols in\n  the binding-forms are bound to their respective init-exprs or parts\n  therein.",
 :forms [(let [bindings*] exprs*)]}
```

Note that `:special-form true` *and* `:macro true`.

The other ones in `clojure.core` are:

```clojure
(->> (vals (ns-publics 'clojure.core)) 
     (map meta)
     (filter :special-form)
     (map :name))
     
;; => (let fn letfn loop)      
```

bbatsov:
Yeah, I know this, but I just don't understand how is it possible and whether we want to show this to the users. Might be a bit confusing...

@arrdem As a documentation expert what's your take on this?

arrdem:
So my $0.02 is that `let` is a macro. Technically it's `let*` which is the special form (ref. [Compiler.java](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Compiler.java#L44), [impl](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Compiler.java#L6259), [bootstrap let](https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj#L31-L35), [destructuring let](https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj#L4433-L4444)).

The fact that underneath `let` is the `let*` special form which actually establishes bindings is, to me, fairly worthless. `let*` is a base case, required by the fact that Clojure (and most traditional Lisps) macro systems depend on symbol equality/identity to determine whether to continue macroexpansion or not.

This wasn't done deliberately but it turns out that Grimoire doesn't document `let*`, just the destructuring let macro we all know and love. Likewise `fn*`.

xiongtx:
My view is that the user should be given the information that the developer provides. If `let` if marked as both macro *and* special form, that's what should be displayed.

> `let*` is total magic. It doesn't resolve. It isn't namespaced. It only exists in `Compiler.java`.

Right...that's why it's a special form after all.

xiongtx:
Added changelog entry.

PR #2021, opened 2017-06-28
Handle ANSI REPL evaluation created by Puget.

plandes:
When `cider-pprint-fn'  is set to `puget' and puget is configured to create color an ANSI string is given in the REPL output after evaluating any expression.  This change uses the (already included by at least 25.1.1 Emacs dist) `ansi-color' library.

## More comprehensive possible future change

By default Puget allows configuration *only* through dynamic variables, which would make configuration hard via Cider so I wrote [this package](https://github.com/plandes/clj-nrepl-puget) to allow this.  I'd be happy to work with someone to get this folded into Cider if that interests the maintainer(s).

I could attempt to do it myself if you'd like.

Thanks and regards,
- Paul

-----------------

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -258,7 +258,7 @@ This buffer is not designed to display anything to the user.  For that, use
 (defun cider-font-lock-as (mode string)
   "Use MODE to font-lock the STRING."
   (let ((string (if (cider-ansi-color-string-p string)
-                    (ansi-color-apply string)
+                    (substring-no-properties (ansi-color-apply string))
                   string)))
     (if (or (null cider-font-lock-max-length)
             (< (length string) cider-font-lock-max-length))
--- a/cider-util.el
+++ b/cider-util.el
@@ -36,6 +36,7 @@
 (require 'subr-x)
 (require 'cider-compat)
 (require 'nrepl-dict)
+(require 'ansi-color)
 
 (defalias 'cider-pop-back 'pop-tag-mark)
 
@@ -250,16 +251,23 @@ This buffer is not designed to display anything to the user.  For that, use
           (funcall mode))
         b)))
 
+(defun cider-detect-ansi-p (string)
+  "Return non-nil if STRING is an ANSI string."
+  (string-match "^\\[" string))
+
 (defun cider-font-lock-as (mode string)
   "Use MODE to font-lock the STRING."
-  (if (or (null cider-font-lock-max-length)
-          (< (length string) cider-font-lock-max-length))
-      (with-current-buffer (cider--make-buffer-for-mode mode)
-        (erase-buffer)
-        (insert string)
-        (font-lock-fontify-region (point-min) (point-max))
-        (buffer-string))
-    string))
+  (let ((string (if (cider-detect-ansi-p string)
+                    (ansi-color-apply string)
+                  string)))
+    (if (or (null cider-font-lock-max-length)
+            (< (length string) cider-font-lock-max-length))
+        (with-current-buffer (cider--make-buffer-for-mode mode)
+          (erase-buffer)
+          (insert string)
+          (font-lock-fontify-region (point-min) (point-max))
+          (buffer-string))
+      string)))
 
 (defun cider-font-lock-region-as (mode beg end &optional buffer)
   "Use MODE to font-lock text between BEG and END.
--- a/cider-util.el
+++ b/cider-util.el
@@ -257,7 +257,7 @@ This buffer is not designed to display anything to the user.  For that, use
 
 (defun cider-font-lock-as (mode string)
   "Use MODE to font-lock the STRING."
-  (let ((string (if (cider-ansi-color-string-p) string
+  (let ((string (if (cider-ansi-color-string-p string)
                     (ansi-color-apply string)
                   string)))
     (if (or (null cider-font-lock-max-length)
--- a/cider-util.el
+++ b/cider-util.el
@@ -1,4 +1,4 @@
-;;; cider-util.el --- Common utility functions that don't belong anywhere else -*- lexical-binding: t -*-
+;; cider-util.el --- Common utility functions that don't belong anywhere else -*- lexical-binding: t -*-
 
 ;; Copyright © 2012-2013 Tim King, Phil Hagelberg, Bozhidar Batsov
 ;; Copyright © 2013-2017 Bozhidar Batsov, Artur Malabarba and CIDER contributors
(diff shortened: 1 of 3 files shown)


cider-util.el
```diff
@@ -250,16 +251,23 @@ This buffer is not designed to display anything to the user.  For that, use
           (funcall mode))
         b)))
 
+(defun cider-detect-ansi-p (string)
```
bbatsov: Probably `cider-contains-ascii-p` would be a better name or. Or `cider-ansi-color-string-p`?

bbatsov:
> By default Puget allows configuration only through dynamic variables, which would make configuration hard via Cider so I wrote this package to allow this. I'd be happy to work with someone to get this folded into Cider if that interests the maintainer(s).
> 
> I could attempt to do it myself if you'd like.

Sure. I just don't understand the need for the custom middleware. Seems to me you just want to have some simple API/UI for setting dynamic vars, which doesn't force you to use a REPL.

plandes:
That was also my initial thought but I'm a little concerned about having to lexically wrap each request, particularly when a binding could get a little beefy.  It could look something like this:

```clojure
(let [puget/*options*
      {:print-color true
       :color-scheme {:keyword [:none]
                      :delimiter [:blue]
                      :number [:magenta]
                      :tag [:black]
                      :symbol [:red]
                      :string [:green]
                      :character [:bold :green]
                      :boolean [:black]
                      :nil [:black]}}]
  (user-eval-sexps))
```

cider-util.el
plandes: I like the second function name as well.

cider-util.el
```diff
@@ -250,16 +251,23 @@ This buffer is not designed to display anything to the user.  For that, use
...
   "Use MODE to font-lock the STRING."
-  (if (or (null cider-font-lock-max-length)
-          (< (length string) cider-font-lock-max-length))
-      (with-current-buffer (cider--make-buffer-for-mode mode)
-        (erase-buffer)
-        (insert string)
-        (font-lock-fontify-region (point-min) (point-max))
-        (buffer-string))
-    string))
+  (let ((string (if (cider-ansi-color-string-p) string
```
bbatsov: Can you spot the mistake here? :-) `string` should actually be the argument to `cider-ansi-color-string-p`.

plandes: Ooops, paredit fat finger mistake.  Sorry about that.

cider-util.el
```diff
@@ -250,16 +251,23 @@ This buffer is not designed to display anything to the user.  For that, use
           (funcall mode))
         b)))
 
+(defun cider-ansi-color-string-p (string)
```
bbatsov: One more thing - you should a unit test for this function and some extra unit test for `cider-font-lock-as`.

plandes: I've added a test for the color detection.  However, I'm not really sure how to test `cider-font-lock-as`.  Looking at properties of the buffer that comes back is probably doable, but that seems overkill.  What were you thinking specifically there?

bbatsov: I was thinking that you can simply check it the result whether there are still ansi sequences or not, but I can live without this.

bbatsov:
Btw, we applying those colors and we're immediately discarding them with the `clojure-mode` font-locking. Does this mess up the font-locking? If so you might also have to use some function to remove the text properties. I'm also curious if there's not some function that can simply strip those ansi color markers from the string. At any rate - the better long term solution is certainly some way to control the configuration of puget.

cider-util.el
plandes: Sounds good.  Let me know if you change your mind.

plandes:
I don't think clojure-mode is clobbering what `ansi-color-apply` is fontifying.  However, the colors given in the REPL are a little different from what's given with `cider-pprint-eval-last-sexp`.  Note I'm apply `:none` to some of these because I like the default given with `cider-pprint-eval-last-sexp` and `clojure-mode` must be using some other (default) font.

See attached.

snap

plandes:
Regarding how to control puget, take a look at [my solution](https://github.com/plandes/clj-nrepl-puget) and recommend your more dynamic approach.  If that doesn't work we can talk about using bindings.  Whichever works for you I'll make the changes and I'll fold it in.

Let me know.

Thanks.

bbatsov:
> I don't think clojure-mode is clobbering what ansi-color-apply is fontifying. However, the colors given in the REPL are a little different from what's given with cider-pprint-eval-last-sexp.

Well, that's why I think it's better to strip the ansi coloring completely - it creates some odd output that doesn't look like something produced by CIDER and clojure-mode. 

> I'm apply :none to some of these because I like the default given with cider-pprint-eval-last-sexp and clojure-mode must be using some other (default) font.

I don't understand what you mean here.

cider-util.el
```diff
@@ -250,16 +251,23 @@ This buffer is not designed to display anything to the user.  For that, use
...
-  (if (or (null cider-font-lock-max-length)
-          (< (length string) cider-font-lock-max-length))
-      (with-current-buffer (cider--make-buffer-for-mode mode)
-        (erase-buffer)
-        (insert string)
-        (font-lock-fontify-region (point-min) (point-max))
-        (buffer-string))
-    string))
+  (let ((string (if (cider-ansi-color-string-p string)
+                    (ansi-color-apply string)
```
bbatsov: Basically I'd wrap this in `substring-no-properties`, so whatever got added by `puget` would be immediately discarded. We don't really want third party tools to mess up the clojure-mode font-locking. What they in terms of highlighting do makes sense in an external REPL, but doesn't make sense anywhere else IMO.

plandes:
I took another look at the emacs code I wrote in my Clojure package and was wrong about pugent clobbering `clojure-mode`'s fontification.  After a bug fix I was able to get puget to not clobber clojure-mode's fortification so now the REPL output looks exactly the same as `cider-pprint-eval-last-sexp`.  This sounds like a pretty cool enhancement since you get your pie (puget) and get to eat it too (pick fonts to "leave alone" for `clojure-mode` to highlight), but I will wrap `ansi-color-apply` with `substring-no-properties`.

You could configure puget to only add colors for which `clojure-mode` doesn't give you (i.e. colors for symbols, numbers, booleans are different color classes).

bbatsov:
> This sounds like a pretty cool enhancement since you get your pie (puget) and get to eat it too (pick fonts to "leave alone" for clojure-mode to highlight), but I will wrap ansi-color-apply with substring-no-properties.

The problem is that this creates inconsistent font-locking - in an editor font-lock should not be handled by external tools. Add the final change and let's wrap this.

plandes:
OK.  Done.

bbatsov:
👍

plandes:
@bbatsov For the more comprehensive solution: what did you decide?

PR #2031, opened 2017-07-07
Add a section in the manual for the spec browser

jpmonettas:
Added a section under the miscellaneous features for the spec browser.


doc/miscellaneous_features.md
```diff
@@ -193,6 +193,36 @@ Keyboard shortcut               | Description
...
+If you are using  Clojure 1.9.0-alpha16 or newer you can browse the Clojure specs registry.
+
+If you know what you are looking for, you can type <kbd>M-x</kbd> `cider-browse-spec`.
+It will prompt you for a spec name to browse to, hit <kbd>RET</kbd> and you will find yourself
+at the spec browser.
+
+![Spec Browser](images/spec_browser.png)
+
+You can also type the command <kbd>M-x</kbd> `cider-browse-spec-all`. This command will prompt you for 
+a regex you can use to filter out the specs you are interested in, who will also take you to the spec browser.
```
bbatsov: , who -> and

doc/miscellaneous_features.md
```diff
@@ -193,6 +193,36 @@ Keyboard shortcut               | Description
...
+You can also type the command <kbd>M-x</kbd> `cider-browse-spec-all`. This command will prompt you for 
+a regex you can use to filter out the specs you are interested in, who will also take you to the spec browser. 
+
+![Spec Browser](images/spec_browser_all.png)
+
+Once in the browser you can use your mouse or the keybindings below to navigate deeper into sub specs.
+
+Keyboard shortcut               | Description
+--------------------------------|-------------------------------
+<kbd>RET</kbd>                  | Browse to the spec at point.
```
bbatsov: Browse the

doc/miscellaneous_features.md
```diff
@@ -193,6 +193,36 @@ Keyboard shortcut               | Description
...
+<kbd>^</kbd>                    | Go up in the navigation stack.
+<kbd>n</kbd>                    | Go to next spec.
+<kbd>p</kbd>                    | Go to previous spec.
+<kbd>e</kbd>                    | Generate an example for the current browser spec.
```
bbatsov: current browser spec -> spec at point

doc/miscellaneous_features.md
```diff
@@ -193,6 +193,36 @@ Keyboard shortcut               | Description
...
+
+If your project contains a version of org.clojure/test.check, you can type <kbd>e</kbd> when browsing
```
bbatsov: Surround the name of the library with backticks.

bbatsov:
@xiongtx made some improvements to the browser and spec view visuals, so it'd be nice if you updated the screenshots to feature them. Also - if you can use zenburn or solarized theme for the screenshots that would be ideal as most of the screenshots in the manual use of those two themes.

doc/miscellaneous_features.md
jpmonettas: In this case you hit e to print a example for the spec you are visiting in the browser, independently of the point

jpmonettas:
@bbatsov just pushed all your suggestions + the new screenshots with zenburn and @xiongtx improvements.

bbatsov:
Thanks!

doc/miscellaneous_features.md
bbatsov: I don't quite understand this, btw - if you're in the browser this can't be independently of the point (otherwise you don't know what you're visiting), but I understand this makes sense in the spec view buffer, which is not the spec browser as far as I'm concerned.

jpmonettas: Yeah I think its just the mining of the words. I'm calling the spec browser to the entire thing, that can browse deep into other specs, not just the first listing.
So hitting e in a listing doesn't do anything independent of where the point is. Once the browser is visiting a spec, you hit e, and doesn't matter where the point is you see the current browser spec, which is the top of the navigation stack.
Does it make sense now?

PR #2033, opened 2017-07-08
Improvements to spec browser

xiongtx:
Some improvements to the spec browser.

All specs:

Single spec:

/cc @jpmonettas

### DIFF
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -91,8 +91,7 @@ For example: (\":ring.request/header-name\" \":ring.request/headers\" \":ring/re
 
 (defun cider-browse-spec--propertize-keyword (kw)
   "Add properties to KW text needed by the spec browser."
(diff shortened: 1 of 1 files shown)


cider-browse-spec.el
```diff
@@ -338,7 +340,8 @@ If FILTER-REGEX is empty, list all specs in the registry."
 (defun cider-browse-spec-handle-mouse (event)
   "Handle mouse click EVENT."
   (interactive "e")
-  (cider-browse-spec--browse-at-point))
+  (when (eq 'highlight (get-text-property (point) 'mouse-face))
```
bbatsov: I'm reasonably sure the other similar browsers have the same issue, so you might take a look at them as well.

bbatsov:
I'd use "improve" instead of "fix", but apart from that - great work! 👍

PR #2034, opened 2017-07-08
Make `filter-regex` for `cider-browse-spec-all` optional

xiongtx:
It's tedious to call `(cider-browse-spec-all "")` each time.


bbatsov:
This should probably be taken a step further:

* the interactive command shouldn't prompt about the filter without a prefix
* you should be able to apply the filter once you're inside the browser

xiongtx:
I'll get to it...but first I'll need to finish these 😉 :

PR #2037, opened 2017-07-08
Filter stacktrace to just project frames

dpsutton:
Clojure stacktraces can often be quite difficult to read. This follows a recent addition to cider-nrepl which marked stackframes as coming from the project. This allows us to filter to the project stackframes.

The old filtering system is a "negative" filtering system. This would hide lines with the tags, and therefore our project would be hidden as it was marked with clj and repl and other tags. The solution is to add a "positive filter", ie, if a tag is found, show it, regardless if it overlaps with a negative filter as well.

So now in a stacktrace buffer, hit p to narrow down to just the project stakframes and then hit p again to go back to all stack frames.

normal-errors

project-errors-only

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -274,6 +274,12 @@ searching and update the hidden count text."
           (replace-match
            (number-to-string cider-stacktrace-hidden-frame-count)))))))
 
+(defun cider-stacktrace-frame-p ()
+  (get-text-property (point) 'cider-stacktrace-frame))
+
+(defun cider-stacktrace-collapsed-p ()
+  (get-text-property (point) 'collapsed))
+
 (defun cider-stacktrace-apply-filters (filters)
   "Set visibility on stack frames using FILTERS.
 Update `cider-stacktrace-hidden-frame-count' and indicate filters applied.
@@ -285,11 +291,13 @@ hidden count."
       (let ((inhibit-read-only t)
             (hidden 0))
         (while (not (eobp))
-          (unless (get-text-property (point) 'collapsed)
+          (when (and (cider-stacktrace-frame-p)
+                     (not (cider-stacktrace-collapsed-p)))
             (let* ((flags (get-text-property (point) 'flags))
                    (hide (if (seq-intersection filters flags) t nil)))
               (when hide (cl-incf hidden))
-              (put-text-property (point) (line-beginning-position 2) 'invisible hide)))
+              (put-text-property (point) (line-beginning-position 2)
+                                 'invisible hide)))
           (forward-line 1))
         (setq cider-stacktrace-hidden-frame-count hidden)))
     (cider-stacktrace-indicate-filters filters)))
@@ -607,7 +615,9 @@ This associates text properties to enable filtering and source navigation."
                 (p2 (search-forward "/"))
                 (p3 (search-forward-regexp "[^/$]+")))
             (put-text-property p1 p4 'font-lock-face 'cider-stacktrace-ns-face)
-            (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
+            (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)
+            (put-text-property (line-beginning-position) (line-end-position)
+                               'cider-stacktrace-frame t)))
         (insert "\n")))))
 
 (declare-function cider-jump-to "cider-interaction")
@@ -717,6 +727,7 @@ through the `cider-stacktrace-suppressed-errors' variable."
             (setq num (1- num))))))
     (cider-stacktrace-initialize causes)
     (font-lock-refresh-defaults)))
+
 
 (provide 'cider-stacktrace)
 
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -117,23 +117,17 @@ The error types are represented as strings."
   :group 'cider-stacktrace
   :package-version '(cider . "0.7.0"))
 
-(defface cider-stacktrace-filter-shown-face
+(defface cider-stacktrace-filter-active-face
   '((t (:inherit button :underline t :weight normal)))
   "Face for filter buttons representing frames currently visible"
   :group 'cider-stacktrace
   :package-version '(cider . "0.6.0"))
 
-(defface cider-stacktrace-filter-hidden-face
+(defface cider-stacktrace-filter-inactive-face
   '((t (:inherit button :underline nil :weight normal)))
   "Face for filter buttons representing frames currently filtered out"
   :group 'cider-stacktrace
   :package-version '(cider . "0.6.0"))
-
-(defface cider-stacktrace-filter-positive-face
-  '((t (:inherit button :underline t :weight normal)))
-  "Face for filter buttons representing frames currently filtered out"
-  :group 'cider-stacktrace
-  :package-version '(cider . "0.15.0"))
 
 (defface cider-stacktrace-face
   '((t (:inherit default)))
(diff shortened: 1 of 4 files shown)


cider-stacktrace.el
```diff
@@ -550,6 +600,11 @@ prompt and whether to use a new window.  Similar to `cider-find-var'."
                           'help-echo (format "Toggle %s stack frames"
                                              (car filter)))
       (insert " "))
+    (insert-text-button "Project-Only"
```
xiongtx: `Project` instead of `Project-Only` would be more in line with the naming of other options.

dpsutton: The problem is that this filter works differently than the others. "Clojure" will remove or not remove all clojure namespace stackframes. This removes all _other_ frames but your project, which is why i named it Project-Only

xiongtx: Oh, that makes sense.

Nonetheless I think `Project-Only` should be the first filter.

dpsutton: moved it first. looks good there

bbatsov:
The feature looks pretty great overall, but I'm worried people would be confused by it's different behaviour. Should we also have some version of this that actually fits into the current model of work? And how can we make it more obvious that this is intentionally behaving pretty differently from the rest of the filters?

bbatsov:
Btw, I haven't read the code yet, but I assume that one you turn this on it turn off all the other filters, right?

dpsutton:
> The feature looks pretty great overall, but I'm worried people would be confused by it's different behaviour. Should we also have some version of this that actually fits into the current model of work?

If you have one that mimics the current behavior of the other filters, it would remove all stackframes from the project, which seems almost useless to me, but it should be basically trivial to add.

> And how can we make it more obvious that this is intentionally behaving pretty differently from the rest of the filters?

I thought about that a little and it's kinda difficult. I haven't come up with anything good so I'm open to feedback or to user reports if there's confusion. Perhaps one thing we can do is to make the `project` tag exclusive to all others. Right now it overlaps so removing `repl` or `clj` can remove your project. But if project would "dominate" in that sense, perhaps it would behave like the others and we wouldn't need the "positive" filter. That notion only exists because we have to ensure the frames are shown when a filter says to remove it.

> Btw, I haven't read the code yet, but I assume that one you turn this on it turn off all the other filters, right?

Yeah, there's a defvar to hold the previous filters and i stick the current in there and then set all to filter and set the "positive" filters to `project.

One caveat is the `All` tag has a little bug in it in that it's not being underlined right now. I think there's some shenanigans because it's kind of a "pseudo" tag and not a real tag at all. The feature still works but only the underlining isn't flushed out. It's limited to the logic in determining what face to show in a little function so it should be easy to correct and test.

bbatsov:
> If you have one that mimics the current behavior of the other filters, it would remove all stackframes from the project, which seems almost useless to me, but it should be basically trivial to add.

I guess you're right. We'll have to think about this down the road. 

> One caveat is the All tag has a little bug in it in that it's not being underlined right now. I think there's some shenanigans because it's kind of a "pseudo" tag and not a real tag at all. The feature still works but only the underlining isn't flushed out. It's limited to the logic in determining what face to show in a little function so it should be easy to correct and test.

Well, I guess we should improve this then. 

Overall the big goal for this feature would be to avoid confusion. Seems that we need some combination of turning off the other filters and a information message. Perhaps we can also add some configuration about the default filters for the stacktrace buffer.

xiongtx:
The objection, as I understand it, is that `Project-Only` will hide *non-project* stack frames, whereas other filters only show/hide frames they name.

I don't think it's too confusing to have a `Project-Only` filter, but perhaps it'd be sufficient to have a `Project` filter that behaves like the other filters.

bbatsov:
> I don't think it's too confusing to have a Project-Only filter, but perhaps it'd be sufficient to have a Project filter that behaves like the other filters.

Yeah, that's what I thought initially myself.

dpsutton:
We could make the project tag exclusive of all others. The problem with this approach is that you probably only want to see project which means you want every _other_ filter than project. So for the project filter to be meaningful it would need to really set the others.

If this is the preferred way I'm happy to do that but I think it makes the project button rather irrelevant. As its setup now, on the stacktrace just hit `p` and it shows project only, and then hit `p` again and it restores the previous view.

If we prefer the standard filter the tags must be made mutually exclusive as the `project` tag overlaps with the other tags.

dpsutton:
> Some dependencies were not available: spinner, queue
> make: *** [elpa] Error 255
> The command "make elpa" failed and exited with 2 during .

Transient bug in the CI. They pass for me. I'll make a minor change and push later to retrigger but otherwise i think we are good here.

bbatsov:
Seems you forgot to have a separate commit for the actual project filter. It would also be nice if the commit messages complied with the project's commit message style.

dpsutton:
I can update that tonight

bbatsov:
OK

vspinu:
How about a separator before ALL to emphasize that ALL  and Project-Only have different semantics?

```
  Show: Clojure Java REPL Tooling Duplicates  |  All Project-Only  (21 frames hidden)
```

bbatsov:
> How about a separator before ALL to emphasize that ALL and Project-Only have different semantics?

I like the idea.

In the commit messages:

- "Only show" -> "Show only"
- "Adds" -> "Add"
- "Introduces" -> "Introduce"

dpsutton:
with-separator

A separator like this?

@bbatsov sorry about the present tense. I thought that this project liked that but then i read the document about commit messages by tpope after. I'll update that now

dpsutton:
or with earmuffs
with-earmuffs

bbatsov:
I think the idea was to have both "All" and "Project Only" in a separate section (separated with a vertical bar for instance), as pressing them eradicates all the other filters so in a way they are similar/special.

dpsutton:
Gotcha.  It sounds silly but that will require some more programming. The "(11 frames hidden)" serves to separate the top of this buffer from the bottom. I started getting rid of some of this questionable stuff with `cider-stacktrace-frame-p` rather than it hiding and unhiding every line of text on the buffer.

bbatsov:
Got it. Well - let's make it better then!

dpsutton:
showing-normal

showing-project-only
I think this UI is working for me. The other filters kinda don't show anything they allow you to hide (negative filter). Whereas the others are show project-only or show all. Seems to indicate that they are distinct mechanisms.

dpsutton:
although I've inverted the underlining to make it clear which are hidden now

indicate-hiding

bbatsov:
Looks good to me! 👍 Thanks for all the work you've put into this!

vspinu:
The show/hide semantics is unfortunate IMO. Is it the underlined keys the one which is shown or hidden? Show/hide suggests that those two lines are opposite of each other, which is obviously not the case.

"Hide" makes things disappear, it's a negative concept, so if you press on a key (an "yes" action) you have to think about which stacks will disappear. It's confusing. Pressing on "yes" with "show" and thinking about which things appear instead is much more straightforward.

How about this:

```
  Show: Project-Only All 
  Type: Java REPL Tooling Duplicates  (25 frames hidden)
```   

"Type" means type of the stacktrace entry to show. It will also align as "type" has same number of letters as "show".

BTW, when clicking on `Project-Only` it is not getting underlined for me.

vspinu:
Could be also 

```
  Filter:  Project-Only 
  Show: Clojure Java REPL Tooling Duplicates All  (19 frames hidden)
```
Entirely backward compatible.

dpsutton:
ah, yes, you're right about the underlining with Project only. I was always using the `p` shortcut which works correctly. I'll figure that one out.

As to my thinking about the show/hide naming, I started from a stacktrace having a bunch of frames in it I don't want to see, so I suppress some of them. The underline means that those types are hidden. But like all things, I've been looking at it for a while so if it looks strange to you I'm inclined to accept that. I have no strong opinions one way or the other so if there's confusion I'm happy to change it.

PR #2038, opened 2017-07-09
[Fix #2023] Make popup-buffer sexp indentation optional

### LINKED ISSUE #2023: Don't indent already indented output when pretty-printing
Note: Currently I believe the issue is with Cider, but it also relates
to recent changes on the master branch of Emacs, so I'm not entirely
sure this is the right place to post this report. I think this is more
of a feature request anyway.

Emacs has introduced lisp-indentation [changes](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=4713dd425beac5cb459704e67dcb8f6faf714375) which don't reparse
the code as often as before and rely on previous parse-state to
improve performance. I believe this change is causing performance
issues with `cider-emit-into-popup-buffer`, when pretty-printing
non-trivial amounts of data.

I profiled Emacs and found that it spends most of the time garbage
collecting and `lisp-indent-calc-next` during pretty-printing.

I've also found that if I comment out the `indent-sexp` call from the
`cider-emit-into-popup-buffer` function, the issue is gone.

I'm not really sure what other features `cider-emit-into-popup-buffer`
is used for (except pretty-printing), but don't pretty-print functions
already indent their output? Why is there a need for Emacs to indent
the output again?

If you agree, I can submit a pull request for
`cider-emit-into-popup-buffer` with an optional parameter to inhibit
indentation.

ak-coram:
- Add optional inhibit-indent argument to cider-emit-into-popup-buffer.

- Inhibit indentation where output is already indented (pprint) or not sexps.

-----------------

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -806,15 +806,18 @@ COMMENT-PREFIX is the comment prefix to use."
   "Make a handler for evaluating and printing stdout/stderr in popup BUFFER.
 
 This is used by pretty-printing commands and intentionally discards their results."
-  (nrepl-make-response-handler (or buffer (current-buffer))
-                               '()
-                               ;; stdout handler
-                               (lambda (buffer str)
-                                 (cider-emit-into-popup-buffer buffer (ansi-color-apply str)))
-                               ;; stderr handler
-                               (lambda (buffer str)
-                                 (cider-emit-into-popup-buffer buffer (ansi-color-apply str)))
-                               '()))
+  (cl-flet ((popup-handler (buffer str)
+                           (cider-emit-into-popup-buffer buffer
+                                                         (ansi-color-apply str)
+                                                         nil
+                                                         t)))
+   (nrepl-make-response-handler (or buffer (current-buffer))
+                                '()
+                                ;; stdout handler
+                                #'popup-handler
+                                ;; stderr handler
+                                #'popup-handler
+                                '())))
 
 (defun cider-visit-error-buffer ()
   "Visit the `cider-error-buffer' (usually *cider-error*) if it exists."
@@ -1483,7 +1486,7 @@ Defaults to the current ns.  With prefix arg QUERY, prompts for a ns."
   "Refresh LOG-BUFFER with RESPONSE."
   (nrepl-dbind-response response (out err reloading status error error-ns after before)
     (cl-flet* ((log (message &optional face)
-                    (cider-emit-into-popup-buffer log-buffer message face))
+                    (cider-emit-into-popup-buffer log-buffer message face t))
 
                (log-echo (message &optional face)
                          (log message face)
@@ -1577,7 +1580,10 @@ refresh functions (defined in `cider-refresh-before-fn' and
          (when cider-refresh-show-log-buffer
            (cider-popup-buffer-display log-buffer))
          (when inhibit-refresh-fns
-           (cider-emit-into-popup-buffer log-buffer "inhibiting refresh functions\n"))
+           (cider-emit-into-popup-buffer log-buffer
+                                         "inhibiting refresh functions\n"
+                                         nil
+                                         t))
          (when clear?
            (cider-nrepl-send-sync-request '("op" "refresh-clear") conn))
          (cider-nrepl-send-request
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -806,18 +806,18 @@ COMMENT-PREFIX is the comment prefix to use."
   "Make a handler for evaluating and printing stdout/stderr in popup BUFFER.
 
 This is used by pretty-printing commands and intentionally discards their results."
-  (cl-flet ((popup-handler (buffer str)
-                           (cider-emit-into-popup-buffer buffer
-                                                         (ansi-color-apply str)
-                                                         nil
-                                                         t)))
-   (nrepl-make-response-handler (or buffer (current-buffer))
-                                '()
-                                ;; stdout handler
-                                #'popup-handler
-                                ;; stderr handler
-                                #'popup-handler
-                                '())))
+  (cl-flet ((popup-output-handler (buffer str)
+                                  (cider-emit-into-popup-buffer buffer
+                                                                (ansi-color-apply str)
+                                                                nil
+                                                                t)))
+    (nrepl-make-response-handler (or buffer (current-buffer))
+                                 '()
+                                 ;; stdout handler
+                                 #'popup-output-handler
+                                 ;; stderr handler
+                                 #'popup-output-handler
+                                 '())))
 
 (defun cider-visit-error-buffer ()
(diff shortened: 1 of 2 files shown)


ak-coram:
I didn't mark "All tests are passing" because I get 3 failing ones (the same tests fail with master for me, so I don't think they relate to these changes).

dpsutton:
Which tests are failing for you? I'm able to run all tests as is the continuous integration setup.

ak-coram:
This is what I get:

nrepl--bdecode-message when bencode structures are complete decodes queues
FAILED: Expected (#s(queue nil nil) . #s(nrepl-response-queue (((1 2 3) (5 6))) (((1 2 3) (5 6))) nil)) to `equal' ([cl-struct-queue nil nil] . [cl-struct-nrepl-response-queue (((1 2 3) (5 6))) (((1 2 3) (5 6))) nil])

nrepl--bdecode-message when bencode structures are complete decodes nrepl doc output
FAILED: Expected (#s(queue nil nil) . #s(nrepl-response-queue ((dict "id" "18" "out" "clojure.core/reduce
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "out" "([f coll] [f val coll])
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "out" "  f should be a function of 2 arguments. If val is not supplied,
  returns the result of applying f to the first 2 items in coll, then
  applying f to that result and the 3rd item, etc. If coll contains no
  items, f must accept no arguments as well, and reduce returns the
  result of calling f with no arguments.  If coll has only 1 item, it
  is returned and f is not called.  If val is supplied, returns the
  result of applying f to val and the first item in coll, then
  applying f to that result and the 2nd item, etc. If coll contains no
  items, returns val and f is not called.
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "ns" "user" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "value" "nil") (dict "id" "18" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "status" ("done"))) ((dict "id" "18" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "status" ("done"))) nil)) to `equal' ([cl-struct-queue nil nil] . [cl-struct-nrepl-response-queue ((dict "id" "18" "out" "clojure.core/reduce
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "out" "([f coll] [f val coll])
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "out" "  f should be a function of 2 arguments. If val is not supplied,
  returns the result of applying f to the first 2 items in coll, then
  applying f to that result and the 3rd item, etc. If coll contains no
  items, f must accept no arguments as well, and reduce returns the
  result of calling f with no arguments.  If coll has only 1 item, it
  is returned and f is not called.  If val is supplied, returns the
  result of applying f to val and the first item in coll, then
  applying f to that result and the 2nd item, etc. If coll contains no
  items, returns val and f is not called.
" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57") (dict "id" "18" "ns" "user" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "value" "nil") (dict "id" "18" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "status" ("done"))) ((dict "id" "18" "session" "6fc999d0-3795-4d51-85fc-ccca7537ee57" "status" ("done"))) nil])

nrepl--bdecode-message when bencode strings have partial structures decodes queues
FAILED: Expected (#s(queue ("i6") ("i6")) . #s(nrepl-response-queue (((1 2 3) (5 6))) (((1 2 3) (5 6))) ((5) ((1 2 3))))) to `equal' ([cl-struct-queue ("i6") ("i6")] . [cl-struct-nrepl-response-queue (((1 2 3) (5 6))) (((1 2 3) (5 6))) ((5) ((1 2 3)))])

Ran 250 specs, 3 failed, in 2.2 seconds.
`

cider-interaction.el
```diff
@@ -806,15 +806,18 @@ COMMENT-PREFIX is the comment prefix to use."
...
-  (nrepl-make-response-handler (or buffer (current-buffer))
-                               '()
-                               ;; stdout handler
-                               (lambda (buffer str)
-                                 (cider-emit-into-popup-buffer buffer (ansi-color-apply str)))
-                               ;; stderr handler
-                               (lambda (buffer str)
-                                 (cider-emit-into-popup-buffer buffer (ansi-color-apply str)))
-                               '()))
+  (cl-flet ((popup-handler (buffer str)
```
bbatsov: I guess this should be named - `popup-output-handler`.

cider-interaction.el
```diff
@@ -806,15 +806,18 @@ COMMENT-PREFIX is the comment prefix to use."
...
+                           (cider-emit-into-popup-buffer buffer
```
bbatsov: The indentation here doesn't seem right. In general I'm not a big of fan of inline definitions - probably this can be just a private function. `nrepl---popup-output-handler` or something like this.

cider-popup.el
```diff
@@ -103,8 +103,9 @@ and automatically removed when killed."
                 nil 'local))
     (current-buffer)))
 
-(defun cider-emit-into-popup-buffer (buffer value &optional face)
-  "Emit into BUFFER the provided VALUE optionally using FACE."
+(defun cider-emit-into-popup-buffer (buffer value &optional face inhibit-indent)
+  "Emit into BUFFER the provided VALUE optionally using FACE. Indents
```
bbatsov: The second sentence should start on the second line.

cider-interaction.el
ak-coram: > The indentation here doesn't seem right. In general I'm not a big of fan of inline definitions

I've seen a similar pattern in `cider-refresh--handle-response` (same file), so I thought it would be okay.

> probably this can be just a private function. nrepl---popup-output-handler or something like this.

Ok, I'm guessing that would go into `nrepl-client.el`, right? I could also put back the original two calls to `cider-emit-into-popup-buffer` (not that badly redundant), if you prefer.

bbatsov: I don't pay close attention to every PR. :-) If this was supported directly by Elisp I might use it, but I don't like `cl-flet` that much. And obviously non-inlined functions can be used at more places.

bbatsov: And in this particular case I was more concerned about the name; the `cl-flet` thing I mentioned just in passing.

ak-coram: Ok, which one should we go for? Personally I wouldn't make a new
global function out of this, because it would be almost the same as
`cider-emit-into-popup-buffer` (with fewer options).

I agree with the name change and can push that if you decide to keep
it inline. Thanks!

bbatsov:
👍

PR #2041, opened 2017-07-13
Two fixes for cljc files

vspinu:
First commit finishes #1875 by fixing the interactive eval which suffered from the same issues as described there.

Second makes `cider-switch-to-repl-buffer` to switch to the most recently used repl. Current behavior is to always switch to clj repl.  To replicate, go to cljs repl, pres C-c C-z to get into cljc buffer, then C-c C-z and you will land in clj repl which covers your cljs repl.


bbatsov:
👍

PR #2042, opened 2017-07-13
[Fix #2040] Fix fontification of conditional expressions in cljc files

### LINKED ISSUE #2040: All #?(:cljs code in cljc is commented out (cider-reader-conditional-face)
I am seeing the following with the latest CIDER. 

cljs

This happens in `.cljc` files irrespectively of clojurescript repl running. AFAICU from the description it should show up on cljs connections.

For what it's worth, I thinking commenting conditionals is a bit too much.  Slight deeming might work better. I still read a lot of cljs code when cljs repl is not running, and this feature becomes quickly a nuisance.

vspinu:
This one turned to be slightly more involved than I thought. But this is the minimal incision to be able to fix it, and it seems to be the right design choice anyhow. 

I am changing  `cider-connection-type-for-buffer` to return nil if connection type cannot
be decided. The current default to "clj" strikes me as arbitrary. I think it's much better to be explicit when you cannot decide on something and let the consumer decide what to do with nil value.


vspinu:
Please give me a moment on this one, don't merge for now.

vspinu:
Ok, good to go.

cider-client.el
```diff
@@ -222,12 +222,16 @@ such a link cannot be established automatically."
   (kill-local-variable 'cider-connections))
 
 (defun cider-connection-type-for-buffer ()
-  "Return the matching connection type (clj or cljs) for the current buffer."
+  "Return the matching connection type (clj or cljs) for the current buffer.
+In cljc files with multiple connected REPLs return nil."
```
bbatsov: Might be better to invent some type for such cases - e.g. `multi`, `dual`, etc.

cider-repl.el
```diff
@@ -305,6 +306,7 @@ client process connection.  Unless NO-BANNER is non-nil, insert a banner."
   "Generate the welcome REPL buffer banner."
   (let ((host (cider--connection-host (current-buffer)))
         (port (cider--connection-port (current-buffer))))
+    (message "Current buffer: %s" (current-buffer))
```
bbatsov: Some debug message?

cider-client.el
vspinu: The idea of nil is that it's "undecided", testing on nil is easier and bugs with nil are likely to surface earlier. I will check the usage of the function to see if it has any further implications.

cider-repl.el
vspinu: Arh ... indeed.

vspinu:
I ended up with 3 separate things on this:

- I left `cider-connection-type-for-buffer` do the fast job of inferring the type from buffer's mode. This seems to the original intention of that function. But I added `cider-project-connections-types` which does what the name suggests.
- Fixed the issue of conditional fontification.
- Made `cider-current-connection` do what people expect - pick the most recently used connection (not the most recently created connection it is now). Funny enough, the unit tests' intention was always to pick the most recent buffer, but due to the test construction it was failing to detect that.

cider-client.el
```diff
@@ -178,6 +178,23 @@ connections are returned, instead of just the most recent."
...
+type based on the major mode. See `cider-project-connections-types' for a
+list of types of actual connections within a project."
+  (cond
+   ((derived-mode-p 'clojurescript-mode) "cljs")
+   ((derived-mode-p 'clojurec-mode) "multi")
+   ((derived-mode-p 'clojurex-mode) "multi")
+   ((derived-mode-p 'clojure-mode) "clj")
+   (cider-repl-type)))
+
+(defun cider-project-connections-types ()
```
bbatsov: Maybe we should have both `project-connection` and `project-types` functions? Seems reasonable to need to list the connections for a project, pict a default project connection (in the absence of smart dispatch and so on).

vspinu: Do you mean a function what would do the same thing that `cider-find-connection-buffer-for-project-directory` does with `all-connections` argument.  I am not a big fan of that function API btw.

cider-client.el
```diff
@@ -178,6 +178,23 @@ connections are returned, instead of just the most recent."
...
+  "Return a list of types of connections within current project."
+  (let ((connections (cider-find-connection-buffer-for-project-directory nil :all-connections)))
```
vspinu: The same thing that `cider-find-connection-buffer-for-project-directory` returns - nil.  AFAICS there is still no systematic way to associated stand-alone buffers to specific connections, so it's little that could be done at the moment. But there is probably nothing to be done anyway. The function returns project-connection types - no project, no connections.

vspinu: > cider-find-connection-buffer-for-project-directory

Let me streamline this sausage as I am already into this stuff till my ears.

dpsutton:
Will this affect the `cider-map-connections` function? This function seems to not be that used any longer (only in some namespace functions and in load-buffer). And if we are pulling the notion of a project's connections out to be easier to use, perhaps this wouldn't look up connections themselves but rather take them as input so it would look like a proper map command:

`(cider-map-connections cmd (project-connections))` etc

vspinu:
> Will this affect the cider-map-connections function? 

No. I am keeping an eye on that.

> (cider-map-connections cmd (project-connections)) etc

It's a bit more tricky. There is quite some confusion going on in the code due to buffer <> connection confounding in cider. What that function does is actually executes `cmd` for two buffers. You can have multiple actual connection for one project, each connection can have up to two buffers (clj and cljs). I will see what I can do to streamline all these parts a bit.

PR #2043, opened 2017-07-15
Interactive stacktraces in REPL

vspinu:
Two new features to make REPL location references useful:

 1. Interactive locations in REPL. 
 2. Highlight vars from the current project 

Interactive locations are implemented as help-echo and have 0 overhead on printout in REPL.  Currently understood are stdout stacks, aviso stacks, printouts of exception objects and references in timbre's logging.  

Highlighting of vars currently uses project directory to infer the root namespaces. Would be happy to know a better way without recurring to parsing project files.  

stack

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### New Features
 
 * [#2050](https://github.com/clojure-emacs/cider/pull/2050) Use `view-mode` for `cider-grimoire` buffers
+* Make stacktraces and other location references in REPL clickable.
+* Highlight root namespace in REPL stacktraces.
 * Filter stacktrace to just frames from your project.
 * [#1918](https://github.com/clojure-emacs/cider/issues/1918): Add new commands `cider-browse-spec` and `cider-browse-spec-all` which start a spec browser.
 * [#2015](https://github.com/clojure-emacs/cider/pull/2015): Show symbols as special forms *and* macros in `cider-doc`
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -465,13 +465,6 @@ Invert meaning of `cider-prompt-for-symbol' if PREFIX indicates it should be."
   (if (cider--prefix-invert-prompt-p prefix)
       (not cider-prompt-for-symbol) cider-prompt-for-symbol))
 
-(defun cider-sync-request:ns-path (ns)
-  "Get the path to the file containing NS."
-  (thread-first `("op" "ns-path"
-                  "ns" ,ns)
-    cider-nrepl-send-sync-request
-    (nrepl-dict-get "path")))
-
 (defun cider--find-ns (ns &optional other-window)
   "Find the file containing NS's definition.
 Optionally open it in a different window if OTHER-WINDOW is truthy."
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1065,41 +1065,31 @@ evaluation command.  Honor `cider-auto-jump-to-error'."
       #'identity))
   "Function to translate Emacs filenames to nREPL namestrings.")
 
-(defvar-local cider--ns-form-cache (make-hash-table :test 'equal)
-  "ns form cache for the current buffer.
+(defvar-local cider--last-ns-form nil
+  "Ns-form of the most recent interactive evaluation.")
 
-The cache is a hash where the keys are connection names and the values
-are ns forms. This allows every connection to keep track of the ns
-form independently.")
-
-(defun cider--cache-ns-form ()
-  "Cache the form in the current buffer for the current connection."
-  (puthash (cider-current-connection)
-           (cider-ns-form)
-           cider--ns-form-cache))
-
-(defun cider--cached-ns-form ()
-  "Retrieve the cached ns form for the current buffer & connection."
-  (gethash (cider-current-connection) cider--ns-form-cache))
-
-(defun cider--prep-interactive-eval (form)
-  "Prepare the environment for an interactive eval of FORM.
-
-Ensure the current ns declaration has been
-evaluated (so that the ns containing FORM exists).
-If FORM is a ns declaration it's not processed and cached.
-
-Clears any compilation highlights and kills the error window."
+(defun cider--prep-interactive-eval (form connection)
+  "Prepare the environment for an interactive eval of FORM in CONNECTION.
+Ensure the current ns declaration has been evaluated (so that the ns
+containing FORM exists). Cache ns-form in the current buffer unless FORM is
+ns declaration itself. Clear any compilation highlights and kill the error
+window."
   (cider--clear-compilation-highlights)
   (cider--quit-error-window)
   (let ((cur-ns-form (cider-ns-form)))
     (when (and cur-ns-form
-               (not (string= cur-ns-form (cider--cached-ns-form)))
+               (not (string= cur-ns-form
+                             (buffer-local-value 'cider--last-ns-form connection)))
                (not (cider-ns-form-p form)))
       (when cider-auto-track-ns-form-changes
-        ;; TODO: check for evaluation errors
-        (cider-eval-ns-form 'sync))
-      (cider--cache-ns-form))))
+        ;; The first interactive eval on a file can load a lot of libs. This can
+        ;; easily lead to more than 10 sec.
+        (let ((nrepl-sync-request-timeout 30))
+          ;; TODO: check for evaluation errors
+          (cider-nrepl-sync-request:eval cur-ns-form connection)))
+      ;; cache at the end, in case of errors
+      (with-current-buffer connection
+        (setq cider--last-ns-form cur-ns-form)))))
 
 (defvar-local cider-interactive-eval-override nil
   "Function to call instead of `cider-interactive-eval'.")
@@ -1127,9 +1117,9 @@ arguments and only proceed with evaluation if it returns nil."
                  (functionp cider-interactive-eval-override)
                  (funcall cider-interactive-eval-override form callback bounds))
       (cider-map-connections #'ignore :any)
-      (cider--prep-interactive-eval form)
       (cider-map-connections
        (lambda (connection)
+         (cider--prep-interactive-eval form connection)
          (cider-nrepl-request:eval
           form
           (or callback (cider-interactive-eval-handler nil bounds))
@@ -1624,10 +1614,12 @@ ClojureScript REPL exists for the project, it is evaluated in both REPLs."
     (remove-overlays nil nil 'cider-temporary t)
     (cider--clear-compilation-highlights)
     (cider--quit-error-window)
-    (cider--cache-ns-form)
-    (let ((filename (buffer-file-name buffer)))
+    (let ((filename (buffer-file-name buffer))
+          (ns-form  (cider-ns-form)))
       (cider-map-connections
        (lambda (connection)
+         (with-current-buffer connection
+           (setq cider--last-ns-form ns-form))
          (cider-request:load-file (cider-file-string filename)
                                   (funcall cider-to-nrepl-filename-function
                                            (cider--server-filename filename))
@@ -1795,11 +1787,7 @@ START and END represent the region's boundaries."
 (defun cider--quit-connection (conn)
   "Quit the connection CONN."
   (when conn
-    (cider--close-connection-buffer conn)
-    ;; clean the cached ns forms for this connection in all Clojure buffers
-    (dolist (clojure-buffer (cider-util--clojure-buffers))
-      (with-current-buffer clojure-buffer
-        (remhash conn cider--ns-form-cache)))))
+    (cider--close-connection-buffer conn)))
 
 (defun cider-quit (&optional quit-all)
   "Quit the currently active CIDER connection.
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1065,9 +1065,6 @@ evaluation command.  Honor `cider-auto-jump-to-error'."
       #'identity))
   "Function to translate Emacs filenames to nREPL namestrings.")
 
-(defvar-local cider--last-ns-form nil
-  "Ns-form of the most recent interactive evaluation.")
-
 (defun cider--prep-interactive-eval (form connection)
   "Prepare the environment for an interactive eval of FORM in CONNECTION.
 Ensure the current ns declaration has been evaluated (so that the ns
@@ -1078,10 +1075,8 @@ window."
   (cider--quit-error-window)
   (let ((cur-ns-form (cider-ns-form)))
     (when (and cur-ns-form
-               (not (string= cur-ns-form
-                             (buffer-local-value 'cider--last-ns-form connection)))
-               (not (cider-ns-form-p form)))
-      (cider-repl--cache-ns-roots cur-ns-form connection)
+               (not (cider-ns-form-p form))
+               (cider-repl--ns-form-changed-p cur-ns-form connection))
       (when cider-auto-track-ns-form-changes
         ;; The first interactive eval on a file can load a lot of libs. This can
         ;; easily lead to more than 10 sec.
@@ -1089,8 +1084,7 @@ window."
           ;; TODO: check for evaluation errors
           (cider-nrepl-sync-request:eval cur-ns-form connection)))
       ;; cache at the end, in case of errors
-      (with-current-buffer connection
-        (setq cider--last-ns-form cur-ns-form)))))
+      (cider-repl--cache-ns-form cur-ns-form connection))))
 
 (defvar-local cider-interactive-eval-override nil
   "Function to call instead of `cider-interactive-eval'.")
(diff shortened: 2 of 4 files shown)


bbatsov:
> Highlighting of vars currently uses project directory to infer the root namespaces. Would be happy to know a better way without recurring to parsing project files.

Well, we're already doing this in the middleware somewhere (I think in the test running code), so it just needs to be extracted into some op and the results can be cached by project as this is not really going to change.

vspinu:
> we're already doing this in the middleware somewhere 

I have run overtime on this one. Would appreciate if you can dig specific location of the code and I will look into middleware in in due time.

> project as this is not really going to change.

Right, but the code could be executed from different projects. Also there is some functionality in cider to associate a connection with a project. 

This is how I see it. Store globaly a hash of project.dir->root-ns when code is evalled from a project.dir, highlight corresponding ns. The trick is to figure out which output comes form which project. So probably a much better idea is to cache all project's root namespace within a connection and highlight all of them.  

All this part is a bit more involved so I would suggest that you merge this the way it is and I will add the middleware within a week or two.

dpsutton:
I added `directory-namespaces` and `ns-common-prefix` to cider-nrepl at https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/stacktrace.clj#L82

I'm not precisely sure I'm following as this seems applicable only to cljs and I'm not sure how much info comes out of the middleware when using cljs. But this is the only thing that I see relating to namespaces for the current project.

bbatsov:
Just for the protocol - that's the code I had in mind as well. 

> I'm not precisely sure I'm following as this seems applicable only to cljs and I'm not sure how much info comes out of the middleware when using cljs. But this is the only thing that I see relating to namespaces for the current project.

Knowing how many things are broken with Cljs I wouldn't fret too much over this.

> All this part is a bit more involved so I would suggest that you merge this the way it is and I will add the middleware within a week or two.

Fine by me.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New Features
 
+* Allow mouse interaction with stacktraces in REPL.
```
bbatsov: Seems to me this works with more than just stacktraces.

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
     (insert-before-markers (cadr ansi-color-context))
     (setq ansi-color-context nil)))
 
+(defvar cider-repl--root-ns-highlitht-template "\\<%s[^$/: \t\n]+"
+  "Regexp used to highlight root ns in REPL buffers.")
+
+(defvar-local cider-repl--root-ns-regexp nil
+  "Cache of root ns regexp in REPLs")
```
bbatsov: You forgot the final `.` here.

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
...
+
+(defun cider-repl--apply-current-project-color (string)
+  "Fortify project's root namespace to make stacktraces more readable.
```
bbatsov: Fontify

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
...
+Foreground of `cider-stacktrace-ns-face' is used to propertize matched
+namespaces."
```
bbatsov: You forgot to mention STRING in the docstring.

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
...
+  (if (null nrepl-project-dir)
+      string
+    (unless nil; cider-repl--root-ns-regexp
```
bbatsov: I'd add a space after `nil`.

vspinu:
> So probably a much better idea is to cache all project's root namespace within a connection and highlight all of them.

I actually don't need any middleware for this. Each time an eval happens the root ns is simply added to the cache. Will do that.

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
...
+      (let ((root (file-name-nondirectory (directory-file-name nrepl-project-dir))))
```
bbatsov: `when-let` maybe?

CHANGELOG.md
vspinu: It does, but the doc becomes too long and with little extra informational context. Most people care about stacktraces. I can expand of course.

cider-repl.el
```diff
@@ -550,6 +549,32 @@ When there is a possible unfinished ansi control sequence,
...
+        (setq cider-repl--root-ns-regexp
+              (format cider-repl--root-ns-highlitht-template
+                      (replace-regexp-in-string "[_-]+" "[_-]+" root)))))
+    (let ((start 0)
```
bbatsov: Some explanations of this part of the code might be useful.

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
+    (print  "\\[\\([^][$ \t]+\\).* +\\([^ \t]+\\) +\\([0-9]+\\)\\]" 0 1 2 3)
+    (timbre "\\(TRACE\\|INFO\\|DEBUG\\|WARN\\|ERROR\\) +\\(\\[\\([^:]+\\):\\([0-9]+\\)\\]\\)" 2 3 nil 4))
+  "Alist holding regular expressions for inline location references.
+Each element in the alist has the form (NAME REGEXP HIGHLIGHT VAR FILE
+LINE), where NAME is the identifier of the regexp, REGEXP - regexp matching
+a location, HIGHLIGHT - sub-expression matching region to highlight on
+mouse-over, VAR - sub-expression giving Clojure VAR to look up. FILE is
+currently used for help-echo and can be nil.")
+
+(defun cider--locref-at-point-1 (reg-list &optional pos)
```
bbatsov: Interactive private function? It should also have a docstring.

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
     (cider-stacktrace-initialize causes)
     (font-lock-refresh-defaults)))
 
+
+;;; Plain Text Stacktraces
```
bbatsov: and logs I guess

cider-repl.el
vspinu: That nil is for debugging purpose. I forgot to delete it.

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
 
+

+;;; Plain Text Stacktraces
+
+(defvar cider-locref-regexp-alist
+  `((stdout "[ \t]\\(at \\([^$(]+\\).*(\\([^:()]+\\):\\([0-9]+\\))\\)" 1 2 3 4)
+    (aviso  "^[ \t]*\\(\\([^$/ \t]+\\).*? +\\([^:]+\\): +\\([0-9]+\\)\\)" 1 2 3 4)
+    (print  "\\[\\([^][$ \t]+\\).* +\\([^ \t]+\\) +\\([0-9]+\\)\\]" 0 1 2 3)
+    (timbre "\\(TRACE\\|INFO\\|DEBUG\\|WARN\\|ERROR\\) +\\(\\[\\([^:]+\\):\\([0-9]+\\)\\]\\)" 2 3 nil 4))
```
bbatsov: Probably log4j and whatever else popular log formats are there should be supported out of the box as well.

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
+currently used for help-echo and can be nil.")
+
+(defun cider--locref-at-point-1 (reg-list &optional pos)
+  (interactive)
+  (save-excursion
+    (let ((pos (or pos (point))))
+      (goto-char pos)
+      (beginning-of-line)
+      (when (re-search-forward (nth 1 reg-list) (point-at-eol) t)
+        (let ((ix-highlight (or (nth 2 reg-list) 0))
```
bbatsov: ix?

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
+number.  See `cider-locref-regexp-alist' for how to specify regexes for
+locref look up."
+  (interactive)
+  (seq-some (lambda (rl) (cider--locref-at-point-1 rl pos))
+            cider-locref-regexp-alist))
+
+(defun cider-jump-to-locref-at-point (&optional pos)
+  "Identify loction reference at point and navigate to it.
+This function is used from help-echo property inside REPL buffers and uses
+regexes from `cider-locref-regexp-alist' to infer locations at point."
```
bbatsov: You haven't mentioned POS in the docstring.

cider-stacktrace.el
vspinu: Index :)

cider-stacktrace.el
```diff
@@ -789,6 +789,93 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
+                 (let ((map (make-sparse-keymap)))
+                   (define-key map [return]  'cider-jump-to-locref-at-point)
+                   (define-key map [mouse-2] 'cider-jump-to-locref-at-point)
+                   map))
+    o)
+  "Overlay used during hoovering on location references in REPL buffers.
+One for all REPLs.")
+
+(defun cider-locref-help-echo (win obj pos)
+  "Function for help-echo property in REPL buffers."
```
bbatsov: You haven't mentioned the params in the docstring.

cider-stacktrace.el
vspinu: For log4j you need to configure the appender's patter and there is no default configuration. The user is expected to customize this as needed. I probably should make it a custom.

bbatsov: Sounds like a good idea to me.

vspinu:
Ok. Fixed all doc strings and implemented root ns tracking without middleware complications. Needed to refactor ns-chaching a bit for this.

cider-interaction.el
```diff
@@ -1065,22 +1065,17 @@ evaluation command.  Honor `cider-auto-jump-to-error'."
       #'identity))
   "Function to translate Emacs filenames to nREPL namestrings.")
 
-(defvar-local cider--ns-form-cache (make-hash-table :test 'equal)
```
bbatsov: I don't think you realize what this is used for - it tracks the state of the ns form in the source buffer and it can be different for different connections. E.g. something might have been evaluated in the cljs connection, but not in the clojure connection. So, the change you made is the definitely not equivalent and would mean out-of-sync evaluations for different connections.

cider-interaction.el
```diff
@@ -1065,22 +1065,17 @@ evaluation command.  Honor `cider-auto-jump-to-error'."
...
-are ns forms. This allows every connection to keep track of the ns
-form independently.")
-
-(defun cider--cache-ns-form ()
-  "Cache the form in the current buffer for the current connection."
-  (puthash (cider-current-connection)
-           (cider-ns-form)
-           cider--ns-form-cache))
+(defun cider--cache-ns-form (ns-form)
+  "Cache NS-FORM in the current connection."
```
bbatsov: Once again - that's for the source buffer, not for the connection, so you can't really cache in the connection...

cider-interaction.el
vspinu: I missed that it's buffer local indeed. Will have to revert that part.

vspinu: >  E.g. something might have been evaluated in the cljs connection, but not in the clojure connection. 

Do you mean that you can get two forms per buffer, one for cljs and another for clj connection? The old code was clearly not doing  for a range of reasons.

If we go with the assumption that one buffer can have one ns form, common for both cljs and clj then the hash table is not necessary.

bbatsov: My bad - I actually meant that several connections of the same type can exist for a particular project and the state of the evaluation of the project buffers would be different for them.

vspinu: > project and the state of the evaluation of the project buffers would be different for them.

How can they be dissynchronized by user actions?  AFAICS there is no way to create clj connection and then add up cljs connection. So any evaluation in cljc buffers should end up in both buffers, otherwise it's a bug. In my mind two ns forms can get dissynchronized only when error occurred in ns form itself on cljs but not on clj. But this situation is orthogonal to hash-map issue, nor it is handled currently in the master code.

vspinu: And BTW, there is a bug in current implementation because ns is sent only to one connection. If you open a cljc and eval a form you will get an error in cljs side but not on cljc. That because ns was evaluated in clj only.  Will try to fix that within this pull.

vspinu:
> Once again - that's for the source buffer, not for the connection, so you can't really cache in the connection...

It is for connection actually. Current ns is property of the connection not of the buffer where evaluation happened.  Even if you perhaps wanted a buffer local implementation you still had a global one all this time because of the how hash maps work in emacs. In emacs local hash tables behave as if global.

You derailed me quite a bit on this one, but well ... I ended up fixing a bunch of things, so I guess it was worth it :)

cider-interaction.el
bbatsov: I meant you can have 3 clj connection for the same project with different evaluation state for instance. There's not way to know whether a ns form is in sync for every project without some connection mapping.

vspinu: 1 project - 3 connection. All 3 store buffer local value of ns-form, this gives the last evaluation state of the project for each connection. On each evaluation ns-form is matched; if different, teh connection state is outdated and ns-form is re-evaluated. What other mapping do you need?

bbatsov:
> It is for connection actually. Current ns is property of the connection not of the buffer where evaluation happened. Even if you perhaps wanted a buffer local implementation you still had a global one all this time because of the how hash maps work in emacs. In emacs local hash tables behave as if global.

That has nothing to do with the connection current ns - the purpose of this is to auto-evaluate the ns form of the source buffer in case it has changed since last time. Basically many people were forgetting to evaluated their ns forms after changing them and were getting strange errors (at least for them). Apart from this there's no need to cache the ns form.

vspinu:
In my implementation whenever there is a change in ns it will be detected, irrespective of the number of connections. The only difference between what you seem to want and what I did, is that in my implementation there will be one extra ns-eval when user switches from one buffer to another, but that's a triviality and it was _really_ working like this all this time with "local" hash table. 

But I guess we started going in circles by now. Let's get on this freshly tomorrow by you providing a use pattern when my implementation breaks.

bbatsov:
> In my implementation whenever there is a change in ns it will be detected, irrespective of the number of connections. The only difference between what you seem to want and what I did, is that in my implementation there will be one extra ns-eval when user switches from one buffer to another, but that's a triviality and it was really working like this all this time with "local" hash table.

I'll have to read this again more carefully, but from what I saw now you just check if the ns hasn't changed and re-evaluate it otherwise. And for some reason you also put the last eval ns in the connection buffer which seems extremely strange to me. IMO this is all just source-buffer state. From time to time I actually regret adding this, as it made something pretty simple relatively complex just because some people would not remember to evaluate their ns changes... I recall I considered disabling it by default, but I think I never got to adding a defcustom for this. 

> But I guess we started going in circles by now. Let's get on this freshly tomorrow by you providing a use pattern when my implementation breaks.

Perhaps. Was pretty tired yesterday, I might have missed something in the implementation.

bbatsov:
Ah, I see I made this configurable, but decided not to disable it by default. 

Anyways, see my new inline comment. I finally understand your idea.

vspinu:
> it made something pretty simple relatively complex 

It's simple indeed but with a complex implementation.

>  I myself would be pretty surprised to see some extra ns evaluation just because I switched ns.

Still, you were not surprised (or even aware of it) all this time ;)

>  And I generally don't understant what's wrong with the hashmap approach 

It's heavy, complex and too buggy for the benefits it adds. Try 

```el
(defvar-local tt (make-hash-table :test 'equal))
;; from buffer A
(puthash :A 1 tt)
;; from buffer B
(puthash :B 1 tt)
;; from buffer C
;; C-h v tt RET to see the state of tt from arbitrary buffer
;; => no local state!
```
So all this time you relied on a global hash table of `cider--ns-form-cache` and the system exhibited precisely the same behavior as the code in my PR. Everyone was just fine, right? So, I guess, it should prove my (and your) point that the current implementation is too complex for this simple feature.

In short, the complexities of buffer local hash-maps are 1) need to initialize hash map in each buffer 2) need to destroy in each buffer after connection is dropped 3) need to use complex reference objects (buffers as keys; buggy hash-tables) 4) non-trivial mapping semantics (we wouldn't have this discussion if it wasn't)

The current PR solves all of the above but keeps a small redundancy in evaluation of the ns forms. 

> after all you can easily have the ns mappings on the connection if you prefer to do it this way.

In this case "centralized" is better than "decentralized". If you really think it's worth to improve on what I already have, I can easily add a local plist within each connection with `ns -> ns-form` mapping.  No hash tables, no init/cleanup of source buffers, no complex objects - just strings and lists.

bbatsov:
Fair enough. I'd still prefer to avoid the redundant evaluation, but I can live with this. One thing, though - probably it makes sense to insert a couple of debug-type messages regarding evaluating out-of-sync namespaces, so users have a better idea what's going on.

And you need to rebase.

vspinu:
Done. I have added ns->ns-form plist cache within each connection. It's simple enough to be worth it.

bbatsov:
Excellent! 👍  Thanks for another awesome contribution!

P.S. Now that this is merged tomorrow or on Friday I might cut a release to celebrate EuroClojure.

PR #2046, opened 2017-07-16
Filter stacktrace

dpsutton:
invoking function from button was not working the way I expected. It was not
seeing new values for defvar's. Now it's called by the button handler and it is
working correctly.

quick update to fix the "Project-Only" not underlining when clicked bug. See https://github.com/clojure-emacs/cider/pull/2037

Also, @vspinu was mentioning possible changes to the wording. I'm happy to do that here on this before merging if that seems like the way to go. As I said in the PR thread I've been thinking in these terms so long that they seem right to me but that means someone with fresh eyes has a much more valuable opinion than mine.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -244,13 +244,21 @@ The error types are represented as strings."
 
 ;; Stacktrace filtering
 
+(defvar cider-stacktrace--all-negative-filters
+  '(clj tooling dup java repl)
+  "Filters that remove stackframes")
+
+(defvar cider-stacktrace--all-positive-filters
+  '(project)
+  "Filters that ensure stackframes are shown")
+
 (defun cider-stacktrace--face-for-filter (filter neg-filters pos-filters)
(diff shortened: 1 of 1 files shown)


bbatsov:
> Also, @vspinu was mentioning possible changes to the wording. I'm happy to do that here on this before merging if that seems like the way to go. As I said in the PR thread I've been thinking in these terms so long that they seem right to me but that means someone with fresh eyes has a much more valuable opinion than mine.

We can always adjust this down the road.

vspinu:
> We can always adjust this down the road.

 @dpsutton would you mind proposing a pull to fix it into 

```
  Filter: Project-Only 
    Show: Clojure Java REPL Tooling Duplicates All  (19 frames hidden)
```

At least for the sake of backward compatibility if you don't buy semantic, cognitive and perceptual aspects of it.

Then you can improve it "down the road" howether you wish, but at least avoid frustration of some users (myself included) in the meanwhile.

Issue #2049, opened 2017-07-16
Better mode for cider grimoire


AlainPicard-at-gocatch:
## *cider-grimoire* buffer is in Fundamental mode.

Dear cider maintainers,

This is a (minor) feature request.  I think it would be useful
if, when invoking (cider-grimoire), one would be left in a 
mode more useful than fundamental.  Perhaps something
based off view-mode, at the very least, so one could 
scroll with space bar and backspace.

For extra credit, one could parse the result page from grimoire,
which appears to be formatted in some sort of markdown, and use
the markdown mode fontification regexes to make the page fontified.

This is definitely just a nice to have, in case anyone feels like hacking
on this; no extra functionality would be added by this, just...  elegance.  :-)

Thanks for all the hard work on CIDER.
          Alain Picard

Cider 0.14.0

xiongtx:
This is already the case as of 238c60c6, from #2007.

AlainPicard-at-gocatch:
Thanks!  

Yes, I've just tried the current tip of master, and indeed, the grimoire
is fontified.

It is not, alas, in a view-mode, so hitting spacebar yields:
and error and prints:  Buffer is read-only: #<buffer *cider-grimoire*>

However, I've just noticed that view-mode already runs as a minor mode,
and can just be invoked in a grimoire buffer, so all we're missing now is
a hook we can run when entering the grimoire mode.  (Although I suppose
I could simply use the existing markdown-mode-hook, since I don't use markdown
myself).

Or do you have a better suggestion altogether?  

Thanks again,
  Alain

xiongtx:
Using `view-mode` seems like a good idea. Doesn't seem to clash w/ `markdown-mode`'s bindings either. #2050 should address this. You can also try it out yourself; it's a extremely simple change.

PR #2050, opened 2017-07-17
Use `view-mode` for `cider-grimoire` buffer

xiongtx:
Address [comment](https://github.com/clojure-emacs/cider/issues/2049#issuecomment-315655822) by @AlainPicard-at-gocatch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* [#2050](https://github.com/clojure-emacs/cider/pull/2050) Use `view-mode` for `cider-grimoire` buffers
 * Filter stacktrace to just frames from your project.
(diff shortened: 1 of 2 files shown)


bbatsov:
Please refer the ticket that this fixes in the commit message and update the changelog. 

I wonder if we don't have other similar cases that might benefit from `view-mode`...

xiongtx:
> Please refer the ticket that this fixes in the commit message and update the changelog.

Fixed.

> I wonder if we don't have other similar cases that might benefit from `view-mode`...

Not as far as I can tell, unless you want [*all* popup buffers](https://github.com/clojure-emacs/cider/blob/5854f3a5a1b8ba9bda48e635a79ff8eb41a29bce/cider-popup.el#L98) to use `view-mode`, which I don't think should be the case.

PR #2051, opened 2017-07-17
[Fix #390] Workaround for orphaned java process on windows

### LINKED ISSUE #390: Residual JVM After Quitting
Whenever I quit emacs or run nrepl-quit I get an orphaned JVM process remaining.

Attached should be an image showing the emacs process tree after running nrepl-jack-in.

I also tested with CIDER 20131018.1553 from a fresh .emacs.d with just cider/clojure-mode and had the same problem.

nrepl-jvm-bug

benedekfazekas:
after quitting the REPL. Use `interrupt-process` instead of `kill-process` if OS
is windows.

-----------------

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -616,6 +616,13 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 
 
 ;;; Client: Process Handling
+(defun nrepl--kill-process (proc)
+  "Kill PROC using the appropriate, os specific way.
+Implement a workaround to clean up an orphaned JVM process left around
+after exiting the REPL on some windows machines."
+  (if (memq system-type '(cygwin windows-nt))
+      (interrupt-process proc)
+    (kill-process proc)))
 
 (defun nrepl--maybe-kill-server-buffer (server-buf)
   "Kill SERVER-BUF and its process, subject to user confirmation.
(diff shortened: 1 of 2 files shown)


jonpither:
This will be extremely useful for us poor developers working in a Windows environment.

nrepl-client.el
```diff
@@ -616,6 +616,11 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 
 
 ;;; Client: Process Handling
+(defun nrepl--kill-process (proc os)
```
bbatsov: What's the point in passing a dynamic variable here? I'd just use `system-type` internally.

nrepl-client.el
```diff
@@ -616,6 +616,11 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
...
+  "Kill PROC using the appropriate, OS specific way."
+  (if (memq os (list 'cygwin 'windows-nt))
```
bbatsov: Why don't you simply quote this list?

nrepl-client.el
```diff
@@ -616,6 +616,11 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 
 
 ;;; Client: Process Handling
+(defun nrepl--kill-process (proc os)
+  "Kill PROC using the appropriate, OS specific way."
```
bbatsov: I'd explain here the need for this, so it's clear it's a form of a workaround.

PR #2052, opened 2017-07-18
Different major modes for spec browser buffers

### LINKED ISSUE #2035: `cider-browse-spec-mode` should allow navigating back and forth
## Expected behavior

`cider-browse-spec-mode` should allow navigating back and forth between specs, like `cider-doc view-mode` does.

## Actual behavior

`cider-browse-spec-mode` only allows navigation to the top-level `"All specs in registry"` buffer.

### Analysis

`cider-browse-spec` is modeled on `cider-browse-ns`, but there's a critical difference: it is possible to navigate from one spec to another from the spec browser, but not from one symbol to another from the ns browser. This is because namespaces and symbols are a tree, while specs are a graph.

`cider-docview-mode` is [derived from `help-mode`](https://github.com/clojure-emacs/cider/blob/ff4c0ed9bc70c3f48ad74d08c9f7c97e91b2553c/cider-doc.el#L177), which has built-in support for navigation via `help-go-back` and `help-go-forward`. `cider-browse-spec-mode` should do the same.

1. View a spec that uses another spec in its definition via `cider-browse-spec`
2. Navigate to the other spec (by clicking on it)
3. Note that you cannot navigate back to the first spec, only to the top-level all specs browser

### CIDER version information

Commit ff4c0ed9bc70c3f48ad74d08c9f7c97e91b2553c.

### LINKED ISSUE #2036: `cider-browse-spec-navigation` is always `nil`
## Expected behavior

`cider-browse-spec-navigation`'s stated purpose is to keep track of the spec browser navigation stack. However, it is always `nil`.

## Actual behavior

`cider-browse-spec-navigation` is always `nil`. This prevents functions like `cider-browse-spec--print-curr-spec-example` from working correctly.

1. View a spec that uses another spec in its definition via `cider-browse-spec`
2. Navigate to the other spec (by clicking on it)
3. See that `cider-browse-spec-navigation` remains `nil`

### CIDER version information

Commit ff4c0ed9bc70c3f48ad74d08c9f7c97e91b2553c.

xiongtx:
Fixes #2035 and fixes #2036.

### DIFF
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -25,7 +25,7 @@
 ;;
 ;; Display a spec description you can browse.
 ;; Pressing <enter> over a sub spec will take you to the description of that sub spec.
-;; Pressing ^ will take you up in the navigation stack.
+;; Pressing ^ takes you to the list of all specs.
 
 ;; M-x cider-browse-spec-all
 ;;
@@ -47,12 +47,6 @@
 (defconst cider-browse-spec-buffer "*cider-spec-browser*")
 (add-to-list 'cider-ancillary-buffers cider-browse-spec-buffer)
 
-(defvar cider-browse-spec-navigation '()
-  "Keeps the cider spec browser navigation stack.
-A list of strings which are specs or specs searches.
-First of the list is the top of the stack.
-Specs searches are encoded as (search \"regex\")
-For example: (\":ring.request/header-name\" \":ring.request/headers\" \":ring/request\" (search \"ring.+\"))")
 (defconst cider-browse-spec-example-buffer "*cider-spec-example*")
 (add-to-list 'cider-ancillary-buffers cider-browse-spec-example-buffer)
 
@@ -62,11 +56,10 @@ For example: (\":ring.request/header-name\" \":ring.request/headers\" \":ring/re
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map cider-popup-buffer-mode-map)
     (define-key map (kbd "RET") #'cider-browse-spec--browse-at-point)
-    (define-key map "^" #'cider-browse-spec--navigate-back)
     (define-key map "n" #'cider-browse-spec--next-spec)
     (define-key map "p" #'cider-browse-spec--prev-spec)
-    (define-key map "e" #'cider-browse-spec--print-curr-spec-example)
-    map))
+    map)
+  "Keymap for `cider-browse-spec-mode'.")
 
 (defvar cider-browse-spec-mouse-map
   (let ((map (make-sparse-keymap)))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New Features
 
+* [#2052](https://github.com/clojure-emacs/cider/pull/2052) Use different major modes for spec browser.
```
bbatsov: This functionality is unreleased, so the changelog entry is going to be confusing.

cider-browse-spec.el
```diff
@@ -311,32 +338,23 @@ default value, \"\", matches all specs in the registry."
...
       (cider-browse-spec--draw-list-buffer (current-buffer)
                                            (if (string-empty-p filter-regex)
-                                   "All specs in registry"
-                                 (format "All specs matching regex `%s' in registry" filter-regex))
-                               specs))))
+                                               "All specs in registry"
+                                             (format "All specs matching regex `%s' in registry" filter-regex))
+                                           specs))))
+
+(defun cider-browse-spec-all* ()
```
bbatsov: Why not the other command with a prefix argument?

CHANGELOG.md
xiongtx: Not sure what you mean. Isn't every PR supposed to be accompanied by a changelog entry?

cider-browse-spec.el
xiongtx: Just didn't want to do `(lambda () (interactive) (cider-browse-spec-all))` multiple times, e.g. [here](https://github.com/clojure-emacs/cider/pull/2052/files#diff-3fa4f149fe6f732da43bf2af6b166da1R83).

CHANGELOG.md
bbatsov: Only user-visible changes between stable releases. As this feature has never a part of a stable release as far as the users are concerned nothing changed.

cider-browse-spec.el
bbatsov: Well, I'd simply mark the regexp optional and remove it from the interactive spec (like most of the codebase does). I just don't see the point in having oddly named functions just to account for an optional param.

cider-browse-spec.el
```diff
@@ -45,17 +45,16 @@
 
 ;; The buffer names used by the spec browser
 (defconst cider-browse-spec-buffer "*cider-spec-browser*")
-(defconst cider-browse-spec-example-buffer "*cider-spec-example*")
-
-(push cider-browse-spec-buffer cider-ancillary-buffers)
-(push cider-browse-spec-example-buffer cider-ancillary-buffers)
+(add-to-list 'cider-ancillary-buffers cider-browse-spec-buffer)
```
bbatsov: Not sure what the reasoning is behind this chage, but I don't mind it. I do mind, however, that most of the code uses `push` and we're using `add-to-list` just here. Maybe we should just add some wrapper like `cider-register-ancillary-buffer` or something to be able to easily make changes to this globally.

cider-browse-spec.el
```diff
@@ -81,11 +74,47 @@ For example: (\":ring.request/header-name\" \":ring.request/headers\" \":ring/re
...
+    (define-key map (kbd "RET") #'cider-browse-spec--browse-at-point)
+    (define-key map "^" #'cider-browse-spec-all*)
+    (define-key map "e" #'cider-browse-spec--print-curr-spec-example)
+    (define-key map "n" #'cider-browse-spec--next-spec)
+    (define-key map "p" #'cider-browse-spec--prev-spec)
+    map)
+  "Keymap for `cider-browse-spec-view-mode'.")
+
+(define-derived-mode cider-browse-spec-view-mode help-mode "Spec view"
+  "Major mode for displaying CIDER spec.
```
bbatsov: CIDER spec -> Clojure spec

cider-browse-spec.el
```diff
@@ -81,11 +74,47 @@ For example: (\":ring.request/header-name\" \":ring.request/headers\" \":ring/re
...
+  (setq-local cider-browse-spec--current-spec nil)
+  (setq-local electric-indent-chars nil)
+  (when cider-special-mode-truncate-lines
+    (setq-local truncate-lines t)))
+
+(defvar cider-browse-spec-example-mode-map
+  (let ((map (make-sparse-keymap)))
+    (set-keymap-parent map cider-popup-buffer-mode-map)
+    (define-key map "^" #'cider-browse-spec-all*)
+    (define-key map "e" #'revert-buffer)
```
bbatsov: This is probably a typo, right? Usually just `g` is revert.

CHANGELOG.md
xiongtx: :white_check_mark:

cider-browse-spec.el
xiongtx: Not a typo. `e` is bound to `cider-browse-spec--print-curr-spec-example` in `cider-browse-spec-view-mode-map`, and I wanted `e` to do the same in `cider-browse-spec-example-mode`.

I'll bind `e` to `cider-browse-spec--print-curr-spec-example` directly instead of to `revert-buffer`.

cider-browse-spec.el
xiongtx: Not sure whether most the code uses `push`, but [`cider-doc` uses `add-to-list`](https://github.com/clojure-emacs/cider/blob/7daf4e968de21e7eb7bb2dc1b6ca30ae4e990532/cider-doc.el#L261), which is the right thing to use b/c we don't want repeated elements.

Something like `cider-register-ancillary-buffer` would be outside the scope of this PR and, IMO, unnecessary. We should find places where we use `push` and switch them to `add-to-list` instead.

cider-browse-spec.el
xiongtx: :white_check_mark:

xiongtx:
@bbatsov By the way, can you make it so that contributors can select reviewers? That'd be useful.

bbatsov:
> @bbatsov By the way, can you make it so that contributors can select reviewers? That'd be useful.

Where can I configure this? I looked at the repo setting but didn't see anything about PR review settings. I always thought this is available just to people with commit access.

xiongtx:
I suppose I don't have commit access then.

PR #2057, opened 2017-07-20
Use `cider--font-lock-ensure` for compatibility with Emacs 24.5

### LINKED ISSUE #2056: Can't use cider-browse-spec in emacs24.5
## Expected behavior
`cider-browse-spec` displays spec data
## Actual behavior
`cider-browse-spec` opens empty buffer and displays message:
`Symbol's function definition is void: font-lock-ensure`

- use emacs24.5.1
- enable cider
- `cider-browse-spec`
- select some spec
- press enter

### CIDER version information
```
;; CIDER 0.15.0 (package: 20170720.248) (London), nREPL 0.2.12
;; Clojure 1.9.0-alpha15, Java 1.8.0_121
```

### Lein/Boot version

Leiningen 2.7.1 on Java 1.8.0_121 Java HotSpot(TM) 64-Bit Server VM  

### Emacs version

GNU Emacs 24.5.1 (x86_64-pc-linux-gnu, GTK+ Version 3.18.9) of 2016-04-17 on lgw01-04, modified by Debian

### Operating system

Ubuntu 16.04.2 LTS

xiongtx:
Fixes #2056.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## master (unreleased)
 
+### New Features
+
+### Changes
+
+### Bugs Fixed
+
+* [#2057](https://github.com/clojure-emacs/cider/pull/2057) Use `cider--font-lock-ensure` for compatibility with Emacs 24.5.
+
 ## 0.15.0 (2017-07-20)
 
 ### New Features
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -256,7 +256,7 @@ a more user friendly representation of SPEC-FORM."
                 (clojure-mode)
                 (insert (cider-browse-spec--pprint spec-form))
                 (indent-region (point-min) (point-max))
-                (font-lock-ensure)
+                (cider--font-lock-ensure)
                 (buffer-string)))
       (cider--make-back-forward-xrefs)
       (current-buffer))))


bbatsov:
This should be in the changelog as it's a bugfix for a stable release. 

We should probably update travis to run the byte-compilation checks for all supported Emacs version (https://github.com/clojure-emacs/cider/blob/master/.travis.yml#L8) to avoid such issues down the road.

xiongtx:
Add changelog and full test matrix for Emacs versions and tests. 

I'm not familiar enough w/ Travis to know whether [this line under `matrix`](https://github.com/clojure-emacs/cider/blob/master/.travis.yml#L19) is necessary.

xiongtx:
Looks like filling out the test matrix is already paying off--just discovered we're using `directory-files-recursively`, which is introduced by Emacs 25, without defining it in `cider-compat`.

bbatsov:
I see that the tests are still failing because of `directory-files-recursively`. I'm also not sure is the checkdoc task working properly. I remember a few PRs with checkdoc errors that weren't flagged for some reason. Also - checkdoc hasn't changed much in a long time, do we need to run it on every supported Emacs?

xiongtx:
I'm mystified as to why `directory-files-recursively` is causing the test to fail. I've posted a question on [Emacs StackExchange](https://emacs.stackexchange.com/questions/34319/function-might-not-be-defined-at-runtim).

> I'm also not sure is the checkdoc task working properly. I remember a few PRs with checkdoc errors that weren't flagged for some reason. 

`checkdoc` produces [warnings](https://travis-ci.org/clojure-emacs/cider/jobs/255846473#L865), not errors, so the tests still pass. 

There's no option to make `checkdoc` use `error` instead of `warn`, but we can run `checkdoc-file` on all source files and check whether the `*Warnings*` buffer exists after that. If it does, we can exit w/ an error.

>  `checkdoc` hasn't changed much in a long time, do we need to run it on every supported Emacs?

I don't see the harm. Travis CI has a 200 build limit IIRC and we're not anywhere near that.

xiongtx:
I'm going to open another PR w/ the TravisCI stuff. We'll keep this PR focused on font lock.

PR #2059, opened 2017-07-22
Fix some TravisCI issues

xiongtx:
@bbatsov originally raised [this issue](https://github.com/clojure-emacs/cider/pull/2057#issuecomment-316832459) when discussing #2057.


test/cider-checks.el
```diff
@@ -20,5 +20,9 @@
                               "\\`[^.].*\\.el\\'" t)))
   (dolist (file files)
     (checkdoc-file file))
+  (when (get-buffer "*Warnings*")
+    (message "Failing due to checkdoc warnings...")
```
bbatsov: I think we should dump the contents of the warnings buffer as well, otherwise it would be really hard for people to figure out what exactly is wrong.

.travis.yml
```diff
@@ -2,11 +2,15 @@ sudo: false
 language: emacs-lisp
 env:
   - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test
+  - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test-bytecomp
   - EMACS_BINARY=emacs-24.5-travis MAKE_TEST=test
+  - EMACS_BINARY=emacs-24.5-travis MAKE_TEST=test-bytecomp
   - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test
+  - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test-bytecomp
+  - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test-checks
```
bbatsov: We should add 25.2 here as well, as it was recently released. Probably in a couple of releases we should drop support for Emacs 24 completely, but it seems too early now.

test/cider-checks.el
xiongtx: Warnings are already [dumped line-by-line](https://travis-ci.org/clojure-emacs/cider/jobs/256287471#L865).

test/cider-checks.el
```diff
@@ -6,6 +6,9 @@
 ;; This assumes that all CIDER dependencies are already on the package dir
 ;; (probably from running `cask install').
 
+;; NOTE: `checkdoc-file' is introduced in Emacs 25, so do not run this check with
+;; Emacs 24.x
```
bbatsov: I guess you can also add a version check around the code itself.

test/cider-checks.el
bbatsov: Ah, OK. Thought they were only in the warnings buffer.

test/cider-checks.el
xiongtx: :white_check_mark:

.travis.yml
xiongtx: :white_check_mark:

xiongtx:
Most tests are passing, but there's an [outstanding issue](https://emacs.stackexchange.com/questions/34341/error-called-with-1-args-to-fill-0-format-fields) with byte-compilation tests for Emacs 24.x.

bbatsov:
> Most tests are passing, but there's an outstanding issue with byte-compilation tests for Emacs 24.x.

Yeah, probably this is a bug, or something changed in Emacs 25. I guess you can simply replace `cl-assert` with some check + `error`. Seems like the simple solution to this problem.

xiongtx:
R.e. the `cl-assert` issue, it's actually a bug in CIDER. See [explanation](https://emacs.stackexchange.com/a/34350/10269).

xiongtx:
@bbatsov [All tests pass](https://travis-ci.org/clojure-emacs/cider/builds/256458182).

xiongtx:
@bbatsov All tests pass.

bbatsov:
> R.e. the cl-assert issue, it's actually a bug in CIDER. See explanation.

Quite interesting. 

Thanks for tackling all of this! Great work!

bbatsov:
P.S. It would be great if you can do all of this for `clojure-mode` as well.

vspinu:
@bbatsov would you be so kind and give some more consideration to the parameter order and double space "conventions"? 

Double space is unnatural and never used in other context. So you are bound to forget those and run into `make test` issues.  Each time you forget about it - go and figure location, fix, re-commit, re-check etc. Good 5-10 minutes of wasted time with 0 benefits. Plus flow disruption.

Same goes with order of arguments, wasting precious minutes of how to twist untwistable and forcing users to read pseudo-english is rather silly.

As to the "conventions" argument, no-one seems to care about that in the Emacs proper. Pretty much every core file I tried has checkdoc issues much more serious than double spaces and order of arguments.

Conventions which don't benefit the user or the programmer are silly and deserve to be disregarded anyways. These two are clearly of that sort.

xiongtx:
The order of arguments check is already [turned off](https://github.com/clojure-emacs/cider/blob/v0.15.0/test/cider-checks.el#L16).

The two-space rule is an Emacs convention that should be followed. We *could* turn it off by setting `sentence-end-double-space` to `nil` while running `checkdoc`, but I don't think we should.

vspinu:
> should be followed

Why is that exactly? Even auto-fill doesn't respect it why should you?

Double space adds more inefficiencies because it happens more often. So, if you already disabled order of args you _should_ consider disabling double space.

bbatsov:
> Double space adds more inefficiencies because it happens more often. So, if you already disabled order of args you should consider disabling double space.

We've maintained the codebase consistent in this regard since day 1 of the project and I see little point to change this now. It's not a big effort and you simply get used to this after a while. Most Elisp projects stick to this convention so it stopped bothering me after a while. I'm also using `flycheck`, so I get immediate visual feedback if something is somewhere there's a missing space. (and I think many Elisp hackers would benefit from adding it)

> Double space is unnatural and never used in other contexts. So you are bound to forget those and run into make test issues. Each time you forget about it - go and figure location, fix, re-commit, re-check etc. Good 5-10 minutes of wasted time with zero benefit. Plus flow disruption.

I certainly don't want to inconvenience anyone, but I'm big on consistency in the codebase of a project. Can't really go back on this given how much code was written following this style - "fixing" this would create a diff that would obscure half the codebase and would create inconsistency with all of my other projects, that I don't want to introduce.

bbatsov:
> Why is that exactly? Even auto-fill doesn't respect it, why should you?

auto-fill doesn't change spacing around punctuation, so I don't see how this is relevant.

vspinu:
> I'm also using flycheck, so I get immediate visual feedback if something is somewhere there's a missing space.

I was never happy with flycheck for emacs. Most of the libraries have checkdoc issues and that would mean constantly seeing those annoying highlights. Cider could be honorable exception where I can keep it active though, but ... it doesn't work in Cider project. I am getting

```
Suspicious state from syntax checker emacs-lisp-checkdoc: Flycheck checker emacs-lisp-checkdoc returned non-zero exit code 255, but its output contained no errors: Invalid function: "top-level"
```
All other projects are fine. Any ideas?

> Can't really go back on this given how much code was written following this style 

Fair enough.

> auto-fill doesn't change spacing around punctuation,

It used to be the case in previous emacses, when you could end up with a sentence starting a new para. But I think that was fixed in 25.x as I haven't encountered the issue for a while.

vspinu:
I am unable to run test-bytecomp and test-checks on my local machine with emacs 25.2 and 26. I am getting cryptic stuff like:

```
cider$ make test-checks
emacs-25.2 --version
GNU Emacs 25.2.2
Copyright (C) 2017 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
cask exec emacs-25.2 -q --no-site-file --no-site-lisp --batch \
	-l test/cider-checks.el ./
Autoloading failed to define function checkdoc-file
Makefile:31: recipe for target 'test-checks' failed
make: *** [test-checks] Error 255
```

and 

```
cider$ make test-bytecomp
Compute dependencies
emacs25 --version
GNU Emacs 25.2.2
Copyright (C) 2017 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
cask exec emacs25 --no-site-file --no-site-lisp --batch \
	-l test/cider-bytecomp-warnings.el cider-test.el
Making version-control local to cider-autoloads.el while let-bound!
Loading /home/vspinu/VC/cider/cider-autoloads.el (source)...

In toplevel form:
cider-test.el:736:1:Error: the function `seq-filter' is not known to be defined.
Makefile:37: recipe for target 'cider-test.elc-test' failed
make: *** [cider-test.elc-test] Error 1
```

Any pointers on this?

bbatsov:
The second failure looks like some missing require in the tests - `(require 'seq)`, but the first puzzles me. Seems for some reason `check-file` is missing, but it's a built-in, so I can't imagine how this can happen.

xiongtx:
Cannot reproduce on Emacs 25.1.1. Have you run [`make elpa`](https://github.com/clojure-emacs/cider/blob/71f737d444da95c6175b513d500c56ae9711bbd0/.travis.yml#L20) beforehand?

vspinu:
It's on the cask's side (cask/cask#335, cask/cask#295, cask/cask#260 seem to be relevant). I am not following all of that but `export EMACS = emacs25` does seem to work, while `export CASK_EMACS = emacs25` doesn't.

PR #2063, opened 2017-07-23
Bump the development version to 16-SNAPSHOT

dpsutton:
this usually happens right after a release, was working on an issue in cider-nrepl and was surprised it kept fetching the 0.15 version and not snapshot.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -90,7 +90,7 @@ project inference will take place."
 
 (require 'seq)
 
-(defconst cider-version "0.15.0"
+(defconst cider-version "0.16.0-SNAPSHOT"
   "Fallback version used when it cannot be extracted automatically.
 Normally it won't be used, unless `pkg-info' fails to extract the
 version from the CIDER package or library.")


dpsutton:
i think this will also need an equivalent bump in cider-nrepl to prevent breaking

xiongtx:
You mean `16-SNAPSHOT`, right?

dpsutton:
yeah. i had it backwards. the code has been corrected but the title remains. i'll edit it now

cider.el
```diff
@@ -90,7 +90,7 @@ project inference will take place."
 
 (require 'seq)
 
-(defconst cider-version "0.15.0"
+(defconst cider-version "0.16.0-SNAPSHOT"
```
bbatsov: The next release is going to be a bugfix probably, that's why I didn't bump the version right away. I don't really want to bother with a separate branch and cherry-picking commits.

bbatsov: And just for the protocol - here (and in the file's header) it should be `snapshot`, not `SNAPSHOT`.

dpsutton: oh my bad. I didn't think about the bugfix release. I'll close this one. I just ran into this issue trying to look at something in the nrepl side and it kept pulling the clojar release.

PR #2064, opened 2017-07-23
When viewing project only, hide all others

dpsutton:
when showing project only, set all filters to remove all frames and then the project frames remain. I accidentally didn't clobber all of the old ones.


bbatsov:
Seems like a bugfix that should have a changelog entry.

dpsutton:
changelog updated

PR #2067, opened 2017-07-31
[Fix #2045] *cider-scratch* is no longer an ancillary buffer

### LINKED ISSUE #2045: *cider-scratch* is killed on connection exit
If this is by "design" it is a bad idea. You might have quite some code in *scratch* and want to restart connection for whatever reason, and then boom, everything is gone.

vspinu:
AFAICS the only function of ancillary buffers right now is to be killed on `cider-quit`, so this simple change seems to be just what's needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changes
 
+* [#2045](https://github.com/clojure-emacs/cider/issues/2045) `*cider-scratch*` buffers are no longer automatically killed on connection quit.
+
 ### Bugs Fixed
 
 * [#2004](https://github.com/clojure-emacs/cider/issues/2004), [#2039](https://github.com/clojure-emacs/cider/issues/2039), [cider-nrepl#420](https://github.com/clojure-emacs/cider-nrepl/issues/420): Fix namespace issues in instrumentation and debugging commands.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 * Fix interactive evaluation in cljc buffers with only one connection.
 * [#2058](https://github.com/clojure-emacs/cider/pull/2058): Don't cache ns-forms in buffers with no such forms.
 * [#2057](https://github.com/clojure-emacs/cider/pull/2057): Use `cider--font-lock-ensure` for compatibility with Emacs 24.5.
+* [cider-nrepl#436](https://github.com/clojure-emacs/cider-nrepl/pull/436): Ensure that `*print-right-margin*` is not ignored by cider-nrepl middleware. 
+* [cider-nrepl#435](https://github.com/clojure-emacs/cider-nrepl/pull/435): Allow debugging of forms with `#?(:cljs ... :clj ..)` conditionals.
 * [cider-nrepl#432](https://github.com/clojure-emacs/cider-nrepl/pull/432): Ensure `pprint` is after `load-file`.
 
 ## 0.15.0 (2017-07-20)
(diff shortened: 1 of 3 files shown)


vspinu:
One test failed for unrelated reason.

bbatsov:
Seems to me after this they won't also be killed on `quit-all`. It's also unclear what should we do on restart or with buffers without connections in general.

vspinu:
I think scratch buffer should never be killed. It's up to the user to decide what to do with that buffer. Just like with emacs' `*scratch*`, or any other buffer that contains user's work of some kind.

> or with buffers without connections in general.

Which are those? You mean all those buffers now called "ancillary"?

bbatsov:
> I think scratch buffer should never be killed. It's up to the user to decide what to do with that buffer. Just like with emacs' *scratch*, or any other buffer that contains user's work of some kind.

But in Emacs this buffer will always be working. On complete CIDER quit we should not leave anything around I think. 

> Which are those? You mean all those buffers now called "ancillary"?

I meant scratch buffers - IMO if we don't kill them they should be treated similarly to REPL buffers and get "reconnected" (or something like this). In general I don't like it that we don't explicitly link scratch buffers to some project and REPL type and things with them become quickly confusing for some people.

vspinu:
>  On complete CIDER quit we should not leave anything around I think.

You can quit all connections and restart them later and be willing to re-use the scratch buffer. For me scratch buffers are just like ordinary files with a mode providing some extra evaluation functionality. It's entirely up to the user what to do with it. I don't see it as part of the CIDER connection ecosystem per se, more like an ordinary clojure buffer.

> f we don't kill them they should be treated similarly to REPL buffers and get "reconnected" (or something like this). 

They are loosely connected in the sense that they will send code to the most recent connection.  I think people using scratch buffers expect precisely this semantics.  

> In general I don't like it that we don't explicitly link scratch buffers to some project and REPL type and things with them become quickly confusing for some people.

I think this is a more general issue with cider connection management. There must be a more user friendly way to associated buffers, directories and entire projects to connections. I hope to provide a PR on this in the following weeks, so we can discuss this in more concrete terms.

As far as scratch buffer is concerned, if scratches are allowed to be associated with projects, their implementation would inevitably become more complex. At least, one should be able to create multiple scratch-buffers.

vspinu:
Uff, I accidentally deleted this branch. Restoring.

bbatsov:
> They are loosely connected in the sense that they will send code to the most recent connection. I think people using scratch buffers expect precisely this semantics.

Based on some tickets filed by users I'm not quite certain that's true. 😃 

> I think this is a more general issue with cider connection management. There must be a more user friendly way to associated buffers, directories and entire projects to connections. I hope to provide a PR on this in the following weeks, so we can discuss this in more concrete terms.

Looking forward to it. 

As for this PR - let's just kill the buffer on `cider-quit-all` and I'll accept it. I really don't want to leave CIDER buffers laying around after a complete shutdown.

vspinu:
> As for this PR - let's just kill the buffer on cider-quit-all and I'll accept it.

I am afraid this one has to wait a bit till #2069 is sorted out.

vspinu:
I have reabased this one for 0.15.1. As per your suggestion, killing the scratch on quit-all but with obligatory user's consent.

vspinu:
Also documenting here my recent changes to cider-nrepl to avoid changelog conflicts. I asume you will be pulling clojure-emacs/cider-nrepl#435 for 0.15.1. 

With this, I have completed my part for 0.15.1.

bbatsov:
👍

PR #2061, opened 2017-07-22
[Fix #2058] Don't cache non-existing ns-forms

vspinu:

PR #2068, opened 2017-07-31
Move test utilities in specialized files and move interaction tests out of cider-client-tests.el  

vspinu:
As per [previous discussion](https://github.com/clojure-emacs/cider/pull/2061#issuecomment-317636286) I am placing utility macros into specialized utilis file and moving a couple of interaction tests where they belong. The idiomatic way to add utility files in buttercup is by including them in separate directories with -L flag. See jorgenschaefer/emacs-buttercup#94.


bbatsov:
👍

PR #2070, opened 2017-08-03
Makefile and tests improvement

vspinu:
I was having hard time running tests locally for different versions of emacs.  First it was because I was setting EMACS directly in the Makefile and Cask was no picking it up. So I am exporting EMACS in case someone else does the same.  

Then `export EMACS=emacs25 && make test` would not update elpa for that emacs. Adding a commit to fix that. 

Finally, isolating test scripts to ./scripts/ sub-folder and adding test-all target.


test/scripts/cider-checks.el
```diff
@@ -9,16 +9,18 @@
...
 
 ;; disable some annoying (or non-applicable) checkdoc checks
 (setq checkdoc-package-keywords-flag nil)
 (setq checkdoc-arguments-in-order-flag nil)
 (setq checkdoc-verb-check-experimental-flag nil)
 
-(let ((files (directory-files default-directory t
-                              "\\`[^.].*\\.el\\'" t)))
-
+(let ((files (seq-filter
```
xiongtx: `cl-remove-if` also works.

vspinu: Why is that important? seq-filter is idiomaticaly used throughout cider.

bbatsov:
👍

PR #2078, opened 2017-08-22
Faster REPL - don't load extra middleware on first sync eval request

vspinu:
This saves a couple of extra seconds and brings cider repl startup time to barebones empty-project `lein repl` of 4s for me.

----

  - Send :inhibit-cider-middleware condition on first request. Relies on a feature added in clojure-emacs/cider-nrepl#438.

  - Combine `require-repl-utils` and `set-initial-ns` into one sync request for efficiency. Note that `require-repl-utils` needs to be a sync request. It could happen that the following init-debuger request would occur before it has finished, resulting in "reader/reader not found error".

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -255,26 +255,25 @@ ENDPOINT is a plist as returned by `nrepl-connect'."
       (add-hook 'nrepl-disconnected-hook 'cider--disconnected-handler nil 'local)
       (current-buffer))))
 
-(defun cider-repl-require-repl-utils ()
-  "Require standard REPL util functions into the current REPL."
-  (interactive)
-  (cider-nrepl-request:eval
-   "(when (clojure.core/resolve 'clojure.main/repl-requires)
-      (clojure.core/map clojure.core/require clojure.main/repl-requires))"
-   (lambda (_response) nil)))
-
 (declare-function cider-set-buffer-ns "cider-mode")
-(defun cider-repl-set-initial-ns (buffer)
-  "Set the REPL BUFFER's initial namespace (by altering `cider-buffer-ns').
-This is \"user\" by default but can be overridden in apps like lein (:init-ns)."
+(defun cider-repl-require-repl-utils-and-set-ns (buffer)
+  "Require standard REPL util functions and set the ns of the REPL's BUFFER.
+Namespace is \"user\" by default, but can be overridden in apps like
+lein (:init-ns).  Both of these operations need to be done as a sync
+request at the beginning of the session.  Bundling them together for
+efficiency."
   ;; we don't want to get a timeout during init
   (let ((nrepl-sync-request-timeout nil))
     (with-current-buffer buffer
-      (let ((initial-ns (or (read
-                             (nrepl-dict-get
-                              (cider-nrepl-sync-request:eval "(str *ns*)")
-                              "value"))
-                            "user")))
+      (let* ((command "(do (when (clojure.core/resolve 'clojure.main/repl-requires)
+                            (clojure.core/map clojure.core/require clojure.main/repl-requires))
+                           (str *ns*))")
+             (response (nrepl-send-sync-request
+                        (lax-plist-put (nrepl--eval-request command)
+                                       "inhibit-cider-middleware" "true")
+                        (cider-current-connection)))
+             (initial-ns (or (read (nrepl-dict-get response "value"))
+                             "user")))
         (cider-set-buffer-ns initial-ns)))))
 
 (defvar cider-current-clojure-buffer nil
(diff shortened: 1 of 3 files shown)


vspinu:
From #2081:
> I was just wondering whether to try to include there some of @vspinu's performance improvement patches or push those back to 0.16.

It depends how long would it take to release 0.16.  I would like to see #2069 in 0.16 and that change with associated features/fixes might take quite some months. If we can pull it in within a couple of months then this change can also go in 0.16, if not then it might be better included now.

As noted in 721440b99a23947, there is one side effect of dynamic loading of middlware - it can occur concurrently. But Clojure seems to have trouble loading namespaces in parallel.

I haven't investigated this thoroughly but I run into issues when debug middleware was requested before `out` middleware on emacs side in async requests. As debug is heavy, `out` middleware ends up loaded simultaneously with `debug` and didn't load correctly.   This is why I changed the order of those two requests on emacs side. `out` middleware is very light and doesn't cause issues with the debug. 

Fixing this on cider's side is easy, just implement a kind of deffered chain to load cider's middleware. But then, some other packages (nrepl-refactor?) might load same namespaces in parallel. One partial solution is to run `cider-connected-hook` at the end of the deffered chain and request external packages to rely on `cider-connected-hook` (which they should do anyways).

xiongtx:
> But Clojure seems to have trouble loading namespaces in parallel.

vspinu:
> Is this due to dependencies between namespaces? Needs a topological sort?

No. It's just that if two actions simultaneously try to load same namespace, something goes wrong and one of them is not loaded correctly.

xiongtx:
Can you put up a minimal repo that can reproduce this problem? Seems like a Clojure level issue, not `cider-nrepl`.

vspinu:
Yes, it's a Clojure level issue. If I had a minimal rep. example I would have already reported it on JIRA.

bbatsov:
> As noted in 721440b, there is one side effect of dynamic loading of middlware - it can occur concurrently. But Clojure seems to have trouble loading namespaces in parallel.

Things are these are generally my concern - don't like introducing potentially breaking changes in a bugfix release. 

> It depends how long would it take to release 0.16. I would like to see #2069 in 0.16 and that change with associated features/fixes might take quite some months. If we can pull it in within a couple of months then this change can also go in 0.16, if not then it might be better included now.

There's never a precise roadmap for releases. Generally I cut them when I feel we've added something important enough to warrant a release.

cider-repl.el
```diff
@@ -255,26 +255,24 @@ ENDPOINT is a plist as returned by `nrepl-connect'."
...
-   "(when (clojure.core/resolve 'clojure.main/repl-requires)
-      (clojure.core/map clojure.core/require clojure.main/repl-requires))"
-   (lambda (_response) nil)))
-
 (declare-function cider-set-buffer-ns "cider-mode")
-(defun cider-repl-set-initial-ns (buffer)
-  "Set the REPL BUFFER's initial namespace (by altering `cider-buffer-ns').
-This is \"user\" by default but can be overridden in apps like lein (:init-ns)."
+(defun cider-repl-require-repl-utils-and-set-ns (buffer)
+  "Require standard REPL util functions and set the REPL BUFFER NS.
```
bbatsov: NS -> ns (it's not a param).

cider-repl.el
```diff
@@ -255,26 +255,24 @@ ENDPOINT is a plist as returned by `nrepl-connect'."
...
+Namespace is \"user\" by default but can be overridden in apps like
```
bbatsov: default, but

cider-repl.el
```diff
@@ -255,26 +255,24 @@ ENDPOINT is a plist as returned by `nrepl-connect'."
...
+lein (:init-ns). Booth of these operations need to be done as a sync
```
bbatsov: booth -> both

cider-repl.el
```diff
@@ -255,26 +255,24 @@ ENDPOINT is a plist as returned by `nrepl-connect'."
...
+request at the beginning of the session. Bundling them for efficiency."
```
bbatsov: There's one space missing here before the final sentence. 
them -> them together

cider-repl.el
```diff
@@ -289,18 +287,29 @@ to call `cider-remember-clojure-buffer'.")
...
-  (cider-repl-set-initial-ns buffer)
-  (cider-repl-require-repl-utils)
-  (unless no-banner
-    (cider-repl--insert-banner-and-prompt buffer))
   (when cider-repl-display-in-current-window
     (add-to-list 'same-window-buffer-names (buffer-name buffer)))
   (pcase cider-repl-pop-to-buffer-on-connect
     (`display-only (display-buffer buffer))
     ((pred identity) (pop-to-buffer buffer)))
+  (cider-repl-require-repl-utils-and-set-ns buffer)
```
bbatsov: Why did you move those down?

cider.el
```diff
@@ -892,10 +892,15 @@ buffer."
     (cider--check-required-nrepl-version)
     (cider--check-clojure-version-supported)
     (cider--check-middleware-compatibility)
-    (cider--debug-init-connection)
     (cider--subscribe-repl-to-server-out)
     (when cider-auto-mode
       (cider-enable-on-existing-clojure-buffers))
+    ;; Middleware on cider-nrepl side is loaded dynamically on first
+    ;; usage. Loading middleware concurrently can lead to confusing issues
+    ;; (likely a clojure bug). Debug middleware is heavy, so enable it at the
```
bbatsov: I recall you verified this is a bug, so I think you should mention it here. The next sentence doesn't sound very reassuring. 😃

bbatsov:
I guess you can also add to this PR a changelog entry, as users are going to be quite excited about the speedup.

bbatsov:
@vspinu ping :-)

vspinu:
Sorry for being slow on this. My kid was just born a couple of weeks ago, so my bandwidth is 1/10th of what it was before :) Will try to fix all my PRs by Saturday.

bbatsov:
> Sorry for being slow on this. My kid was just born a couple of weeks ago, so my bandwidth is 1/10th of what it was before :) Will try to fix all my PRs by Saturday.

Thanks and congratulations! 🎉 🎂

cider-repl.el
vspinu: The REPL shows a bit earlier (.5-1 sec for me). Also, the repl is always shown, even when errors occur with those subsequent requests.

vspinu:
@bbatsov, I have addressed your comments in all of my PRs (also the one on cider-nrepl side). Added all change-log entries here to avoid conflicts.

bbatsov:
👍

PR #2079, opened 2017-08-22
cider-doc: make links in "Also see" section to be text buttons

madand:
This minor change is for compatibility with [ace-link](https://github.com/abo-abo/ace-link)
package, which cannot detect links created with `insert-button`.

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -484,9 +484,9 @@ Tables are marked to be ignored by line wrap."
                          ;; if the var belongs to the same namespace,
                          ;; we omit the namespace to save some screen space
                          (symbol (if (equal ns see-also-ns) see-also-sym ns-sym)))
-                    (insert-button symbol
-                                   'type 'help-xref
-                                   'help-function (apply-partially #'cider-doc-lookup symbol)))
+                    (insert-text-button symbol
+                                        'type 'help-xref
+                                        'help-function (apply-partially #'cider-doc-lookup symbol)))
                   (insert " "))
                 see-also))
         (cider--doc-make-xrefs)


vspinu:
Wouldn't it make more sense to improve `ace-link` instead?

madand:
I investigated a bit deeper and can answer, certainly no. In a few words, this change is needed not because `ace-link` is broken, but because modes derived from `help-mode` should use `insert-text-button` for links.

Here is the long story.

A quote from `make-text-button` docs:
```
This function is like `make-button', except that the button is actually
part of the text instead of being a property of the buffer.  That is,
this function uses text properties, the other uses overlays.
Creating large numbers of buttons can also be somewhat faster
using `make-text-button'.
```

In buffers, with `help-mode` major mode, `ace-link` scans the only the **buffer text** for links, thus it cannot see links added as overlays (since overlay is not a part of a buffer text). `cider-docview-mode` is derived from `help-mode`. In file `help-mode.el.gz` (part of Emacs), where `help-mode` is implemented, only `insert-text-button` is used and never `insert-button`, which explains lack of scanning in overlays in `ace-link`.

cider-doc.el
```diff
@@ -484,7 +484,7 @@ Tables are marked to be ignored by line wrap."
                          ;; if the var belongs to the same namespace,
                          ;; we omit the namespace to save some screen space
                          (symbol (if (equal ns see-also-ns) see-also-sym ns-sym)))
-                    (insert-button symbol
+                    (insert-text-button symbol
                                    'type 'help-xref
```
bbatsov: The indentation should be fixed here.

bbatsov:
Two things:

* you should check what's wrong with the CI build
* you should check if this needs to be fixed in other parts of the codebase

madand:
* CI build failure is not related to this PR. The offending change is among the recent commits in `master`: https://travis-ci.org/clojure-emacs/cider/builds/265448569.
* Fixed.

vspinu:
That's a doc commit, my guess is that something broke on buttercup side.

vspinu:
Broken on the [buttercup side](https://github.com/jorgenschaefer/emacs-buttercup/issues/104).

PR #2082, opened 2017-08-26
Add stacktrace navigator for clojure.spec/asserts

vspinu:
`clojure.spec/assert`s are hard to read, especially in cider stacktrace browser. Adding specialized support for spec/asserts with the ability to toggle visibility of multi-line values.

spec-stacktrace
   
The visibility toggling is not fully integrated with the default cause-visibility-cycling in the stacktrace browser, mostly because the non-idiomatic usage of invisibility properties by `cider-stacktrace.el`. Instead of modifying `buffer-invisibility-spec`, as intended by emacs design, it modifies buffer text properties directly. I might rewrite that part some day but it's low priority. For now it's unlikely that anyone will be bothered by this.

Assuming that this will go into 0.16, so no doc entry as yet.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -596,11 +596,22 @@ prompt and whether to use a new window.  Similar to `cider-find-var'."
 
 ;; Rendering
 
-(defun cider-stacktrace-emit-indented (text indent &optional fill)
-  "Insert TEXT, and INDENT and optionally FILL the entire block."
-  (let ((beg (point)))
+(defun cider-stacktrace-emit-indented (text &optional indent fill fontify)
+  "Insert TEXT, and optionally FILL and FONTIFY as clojure the entire block.
+INDENT is a string to insert before each line.  When INDENT is nil, first
+line is not indented and INDENT defaults to a white-spaced string with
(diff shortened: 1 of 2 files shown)


cider-stacktrace.el
```diff
@@ -727,10 +726,53 @@ This associates text properties to enable filtering and source navigation."
         (insert (propertize (format " at (%d:%d)" line column)
                             'font-lock-face message-face))))))
 
+(defun cider-stacktrace--insert-keyed-value (indent key &rest args)
+  "Insert KEY and ARGS prefixed by INDENT."
+     (let ((str (when args
```
bbatsov: The indentation is wrong here. 

What's keyed value? From the usages of this it seems you're just passing some labels as "keys".

vspinu: That's the best name I could think of, but I hardly spend any time thinking about it. The idea is to be able to insert into the buffer a pair of objects - key and value. For example in the screenshot above `in:` is key and `[:cljproj.core/email]` is a value.  If "value" is multi-line then all of those line should be indented. 

In any case, will think of something better once I find time refresh on this.

PR #2083, opened 2017-08-28
Improvements for REPL output pre-processing and locref navigation 

vspinu:
Improving on #2043 a bit for navigation. Pass through file metatdata when available from location references and no info from var names could be retrieved. `cider-find-file`  does a good job  in those cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,17 @@
 
 ### New Features
 
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): New utility function `cider-add-face`.
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): New utility function `cider-run-chained-hook`.
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): New `cider-repl-preoutput-hook` that allows custom output processing. 
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): Highlight clojure.spec keywords in REPL (`cider-repl-highlight-spec-keywords` pre-output processor).
+
 ### Changes
 
 * [#2045](https://github.com/clojure-emacs/cider/issues/2045) `*cider-scratch*` buffers are no longer automatically killed on connection quit.
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): Jump to other window when clicking on location references in REPL.
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): Improve project namespace highlighting in REPLs.
+* [#2083](https://github.com/clojure-emacs/cider/pull/2083): Find locations in more cases when clicking on references in REPL.
 
 ### Bugs Fixed
 
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -589,24 +589,30 @@ When there is a possible unfinished ansi control sequence,
               (setq cider-repl--root-ns-regexp
                     (format cider-repl--root-ns-highlight-template roots)))))))))
 
-(defun cider-repl--apply-current-project-color (string)
+(defun cider-repl-highlight-current-project (string)
   "Fontify project's root namespace to make stacktraces more readable.
 Foreground of `cider-stacktrace-ns-face' is used to propertize matched
 namespaces.  STRING is REPL's output."
-  (if cider-repl--root-ns-regexp
-      (let ((start 0)
-            (end 0))
-        (while (setq start (string-match cider-repl--root-ns-regexp string end))
-          (setq end (match-end 0))
-          (let ((face-spec (list (cons 'foreground-color
-                                       (face-attribute 'cider-stacktrace-ns-face :foreground nil t)))))
-            (font-lock-prepend-text-property start end 'face face-spec string)))
-        string)
-    string))
+  (cider-add-face cider-repl--root-ns-regexp 'cider-stacktrace-ns-face
+                  t nil string)
+  string)
+
+(defun cider-repl-add-locref-help-echo (string)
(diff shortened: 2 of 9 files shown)


cider-repl.el
```diff
@@ -1046,10 +1046,10 @@ regexes from `cider-locref-regexp-alist' to infer locations at point."
   (if-let ((loc (cider-locref-at-point pos)))
       (let* ((var (plist-get loc :var))
              (line (plist-get loc :line))
-             (file (if var
-                       (or (cider-sync-request:ns-path var)
-                           (nrepl-dict-get (cider-sync-request:info var) "file"))
-                     (plist-get loc :file))))
+             (file (or (when var
```
bbatsov: Who does this improve the code? Seems exactly the same to me.

cider-repl.el
```diff
@@ -561,7 +561,7 @@ When there is a possible unfinished ansi control sequence,
                    (buffer-local-value 'cider-repl--ns-forms-plist connection)
                    ns)))))
 
-(defvar cider-repl--root-ns-highlight-template "\\<\\(%s\\)[^$/: \t\n]+"
+(defvar cider-repl--root-ns-highlight-template "\\<\\(%s\\)[^$/: \t\n()]+"
```
bbatsov: What's the purpose of this change?

cider-repl.el
```diff
@@ -1051,7 +1051,7 @@ regexes from `cider-locref-regexp-alist' to infer locations at point."
                              (nrepl-dict-get (cider-sync-request:info var) "file")))
                        (plist-get loc :file))))
         (if file
-            (cider--jump-to-loc-from-info (nrepl-dict "file" file "line" line))
+            (cider--jump-to-loc-from-info (nrepl-dict "file" file "line" line) t)
```
bbatsov: I think it's subjective whether this is better or worse - in such cases I always add a defcustom. 

P.S. Mind the commit message style. 😃

vspinu: It's not subjective. These pseudo links now behave like links in cider stacktrace buffer. Also like links in every other buffer which has a bunch of links in it (occur, grep, compilation etc) behave like that. It's common to click on multiple error references till you get what you want. 

> Mind the commit message style. :smiley:

What do you mean more concretely?

cider-repl.el
vspinu: Improve original regexp. Currently things like ns1.foo.bar(file.clj:123) are highlighted all the way to 123. This change stops them at (. 

Originally I thought all those reference have $ in their name, but that's not always the case.

cider-repl.el
xiongtx: > What do you mean more concretely?

Follow [this guide](https://chris.beams.io/posts/git-commit/). 

> When clicked on locref in REPL, jump to other window.

doesn't adhere to #5.

cider-repl.el
vspinu: The difference is that now `:file` metadata is picked up when first two resolutions fail (those within or). This happens when a var is a protocol method. This was also the original intention btw.

cider-repl.el
vspinu: Ohh, dot at the end of the message. Fixing that :) The subject line is already imperative mood.

xiongtx: > The subject line is already in imperative mood.

I'm not a stickler for this stuff, but no, it's not. 

> Jump to other window when clicking on locref in REPL

would be.

cider-repl.el
bbatsov: Ah, I see your point now - `var` might be present, but its branch might return `nil`.  Add some comments as it's somewhat hard to follow the what the assumptions here otherwise.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Bugs Fixed
 
+* Improve file retrieval when jumping to REPLs location references and Clojure vars cannot be resolved.
```
bbatsov: How come you can just to something that's not resolved? I guess you made some mistake in the description here.

cider-repl.el
vspinu: Yes, I used "var" loosely to refer to the variable at point, which need not correspond to clojure's var. Will add a couple of notes.

CHANGELOG.md
vspinu: That code is recent. I added it in  #2043 and now I am just improving on it. 

I just said what I did there - when locrefs cannot be referenced from variables at point, use file metadata from the locrefs themselves. I can state it more generically "Improve locations lookup by using file metadata when available". Not accurate but conveys the improvement. 

AFAIC, this need not be documented at all as being a small and technical fix on a recent change, but knowing how zealous you are on change-logs I added it.

vspinu:
I am piling a bunch of related changes and improvements here. With a bit more doc this time, so should be self explanatory. 

Also fixing failing tests to account for recent buttercup changes.

Two notable new features:

 - `cider-preoutput-hook` which implements a customizable pipeline of output pre-processing and includes most of our output pre-processing as "middleware". 
 - `cider-repl-highlight-spec-keywords` pre-output processor which highlights clojure.spec "explain" keywords. Clojure spec explain output is extremely difficult to read.. Make it a bit easy by showing this:

spec-asert2
  
instead of

spec-asert1

xiongtx:
> I am piling a bunch of related changes and improvements here. 

It'd be better to open another PR. A bunch of unrelated changes in a single PR is hard to review.

cider-repl.el
bbatsov: Indeed. The commit messages should ideally always start with a verb.

PR #2086, opened 2017-09-18
Fix fns defined with def not font locked as fns

oskarkv:
See https://github.com/clojure-emacs/cider-nrepl/pull/442 which is needed too.


cider-mode.el
```diff
@@ -580,7 +580,7 @@ with the given LIMIT."
                          (push sym deprecated))
                        (cond ((and do-macro (nrepl-dict-get meta "macro"))
                               (push sym macros))
-                             ((and do-function (nrepl-dict-get meta "arglists"))
+                             ((and do-function (nrepl-dict-get meta "fn"))
```
Malabarba: Might be worth keeping the old behaviour too (just in case). That would be `(or (nrepl-dict-get meta "fn") (nrepl-dict-get meta "arglists"))`.

bbatsov:
Please, mention the CIDER issue this fixes in the commit message and also add a changelog entry.

oskarkv:
I don't think there was an issue for it. Should I create one?

oskarkv:
Made the changes. Fixes #2088.

PR #2089, opened 2017-09-22
Allow loading buffers without associated files

AdamNiederer:
This lets us load `*scratch*` et al into a cider repl by saving the file to the user's temp directory. Before, it would throw an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#2083](https://github.com/clojure-emacs/cider/pull/2083): New utility function `cider-run-chained-hook`.
 * [#2083](https://github.com/clojure-emacs/cider/pull/2083): New `cider-repl-preoutput-hook` that allows custom output processing.
 * [#2083](https://github.com/clojure-emacs/cider/pull/2083): Highlight clojure.spec keywords in REPL (`cider-repl-highlight-spec-keywords` pre-output processor).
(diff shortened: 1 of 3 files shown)


xiongtx:
What is the purpose of this? Why not just create a temp file in your project dir? The location of Emacs-generated temp file is not transparent to the user.

AdamNiederer:
Using buffers instead of files allows one to experiment without having to delete a temporary file later. It's also useful for quick experimentation when cider-scratch isn't enough.

The emacs-generated temp file is deleted almost immediately after it's created; the user won't know it exists unless they're watching /tmp.

xiongtx:
> The emacs-generated temp file is deleted almost immediately after it's created;

That seems like it'd break `cider-find-var` and other commands that jump to source, as well as commands that can reload files, e.g. `cider-refresh`.

AdamNiederer:
Looks like `cider-find-var` doesn't work on code loaded from a buffer without an associated file. Does nREPL require the filename passed via a `load-file` command to map directly to a file on disk? It looks like the contents of the file are also sent with the request.

xiongtx:
`cider-find-var` and friends need to be able to jump to a position in a file. If the file is deleted, then it can no longer jump to that location.

bbatsov:
I don't think that's a good idea - if you want to work without files you can simply evaluate an entire buffer.

AdamNiederer:
That's fair; it looks like the benefits of this change are outweighed by the cost of changing a lot of code to be buffer-oriented rather than file-oriented.

PR #2101, opened 2017-10-24
Created cider-eval-last-sexp-to-comment

marcelotoledo:
Created cider-eval-last-sexp-to-comment which is more flexible and useful than cider-eval-defun-to-comment. You're able to eval to comment any sexp and not only functions. Also rebinded the shortcut C-c M-; to this new func.

PR #2111, opened 2017-11-10
Implement cider-pprint-eval-last-sexp-to-comment

gonewest818:
This function has similar functionality to the other pprint interactions;
however this variant inserts the result of the evaluation directly in the
current buffer.

The comment prefix used in the output is user-configurable, and for consistency
we also use same prefix in cider-eval-defun-to-comment.

By default the output will be commented like so
```
(for [i (range 1 10)]
  (range i))
;; => ((0)
;;     (0 1)
;;     (0 1 2)
;;     (0 1 2 3)
;;     (0 1 2 3 4)
;;     (0 1 2 3 4 5)
;;     (0 1 2 3 4 5 6)
;;     (0 1 2 3 4 5 6 7)
;;     (0 1 2 3 4 5 6 7 8))
```
but by customizing the `cider-pprint-comment-*` variables you can also get output like this

```
#_((0)
   (0 1)
   (0 1 2)
   (0 1 2 3)
   (0 1 2 3 4)
   (0 1 2 3 4 5)
   (0 1 2 3 4 5 6)
   (0 1 2 3 4 5 6 7)
   (0 1 2 3 4 5 6 7 8))
```

or even

```
(comment
  ((0)
   (0 1)
   (0 1 2)
   (0 1 2 3)
   (0 1 2 3 4)
   (0 1 2 3 4 5)
   (0 1 2 3 4 5 6)
   (0 1 2 3 4 5 6 7)
   (0 1 2 3 4 5 6 7 8))
)
```
-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New Features
 
 * [#2082](https://github.com/clojure-emacs/cider/pull/2082), [cider-nrepl#440](https://github.com/clojure-emacs/cider-nrepl/pull/440): Add specialized stacktraces for clojure.spec assertions.
+* [#2111](https://github.com/clojure-emacs/cider/pull/2111): Add `cider-pprint-eval-last-sexp-to-comment` and `cider-pprint-eval-defun-to-comment`.
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -244,6 +244,24 @@ namespace-qualified function of zero arity."
   "Face used to highlight compilation warnings in Clojure buffers."
   :group 'cider)
 
+(defcustom cider-comment-prefix ";; => "
+  "The prefix to insert before the first line of commented output."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.16.0"))
+
+(defcustom cider-comment-continued-prefix ";;    "
+  "The prefix to use on the second and subsequent lines of commented output."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.16.0"))
+
+(defcustom cider-comment-postfix ""
+  "The postfix to be appended after the final line of commented output."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.16.0"))
+
 (defconst cider-clojure-artifact-id "org.clojure/clojure"
   "Artifact identifier for Clojure.")
 
@@ -795,6 +813,30 @@ COMMENT-PREFIX is the comment prefix to use."
                                  (cider-emit-interactive-eval-err-output err))
                                '()))
 
+(defun cider-eval-pprint-with-multiline-comment-handler (buffer location comment-prefix continued-prefix comment-postfix)
+  "Make a handler for evaluating and inserting results in BUFFER.
+The inserted text is pretty-printed and region will be commented.
+LOCATION is the location at which to insert.
+COMMENT-PREFIX is the comment prefix for the first line of output.
+CONTINUED-PREFIX is the comment prefix to use for the remaining lines.
+COMMENT-POSTFIX is the text to output after the last line."
+  (cl-flet ((multiline-comment-handler (buffer value)
+             (with-current-buffer buffer
+               (save-excursion
+                 (goto-char location)
+                 (let ((lines (split-string value "[\n]+" t)))
+                   ;; only the first line gets the normal comment-prefix
+                   (insert (concat comment-prefix (pop lines)))
+                   (dolist (elem lines)
+                     (insert (concat "\n" continued-prefix elem)))
+                   (unless (string= comment-postfix "")
+                     (insert comment-postfix)))))))
+    (nrepl-make-response-handler buffer
+                                 '()
+                                 #'multiline-comment-handler
+                                 #'multiline-comment-handler
+                                 '())))
+
 (defun cider-popup-eval-out-handler (&optional buffer)
   "Make a handler for evaluating and printing stdout/stderr in popup BUFFER.
 
@@ -1132,12 +1174,12 @@ arguments and only proceed with evaluation if it returns nil."
   (interactive "r")
   (cider-interactive-eval nil nil (list start end)))
 
-(defun cider-eval-last-sexp (&optional prefix)
+(defun cider-eval-last-sexp (&optional output-to-current-buffer)
   "Evaluate the expression preceding point.
-If invoked with a PREFIX argument, print the result in the current buffer."
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current buffer."
   (interactive "P")
   (cider-interactive-eval nil
-                          (when prefix (cider-eval-print-handler))
+                          (when output-to-current-buffer (cider-eval-print-handler))
                           (cider-last-sexp 'bounds)))
 
 (defun cider-eval-last-sexp-and-replace ()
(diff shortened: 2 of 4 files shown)


gonewest818:
AFAICT, some of those travis errors look like transient network issues on their end (e.g. "Opening TLS connection to ‘melpa.org’...failed"), while the other errors refer to a file I didn't change (cider-stacktrace.el).

bbatsov:
> AFAICT, some of those travis errors look like transient network issues on their end (e.g. "Opening TLS connection to ‘melpa.org’...failed"),

Or something to do with changes in Emacs 26. We have to investigate this.

cider-interaction.el
```diff
@@ -244,6 +244,24 @@ namespace-qualified function of zero arity."
   "Face used to highlight compilation warnings in Clojure buffers."
   :group 'cider)
 
+(defcustom cider-pprint-comment-prefix ";; => "
+  "The prefix to insert before the first line of pprint output."
+  :type 'string
+  :group 'cider
+  :package-version '(cider . "0.15.2"))
```
bbatsov: The version should 0.16 everywhere. There won't be 0.15.2.

gonewest818: Ok, I’ll fix that.

gonewest818:
With respect to the TLS error in Travis...  I think the difference is emacs versions <=25.2 fall back to `s_client` when `gnutls-cli` fails, but emacs 26 does not (see discussion in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=766397 and commit
https://github.com/emacs-mirror/emacs/commit/6e45de6bacc508db11b15b2c8ba86aad8c0570df)

Even in the failed Travis jobs, there's a point where evm successfully bootstraps itself using emacs-24.3-travis: https://travis-ci.org/clojure-emacs/cider/jobs/307715504#L485-L492 
where you see it's the `s_client` attempt that actually succeeds.

The build failure occurs later, as cask is installing dependencies for emacs-git-snapshot-travis:
https://travis-ci.org/clojure-emacs/cider/jobs/307715504#L879-L885
and you can see there is no `s_client` fallback attempt.

So I'm not sure why `gnutls-cli` is failing in Travis, perhaps the cert isn't valid (or isn't where it's supposed to be)?  Or perhaps something else.

bbatsov:
Thanks for looking into this. I guess the easiest think we can do is to switch to just `http` to avoid those errors, although I don't like this that much.

Please mention your changes in the changelog and in the relevant parts of the manual.

bbatsov:
You should probably add the new command to some of the mode menus as well, otherwise no one will find it.

gonewest818:
Ok, will do.  Want me to squash commits when all is said and done?

bbatsov:
@gonewest818 Yep, that'd be best.

gonewest818:
I made this feature invokable with a prefix argument on `cider-pprint-eval-last-sexp`. In other words, `C-u C-c C-p` pretty-prints the last sexp and inserts the resulting text as as comment in the current buffer.

I updated the changelog and the docs to indicate this is the case.  I didn't change README.md, as it seems more like a landing page with links off to those other docs I already changed.

And I haven't forgottem, can still do the squash commit if/when you approve.

cider-interaction.el
```diff
@@ -1167,9 +1209,28 @@ With a prefix arg, LOC, insert before the form, otherwise afterwards."
          (insertion-point (nth (if loc 0 1) bounds)))
     (cider-interactive-eval nil
                             (cider-eval-print-with-comment-handler
-                             (current-buffer) insertion-point ";; => ")
+                             (current-buffer)
+                             insertion-point
+                             cider-pprint-comment-prefix)
                             bounds)))
 
+(defun cider-pprint-eval-last-sexp-to-comment (loc)
```
bbatsov: Guess you should make `loc` optional. And maybe name it `insert-result-before` or something more meaningful.

cider-interaction.el
```diff
@@ -1167,9 +1209,28 @@ With a prefix arg, LOC, insert before the form, otherwise afterwards."
...
+  "Evaluate the last sexp and insert result as comment at LOC.
+
+With a prefix arg, LOC, insert before the form, otherwise afterwards."
```
bbatsov: Maybe you should mention here the configuration options related to this command.

cider-interaction.el
```diff
@@ -1211,10 +1271,13 @@ Print its value into the current buffer."
                             (when (consp form) form)
                             (cider--nrepl-pprint-request-plist (cider--pretty-print-width)))))
 
-(defun cider-pprint-eval-last-sexp ()
-  "Evaluate the sexp preceding point and pprint its value in a popup buffer."
-  (interactive)
-  (cider--pprint-eval-form (cider-last-sexp 'bounds)))
+(defun cider-pprint-eval-last-sexp (&optional prefix)
```
bbatsov: Normally I avoid the name `prefix` and go for something like `print-in-current-buffer` (or something).

cider-interaction.el
gonewest818: That matches the implementation of `cider-eval-defun-to-comment` so I'l fix both.

gonewest818:
ok @bbatsov, code review comments are addressed and I squashed the commits again.

xiongtx:
This introduces a customizable variable called `cider-pprint-comment-prefix`, but it's also [used in `cider-eval-defun-to-comment`](https://github.com/clojure-emacs/cider/pull/2111/files#diff-e39c33031dc8d3e78b32ebee83293cd6R1218).

Seems that the variables can be used in comment functions in general, not just the `pprint` variants, so we can drop the `pprint` from the names:

- `cider-comment-prefix`
- `cider-comment-continued-prefix` 
- `cider-comment-postfix`

cider-interaction.el
```diff
@@ -1150,26 +1192,62 @@ If invoked with a PREFIX argument, print the result in the current buffer."
...
   (interactive "P")
   (let* ((bounds (cider-defun-at-point 'bounds))
-         (insertion-point (nth (if loc 0 1) bounds)))
+         (insertion-point (nth (if insert-before 0 1) bounds)))
     (cider-interactive-eval nil
                             (cider-eval-print-with-comment-handler
-                             (current-buffer) insertion-point ";; => ")
+                             (current-buffer)
+                             insertion-point
+                             cider-pprint-comment-prefix)
```
xiongtx: This introduces a customizable variable called `cider-pprint-comment-prefix`, but it's also [used in `cider-eval-defun-to-comment`](https://github.com/clojure-emacs/cider/pull/2111/files#diff-e39c33031dc8d3e78b32ebee83293cd6R1218).

Seems that the variables can be used in comment functions in general, not just the `pprint` variants, so we can drop the `pprint` from the names:

- `cider-comment-prefix`
- `cider-comment-continued-prefix` 
- `cider-comment-postfix`

cider-interaction.el
```diff
@@ -1150,26 +1192,62 @@ If invoked with a PREFIX argument, print the result in the current buffer."
...
                             bounds)))
 
+(defun cider-pprint-eval-last-sexp-to-comment (&optional insert-before)
```
xiongtx: Currently, CIDER has only has `cider-eval-defun-to-comment`. Seems asymmetric to add `cider-pprint-eval-last-sexp-to-comment` instead of `cider-pprint-eval-defun-to-comment`. Could we implement the latter instead?

cider-interaction.el
gonewest818: yes, that's a fair point.  I'll do that.

cider-interaction.el
gonewest818: I can, but symmetry is arguably a larger issue than this, considering:

- cider-eval-sexp-at-point
- cider-eval-last-sexp
- cider-eval-print-last-sexp
- cider-eval-last-sexp-to-repl
- cider-eval-last-sexp-to-comment (#2101)
- cider-eval-defun-at-point
- cider-eval-defun-to-comment
- cider-pprint-eval-last-sexp
- cider-pprint-eval-last-sexp-to-repl
- cider-pprint-eval-last-sexp-to-comment (#2111)
- cider-pprint-eval-defun-at-point

xiongtx: OK, I didn't see #2101. But we'd still have no `cider-pprint-eval-defun-to-comment` :man_shrugging:. Maybe add it in the next PR?

gonewest818: Let me take a look. I doubt the defun case is really all that different from what I have.

But, for the record there's a lot of overlapping functionality in this set. Everything amounts to "eval <some selected form> and send result to <some destination> <with/without pprint>". It seems clear to me this could be more aggressively refactored such that you don't actually have to write distinct functions for each of the combinatorial possibilities.  I'm not proposing to do that now, mind you, that's too much scope for what was supposed to be an incremental feature.

gonewest818: done. see the latest commit.

cider-interaction.el
gonewest818: done. see commit.

gonewest818:
I found a bug in this implementation but I can't figure out how to resolve it.

If I invoke these new functions by name, e.g. `M-x cider-pprint-eval-last-sexp-to-comment` the function executes and silently exits without seeming to modify the buffer. It's the *next* character I type that causes the result to appear in the buffer.

Whereas, if I invoke the same function through its keybinding `C-u C-c C-p` the function executes and the buffer updates immediately.  Any hints why?

cider-interaction.el
```diff
@@ -1211,10 +1320,13 @@ Print its value into the current buffer."
...
-(defun cider-pprint-eval-last-sexp ()
-  "Evaluate the sexp preceding point and pprint its value in a popup buffer."
-  (interactive)
-  (cider--pprint-eval-form (cider-last-sexp 'bounds)))
+(defun cider-pprint-eval-last-sexp (&optional output-to-current-buffer)
+  "Evaluate the sexp preceding point and pprint its value in a popup buffer.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, insert as comment in the current buffer."
+  (interactive "P")
+  (if output-to-current-buffer
+      (cider-pprint-eval-last-sexp-to-comment nil)
```
bbatsov: You don't have to pass `nil` here - now it's optional, which means the param is going to be nil by default.

gonewest818: right.  thanks!

bbatsov:
> Whereas, if I invoke the same function through its keybinding C-u C-c C-p the function executes and the buffer updates immediately. Any hints why?

See the debugging section of the manual. Using the Elisp debugger is the best way to see what's going wrong.

gonewest818:
I couldn't reproduce the problem in the debugger, but I managed to find it anyway. In emacs 25.1 and newer (simple.el, in the function `execute-extended-command`) there's a new UX feature that suggests faster ways to invoke a command:

    You can run the function 'cider-pprint-eval-defun-to-comment' with M-x -p-d-t RET

https://github.com/emacs-mirror/emacs/blob/emacs-25/lisp/simple.el#L1722-L1730

A comment in the code mentions a caveat that the search for a shorter-but-unique string "can be slow".  And so it seems, at least in this particular case. If the function has a keybinding, emacs suggests the available keybinding rather than trying to find a shorter string. And that is fast.

bbatsov:
Ah, yeah. Now I remembered about that feature. However, I don't recall ever using it in practice. I guess I don't use `M-x` very often. :-)

bbatsov:
Seems we're finally ready to merge then. Address my tiny remark and update your commit message in light of the final state of the PR.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New Features
 
 * [#2082](https://github.com/clojure-emacs/cider/pull/2082), [cider-nrepl#440](https://github.com/clojure-emacs/cider-nrepl/pull/440): Add specialized stacktraces for clojure.spec assertions.
+* [#2111](https://github.com/clojure-emacs/cider/pull/2111): Add `cider-pprint-eval-last-sexp-to-comment` and `cider-pprint-eval-defun-to-comment`
```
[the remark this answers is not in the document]
gonewest818: done

gonewest818:
Done. Thanks for your helpful (and patient) feedback.

bbatsov:
🎉 

You're welcome! Looking forward to more contributions from you down the road! 😉

PR #2119, opened 2017-12-03
add reverse destructuring form function

edvorg:
This change adds a function that reads clojure destructuring form and gives user a clojure structure that matches that form.

Use case:
Let's say user has a function:
```clojure
(defn f [{:keys [a] {[c] :c} :b}]
  (timbre/info a c))
```
He would like to test that function using repl. He can simply select a region `{:keys [a] {[c] :c} :b}` and run command `cider-reverse-destructuring-form-region`. It will give him a buffer with resulting `{:a a, :b {:c [c]}}`. Which he can then use to write a function call like this:
```clojure
(f {:a 1, :b {:c [2]}}
```
-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New Features
 
 * [#2082](https://github.com/clojure-emacs/cider/pull/2082), [cider-nrepl#440](https://github.com/clojure-emacs/cider-nrepl/pull/440): Add specialized stacktraces for clojure.spec assertions.
+* [#2119](https://github.com/clojure-emacs/cider/pull/2119): Add reverse destructuring form function
 
 ### Changes
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1871,6 +1871,33 @@ With a prefix argument, prompt for function to run instead of -main."
                        name))))
         (user-error "No %s var defined in any namespace" (cider-propertize name 'fn))))))
 
+(defun cider-reverse-destructuring-form-region ()
+  "Reverse the destructuring map in current region.
+Display the result in a separate buffer.
+Example: {:keys [a] {[c] :c} :b} will result in {:a a, :b {:c [c]}}."
+  (interactive)
+  (let* ((s (thread-last (buffer-substring (mark) (point))
+                         (format "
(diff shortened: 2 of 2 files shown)


bbatsov:
> He would like to test that function using repl. He can simply select a region {:keys [a] {[c] :c} :b} and run command cider-reverse-destructuring-form-region. It will give him a buffer with resulting {:a a, :b {:c [c]}}. Which he can then use to write a function call like this:

Is this something that would work for any possible destructuring?

Also - you've implemented this with some inline Clojure code, but we don't really do thing like this in CIDER. If something requires complex Clojure code we move it to the middleware layer. 

Right now it's really hard for me to gauge how generic your approach is, as only something that covers lots of use-cases is likely going to be useful to the CIDER users in general.

edvorg:
> Is this something that would work for any possible destructuring?
Should work any possible destructuring. Works for sequences and maps recursively.

> we don't really do thing like this in CIDER
Got it. Should I commit the functionality to https://github.com/clojure-emacs/cider-nrepl and call it from cider?

> Right now it's really hard for me to gauge how generic your approach is, as only something that covers lots of use-cases is likely going to be useful to the CIDER users in general.
I currently have only one use case for this -> when you need to get data structure that is conforming destructuring form. I need something like this quite often - when creating a test case or when calling function with complex arguments destructuring. If it's something that's going to be welcome in cider. I will proceed and update the pr.

bbatsov:
> Got it. Should I commit the functionality to https://github.com/clojure-emacs/cider-nrepl and call it from cider?

Yep.

> I currently have only one use case for this -> when you need to get data structure that is conforming destructuring the form. I need something like this quite often - when creating a test case or when calling function with complex arguments destructuring. If it's something that's welcome in cider. I will proceed and update the pr.

I'd probably change the interface of this a bit, though. Call it possible to call it from the body of the function in question and would just insert a function invocation stub in the REPL. I guess ideally this should work for functions without destructing as well. For them it can just insert the names of the positional params or something like this.

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

Issue #2084, opened 2017-09-08
cider-switch-to-repl-buffer should switch to relevant (not most recent


mwfogleman:
A few people (@magnars @dpsutton @jumar) were talking on Clojurians yesterday about an issue that I've noticed. I don't have all the details, but I borrowed from their descriptions, but will try to help report.

The documentation says "To switch to the relevant REPL buffer based on the Clojure namespace in the current Clojure buffer, use: `C-c C-z`." From cljs-mode, it should switch to the project's cljs-repl; from clojure-mode it should switch to the project's clj-repl. 

Instead, it switches to the most recent buffer.

@dpsutton says it is related to: To switch to the relevant REPL buffer based on the Clojure namespace in the current Clojure buffer, use: `C-c C-z`.

To reproduce the problem:

- Create a project with CLJ and CLJS.
- Navigate to a file in that project, and run cider-jack-in-clojurescript.
- Switch to the CLJ REPL buffer.
- Navigate to a CLJS file buffer.
- Hit C-c C-z (cider-switch-to-repl-buffer). It will switch to the CLJ REPL buffer, not the CLJS REPL buffer. 

Leiningen 2.7.1 

Emacs: 25.2.1

Arch/Antergos Linux

dpsutton:
My thinking is that it is the repl resolution that is the problem. In `cider-current-connection`, when there are more than one connection it just banks on the most recent connection, which seems wrong to me. This sounds like a resolution strategy that should be further down. I've had instances of cljs buffers loading in the clj repl, and clj files loading in other project repls.

      (cond
       ((not connections) nil)
       ;; if you're in a REPL buffer, it's the connection buffer
       ((and (derived-mode-p 'cider-repl-mode) (right-type-p (current-buffer) type)))
       ((eq cider-request-dispatch 'static) (car connections))
       ((= 1 (length connections)) (right-type-p (car connections) type))
       (t (let ((project-connections (cider-find-connection-buffer-for-project-directory
                                      nil :all-connections))
                (guessed-type (or type (cider-connection-type-for-buffer))))
            (or
             ;; cljc
             (and (equal guessed-type "multi")
                  (most-recent-buf project-connections nil))
             ;; clj or cljs
             (and guessed-type
                  (or (most-recent-buf project-connections guessed-type)
                      (most-recent-buf connections guessed-type)))
             ;; when type was not specified or guessed
             (most-recent-buf project-connections type)
             (most-recent-buf connections type)))))

[...]

PR #2127, opened 2017-12-10
Switch to REPL of correct dialect  

vspinu:
Fixing #2084 and fixing `cider-switch-to-last-clojure-buffer` on this occasion. 

The problem with `cider-switch-to-last-clojure-buffer` is that if the buffer which was remembered with`cider-remember-clojure-buffer` (on `C-c C-z` from the clojure buffer) no longer exists, it would throw an error.  With this fix, most appropriate recent clj, cljs or cljc buffer is now always found.

Frankly I quite dislike this `cider-remember-clojure-buffer`/`cider-last-clojure-buffer`. It's frustrating when after manually switching between clojure buffers, `C-c C-z` from the repl brings you back to some unrelated, deeply buried `cider-last-clojure-buffer`.  

I think the current implementation is the way it is because it was "easy" to do so, not because it's useful. I would rather remove it in favor of the more predictable "bring me the most recent buffer of the appropriate dialect". What do you think about this?


bbatsov:
> I think the current implementation is the way it is because it was "easy" to do so, not because it's useful. I would rather remove it in favor of the more predictable "bring me the most recent buffer of the appropriate dialect". What do you think about this?

I agree. I think this was just some legacy from SLIME.

vspinu:
> I agree.

Ok, added a commit for this.

manuel-uberti:
Thanks everyone for this. Going to test it as soon as it is available on MELPA.

Issue #2120, opened 2017-12-04
Travis CI errors with EMACS_BINARY=emacs-git-snapshot-travis


gonewest818:
## Expected behavior

Travis tests should succeed

## Actual behavior

Travis tests are failing in the "before_script".

## Steps to reproduce the problem

See a recent build, for example
https://travis-ci.org/clojure-emacs/cider/builds/307426738

I noticed this issue when submitting #2111.  Copying some of that discussion into a new issue to keep the other ticket clean.

## Analysis

With respect to the TLS error in Travis... I think the difference is emacs versions <=25.2 fall back to s_client when gnutls-cli fails, but emacs 26 does not (see discussion in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=766397 and commit
emacs-mirror/emacs@6e45de6)

Even in the failed Travis jobs, there's a point where evm successfully bootstraps itself using emacs-24.3-travis: https://travis-ci.org/clojure-emacs/cider/jobs/307715504#L485-L492
where you see it's the s_client attempt that actually succeeds.

The build failure occurs later, as cask is installing dependencies for emacs-git-snapshot-travis:
https://travis-ci.org/clojure-emacs/cider/jobs/307715504#L879-L885
and you can see there is no s_client fallback attempt.

I poked around a little bit and discovered elpa.gnu.org and melpa.org are using certs from Let's Encrypt. However the Travis CI workers happen to be running Ubuntu 14.04, and I haven't checked, but it seems likely they do not have the CA certs required to complete the chain of trust for those Lets Encrypt certs. If so, then the solution would be to do whatever incantations are necessary in the .travis.yml to make that happen.

## Links

For posterity here are some references.

https://bugs.launchpad.net/ubuntu/+source/gnutls26/+bug/1373422
https://tools.ietf.org/html/rfc5246
https://www.ssllabs.com/ssltest/analyze.html?d=elpa.gnu.org
https://cryptoreport.websecurity.symantec.com/checker/

bbatsov:
Thanks for looking into this and opening the ticket. Working against the broken CI is not fun.

gonewest818:
I'm taking some time to look at this (if that's okay?)

I'm starting by pulling the docker image they use for travis builds (they've published the images) with the intent to reproduce the issue locally. If so then the debugging will go much faster, compared to branching pushing .travis.yml over and over.

bbatsov:
It's more than OK - it's much appreciated!

[...]

PR #2128, opened 2017-12-11
[#2120] fix TLS related errors in Travis CI

gonewest818:
Fixes the Travis CI build errors reported in #2120.

Diffs to .travis.yml are as follows:

(1) I've added some explicit package dependencies into `addons.apt.packages`. This is basically the compiler toolchain needed to build gnutls from source

(2) Configured $HOME/local to be cached by Travis between subsequent builds. This is done with the setting `cache.directories`.

See Travis documentation for further information about the caching of data.  Basically, the contents of that folder are rolled up into a tar ball and saved to an S3 bucket whenever the contents change. The tarball is retrieved and installed prior to the start of your CI job.

(3) Added a utility script named `travis-ci/travis-gnutls.sh` which handles the downloading and compiling of the sources.

The script is organized so that the version of gnutls can be easily updated whenever needed. Just edit the version numbers in that script and push the change. The script is able to detect when it doesn't have the specific version requested, and when that happens it deletes the cache and rebuilds gnutls.

(4) While I was at it, I added emacs 25.3-travis and 26-pretest-travis to the build matrix. Note that git-snapshot-travis now reports itself as being version 27. (Also note, emacs 26 and 27 are still broken builds, but at least now it's not because of the build script itself.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,54 @@
 sudo: false
 language: emacs-lisp
+
+addons:
+  apt:
+    packages:
+    - autogen
+    - ca-certificates
+    - curl
+    - gcc
+    - libgmp-dev
+    - m4
+    - make
+    - pkg-config
+    - xz-utils
+
+cache:
+  directories:
+  - $HOME/local
+
 env:
-  - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-24.5-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-24.5-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-25.1-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test-checks
-  - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,28 +1,54 @@
...
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-25.2-travis MAKE_TEST=test-checks
-  - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test
-  - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test-bytecomp
-  - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test-checks
+  global:
+    - PATH=$HOME/local/bin:$PATH
+  matrix:
+    - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test
```
xiongtx: @bbatsov Magit recently announced that it will [stop supporting Emacs 24.4](https://twitter.com/magit_emacs/status/936628292261687296). Perhaps we can do that as well after the next release; the text matrix is getting rather large.

@tarsius any reason you've decided to stop supporting Emacs 24.4 recently? Emacs versions don't have a formal EOL AFAIK.

bbatsov: I was thinking when it comes to bugfix releases, we should have in the matrix just the most recent. The matrix is indeed pretty big already.

tarsius: Eventually I will have to do it anyway. For the upcoming release I have concentrated on doing many things that don't require major refactorings and new abstractions. After the release I will concentrate on things that *are* new abstractions that depend on features only available in > 25.

xiongtx: Do older Emacs versions even get any fixes? Or is it fix-forward only? Since CIDER is an Emacs package, we should maintain support only to the extent that Emacs does, and no more.

tarsius: Once or twice every decade a security fix is backported. Or not - the latest fix only resulted in the release of `25.3`, but I don't know if that's because `24` wasn't affected or because it is no longer supported.

PR #2131, opened 2017-12-11
Improve point restoration of `cider--format-buffer`

xiongtx:
See [discussion](https://github.com/clojure-emacs/cider/commit/c1a500d7da606f48cbde8546da922ca6ddfea9ef).

I found that 32 chars wasn't quite enough to distinguish b/t some code blocks, so I doubled it to 64 :man_shrugging:.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * [cider-nrepl#438](https://github.com/clojure-emacs/cider-nrepl/pull/438): Improve startup time by deferring loading CIDER's middleware until the first usage.
 * [#2078](https://github.com/clojure-emacs/cider/pull/2078): Improve startup time by bundling together sync requests during startup.
 * `cider-rotate-default-connection` will warn if you use it with only a single active connection.
-* `cider-format-buffer` preserves the point position.
+* `cider-format-buffer` tries to preserve the point position.
 
 ### Bugs Fixed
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Just one small remark - can you please extract this to some `save-excursion` type of macro, so we can reuse in other formatting commands?

xiongtx:
I'd imagine that'd be for functions like the proposed `cider-format-defun`. It'd be better to implement those first so we can see what's best extracted.

kommen:
Are you talking about another `cider-format-defun` than this? https://github.com/clojure-emacs/cider/blob/c1a500d7da606f48cbde8546da922ca6ddfea9ef/cider-interaction.el#L1834

bbatsov:
@kommen Damn, seems I wrote those and forgot about them. @xiongtx Guess now you know. :D

xiongtx:
@bbatsov I didn't use a macro (didn't find the need for one) but did remove duplication b/t `cider--format-region` and `cider--format-buffer`. All the formatting and point restoration logic is now within `cider--format-region`, making `cider--format-buffer` just a call to the former.

bbatsov:
That's a decent approach. I suggested the macro only because the thought crossed might mind we mind need this in other places where we're doing similar replacements, but let's not go overboard with this without any concrete reason to do so.

PR #2132, opened 2017-12-11
Replace `if-let` and `when-let` with starred versions

xiongtx:
Emacs 26 [obsoletes `if-let` and `when-let`](http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS?h=emacs-26#n1278), replacing them with `if-let*` and `when-let*`. This raises byte-compilation warnings (treated as errors) when testing against Emacs 26.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * [#2078](https://github.com/clojure-emacs/cider/pull/2078): Improve startup time by bundling together sync requests during startup.
 * `cider-rotate-default-connection` will warn if you use it with only a single active connection.
 * `cider-format-buffer` preserves the point position.
+* [#2132](https://github.com/clojure-emacs/cider/pull/2132): Replace `if-let` and `when-let` with starred versions [per Emacs 26](http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS?h=emacs-26#n1278).
```
bbatsov: That's not a user-visible change, so we shouldn't put in in the changelog.

cider-overlays.el
```diff
@@ -153,9 +153,9 @@ END is the position where the sexp ends, and defaults to point."
                              'before-string cider--fringe-overlay-good)))))
 
 (cl-defun cider--make-result-overlay (value &rest props &key where duration (type 'result)
-                                        (format (concat " " cider-eval-result-prefix "%s "))
-                                        (prepend-face 'cider-result-overlay-face)
-                                        &allow-other-keys)
+                                            (format (concat " " cider-eval-result-prefix "%s "))
```
bbatsov: This change seems unrelated.

xiongtx: We should all get in the good habit of putting `whitespace-cleanup` in `before-save-hook` :stuck_out_tongue:.

CHANGELOG.md
xiongtx: :white_check_mark:

cider-overlays.el
bbatsov: Yeah, I get how this happened, I just don't unrelated changes in commits. Btw, even the fixed indentation seems wrong for some reason. `cl-defun` should be indented as `defun`.

bbatsov:
Thanks for taking care of this, @xiongtx! You rock! 🎸

PR #2134, opened 2017-12-11
Restore missing parens

dpsutton:
from ixup from when-let to when-let*

get end of buffer while reading file errors


bbatsov:
Ops, sorry about this! That's what we get with a broken CI... :-(

Issue #2129, opened 2017-12-11
"Cask could not be bootstrapped" errors in Travis CI


gonewest818:
## Expected behavior

Consistent CI test runs every time.

## Actual behavior

CI jobs intermittently failing with the message:

```
Cask could not be bootstrapped. Try again later, or report an issue at https://github.com/cask/cask/issues

The command "curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh" failed and exited with 1 during .
```

## Steps to reproduce the problem

Unfortunately this bug is intermittent. However in my experience it usually occurs at least once in any given matrix of Travis jobs.

Examples: Travis builds [4907.1](https://travis-ci.org/clojure-emacs/cider/jobs/314609496#L5083-L5088), [4907.7](https://travis-ci.org/clojure-emacs/cider/jobs/314609502), [4907.11](https://travis-ci.org/clojure-emacs/cider/jobs/314609506)

## Workaround

This appears to be transient. If you restart the Travis job, sooner or later it will succeed, provided there aren't other unresolved bugs in the source code.

### CIDER version information

Varies.  Does not seem related to the version of emacs/cider being tested in Travis.

### Lein/Boot version

n/a

### Emacs version

n/a

### Operating system

Linux (Travis runs Ubuntu 14.04)

bbatsov:
Cask is abandonware and causes way more problems than it solves so we should drop it and just replace it with some simple script that fetches the cider dependencies. I'm sick from having to deal with it. Not sure if EVM uses Cask in any way as they have the same author. If that's true maybe we'll have to find a different approach to test on multiple Emacs versions as well.

xiongtx:
> Cask is abandonware

Really? The [last commit](https://github.com/cask/cask/commit/eea660b78c05d70159067f5be8cd3bddb55df4c5) is only one month ago.

bbatsov:
Count the numbers of commits since then https://github.com/cask/cask/issues/376 For me Cask is dead at this point.

xiongtx:
> For me Cask is dead at this point.

:ghost: 

> Not sure if EVM uses Cask in any way as they have the same author.

[Doesn't seem like EVM uses Cask](https://github.com/rejeep/evm/search?utf8=%E2%9C%93&q=cask&type=), for what it's worth.

bbatsov:
Thanks for looking into this!

gonewest818:
A proposal: 

[...]

PR #2139, opened 2017-12-14
[#2129] make evm and cask setup more reliable in travis

gonewest818:
Resolves [#2129] with new scripts to setup evm and cask in Travis builds.

These scripts improve on the prior process by adding error detection with
retries, such that network errors during this setup stage are less likely to
abort the build with mysterious errors.

In addition to the retry logic, we also cache the installations of evm and cask
such that the contents of those directories are restored across successful
builds. This eliminates the requirement to install evm and cask from scratch. It
is possible to force-delete these caches if needed.

The net result of all this is, most incremental builds need only attempt to
update evm and cask (essentially a git pull) and update any needed elisp
packages via cask update.

-----------------

- [ ] [n/a] You've added tests (if possible) to cover your change(s)
- [ ] [n/a] The new code is not generating bytecode or `M-x checkdoc` warnings
- [ ] [n/a] You've updated the changelog (if adding/changing user-visible functionality)
- [ ] [n/a] You've updated the readme (if adding/changing user-visible functionality)
- [ ] [n/a] You've updated the refcard (if you made changes to the commands listed there)

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
 
 env:
   global:
-    - PATH=$HOME/local/bin:$PATH
+    - PATH=$HOME/local/bin:$HOME/local/evm/bin:$HOME/local/cask/bin:$PATH
   matrix:
     - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test
     - EMACS_BINARY=emacs-24.4-travis MAKE_TEST=test-bytecomp
@@ -41,11 +41,11 @@ env:
     - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test-checks
 
 before_script:
-  - sudo sh travis-ci/travis-gnutls.sh
+  - sh travis-ci/install-gnutls.sh
   - gnutls-cli -v
-  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh
+  - sh travis-ci/install-evm.sh
   - evm install $EMACS_BINARY --use --skip
-  - make elpa
+  - sh travis-ci/install-cask.sh
 script:
   - emacs --version
   - make $MAKE_TEST
--- a/travis-ci/install-cask.sh
+++ b/travis-ci/install-cask.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Install cask for Travis CI
+# or if already installed, then check for updates
+
+set -x
+
+WORKDIR=${HOME}/local
+CASKDIR=$WORKDIR/cask
+
+. travis-ci/retry.sh
+
+cask_upgrade_cask_or_reset() {
+    cask upgrade-cask || { rm -rf $HOME/.emacs.d/.cask && false; }
+}
+
+cask_install_or_reset() {
+    cask install || { rm -rf .cask && false; }
+}
+
+# Bootstrap the cask tool and its dependencies
+if [ -d $CASKDIR ]
+then
+    travis_retry cask_upgrade_cask_or_reset
+else
+    git clone https://github.com/cask/cask.git $CASKDIR
+    travis_retry cask_upgrade_cask_or_reset
+fi
+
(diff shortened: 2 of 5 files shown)


gonewest818:
by the way, you have probably been ignoring travis because it's been so broken, but I want to brag about the fact the build errors on this branch are now all completely legitimate issues to be dealt with.

xiongtx:
Can you explain why we need this retry logic? I recall (a long time ago in a galaxy far far away) that CI used to work reliably without this. Is the problem w/ @travis-ci, EVM, or Cask?

gonewest818:
I can’t completely explain why, but I’ve definitely seen the results of broken connections in various stages of the build, and for instance if the “cask install” fails to properly install the elisp package, or the files are corrupted in transit (I’ve seen that happen) then the build is hosed.

The retry logic is really about making the package install idempotent. Either it installs properly or it is deleted without a trace, so that the next install starts from a clean slate.  After three failed retries the job will abort just as before.

I will also say, I’ve never seen a failure with “apt-get” type package installs. I suspect Travis might have mirrors for the deb packages to avoid excessive (and rather repetitive) inbound traffic related to jobs setting themselves up.  So in that regard evm and cask are probably not benefitting from any kind of mirror - they’re just installing themselves plus all their package dependencies over the internet, over and over in each job.

And this change tries to minimize the impact of that.

gonewest818:
Here's a concrete example from the build https://travis-ci.org/clojure-emacs/cider/jobs/316281526

At the tail of the log it says:
```
Symbol's value as variable is void: -
Cask could not be bootstrapped. Try again later, or report an issue at https://github.com/cask/cask/issues
The command "curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh" failed and exited with 1 during .
```

And a little ways before that in the log it says:

```
Compiling /home/travis/.emacs.d/.cask/24.3/bootstrap/commander-20140120.1052/commander.el...

In toplevel form:
commander.el:512:1:Warning: reference to free variable `-'
commander.el:512:3:Warning: reference to free variable `Peer'
commander.el:512:8:Warning: reference to free variable `has'
commander.el:512:12:Warning: reference to free variable `closed'
commander.el:512:19:Warning: reference to free variable `the'
commander.el:512:23:Warning: reference to free variable `GnuTLS'
commander.el:512:30:Warning: reference to free variable `connection'
Wrote /home/travis/.emacs.d/.cask/24.3/bootstrap/commander-20140120.1052/commander.elc
```

Which is sort of amusing if it weren't so annoying.

Does Travis need to fix their infrastructure? Should the underlying emacs packaging system prevent this this sort of file corruption? Should cask be more resilient? Should every project need to patch over this in the way I did? Yes, yes, maybe and definitely no.

Should cask be thrown out because of this? Up to you, but this gives you reliable builds while you sort out your desired strategy for this.

xiongtx:
```
commander.el:512:1:Warning: reference to free variable `-'
commander.el:512:3:Warning: reference to free variable `Peer'
commander.el:512:8:Warning: reference to free variable `has'
commander.el:512:12:Warning: reference to free variable `closed'
commander.el:512:19:Warning: reference to free variable `the'
commander.el:512:23:Warning: reference to free variable `GnuTLS'
commander.el:512:30:Warning: reference to free variable `connection'
```

It seems the problem is still w/ GnuTLS? Is the closed connection a problem w/ @travis-ci? Does it happen every time?

gonewest818:
As mentioned in #2120 and #2129, it’s intermittent. Note that before pull request #2128 (fixed 2130) gnutls wasn’t even installed in your build jobs — so it’s not likely gnutls is the root cause of these intermittent errors.

bbatsov:
> Should cask be thrown out because of this? Up to you, but this gives you reliable builds while you sort out your desired strategy for this.

Likely yes. As I said - the project is mostly abandonware at this point and betting on it, doesn't seem particularly wise. Anyways, I appreciate every step in the direction of stable CI builds!

Thanks for working on this!

PR #2141, opened 2017-12-18
Dynamically build structs in `nrepl-bencode-tests`

### LINKED ISSUE #2133: Print representation of structs in Emacs 26 breaks tests
Some tests are [failing when run against Emacs 26](https://travis-ci.org/clojure-emacs/cider/jobs/314678704#L1868) due to the new `#s` struct print syntax, which seems to be due to the use of [records to represent structs](http://git.savannah.gnu.org/cgit/emacs.git/tree/doc/lispref/records.texi#n27).

xiongtx:
Fixes #2133

### DIFF
--- a/test/nrepl-bencode-tests.el
+++ b/test/nrepl-bencode-tests.el
@@ -28,6 +28,7 @@
 ;;; Code:
 
 (require 'buttercup)
+(require 'cl-lib)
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #2142, opened 2017-12-18
Fix bytecomp errors

xiongtx:
Fix various byte-compilation errors.

@bbatsov I propose we stop supporting Emacs 24.4. after the next release. We should cull the test matrix a bit and avoid accumulating too many version-based conditionals in the code.


bbatsov:
Great work! Thanks!

bbatsov:
> @bbatsov I propose we stop supporting Emacs 24.4. after the next release. We should cull the test matrix a bit and avoid accumulating too many version-based conditionals in the code.

I agree. 0.16 will be final version supporting it.

PR #2146, opened 2017-12-20
Fix nesting of when-let* body in several inspector functions

ak-coram:
The inspector has been broken (not showing up) for me since 8143c4a, this commit fixes it.

See 3a98c1b for a similar fix in cider-repl.el

-----------------

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -142,30 +142,30 @@ Interactively, EXPR is read from the minibuffer, and NS the
 current buffer's namespace."
   (interactive (list (cider-read-from-minibuffer "Inspect expression: " (cider-sexp-at-point))
                      (cider-current-ns)))
-  (when-let* ((value (cider-sync-request:inspect-expr expr ns (or cider-inspector-page-size 32)))
-    (cider-inspector--render-value value))))
+  (when-let* ((value (cider-sync-request:inspect-expr expr ns (or cider-inspector-page-size 32))))
+    (cider-inspector--render-value value)))
 
 (defun cider-inspector-pop ()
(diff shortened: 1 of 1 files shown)


bbatsov:
We really messed this up. Sorry about this! And thanks for the fix! 👍

bbatsov:
@xiongtx I wonder why this wasn't causing issues in the CI - probably because the file never got loaded. We might add some sanity checking test that simply loads all files.

xiongtx:
> We really messed this up.

:grimacing: :bowing_man: 

But in this case byte-compilation didn't raise an error b/c the form is valid--the parens are balanced, they're just in the wrong place.

I see that we have no tests for `cider-inspector` functions--if we had any of _those_, they could've caught this.

bbatsov:
> But in this case byte-compilation didn't raise an error b/c the form is valid--the parens are balanced, they're just in the wrong place.

They are balanced, but they don't look like proper binding forms. I thought this raised some warnings (I have some memory of seeing such in the past).

bbatsov:
And I hope there are no more of those lying around.

xiongtx:
> They are balanced, but they don't look like proper binding forms.

Unfortunately, this is a proper binding form:

```emacs-lisp
(when-let* ((value (cider-sync-request:inspect-refresh))
            (cider-inspector--render-value value)))
```

It's just binding the value of `cider-inspect--render-value` to `value` :man_facepalming:.

bbatsov:
Ah, yeah. Didn't pay much attention to this - seems like the very definition of bad luck. Thanks for checking it!

PR #2149, opened 2018-01-02
[Fix #1913] Allow toggling of current buffer connection

### LINKED ISSUE #1913: cider-toggle-buffer-connection fails

dpsutton:
Cljc buffers send their evals to both clj and cljs repls if available due to
`cider-map-connections`. Toggling a current buffer's connection involves hiding
the other connection.

Previously, when toggling _again_, the original list was not consulted and only
the truncated list, preventing the other connection from being found. This
allows for the full list to be searched for the other buffer.

In addition, a prefix dictates that the local connection list is discarded in
favor of the full list, restoring the evaluation in both clj and cljs
buffers (if both are present).

See https://github.com/clojure-emacs/cider/issues/1913

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -225,14 +225,35 @@ such a link cannot be established automatically."
     (when conn
       (setq-local cider-connections (list conn)))))
 
-(defun cider-toggle-buffer-connection ()
-  "Toggle the current buffer's connection between Clojure and ClojureScript."
-  (interactive)
+(defun cider-toggle-buffer-connection (&optional restore-all)
+  "Toggle the current buffer's connection between Clojure and ClojureScript.
+
+Default behavior of a cljc buffer is to send eval commands to both Clojure
+and ClojureScript.  This function sets a local buffer variable to hide one
+or the other.  Optional argument RESTORE-ALL undo any toggled behavior by
+using the default list of connections."
+  (interactive "P")
   (cider-ensure-connected)
-  (let ((other-conn (cider-other-connection)))
-    (if other-conn
-        (setq-local cider-connections (list other-conn))
-      (user-error "No other connection available"))))
+  (if restore-all
+      (progn
+        (kill-local-variable 'cider-connections)
+        (let ((types (mapcar #'cider--connection-type (cider-connections))))
+          (message (format "Cider connections available: %s" types))))
+    (let ((current-conn (cider-current-connection))
+          (was-local (local-variable-p 'cider-connections))
+          (original-connections (cider-connections)))
+      ;; we set the local variable to eclipse all connections in favor of the
+      ;; toggled connection. to recover the full list we must remove the
+      ;; obfuscation
+      (kill-local-variable 'cider-connections)
+      (if-let* ((other-conn (cider-other-connection current-conn)))
+        (progn
+          (setq-local cider-connections (list other-conn))
+          (message (format "Connection set to %s" (cider--connection-type other-conn))))
+        (progn
+          (when was-local
+            (setq-local cider-connections original-connections))
+          (user-error "No other connection available"))))))
 
 (defun cider-clear-buffer-local-connection ()
   "Remove association between the current buffer and a connection."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs Fixed
+
+* [#1913](https://github.com/clojure-emacs/cider/issues/1913): Fix `cider-toggle-buffer-connection` to allow cycling of connection and restoring all connections in cljc buffers
+
 ## 0.16.0 (2017-12-28)
 
 ### New Features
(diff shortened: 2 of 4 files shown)


dpsutton:
> The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install autogen ca-certificates curl gcc libgmp-dev m4 make pkg-config xz-utils" failed and exited with 100 during .

> An error occurred while generating the build script.

Seems like two transient build errors in travis

cider-client.el
```diff
@@ -225,14 +225,35 @@ such a link cannot be established automatically."
...
+  (interactive "P")
   (cider-ensure-connected)
-  (let ((other-conn (cider-other-connection)))
-    (if other-conn
-        (setq-local cider-connections (list other-conn))
-      (user-error "No other connection available"))))
+  (if restore-all
+      (progn
+        (kill-local-variable 'cider-connections)
+        (let ((types (mapcar #'cider--connection-type (cider-connections))))
```
bbatsov: I don't get what's the purpose of this portion of the code.

cider-client.el
```diff
@@ -225,14 +225,35 @@ such a link cannot be established automatically."
...
+          (message (format "Cider connections available: %s" types))))
+    (let ((current (cider-current-connection))
```
bbatsov: I'd name this `current-conn`.

bbatsov:
> Cljc buffers send their evals to both clj and cljs repls if available due to
cider-map-connections. Toggling a current buffer's connection involves hiding
the other connection.

This reminds me it might be nice to explain this behavior somewhere in the manual.

cider-client.el
dpsutton: The default behavior of a `cljc` buffer is to map evaluations across all connections. Toggling switches to just one or the other. With a prefix arg, (`restore-all`), it deletes the buffer local list of connections and allows the eval mechanism to see the original list of connections.

cider-client.el
dpsutton: done and push.

dpsutton:
documentation updated in `up_and_running.md`

cider-client.el
bbatsov: I get what the code does in general - I was referring to this `let` and the message inside it. You get the types of all connections and you print them in a message basically, which seems strange to me.

dpsutton: Ah yes. The connections stuff can be rather opaque. So when toggling it it tells you which connection it is going to use, and then I give a message listing all connections when all of the connections are "untoggled"

bbatsov: Ah, I finally get it.

bbatsov:
Thanks!

PR #2151, opened 2018-01-02
Improve formatting of spec in `cider-doc` buffer

### LINKED ISSUE #2150: `cider-doc` specs could use better formatting
## Expected behavior

Specs in `cider-doc` buffers should be properly indented and font-locked.

image

## Actual behavior

Specs in `cider-doc` buffers are not properly indented or font-locked.

image

The poor formatting is in part due to attempts to perform formatting on the `cider-nrepl` side rather than the Emacs side. `cider-nrepl` should return _data_ rather than formatted strings (to the extent possible) and let CIDER take care of formatting.

## Note

#2029 also deals with formatting specs in the `cider-doc` buffer. Perhaps the work here could be used to improve the work there.

xiongtx:
Fix #2150

Depends on clojure-emacs/cider-nrepl#465

Font-lock and align spec in `cider-doc` buffer properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#2151](https://github.com/clojure-emacs/cider/pull/2151) Improve formatting of spec in `cider-doc` buffer
+
 ## 0.16.0 (2017-12-28)
 
 ### New Features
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice improvement! Just add a changelog entry for it.

xiongtx:
Added changelog entry

bbatsov:
👍

PR #2154, opened 2018-01-05
cider-doc now renders spec using `cider-browse-spec--pprint-indented`

jpmonettas:
There is also a button to jump to the browser

**This needs to be merged together with!** https://github.com/clojure-emacs/cider-nrepl/pull/472

-----------------

### DIFF
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -36,7 +36,6 @@
 
 (require 'cider-client)
 (require 'cider-compat)
-(require 'cider-interaction)
 (require 'cider-util)
 (require 'cl-lib)
 (require 'nrepl-dict)
@@ -233,15 +232,30 @@ Display TITLE at the top and SPECS are indented underneath."
             ;; prettier (s/fspec )
             ((cider--spec-fn-p form-tag "fspec")
              (thread-last (seq-partition (cl-rest form) 2)
+               (cl-remove-if (lambda (s) (and (stringp (cl-second s))
+                                              (string-empty-p (cl-second s))))) 
                (mapcar (lambda (s)
-                         (concat "\n" (cl-first s) " " (cider-browse-spec--pprint (cl-second s)))))
+                         (format "\n%-11s: %s" (pcase (cl-first s)
+                                                (":args" "arguments")
+                                                (":ret" "returns")
+                                                (":fn" "invariants"))
+                                 (cider-browse-spec--pprint (cl-second s)))))
                (cl-reduce #'concat)
-               (format "(s/fspec \n %s)")))
+               (format "%s")))
             ;; every other with no special management
             (t (format "(%s %s)"
                        (cider-browse-spec--pprint form-tag)
                        (string-join (mapcar #'cider-browse-spec--pprint (cl-rest form)) " "))))))
         (t (format "%s" form))))
+
+(defun cider-browse-spec--pprint-indented (spec-form)
+  "Given a SPEC-FORM returns a string with it pretty printed, indented and clojure font locked."
+  (with-temp-buffer
+    (clojure-mode)
+    (insert (cider-browse-spec--pprint spec-form))
+    (indent-region (point-min) (point-max))
+    (cider--font-lock-ensure)
+    (buffer-string)))
 
 (defun cider-browse-spec--draw-spec-buffer (buffer spec spec-form)
   "Reset contents of BUFFER and draws everything needed to browse the SPEC-FORM.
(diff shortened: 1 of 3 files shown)


jpmonettas:
@bbatsov **This is not ready to be merged yet**, it works with eval-buffer but has a compilation error due to dependency cycle. I'm creating the pull request to discuss the solution.

The thing is I need to create a button in cider-doc.el that when clicked calls a procedure inside cider-browse-spec.el.

Can someone with more experience in elisp help me with this? Thanks

xiongtx:
A screenshot would be helpful:

image

This presentation could be improved. See the [existing](https://github.com/clojure-emacs/cider/issues/2150#issue-285540638):

image

jpmonettas:
@xiongtx Don't know why it looks like that, mines looks different :

latest-screenshot

I prefer it to show anonymous functions as functions and not to remove the #, because they look like the spec is invalid.

xiongtx:
Hmm, I'm not sure what's causing the differences. Perhaps @bbatsov can see what results he gets.

bbatsov:
Perhaps I'm missing something, but the first version does look a bit tidier to me. Perhaps we can make it configurable which version to display in the doc buffer? After all I guess the "Browse" button is the most important thing.

xiongtx:
Which version looks tidier? Couldn't be the first one in [my post](https://github.com/clojure-emacs/cider/pull/2154#issuecomment-355682134)--it doesn't even line up the `args`, `ret`, etc.

Anyways, I'd be fine w/ merging what @jpmonettas [shows](https://github.com/clojure-emacs/cider/pull/2154#issuecomment-355685909), though I don't think the `s/fspec` is necessary--it's just noise. That it's an `fdef` is obvious from the `:args`, `:ret`, etc.

cider-browse-spec.el
```diff
@@ -243,6 +245,14 @@ Display TITLE at the top and SPECS are indented underneath."
                        (string-join (mapcar #'cider-browse-spec--pprint (cl-rest form)) " "))))))
         (t (format "%s" form))))
 
+(defun cider-browse-spec--pprint-indented (spec-form)
```
xiongtx: Seems this can be simplified w/ `cider-font-lock-as-clojure` on the pprinted string.

bbatsov:
By first, I meant the existing one. I like the nice structural distinction between `args`, `ret` and `fn` (and I also think we can use better labels for those, btw). My personal preference would be to show by default something easier to parse by human minds and optionally be able to show the actual spec as it exists in the code.

jpmonettas:
@bbatsov @xiongtx Just removed the surrounding (s/fspec ...) stuff, now it looks like:

latest-screenshot

The cat looks nicer in our current version because it only has two parts, so we can show it all in one line, but as soon as you have a larger cat line I think it looks nicer to print one part in each line.

> My personal preference would be to show by default something easier to parse by human minds and optionally be able to show the actual spec as it exists in the code.

I think this too, that's why I created that elisp spec formatter, so we have control and can fine tune how we print a spec instead of relaying on vanilla s/form pprinting.

Some time ago I created [pretty-spec](https://github.com/jpmonettas/pretty-spec) that uses [fipp](https://github.com/brandonbloom/fipp) to pretty print specs sensitive to a width you have defined. Using something like this will require to move pprinting from elisp to cider-nrepl. (I explored those ideas some time ago when writing [Inspectable](https://github.com/jpmonettas/inspectable)

cider-doc.el
```diff
@@ -461,18 +461,14 @@ Tables are marked to be ignored by line wrap."
...
-              (insert (format "%-4s: " role))
-              (thread-first desc
-                cider-sync-request:format-code
-                cider-font-lock-as-clojure
-                (split-string "\n")
-                insert-rectangle))
-            (insert "\n"))
-          (insert "\n"))
+          (emit "Spec:" 'font-lock-function-name-face)
+          (insert (cider-browse-spec--pprint-indented spec))
```
xiongtx: You should probably `(require 'cider-browse-spec)` at the top of the file.

xiongtx:
Yes, removing the `s/fspec` stuff cleans it up a lot. I'm good w/ how it looks now.

jpmonettas:
Maybe this looks better? What about the naming ?

latest-screenshot

cider-doc.el
xiongtx: Try to figure out what's causing the dependency cycle then, and break it. We probably have some unnecessary `require`s in some files, or maybe some functions need to be shuffled around. 

In any case, you're using `cider-browse-spec--...` in `cider-doc.el` so you should definitely `require` it in `cider-doc.el`.

jpmonettas: Yeah, was cider-browse-spec depending on cider-interaction, it doesn't need to, so this is fixed.

jpmonettas:
@bbatsov @xiongtx if both agree on the latest screenshot now everything is ready to be merged together with https://github.com/clojure-emacs/cider-nrepl/pull/472

bbatsov:
I'm fine with it. Just rebase on top of the current `master`.

bbatsov:
You should also add some changelog entry.

jpmonettas:
@bbatsov added an entry to the changelog and rebased on top of master.

cider-doc.el
```diff
@@ -461,18 +462,14 @@ Tables are marked to be ignored by line wrap."
...
+          (insert "\n\n")
+          (insert-text-button "Browse spec"
+                              'follow-link t
+                              'action (lambda (x)
+                                        (cider-browse-spec (format "%s/%s" ns name))))
```
bbatsov: I noticed at the build output `x` is unused, which I assume is a mistake. It should be just `_` if it's not needed.

PR #2156, opened 2018-01-07
[Fix #2148] 'jump to definition' when remote cider -jack-in and -connect (#2148)

### LINKED ISSUE #2148: 'jump to definition' not working on cider-connect with remote embed repl server.
I don't know it is bug or not, yet. 

## Expected behavior
Jumping to the definition that required by ns.

## Actual behavior
Not jumping. `symbol:` prompted.

Add `[cider/cider-nrepl "0.17.0-SNAPSHOT"]` dependency to project.clj.
Add `[cider.nrepl :refer [cider-nrepl-handler]]` to `ns` in clj file that send its handler to an embed REPL server.
Find remote project folder and files by Emacs Tramp.
`cider-connect` on any remote project file to remote embed REPL server.
Connection established. 'cider-repl buffer' worked fine.
Try jump to definition in any function from required external namespace.
Failed and `symbol:` prompted. 

*This is extremely important! Providing us with a reliable way to reproduce
a problem will expedite its solution.*

### CIDER version information
0.17.0-SNAPSHOT (20171228.1254), nREPL 0.2.12
Clojure 1.8.0, Java 1.8.0_151

### Lein/Boot version
Lein 2.8.1

### Emacs version
24.5

### Operating system
Elementary OS Loki based on Ubuntu 16.04

soulawaker:
1. The `cider-tramp-prefix` does not return only tramp prefix.
     It used the `tramp-make-tramp-file` in the `with-parsed-tramp-file-name` which force it to make a full tramp file path, not only prefix. I placed the `cider-make-tramp-prefix` for making prefix only. It is modified a little bit from `tramp-make-tramp-file`.

2. The `cider--client-tramp-filename` does not remove `file:` scheme for tramp filename.
    It should return the string with tramp prefix and remote source path. But previous version leaves `file:` scheme in a remote source path. I just added a line to remove this scheme in this function.

3. The `cider-find-file` leaves a useless list buffer per opening remote archived source like zip, jar.  
   When continuous tracking, it'll be big trash make us tired to remove. Now it'll be removed automatically.

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -125,13 +125,13 @@ create a valid path."
 
 (defun cider-make-tramp-prefix (method user host)
   "Constructs a Tramp file prefix from METHOD, USER, HOST.
-It originated from Tramp's `tramp-make-tramp-file-name'.  The original be forced
-to make full file name with `with-parsed-tramp-file-name', not providing
-prefix only option.  This function does not support to Emacs 26 or later."
+It originated from Tramp's `tramp-make-tramp-file-name'.  The original be
+forced to make full file name with `with-parsed-tramp-file-name', not providing
+prefix only option."
   (concat tramp-prefix-format
-          (when (not (zerop (length method)))
+          (unless (zerop (length method))
             (concat method tramp-postfix-method-format))
-          (when (not (zerop (length user)))
+          (unless (zerop (length user))
             (concat user tramp-postfix-user-format))
           (when host
             (if (string-match tramp-ipv6-regexp host)
@@ -141,21 +141,16 @@ prefix only option.  This function does not support to Emacs 26 or later."
 
 (defun cider-tramp-prefix (&optional buffer)
   "Use the filename for BUFFER to determine a tramp prefix.
-Defaults to the current buffer.
-Return the tramp prefix, or nil if BUFFER is local."
+Defaults to the current buffer. Return the tramp prefix, or nil
+if BUFFER is local."
   (let* ((buffer (or buffer (current-buffer)))
          (name (or (buffer-file-name buffer)
                    (with-current-buffer buffer
                      default-directory))))
     (when (tramp-tramp-file-p name)
       (with-parsed-tramp-file-name name v
-        ;; `tramp-make-tramp-file-name' was changed to take 6 mandatory
-        ;; parameters in Emacs 26 instead of 4
-        (if (version< emacs-version "26")
-            (with-no-warnings
-              (cider-make-tramp-prefix v-method v-user v-host))
-          (with-no-warnings
-            (tramp-make-tramp-file-name v-method v-user v-domain v-host v-port v-localname)))))))
+        (with-no-warnings
+          (cider-make-tramp-prefix v-method v-user v-host))))))
 
 (defun cider--client-tramp-filename (name &optional buffer)
   "Return the tramp filename for path NAME relative to BUFFER.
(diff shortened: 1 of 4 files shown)


soulawaker:
It is not tested for Emacs 26. So I don't ensure to work in Emacs 26 or later.

cider-common.el
```diff
@@ -123,6 +123,22 @@ create a valid path."
         (match-string 1 filename)
       filename)))
 
+(defun cider-make-tramp-prefix (method user host)
+  "Constructs a Tramp file prefix from METHOD, USER, HOST.
+It originated from Tramp's `tramp-make-tramp-file-name'.  The original be forced
+to make full file name with `with-parsed-tramp-file-name', not providing
+prefix only option.  This function does not support to Emacs 26 or later."
```
bbatsov: I don't get the last part of this comment. What's wrong with Emacs 26 or newer?

cider-common.el
```diff
@@ -202,6 +219,7 @@ existing file ending with URL has been found."
          (when-let* ((entry (match-string 3 url))
                      (file  (cider--url-to-file (match-string 2 url)))
                      (path  (cider--file-path file))
+                     (trash (replace-regexp-in-string "^/.+/" "" path))
```
bbatsov: I guess there should be some explanation about what exactly is this trash buffer in the code.

cider-common.el
```diff
@@ -123,6 +123,22 @@ create a valid path."
         (match-string 1 filename)
       filename)))
 
+(defun cider-make-tramp-prefix (method user host)
+  "Constructs a Tramp file prefix from METHOD, USER, HOST.
+It originated from Tramp's `tramp-make-tramp-file-name'.  The original be forced
+to make full file name with `with-parsed-tramp-file-name', not providing
+prefix only option.  This function does not support to Emacs 26 or later."
+  (concat tramp-prefix-format
+          (when (not (zerop (length method)))
```
bbatsov: I'd use `(unless (string-empty-p ...)` here.

bbatsov:
Apart from my small remarks the changes look good.

cider-common.el
```diff
@@ -123,6 +123,22 @@ create a valid path."
         (match-string 1 filename)
       filename)))
 
+(defun cider-make-tramp-prefix (method user host)
```
bbatsov: It'd be nice if you added a few unit tests for this function and maybe some of the related functions, so we can validate their behavior.

cider-common.el
soulawaker: `cider-tramp-prefix` has different processes for Emacs 26 and lower version.
According to the `cider-tramp-prefix` document and code, Tramp was changed in Emacs 26.
I've used Emacs 24, 25, never 26. So I don't know it would work in Emacs 26.
Hmm. But I think I can try this with emacs-mirror sources.
When I do it, I would push an incremental PR to here some time.

cider-common.el
soulawaker: That trash buffer was made by `(find-file path)` in `cider-find-file`. It's a list of files in archived Zip or Jar. It is useless after jumping to the buffer which has the real definition. That trash buffer has the name like *.jar which is the last part of `path`. So it is enable to targeting that specific buffer.

cider-common.el
soulawaker: It just originated, well, more exactly, copied from `tramp-make-tramp-file-name` in `tramp.el`.
It would be good as if such a making tramp prefix function is prepared in `tramp.el`. 
I thought it was not bad to respect the original maker's thought.

cider-common.el
soulawaker: Gladly, I will do. Maybe I need more time.

cider-common.el
bbatsov: Fine by me, although simply the code was written a long time ago.

cider-common.el
bbatsov: I get this, but future readers of the code won't get it - that's why I suggested adding a comment explaining it.

cider-common.el
bbatsov: Can't you just read the source for the Emacs 26 version to quickly compare the two versions?

soulawaker: Sure, why not.

cider-common.el
soulawaker: Ah, got it. I'll comment it.

cider-common.el
bbatsov: 👍

bbatsov:
Thanks! 🙇

PR #2158, opened 2018-01-09
Fix typo

daveliepmann:
This PR fixes a tiny 1-letter typo in documentation: "test a point" where it should say "test at point".

I did not run tests or any of the other checkboxes because it is a one-character documentation change that does not touch code.


bbatsov:
Thanks!

Issue #2157, opened 2018-01-09
`cider-scale-background-color` produces unusable color codes


daveliepmann:
Everywhere that `cider-scale-background-color` is used—in test output and stacktraces—it creates unusable 12-character color codes. These color codes are ignored, causing the default (white) to be used. Since I have a dark theme, this makes CIDER's stacktraces and test output jarringly difficult to read. 

For instance, in a buffer with a background color of "#161616", executing `(cider-scale-background-color)` evaluates to "#22e322e322e3". I suspect this is unintended behavior.

I currently use a workaround: 
```elisp
(set-variable 'cider-stacktrace-frames-background-color "#161616")
(set-variable 'cider-test-items-background-color "#333333")
```
...but would prefer not to need it. 😃 

## Expected behavior
When running tests, I expect test items to be displayed with a readable background color, scaled from my [dark theme](https://github.com/jackrusher/dotemacs/blob/master/eigengrau/eigengrau-definitions.el)'s existing background color. Similarly, when evaluating code that causes an excepton, I expect items in the stack trace to also be displayed with a scaled version of my background color.

## Actual behavior
Instead, in both cases (`*cider-test-report*` and `*cider-error*`), the background controlled by `cider-scale-background-color` is pure white. This conflicts with the dark theme both aesthetically and in the fact that the off-white text is unreadable.

## Steps to reproduce the problem
This happens when running tests or executing exception-producing statements. My theme is [eigengrau](https://github.com/jackrusher/dotemacs/blob/master/eigengrau/). The issue can be reproduced most directly, however, by simply evaluating `cider-scale-background-color` an seeing the 12-character output. The function detects the current buffer's background color, so try it with the background set to "#161616" or another dark color.

### CIDER version information

```
;; CIDER 0.16.0 (Riga), nREPL 0.2.13
;; Clojure 1.8.0, Java 1.8.0_151
```

### Emacs version

GNU Emacs 25.3.1 (x86_64-apple-darwin17.2.0, Carbon Version 158 AppKit 1561.1) of 2018-01-05

### Operating system

MacOS High Sierra v10.13.2 (17C88)

bbatsov:
@jeffvalk Can you take a look at this?

xiongtx:
Are you sure the 12-digit RGB triplets are not being applied properly? They _should_ work, [according to the Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Colors.html).

[...]

PR #2159, opened 2018-01-09
Use `color-lighten-name` instead of custom color scaling function

xiongtx:
No need to implement our own `cider-scale-color` when there's a built-in `color-lighten-name`.

Note that this does not given the _exact_ same result. `(cider-scale-color "#3F3F3F" 0.05)` gives `#4bcd4bcd4bcd`, shortened to `#4b4b4b`, while `(color-lighten-name "#3F3F3F" 5)` gives `"#4c0b4c0b4c0b"`, shortened to `#4c4c4c`. Close enough, I'd say.

See #2157.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -31,13 +31,16 @@
 
 ;;; Code:
 
-(require 'seq)
-(require 'clojure-mode)
-(require 'subr-x)
-(require 'cider-compat)
-(require 'nrepl-dict)
+;; Built-ins
 (require 'ansi-color)
 (require 'color)
+(require 'seq)
+(require 'subr-x)
+
+;; clojure-mode and CIDER
+(require 'cider-compat)
+(require 'clojure-mode)
+(require 'nrepl-dict)
 
 (defalias 'cider-pop-back 'pop-tag-mark)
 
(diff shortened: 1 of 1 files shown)


cider-util.el
```diff
@@ -31,13 +31,13 @@
 
 ;;; Code:
 
-(require 'seq)
-(require 'clojure-mode)
-(require 'subr-x)
-(require 'cider-compat)
-(require 'nrepl-dict)
 (require 'ansi-color)
+(require 'cider-compat)
```
bbatsov: Btw, the general approach to requires was supposed to be - built-in packages first, 3rd party afterwards, clojure-mode and cider last. Anyways, clearly here this was not the case, so I'm not with changing it.

xiongtx: Don't think the order of `require`s matters for functionality (or should matter, anyway), but that's an 👌  convention.

bbatsov:
I see the builds are mostly failing. Just make sure that's not some real problem.

xiongtx:
Turns out there's no frame background color when running tests, so `color` is set to `"unspecified-bg"`. Need to add a check for that.

PR #2167, opened 2018-01-15
Java files are now linked in stack traces if paths are configured

abo-abo:
cider-stacktrace.el (cider-jdk-src-paths): New defvar.

* cider-stacktrace.el (cider-resolve-java-class): New defun.
(cider-stacktrace-navigate): Update.

On Ubuntu:

    sudo apt install openjdk-8-source

The zip is installed to /usr/lib/jvm/openjdk-8/src.zip.

Download also this one:
https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/clojure-1.8.0-sources.jar

Extract both and configure e.g. like so:

    (setq cider-jdk-src-paths '("~/git/java/clojure-1.8.0-sources"
                                "~/git/java/openjvm-8-src"))

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -576,10 +576,12 @@ Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/.
        'file-exists-p
        (mapcar
         (lambda (d)
-          (if (file-directory-p d)
-              (expand-file-name file-name d)
-            (when (file-exists-p d)
-              (format "zip:file:%s!/%s" d file-name))))
+          (cond ((file-directory-p d)
+                 (expand-file-name file-name d))
+                ((and (file-exists-p d)
+                      (member (file-name-extension d) '("jar" "zip")))
+                 (format "zip:file:%s!/%s" d file-name))
+                (t (error "Unexpected archive: %s" d))))
         cider-jdk-src-paths)))))
 
 (defun cider-stacktrace-navigate (button)
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -561,6 +561,25 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors' 
         (button-put button 'help-echo "Click to promote these stacktraces."))
       (button-put button 'suppressed (not suppressed)))))
 
+(defvar cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
+  "Used by `cider-stacktrace-navigate'.
+Zip files work, but it's better to extract them and put the directory paths here.
+Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/.")
+
+(defun cider-resolve-java-class (class)
+  "Return a zip path to a file that corresponds to CLASS."
+  (when class
+    (let ((file-name (concat (replace-regexp-in-string "\\." "/" class) ".java")))
+      (cl-find-if
+       'file-exists-p
+       (mapcar
+        (lambda (d)
+          (if (file-directory-p d)
+              (expand-file-name file-name d)
+            (when (file-exists-p d)
+              (format "zip:file:%s!/%s" d file-name))))
+        cider-jdk-src-paths)))))
+
 (defun cider-stacktrace-navigate (button)
   "Navigate to the stack frame source represented by the BUTTON."
   (let* ((var (button-get button 'var))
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice!

Can also also add something about this in the manual (e.g. in the `Configuration` section and in the changelog?

cider-stacktrace.el
```diff
@@ -561,6 +561,25 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
         (button-put button 'help-echo "Click to promote these stacktraces."))
       (button-put button 'suppressed (not suppressed)))))
 
+(defvar cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
+  "Used by `cider-stacktrace-navigate'.
+Zip files work, but it's better to extract them and put the directory paths here.
+Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/.")
+
+(defun cider-resolve-java-class (class)
```
bbatsov: Not sure if the name here is a good one as we usually use `resolve` for mapping a name to some object. Although here it makes some sense as well.

cider-stacktrace.el
```diff
@@ -561,6 +561,25 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
         (button-put button 'help-echo "Click to promote these stacktraces."))
       (button-put button 'suppressed (not suppressed)))))
 
+(defvar cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
```
bbatsov: I wonder if for the default we can't do some search of some common folders to make it more useful out of the box.

cider-stacktrace.el
```diff
@@ -561,6 +561,25 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
...
+  "Used by `cider-stacktrace-navigate'.
+Zip files work, but it's better to extract them and put the directory paths here.
+Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/.")
```
bbatsov: Btw, I'd make this this a `defcustom`.

cider-stacktrace.el
abo-abo: We could do it, but I'm only familiar with Debian/Ubuntu setup. Two things are relevant: the source isn't installed by default; and if it's installed via the package manager, it's a zip file.  Zip files are barely OK; I like to unzip them because then I get `dired-jump` and `counsel-ag`.

I think that if the steps to configure it are manual anyway, it doesn't make sense to try too much auto detection.

cider-stacktrace.el
```diff
@@ -561,6 +561,26 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
...
       (button-put button 'suppressed (not suppressed)))))
 
+(defcustom cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
+  "Used by `cider-stacktrace-navigate'.
+Zip files work, but it's better to extract them and put the directory paths here.
+Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/."
+  :type '(list string))
+
+(defun cider-resolve-java-class (class)
+  "Return a zip path to a file that corresponds to CLASS."
```
bbatsov: Btw, I notice this returns a zip path just for zip files, so I guess you should update this docstring.

cider-stacktrace.el
```diff
@@ -561,6 +561,26 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
...
+  (when class
+    (let ((file-name (concat (replace-regexp-in-string "\\." "/" class) ".java")))
+      (cl-find-if
+       'file-exists-p
+       (mapcar
+        (lambda (d)
+          (if (file-directory-p d)
+              (expand-file-name file-name d)
+            (when (file-exists-p d)
```
bbatsov: Perhaps we should raise an error if the file is not `zip` or `jar`?

doc/configuration.md
```diff
@@ -117,6 +117,31 @@ path being `/usr/share/doc/java/api/`, put the following line in
...
+
+When an exception is thrown, e.g. when eval-ing `(. clojure.lang.RT foo)`, a
+stack trace pops up. Some places of the stack trace link to Clojure files,
+others to Java files. By default, you can click the Clojure file links to
+navigate there. If you configure `cider-jdk-src-paths`, you can also click the
+Java file links to navigate there.
+
+Here's how do get the JDK source on Ubuntu:
+
+    sudo apt install openjdk-8-source
```
bbatsov: Btw, what does `openjdk-source` install? Isn't it JDK 8 as well?

cider-stacktrace.el
```diff
@@ -561,6 +561,26 @@ Achieved by destructively manipulating the `cider-stacktrace-suppressed-errors'
         (button-put button 'help-echo "Click to promote these stacktraces."))
       (button-put button 'suppressed (not suppressed)))))
 
+(defcustom cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
+  "Used by `cider-stacktrace-navigate'.
+Zip files work, but it's better to extract them and put the directory paths here.
+Clojure sources here: https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/."
+  :type '(list string))
```
bbatsov: Please, also add here `:package-version '(cider . "0.17.0")`.

doc/configuration.md
abo-abo: Yes, I think so. That's the usual convention for source packages.

bbatsov:
Thanks!

PR #2172, opened 2018-01-21
Render diffs for expected / actual test results

xiongtx:
Renders diffs provided by clojure-emacs/cider-nrepl#478 in test result buffer.

```clojure
(deftest my-test
  (is (= {:a 1 :b 2}
         {:a 2 :b 3}
         {:c "a" :d "b"})))
```

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2172](https://github.com/clojure-emacs/cider/pull/2172): Render diffs for expected / actual test results.
 * [#2167](https://github.com/clojure-emacs/cider/pull/2167): Add new defcustom `cider-jdk-src-paths`. Configure it to connect stack trace links to Java source code.
 * [#2161](https://github.com/clojure-emacs/cider/issues/2161): Add new interactive command `cider-eval-defun-to-point` which is bound to `C-c C-v (C-)z`. It evaluates the current top-level form up to the point.
 * [#2113](https://github.com/clojure-emacs/cider/issues/2113): Add new interactive commands `cider-eval-last-sexp-in-context` (bound to `C-c C-v (C-)c`) and `cider-eval-sexp-at-point-in-context` (bound to `C-c C-v (C-)b`).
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks pretty cool!

Just two notes from me:

* How is this going to look for big data structures? 
* Did you know about the ediff support in `cider-test`? :-)

This change also requires a changelog entry.

xiongtx:
> * How is this going to look for big data structures?'

Arguably better than currently:

1. Forms are now inserted via `insert-rectangle`, so they align properly
2. `actual` values are displayed individually (w/ diffs below) instead of in a single `(not (= ...))`

> Did you know about the ediff support in cider-test? :-)

I was vaguely aware of it, but Ediff:

1.  Is heavy-handed
  - Changes your window configuration or creates a new frame
  - Puts results in 2 buffers which is total overkill for most cases
  - Requires confirmation to exit
2. Diffs only the last value in a multi-`=`:

Try w/:

```clojure
(deftest diff-test
  (is (= {:a 1 :b 2}
         {:a 2 :b 3}
         {:b 4 :c 5})))
```

3. Compares results _textually_, which could be screwed by indentation, map key order, etc.

bbatsov:
@xiongtx I'm sold. :-)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#2172](https://github.com/clojure-emacs/cider/pull/2172): Render diffs for expected / actual test results.
```
bbatsov: This should be under "New Features". :-)

xiongtx: ✅

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### New features
 
+* [#2172](https://github.com/clojure-emacs/cider/pull/2172): Render diffs for expected / actual test results.
+
+### New features
```
bbatsov: I think you meant to remove this - now we have two "New features" sections.

xiongtx: :man_facepalming:

PR #2173, opened 2018-01-21
[Fix #2155] install-gnutls.sh should exit early on error

### LINKED ISSUE #2155: travis-ci/install-gnutls.sh should exit early if libnettle can't be installed
I noticed an error in [a recent Travis CI log](https://travis-ci.org/clojure-emacs/cider/jobs/324610574)

```
configure: error: 
  ***
  *** Libnettle 3.1 was not found.
```

## Expected behavior

`install-gnutls.sh` should download and compile (1) libnettle, and (2) gnutls.

## Actual behavior

When libnettle isn't found (in the example job linked above, it looks like this particular Travis CI worker got borked somehow) this install script carries on and attempts to compile gnutls anyway, but it obviously can't succeed without the necessary dependency.

This is frustrating, but the error is transient and seems to be relatively rare. The faulty behavior is clear from the logs, however. Simple fix.

n/a

### Lein/Boot version

n/a

### Emacs version

n/a

### Operating system

n/a

gonewest818:
When setting up the job in Travis CI, it is possible for libnettle to fail to install properly, and should that actually happen, then there is little point trying to download and compile gnutls. This change detects this case and exits immediately with an error code.

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/travis-ci/install-gnutls.sh
+++ b/travis-ci/install-gnutls.sh
@@ -38,6 +38,8 @@ curl -O https://ftp.gnu.org/gnu/nettle/nettle-${NETTLE_VERSION}.tar.gz \
     && make -j4 install \
     && make distclean
 
+if [ "$?" -ne "0" ]; then exit 1; fi
+
 cd $WORKDIR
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #2185, opened 2018-01-29
Remove `:url`s from the cheatsheet hierarchy

### LINKED ISSUE #2184: `cider-cheatsheet` does not handle `:url`s
4e0458a is a good start, but does not handle sections with `(:url ...)` members.

cider-cheatsheet

The problem is that some sections can look like:

```emacs-lisp
("Vars & Global Environment"
 (:url "Documentation" "http://clojure.org/vars")
 ("Def Variants"
  (:special def)
  (clojure.core ...)))
```

So the `(while (stringp (caar ...)))` logic fails.

I'd be fine with removing the `(:url ...)` stuff. They add little value and complicate the data structure.

xiongtx:
Fix #2184

They add little value and complicate the data structure.

### DIFF
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -83,7 +83,6 @@
        (clojure.core count get subs compare)
        (clojure.string join escape split split-lines replace replace-first reverse re-quote-replacement index-of last-index-of starts-with? ends-with? includes?))
       ("Regex"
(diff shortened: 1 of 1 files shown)


bbatsov:
Fine by me. Supposedly I had added some code to ignore this entries, but I guess I messed something up. The only reason why I didn't kill those myself was that I assumed they were used in the Helm interface. Just noticed you killed them there too.

PR #2188, opened 2018-01-31
Remove last remnants of cljx code causing CI breakage

gonewest818:
Looks like the builds have been broken since a recent commit where cljx was removed.

-----------------

Thanks!

manual][2] extremely useful.*


bbatsov:
> It seems like the bytecompiler in Emacs 26 and newer seems capable of optimizing
away the usage of `section` inside the while loop, but then complains that the
outer let binding is unused.

Nice! It used to be used, before I refactored this code. I just didn't pay this enough attention, and I was getting some weird errors on the CI about undefined `help-mode-map`.

Thanks a bunch for cleaning up the mess I've created!

PR #2195, opened 2018-02-04
Update base image in Dockerfile

gonewest818:
Travis CI engineering publishes new Docker images every now and then. This is the latest.


bbatsov:
Thanks!

PR #2197, opened 2018-02-04
- integrate thunknyc profile to cider

jmayaalv:
This is my attempt to solve clojure-emacs/cider#1993 (comment)

This is the companion PR for `cider-nrepl`: https://github.com/clojure-emacs/cider-nrepl/pull/502

Before submitting a PR make sure the following things have been done:

-----------------

- [] The new code is not generating bytecode or `M-x checkdoc` warnings

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-profile.el
+++ b/cider-profile.el
@@ -1,0 +1,201 @@
+;;; cider-profile.el --- CIDER support for thunknyc/profile
+
+;; Copyright © 2015-2018  Bozhidar Batsov, Artur Malabarba and CIDER contributors
+
+;; Author: Juan E. Maya <jmayaalv@gmail.com>
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provides coarse-grained interactive profiling support.
+
+;;; Code:
+
+(require 'cider-client)
+(require 'cider-mode)
+
+(defconst cider-profile-buffer "*cider-profile*")
+
+;;;###autoload
+(defun cider-profile-message-response-handler (handler &optional buffer)
+  "Default message response handler.
+It takes HANDLER to process the message on an optional BUFFER or the current buffer if not specified."
+  (nrepl-make-response-handler
+   (or buffer (current-buffer))
+   handler
+   '()
+   '()
+   '()))
+
+;;;###autoload
+(defun cider-profile-samples (&optional query)
+  "Displays current max-sample-count.
+If optional QUERY is specified, set max-sample-count and display new value."
+  (interactive "P")
+  (cider-ensure-op-supported "set-max-samples")
+  (cider-ensure-op-supported "get-max-samples")
+  (if (not (null query))
+      (cider-nrepl-send-request
+       (let ((max-samples (if (numberp query) query '())))
+         (message "query: %s" max-samples)
+         (list "op" "set-max-samples" "max-samples" max-samples))
+       (cider-profile-message-response-handler
+        (lambda (_buffer value)
+          (let ((value (if (zerop (length value)) "unlimited" value)))
+            (message "max-sample-count is now %s." value)))))
+    (cider-nrepl-send-request
+     (list "op" "get-max-samples")
+     (nrepl-make-response-handler
+      (current-buffer)
+      (lambda (_buffer value)
+        (let ((value (if (zerop (length value)) "unlimited" value)))
+          (message "max-sample-count is now %s." value)))
+      '()
+      '()
+      '())))
+  query)
+
+;;;###autoload
+(defun cider-profile-var-profiledp (query)
+  "Displays the profiling status of var under point.
+Prompts for var if none under point or QUERY is present."
+  (interactive "P")
+  (cider-ensure-op-supported "is-var-profiled")
+  (cider-read-symbol-name
+   "Report profiling status for var: "
+   (lambda (sym)
+     (let ((ns (cider-current-ns)))
+       (cider-nrepl-send-request
+        (list "op" "is-var-profiled"
+              "ns" ns
+              "sym" sym)
+        (cider-profile-message-response-handler
+         (lambda (_buffer value)
+           (cond ((equal value "profiled")
+                  (message "profiling %s/%s." ns sym))
+                 ((equal value "unprofiled")
+                  (message "not profiling %s/%s." ns sym))
+                 ((equal value "unbound")
+                  (message "%s/%s is not bound." ns sym)))))))))
+  query)
+
+;;;###autoload
+(defun cider-profile-ns-toggle (&optional query)
+  "Toggle profiling for the ns associated with optional QUERY or, if nil current ns."
+  (interactive "P")
+  (cider-ensure-op-supported "toggle-profile-ns")
+  (let ((ns (if query (completing-read
+                       "Toggle profiling for ns: " (cider-sync-request:ns-list))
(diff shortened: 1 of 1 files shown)


jmayaalv:
This PR is still not ready to be merged. I am having some problems displaying the cider-profile buffer, it's just blank. I am not pretty sure how to debug it. Any help would be appreciated :)

bbatsov:
> This PR is still not ready to be merged. I am having some problems displaying the cider-profile buffer, it's just blank. I am not pretty sure how to debug it. Any help would be appreciated :)

Sure thing! I'll take a look at the code and I welcome the other active contributors to take a look as well. :-) /cc @xiongtx @ckoparkar @dpsutton

cider-profile.el
```diff
@@ -0,0 +1,211 @@
...
+    (lambda (_buffer value)
+      (when (equal value "cleared")
+        (message "cleared profile data.")))
+    '()
+    '()
+    '()))
+  query)
+
+;;;###autoload
+(define-minor-mode cider-profile-mode
```
bbatsov: I think we don't need this minor mode. Probably we can just put all the commands under `cider-mode` as usual. 
Not to mention those keys are reserved for users, so we can certainly come up with better ones.

bbatsov:
Btw, should should by running `M-x checdoc` in this buffer and fixing all the offenses in the docstrings, as those fail the build.

cider-profile.el
```diff
@@ -0,0 +1,211 @@
...
+                 ((equal value "unbound")
+                  (message (format "%s/%s is not bound." ns sym)))))
+         '()
+         '()
+         '()))))
+   query))
+
+(defun cider-profile-display-stats (stats-response)
+  (let ((table (nrepl-dict-get stats-response "err")))
+    (if cider-profile-buffer
```
dpsutton: I don't understand why this is a conditional. `cider-profile-buffer` is `defconst`ed to a string.

cider-profile.el
```diff
@@ -0,0 +1,211 @@
...
+         (list "op" "set-max-samples" "max-samples" max-samples))
+       (nrepl-make-response-handler
+        (current-buffer)
+        (lambda (_buffer value)
+          (let ((value (if (zerop (length value)) "unlimited" value)))
+            (message (format "max-sample-count is now %s." value))))
+        '()
+        '()
+        '()))
+    (cider-nrepl-send-request
```
dpsutton: this pattern comes up a lot in this code. i'd probably extract it into some type of `cider-message-response-handler` that only takes the lambda of how to case over the successful message. i saw that you handle errors on the nrepl side but i wasn't sure if those would need an error handler on this side.

cider-profile.el
```diff
@@ -0,0 +1,211 @@
...
+      (cider-repl-emit-err-output (cider-current-repl-buffer) table))))
+
+;;;###autoload
+(defun cider-profile-summary (query)
+  "Display a summary of currently collected profile data."
+  (interactive "P")
+  (cider-ensure-op-supported "profile-summary")
+  (cider-profile-display-stats
+   (cider-nrepl-send-sync-request (list "op" "profile-summary")))
+  query)
```
dpsutton: i'm not sure what `query` is doing here

cider-profile.el
jmayaalv: Thanks a lot @dpsutton 
this was mostly a copy & paste from https://github.com/thunknyc/nrepl-profile/

I am totally  newby on elisp, so i will do my best :)

cider-profile.el
jmayaalv: The `query` argument is not using anywhere but the function always accepts it. What's the patter to do it correctly? should i set the `query` argument optional?
This is being done in quiet a few places.

jmayaalv:
I've fixes the problems with the docs :) sorry. Had no idea that i had to run `checkdocs` on the buffer. Learning a lot here :)

I can now see the profile summary which is exciting! Thank you! 
I've also removed the `cider-profile-mode` and added new keybindings.

cider-profile.el
dpsutton: I don't understand what you are trying to accomplish. I don't know of a reason that you need to allow for an argument if you aren't using it. If you are just putting it there to make it interactive, it's fine to remove the argument and just use `(interactive)`. Otherwise, if there's a reason you need some information the function should act on that information.

cider-profile.el
```diff
@@ -0,0 +1,196 @@
...
+;;; Code:
+
+(require 'cider-client)
+(require 'cider-mode)
+
+(defconst cider-profile-buffer "*cider-profile*")
+
+;;;###autoload
+(defun cider-profile-message-response-handler (handler &optional buffer)
+  "Default message response handler.  It takes HANDLER to process the message on an optional BUFFER or the current buffer if not specified."
```
bbatsov: I'm surprised this pass the checkdoc test. The second sentence should start on the second line. This applies to all docstrings.

bbatsov:
Btw, you might also want to reformat this entire buffer, as right now I see the indent is wrong here and there.

@gonewest818 I've a small idea for a new travis check - whether the files changed have the right indentation or trailing whitespace. Guess we can just compare the committed file to the same file after `indent-region` and some cleanup function (`untabify`, `delete-trailing-whitespace`, etc).

gonewest818:
So, like `cljfmt check` but for elisp.  Yep, that sounds useful.

cider-profile.el
jmayaalv: Got it! it dint' really understood how `interactive` works.  I think i got it know. Thanks :)

dpsutton: awesome! and thanks for adding this feature!

cider-profile.el
```diff
@@ -0,0 +1,201 @@
...
+   "Toggle profiling for var: "
+   (lambda (sym)
+     (let ((ns (cider-current-ns)))
+       (cider-nrepl-send-request
+        (list "op" "toggle-profile"
+              "ns" ns
+              "sym" sym)
+        (cider-profile-message-response-handler
+         (lambda (_buffer value)
+           (cond ((equal value "profiled")
```
xiongtx: `cl-case` or `pcase`

cider-profile.el
```diff
@@ -0,0 +1,201 @@
...
+   "Report profiling status for var: "
+   (lambda (sym)
+     (let ((ns (cider-current-ns)))
+       (cider-nrepl-send-request
+        (list "op" "is-var-profiled"
+              "ns" ns
+              "sym" sym)
+        (cider-profile-message-response-handler
+         (lambda (_buffer value)
+           (cond ((equal value "profiled")
```
bbatsov: Here in Emacs you'd normally use `cl-case` or `pcase` to make this code more compact.

cider-profile.el
```diff
@@ -0,0 +1,201 @@
...
+  (cider-ensure-op-supported "toggle-profile-ns")
+  (let ((ns (if query (completing-read
+                       "Toggle profiling for ns: " (cider-sync-request:ns-list))
+              (cider-current-ns))))
+    (cider-nrepl-send-request
+     (list "op" "toggle-profile-ns"
+           "ns" ns)
+     (cider-profile-message-response-handler
+      (lambda (_buffer value)
+        (cond ((equal value "profiled")
```
bbatsov: Here as well.

bbatsov:
@jmayaalv Seems we're almost ready with this. Apart from the tiny feedback about `cl-case`/`pcase`, the only thing missing is a changelog entry. 

Take a look at my feedback about the Clojure part of the PR as well.

bbatsov:
I guess we can clean this up after we've merged the initial working version. Thanks for your work on this @jmayaalv! Looking forward to future contributions from you! 😉

bbatsov:
I was also reminded that this deserves a page in CIDER's manual - it'd be great if you wrote one when you have some time.

jmayaalv:
Thanks a lot for the help guys! 
I will create a new pr this week with the fixes and the entry in the manual! 

Would love to contribute more. If there is something you have in mind that a elisp newby can do i would be glad to help.

stardiviner:
@jmayaalv Thanks, I waited this for a long time. Really Thanks a lot.

cider-profile.el
xiongtx: Just realized that `cl-case` uses `eql` for comparison, while `pcase` uses `equal`. Strings are _not_ `eql` to each other. So

```emacs-lisp
(cl-case "foo"
  ("foo" "got-foo"))
;; => nil
```

PR #2204, opened 2018-02-12
[docs] Remove a note about company-mode 0.9, it's stable now

alexander-yakushev:
I think it is reasonable to assume that new users will install the latest version of company.


bbatsov:
👍

Issue #2199, opened 2018-02-06
add elisp formatting checks in CI


gonewest818:
As mentioned in #2197 it would be useful to have a basic formatting check running in CI to catch indentation issues, trailing whitespace, etc.

gonewest818:
I'm looking at https://github.com/nschum/elisp-lint as a possible starting point, but it needs some updating.

xiongtx:
We already run `checkdoc` and bytecompile tests. What more do we need? There isn't a standard style for Elisp the way there is for e.g. Go. And I wouldn't want these checks to frustrate new contributors.

gonewest818:
Well, elisp-lint can confirm `indent-region`, verify indents are following `indent-tabs-mode`, no line lengths exceeding `fill-column`, no trailing whitespace, and so on.  Individual checks can be disabled.

Like I said, I'm having to adjust some things in that codebase but what's left to do is the `indent-region` check.  Right now it's flagging compat macros like `if-let*` and `when-let*` because it wants to indent them like regular functions.

gonewest818:
Well, anyway, there are [4 pull requests pending in the elisp-lint project](https://github.com/nschum/elisp-lint/pulls) that boost the functionality enough to be configured and tested.

I ended up taking the `cljfmt` approach to indent rules, meaning I'm not trying to locate and evaluate all the macros (to get the `(declare (indent __))` forms automagically. Instead you need to configure an alist in `.dir.locals.el` with the indent rules you need.

It's fine with me if you collectively decide not to use that linter on `cider`. I'll probably adopt it for personal projects like `dimmer.el` though.

bbatsov:
I think it might have been easier for you to create a similar package. I'm fine with the approach.

> We already run checkdoc and bytecompile tests. What more do we need? There isn't a standard style for Elisp the way there is for e.g. Go. And I wouldn't want these checks to frustrate new contributors.

@xiongtx In practice ELisp mode defines how things are supposed to be indented, so I just want to make sure that code people submit was actually indented according to the mode and clean of redundant whitespace. Not a big deal, that's why I just imagined a tiny script that applies some clean to a buffer and compares this to what was committed.

gonewest818:
Update.

<img width="703" alt="screen shot 2018-02-09 at 2 22 09 pm" src="https://user-images.githubusercontent.com/2460918/36052968-b3d3456a-0da4-11e8-9919-4e4a01512728.png">

gonewest818:
[...]

PR #2209, opened 2018-02-18
[#2199] Implement elisp-lint as an additional CI formatting check

gonewest818:
As per #2199, this pulls in `elisp-lint` as a Cask dependency to check the source code for formatting issues.  The `make lint` target ~runs along with~ now replaces what used to be `make test-checks`. In Travis CI the linter runs during a "check" stage which is positioned prior to the unit tests.

As of the time of this commit ~the source code does not fully pass `make test-lint` but I'm submitting the pull request to show you how things are currently working.~ The most prevalent formatting issue is line widths exceeding the `fill-column`.

`elisp-lint` itself had been dormant up until recently. The package author agreed to pass the repo to me as a maintainer so we can iterate quickly on features if so desired.

-----------------

Not all of the following are relevant.

Thanks!

manual][2] extremely useful.*


.travis.yml
```diff
@@ -32,13 +32,28 @@ env:
...
-    - EMACS_BINARY=emacs-git-snapshot-travis MAKE_TEST=test-checks
+
+stages:
+  - check
+  - test
+
+jobs:
+  include:
+    # code quality "check" jobs
+    - stage: check
```
bbatsov: The byte-compilation warnings can also be qualified as such a check. 

What's `test-checks` right now? Isn't it just `checkdoc`? Maybe we can give those tasks better names. :-)

bbatsov:
Seems like a good start to me. I wonder if things like `checkdoc` and byte-compilation warnings can't be folded into eldoc - that would simplify the CI setup and would be reusable for pretty much any other Elisp project.

gonewest818:
I guess that sort of restructuring would be available no sooner than Emacs 27, as `eldoc` and `checkdoc` are both part of the distribution. However I don't think I have the time to get pulled into Emacs development, I want to hold the line at package maintenance. :)

.travis.yml
gonewest818: Indeed, `test-checks` is "just" `checkdoc` but you've set some configuration in `test/scripts/cider-checks.el` that `elisp-lint` doesn't yet support. So if I were to extend `elisp-lint` to support your configuration needs then `cider-checks.el` can be retired.

bbatsov:
I meant that probably `elint` can invoke them internally and just flag the problems reported by them.

gonewest818:
Ah, that was my misunderstanding.  Yes, elisp-lint already has checkdoc, and a rudimentary ability to byte-compile.

gonewest818:
I told myself "just wrap up the elisp-lint integration, and then move everything to CircleCI", but once again I'm blocked because Travis CI jobs are erroring for no good reason.

gonewest818:
Enabling the `fill-column` check in the linter will require a bunch of reformatting across the code base. The setting is 80 columns. I don't have any problem doing the work per se...  but want to make sure (a) core contributors definitely want long lines to be flagged by the linter, and (b) 80 columns is the desired limit.

xiongtx:
I don't think we should be enforcing the 80-column rule as a matter of principle; certainly not to the extent of failing a build. Sometimes symbol names are long; nothing to do about that.

In any case, Emacs Lisp is not even supposed to wrap at 80 columns--[usually 70 or 72](https://www.gnu.org/software/emacs/manual/html_node/eintr/fill_002dcolumn-Example.html).

gonewest818:
I should have mentioned all linting/validation is advisory only for now. Failures don't block the rest of the build.

The`fill-column` validator is completely disabled. The other validators (byte compile, checkdoc, check-declare, indentation, indent-tabs-mode, trailing whitespace) are running *and* passing on Emacs releases 24, 25, and pre-release 26.  On the emacs.git master branch there is one issue (#2216) arising from the bytecompile check.

bbatsov:
Sorry for the slow response. I think the only real issue we have is with `checkdoc` as it complains about docstrings longer than 80 characters (or maybe that's configurable and I didn't notice that). I don't think we should focus efforts on fixing the line length issues globally, although I always re-flow code longer than 80 chars when I encounter some. After all you have to be able to see two windows side by side fully on a smaller screen. 😉

gonewest818:
I think it's fine to leave the `checkdoc` validation as-is, after all it's widespread in the community and also a validation that MELPA asks for (and enforces? I don't recall.)  Oh, and all this code currently passes checkdoc already.

For the curious, it seems `checkdoc` enforces docstrings need to be < 80 characters wide *and* interesting to know the 80 limit is hard coded. https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/checkdoc.el#L1595-L1615

bbatsov:
> I think it's fine to leave the checkdoc validation as-is, after all it's widespread in the community and also a validation that MELPA asks for (and enforces? I don't recall.) Oh, and all this code currently passes checkdoc already.

MELPA requires other things - mostly proper package metadata. It doesn't care about checkdoc at all.

> For the curious, it seems checkdoc enforces docstrings need to be < 80 characters wide and interesting to know the 80 limit is hard coded. https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/checkdoc.el#L1595-L1615

Ouch. 😆

gonewest818:
> MELPA requires other things - mostly proper package metadata. It doesn't care about checkdoc at all.

Ok. I was referring to a bullet in the [MELPA contributor guidelines](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md#expediting-recipe-reviews): "Use checkdoc to make sure that your package follows the conventions for documentation strings."

gonewest818:
Rebased, fixed up a `check-declare` failure, & updated the github pull request template.

The thing I'm not 100% happy about is the `elisp-lint` indentation check. It's not smart enough to understand, e.g. inside `cl-flet` bindings should be indented like functions. You can get around this by setting indent metadata manually in `.dir-locals.el` but I wish it were better.

bbatsov:
It's OK not to get everything perfect from the first try. 😉

PR #2217, opened 2018-03-01
Correct variable name refactor `cider-cljs-repl-types`

dpsutton:
missed a variable name in the last commit. `cider--cljs-repl-types` => `cider-cljs-repl-types`


dpsutton:
and that travis ci docker image is enormous. i'm on emacs 27 and it fails byte comp due to the tramp changes. (but thanks to gonewest for the ability to crank it up and run the checks). I'm actually still waiting on the download to finish but will test locally after that

PR #2221, opened 2018-03-02
[Fix #2211] Scope .dir-locals.el settings to specific modes

### LINKED ISSUE #2211: Scope .dir-locals.el settings to specific modes
In a [pull request on another project](https://github.com/gonewest818/adafruit-wisdom.el/pull/1) @tarsius called me out on a `.dir-locals.el` file that set configuration like `fill-column` using `nil` as the mode name rather than scoping those settings to a specific mode. Citing, e.g. potential for adverse effects on things like `dired` and `magit`.

Just putting a note here because the same practice occurs in CIDER and clojure-mode, and I wanted that discussion to play out where it matters more (whereas my side project has only a few dozens of downloads).

gganley:
Locals only apply to intended modes.

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * [#2163](https://github.com/clojure-emacs/cider/issues/2163): Add `cider-browse-spec-regex`, and changed `cider-browse-spec-all` to use it.
 * [#2029](https://github.com/clojure-emacs/cider/pull/2154): Make cider-doc use cider-browse-spec functionality to print the spec part of the doc buffer
 * [#2151](https://github.com/clojure-emacs/cider/pull/2151): Improve formatting of spec in `cider-doc` buffer.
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * [#2163](https://github.com/clojure-emacs/cider/issues/2163): Add `cider-browse-spec-regex`, and changed `cider-browse-spec-all` to use it.
 * [#2029](https://github.com/clojure-emacs/cider/pull/2154): Make cider-doc use cider-browse-spec functionality to print the spec part of the doc buffer
 * [#2151](https://github.com/clojure-emacs/cider/pull/2151): Improve formatting of spec in `cider-doc` buffer.
+* [#2211](https://github.com/clojure-emacs/cider/issues/2211): Scope .dir-locals.el settings to specific modes.
```
bbatsov: That's not user visible (it affects just the development of CIDER), so it shouldn't be in the changelog.

PR #2222, opened 2018-03-02
Declare functions to resolve warnings

gganley:
Currently the only warnings that are being thrown by the tests are along the lines of "this function is not defined in the current scope". This squashes those warnings so `make test` runs squeaky clean.

-----------------

Thanks!

manual][2] extremely useful.*


gganley:
I don't think that small changes like these should be included in the change log but if I'm wrong I'd like to know.

bbatsov:
Nicely done! 👍

PR #2225, opened 2018-03-04
[2223] Clarify PR template

xiongtx:
@gganley @bbatsov


gganley:
This is better than my idea, also I'm going to adopt your branch naming conventions.

bbatsov:
I guess that's going to be changed when we integrate @gonewest818's work in #2209, but I'm fine with the change for now.

gonewest818:
I know I'm late to the party, but yes I agree this is fine for now and will probably need a slight adjustment in #2209.

PR #2228, opened 2018-03-05
Bugfix (minor): select-zombie-buffer was slurped into previous fn

josf:
Fixed by moving the function back up to the top-level


josf:
Looks like this is already fixed. Closing.

PR #2235, opened 2018-03-10
Make repl ignore the empty string rather than evaluating ""

gonewest818:
As the subject says. The bug was identified and commented on in https://github.com/clojure-emacs/cider/issues/1115#issuecomment-372051813

I'll push some additional changes to this branch to finish up the checklist below.

-----------------

Thanks!

manual][2] extremely useful.*


cider-repl.el
```diff
@@ -823,37 +823,38 @@ the symbol."
...
-     input
-     (cider-repl-handler (current-buffer))
-     (cider-current-ns)
-     (line-number-at-pos input-start)
-     (cider-column-number-at-pos input-start)
-     (unless (or (not cider-repl-use-pretty-printing)
-                 (string-match-p "\\`[ \t\r\n]*\\'" input))
-       (cider--nrepl-pprint-request-plist (cider--pretty-print-width))))))
+  (let ((input (cider-repl--current-input)))
+    (unless (string= input "")          ; don't evaluate an empty string
```
xiongtx: `string-empty-p`

bbatsov:
LGTM. This doesn't affect printing the next prompt, right? Just want to make sure we weren't doing the printing in an eval handler or something like this.

cider-repl.el
```diff
@@ -823,37 +823,38 @@ the symbol."
...
-     input
-     (cider-repl-handler (current-buffer))
-     (cider-current-ns)
-     (line-number-at-pos input-start)
-     (cider-column-number-at-pos input-start)
-     (unless (or (not cider-repl-use-pretty-printing)
-                 (string-match-p "\\`[ \t\r\n]*\\'" input))
-       (cider--nrepl-pprint-request-plist (cider--pretty-print-width))))))
+  (let ((input (cider-repl--current-input)))
+    (unless (string-empty-p input)      ; don't evaluate an empty string
```
bbatsov: Btw, I'm curious if this will work properly if you type a few spaces and then press Return in the REPL. Likely not, so probably we need a bit more robust check for no input.

bbatsov: I guess `string-blank-p` should do the trick.

gonewest818:
Ok, I adopted `string-blank-p` and added some logic as well.  No the behavior is as follows.

If the input string contains only whitespace (including newlines) then we clear the input without evaluating, print a fresh prompt, and we do NOT evaluate or add the blank input to the history.

Otherwise, if the input is not blank it is evaluated as usual.

bbatsov:
Nicely done! 👍

vspinu:
I think this change is a "usability" regression. The process might be killed or locked for whatever reason, but my returns go through as if nothing happened.  I am so used to sending RETs as a control for hang in all emacs REPLs that I find this new behavior rather disturbing.

What bug does this change fix more concretely?  The referenced comment doesn't shed light on the nature of the bug.

bbatsov:
I considered this a bug because when you press Ret on an empty line technically you haven't entered anything, but you ended up evaluating an empty string. For me the only case in which you should eval an empty string should be typing `""` in the REPL.

bbatsov:
> I think this change is a "usability" regression. The process might be killed or locked for whatever reason, but my returns go through as if nothing happened. I am so used to sending RETs as a control for hang in all emacs REPLs that I find this new behavior rather disturbing.

That's an interesting argument that I didn't think of, though. I generally just type something to test this, so it never came to my mind that someone might be testing their REPL in this manner.

vspinu:
>  technically you haven't entered anything, but you ended up evaluating an empty string

Shouldn't this be taken care by nREPL automatically? To my mind the eval op  should differentiate between `:code ""` and  `:code "\"\""`. 

> I generally just type something to test this

Never came to my mind that someone might be testing their REPL in this manner either :)

PR #2238, opened 2018-03-14
Allow predicates in plugin and middleware lists

raxod502:
This pull request is a follow-up to [discussion on the `clj-refactor.el` issue tracker](https://github.com/clojure-emacs/clj-refactor.el/pull/392#issuecomment-352399191). Basically, the problem is that `clj-refactor.el` wants to inject Leiningen plugins and nREPL middleware, but only for REPLs within a project context. (If the middleware is injected outside a project context, then errors result.) Unfortunately, CIDER provides no way to change `cider-jack-in-lein-plugins` or `cider-jack-in-nrepl-middlewares` without using advice. This pull request adds support for including arbitrary attributes in those variables using keyword arguments. In particular, the `:predicate` keyword allows for conditionally adding to those variables. I designed the interface with the idea of not breaking backwards compatibility and not making it hard to reverse the change later, if desired.

I haven't gone through the checklist yet, since I'd first like feedback on whether this is a reasonable approach to the problem.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -157,8 +157,10 @@ version from the CIDER package or library.")
   :package-version '(cider . "0.17.0"))
 
 (defcustom cider-clojure-cli-parameters
-  "-e '(require (quote cider-nrepl.main)) (cider-nrepl.main/init [\"cider.nrepl/cider-middleware\"])'"
-  "Params passed to clojure to start an nREPL server via `cider-jack-in'."
+  "-e '(require (quote cider-nrepl.main)) (cider-nrepl.main/init %s)'"
+  "Params passed to clojure to start an nREPL server via `cider-jack-in'.
+This is evaluated using `format', with the first argument being the Clojure
+vector of middleware variables as a string."
   :type 'string
   :group 'cider
   :safe #'stringp
@@ -346,7 +348,14 @@ Throws an error if PROJECT-TYPE is unknown.  Known types are
   (pcase project-type
     ("lein" cider-lein-parameters)
     ("boot" cider-boot-parameters)
-    ("clojure-cli" cider-clojure-cli-parameters)
+    ("clojure-cli" (format cider-clojure-cli-parameters
+                           (concat
+                            "["
+                            (mapconcat
+                             (apply-partially #'format "\"%s\"")
+                             (cider-jack-in-normalized-nrepl-middlewares)
+                             ", ")
+                            "]")))
     ("shadow-cljs" cider-shadow-cljs-parameters)
     ("gradle" cider-gradle-parameters)
     (_ (user-error "Unsupported project type `%s'" project-type))))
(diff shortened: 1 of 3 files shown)


cider.el
```diff
@@ -358,17 +358,57 @@ specifying the artifact ID, and the second element the version number."
                        (string :tag "Version"))))
 
 (defvar cider-jack-in-lein-plugins nil
-  "List of Leiningen plugins where elements are lists of artifact name and version.")
+  "List of Leiningen plugins to be injected at jack-in.
+Each element is a list of artifact name and version, followed optionally by
```
bbatsov: I'm reasonably certain without some examples most people would be really confused what are they supposed to put here.

cider.el
```diff
@@ -358,17 +358,57 @@ specifying the artifact ID, and the second element the version number."
...
+keyword arguments.  The only keyword argument currently accepted is
+`:predicate', which should be given a function that takes the list (name,
+version, and keyword arguments) and returns non-nil to indicate that the
+plugin should actually be injected.")
 (put 'cider-jack-in-lein-plugins 'risky-local-variable t)
 (cider-add-to-alist 'cider-jack-in-lein-plugins
                     "cider/cider-nrepl" (upcase cider-version))
 
+(defun cider-jack-in-lein-plugins ()
```
bbatsov: Why isn't this function used anywhere?

bbatsov:
Btw, if you want some project-specific settings isn't `.dir-locals.el` the simple way to do it?

cider.el
raxod502: @bbatsov Do you mean in terms of the format of the keyword arguments, or why you would want to use a predicate?

If the latter, I'm not sure that this is really meant to be an end-user feature. The only current application is in `clj-refactor.el`, after all. But I would not at all be opposed to adding examples.

cider.el
raxod502: @bbatsov Because I forgot to update the places where `cider-jack-in-lein-plugins` is used. If the approach seems reasonable, I'll go and fix that. The idea is just to take any place where the variable `cider-jack-in-lein-plugins` is referenced, and replace the variable with a function call.

raxod502:
@bbatsov Yes, but that doesn't work in the case of `clj-refactor.el`, since a solution using directory-local variables would mean that all users who wish to use `clj-refactor.el` are required to create a `.dir-locals.el` file specifying the internal middleware and Leiningen plugins used by `clj-refactor.el`, in every one of their projects. That would be a terrible user experience.

cider.el
bbatsov: I see. That makes sense, but it's exactly the kind of info that has to be documented. First I thought this might be something users would want to use in general or something.

cider.el
```diff
@@ -358,17 +358,57 @@ specifying the artifact ID, and the second element the version number."
...
+  "Return a normalized list of Leiningen plugins to be injected.
+See `cider-jack-in-lein-plugins' for the format, except that the list
+returned by this function does not include keyword arguments."
+  (mapcar
```
xiongtx: This seem like it can be made more readable (i.e. linear) w/ a `thread-last`.

cider.el
```diff
@@ -358,17 +358,57 @@ specifying the artifact ID, and the second element the version number."
...
 (put 'cider-jack-in-nrepl-middlewares 'risky-local-variable t)
 (add-to-list 'cider-jack-in-nrepl-middlewares "cider.nrepl/cider-middleware")
 
+(defun cider-jack-in-nrepl-middlewares ()
+  "Return a normalized list of middleware variable names.
+See `cider-jack-in-nrepl-middlewares' for the format, except that the list
+returned by this function only contains strings."
+  (mapcar
+   (lambda (spec)
+     (if (listp spec)
```
xiongtx: `car-safe`

raxod502: @xiongtx That would return nil if `spec` is not a cons cell. I want to return `spec`.

cider.el
raxod502: @xiongtx Agreed, fixed.

raxod502:
@bbatsov All comments are addressed and tests are added. Thanks for your feedback. Also, thanks for such a nice test suite. It was actually fun to use :)

cider.el
```diff
@@ -482,14 +525,14 @@ dependencies."
...
-             cider-jack-in-lein-plugins))
+             (cider-jack-in-normalized-lein-plugins)))
     ("boot" (cider-boot-jack-in-dependencies
              global-opts
              params
              (cider-add-clojure-dependencies-maybe
               cider-jack-in-dependencies)
-             cider-jack-in-lein-plugins
-             cider-jack-in-nrepl-middlewares))
+             (cider-jack-in-normalized-lein-plugins)
```
bbatsov: I guess this is some odd legacy, but why are we passing lein plugins to boot? //cc @benedekfazekas

bbatsov:
While you're at this - the hardcoded middleware here should be replaced with the new function you introduced https://github.com/clojure-emacs/cider/blob/master/cider.el#L159

bbatsov:
And this has to be rebased on top of `master`.

cider.el
benedekfazekas: because they are implemented as such (both cider-nrepl and refactor-nrepl). boot either has machinery to handle them or just handles them as ordinary deps.

bbatsov: Yeah, I know why it was done. :-) I'm just saying it looks really strange to be passing something named lein-plugins to boot. For something like tools.deps and shadow-cljs the middleware packages would be regular deps.

benedekfazekas: right. unfortunately lein needs them separately. this is unfortunately the smallest denominator in terms of interface. have not checked the details of this PR yet but I am sure this can be improved.

raxod502:
@bbatsov Regarding `cider-clojure-cli-parameters`, the most straight-forward solution seems to be to replace `[\"cider.nrepl/cider-middleware\"]` with `%s` and then have the variable run through `format` with `(cider-jack-in-normalized-nrepl-middlewares)` substituted in. Is that acceptable?

bbatsov:
Yeah, that should work. Just test that it's working by jacking-in outside a project in with `deps.edn` project. Basically this just needs to build a vector of strings.

cider.el
bbatsov: I guess the right solution would be just add some logic that removes those plugins from the deps if they are present there and add `cider-nrepl` to the list of default deps where it belong. Then we stop using this variable for non lein projects.

raxod502:
@bbatsov Fixed, seems to be working.

bbatsov:
👍

PR #2248, opened 2018-03-24
Add support for displaying various images

### LINKED ISSUE #1510: Display images of various format generated by clojure code in dedicated buffers or inline in the REPL
Bozidar mentioned that he already has this Idea, but I imagine that it is not high priority now. 

I am opening this issue to hear thoughts from other interested users. I would implement that if I wasn't total elisp beginner, but am willing to take on this if in the future (currently I only have tiny time for learning elisp, so that won't be soon! :)

arrdem:
Fixes #1510 

This is my second-pass changeset to implement support for embedding images in the REPL. It's probably got some sharp edges - my elisp isn't that good - but it definitely functions as advertised. I'd appreocate criticism on how the handlers should be made more configurable both on the nREPL side of the changeset (clojure-emacs/cider-nrepl#517) and on the emacs side.

### Demo

2018-03-23-220235_1188x1542_scrot

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -817,33 +817,47 @@ the symbol."
 
 (defcustom cider-repl-image-margin 10
   "Number or pair of numbers encoding either a symmetric margin for REPL
-  visualized images, or the `(x . y)' offset to use in placing the image.")
+  visualized images, or the `(x . y)' offset to use in placing the image."
+  :group 'cider-repl
+  :package-version '(cider . "0.17.0"))
 
 (defun cider-repl--image (image type datap)
   (create-image (if datap (base64-decode-string image) image) type datap
                 :margin cider-repl-image-margin))
 
-(setq cider-repl-content-type-handler-alist
-      '(("image/jpeg" .
-         (lambda (buffer image &optional show-prefix bol)
-           (cider-repl--display-image buffer
-                                      (cider-repl--image image 'jpeg nil)
-                                      show-prefix bol image)))
-        ("image/jpeg;base64" .
-         (lambda (buffer image &optional show-prefix bol)
-           (cider-repl--display-image buffer
-                                      (cider-repl--image image 'jpeg t)
-                                      show-prefix bol)))
-        ("image/png".
-         (lambda (buffer image &optional show-prefix bol)
-           (cider-repl--display-image buffer
-                                      (cider-repl--image image 'png nil)
-                                      show-prefix bol image)))
-        ("image/png;base64" .
-         (lambda (buffer image &optional show-prefix bol)
-           (cider-repl--display-image buffer
-                                      (cider-repl--image image 'png t)
-                                      show-prefix bol)))))
+(defun cider-repl-handle-jpeg (buffer image &optional show-prefix bol)
+  (cider-repl--display-image buffer
+                             (cider-repl--image image 'jpeg nil)
+                             show-prefix bol image))
+
+(defun cider-repl-handle-jpeg64 (buffer image &optional show-prefix bol)
+  (cider-repl--display-image buffer
+                             (cider-repl--image image 'jpeg t)
+                             show-prefix bol))
+
+(defun cider-repl-handle-png (buffer image &optional show-prefix bol)
+  (cider-repl--display-image buffer
+                             (cider-repl--image image 'png nil)
+                             show-prefix bol image))
+
+(defun cider-repl-handle-png64 (buffer image &optional show-prefix bol)
+  (cider-repl--display-image buffer
+                             (cider-repl--image image 'png t)
+                             show-prefix bol))
+
+(defcustom cider-repl-content-type-handler-alist
+  '(("image/jpeg" . #'cider-repl-handle-jpeg)
+    ("image/jpeg;base64" .  #'cider-repl-handle-jpeg64)
+    ("image/png". #'cider-repl-handle-png)
+    ("image/png;base64" . #'cider-repl-handle-png64))
+  "Association list from content-types to handlers for inserting nREPL
+responses of that content type  into the CIDER REPL buffer.
+
+Handlers must be functions of two required and two optional arguments -
+being the REPL buffer to insert into, the value of the given content type
+as a raw string, the REPL's show prefix as any and an end-of-line flag."
+  :group 'cider-repl
+  :package-version '(cider . "0.17.0"))
 
 (defun cider-repl-handler (buffer)
   "Make an nREPL evaluation handler for the REPL BUFFER."
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -815,18 +815,18 @@ the symbol."
     (cider-repl--show-maximum-output)))
 
 (setq cider-repl-content-type-handler-alist
-  '(("image/jpeg" .
-     (lambda (u v s b)
-       (cider-repl--display-image u (create-image v 'jpeg) s b)))
-    ("image/jpeg;base64" .
-     (lambda (u v s b)
-       (cider-repl--display-image u (create-image (base64-decode-string v) 'jpeg 't) s b)))
-    ("image/png".
-     (lambda (u v s b)
-       (cider-repl--display-image u (create-image v 'png) s b)))
-    ("image/png;base64" .
-     (lambda (u v s b)
-       (cider-repl--display-image u (create-image (base64-decode-string v) 'png 't) s v)))))
+      '(("image/jpeg" .
+         (lambda (u v s b)
+           (cider-repl--display-image u (create-image v 'jpeg) s b)))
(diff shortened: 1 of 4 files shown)


cider-repl.el
```diff
@@ -799,23 +799,79 @@ the symbol."
...
+                                     (when (and win force-prompt)
+                                       (with-selected-window win
+                                         (set-window-point win cider-repl-input-start-mark))
+                                       (cider-repl--show-maximum-output))))
+                                 nrepl-err-handler
                                  (lambda (buffer pprint-out)
                                    (cider-repl-emit-result buffer pprint-out (not after-first-result-chunk))
+                                   (setq after-first-result-chunk t))
+                                 (lambda (buffer value content-type)
+                                   (if-let ((handler (cdr (assoc content-type cider-repl-content-type-handler-alist))))
```
xiongtx: `if-let` is obsolete. Emacs >=26.1 uses `if-let*`, as does CIDER (via `cider-compat.el`).

nrepl-client.el
```diff
@@ -764,7 +764,8 @@ to the REPL."
 (defun nrepl-make-response-handler (buffer value-handler stdout-handler
                                            stderr-handler done-handler
                                            &optional eval-error-handler
-                                           pprint-out-handler)
+                                           pprint-out-handler
+                                           content-type-handler)
```
xiongtx: Docstring should mention this and say what it does.

cider-repl.el
```diff
@@ -799,23 +799,79 @@ the symbol."
...
+
+(defcustom cider-repl-image-margin 10
+  "Number or pair of numbers encoding either a symmetric margin for REPL
+  visualized images, or the `(x . y)' offset to use in placing the image.")
+
+(defun cider-repl--image (image type datap)
+  (create-image (if datap (base64-decode-string image) image) type datap
+                :margin cider-repl-image-margin))
+
+(setq cider-repl-content-type-handler-alist
```
xiongtx: `defvar`?

nrepl-client.el
```diff
@@ -781,12 +782,14 @@ EVAL-ERROR-HANDLER is nil, the default `nrepl-err-handler' is used.  If any
...
   (lambda (response)
-    (nrepl-dbind-response response (value ns out err status id pprint-out)
+    (nrepl-dbind-response response (content-type content value ns out err status id pprint-out)
       (when (buffer-live-p buffer)
         (with-current-buffer buffer
           (when (and ns (not (derived-mode-p 'clojure-mode)))
             (cider-set-buffer-ns ns))))
-      (cond (value
+      (cond ((and content content-type content-type-handler)
+             (funcall content-type-handler buffer content content-type))
```
xiongtx: `content-type-handler` is called w/ only 3 args, while the [functions](https://github.com/clojure-emacs/cider/pull/2248/files#diff-25e8a6dee6c1a1b53de3d45cead245e3R827) in `cider-repl-content-type-handler-alist` can take 4.

cider-repl.el
```diff
@@ -799,24 +799,83 @@ the symbol."
...
+         (when (and win force-prompt)
+           (with-selected-window win
+             (set-window-point win cider-repl-input-start-mark))
+           (cider-repl--show-maximum-output))))
+     nrepl-err-handler
+     (lambda (buffer pprint-out)
+       (cider-repl-emit-result buffer pprint-out (not after-first-result-chunk))
+       (setq after-first-result-chunk t))
+     (lambda (buffer value content-type)
+       (if-let ((handler (cdr (assoc content-type
```
xiongtx: `if-let` -> `if-let*`

nrepl-client.el
arrdem: Done.

nrepl-client.el
arrdem: Right. What happens here is that the `content-type-handler` function is called as `(buff content content-type)`, it does its own internal dispatch, chooses a handler implementatiion (if any) and defers to that handler.

Because of the current mechanics of CIDER's various insertion functions, the arguments `&optional show-prefix bol` seem to be required in order to ensure that the user's prefix configuration is respected. The precise behavior of the `bol` flag is ... not obvious so I just retained it blindly.

I'd like to see the way that the REPL handles marks cleaned up to maintain one set of marks per eval requests which would I think simplify some of this logic and fix some interesting behavior where output / errors from threads forked by previous evals can become interleaved with the latest output and its value(s).

cider-repl.el
```diff
@@ -799,24 +800,97 @@ the symbol."
...
+  (cider-repl--display-image buffer
+                             (cider-repl--image image 'png t)
+                             show-prefix bol))
+
+(defcustom cider-repl-content-type-handler-alist
+  '(("image/jpeg" . #'cider-repl-handle-jpeg)
+    ("image/jpeg;base64" .  #'cider-repl-handle-jpeg64)
+    ("image/png". #'cider-repl-handle-png)
+    ("image/png;base64" . #'cider-repl-handle-png64))
+  "Association list from content-types to handlers for inserting nREPL
```
xiongtx: Just `alist` is fine.

stardiviner:
@arrdem in the screenshot, seems it is `gnuplot` to convert image. I'm watching this nREPL support image content-type topic. Because I want to add this feature to [ob-clojure](https://code.orgmode.org/bzg/org-mode/src/master/lisp/ob-clojure.el) and [ob-clojure-literate](https://code.orgmode.org/bzg/org-mode/src/master/contrib/lisp/ob-clojure-literate.el)
But after reading you two post:
  - [ ] https://github.com/clojure-emacs/cider/issues/1510#issuecomment-373988901
  - [ ] https://github.com/clojure-emacs/cider/pull/2248
Can't fully understand (not good on Clojure still)
Can you check out `ob-clojure.el` and `ob-clojure-literate.el`, and point me out how can I use this new feature? And I want to understand the theory how does it works.

nrepl-client.el
bbatsov: Yeah, that'd be nice. Unfortunately there's a ton of legacy in the REPL code - much of it came straight from SLIME and was barely updated afterwards.

nrepl-client.el
bbatsov: I love it how ridiculous this function is at that point. :D It's certainly one of the prime candidates for some massive refactorings.

cider-repl.el
```diff
@@ -799,24 +800,120 @@ the symbol."
...
+          (when (and bol (not (bolp)))
+            (insert-before-markers "\n"))
+          (when show-prefix
+            (insert-before-markers (propertize cider-repl-result-prefix 'font-lock-face 'font-lock-comment-face)))
+          (insert-image image string)
+          (set-marker cider-repl-input-start-mark (point) buffer)
+          (set-marker cider-repl-prompt-start-mark (point) buffer))))
+    (cider-repl--show-maximum-output)))
+
+(defcustom cider-repl-image-margin 10
```
bbatsov: Probably we should also add some defcustom to turn this functionality off in case someone doesn't like it, or the initial implementation causes problems in some cases.

bbatsov:
Overall things look good to me (apart from the failing build). 

> I'd appreocate criticism on how the handlers should be made more configurable both on the nREPL side of the changeset (clojure-emacs/cider-nrepl#517) and on the emacs side.

I guess one way to make things configurable would be set the configuration on the Emacs side and send it to the middleware as part of some "init" message when the connection is being established. 

I dislike in the Emacs handlers that even though they don't have to support text, they current support params that make sense only for text output. 

I've got two questions:

1) What happens now if some expression produces some output and its result is an image. 
2) What happens if the resulting image can't fit in the REPL buffer (e.g. it's too wide)?

I'd be really nice to add support for this for interactive evaluation as well, which probably means some of the image handling code, shouldn't live in the REPL source code. E.g. you can show resulting images in an overlay or a dedicated result buffer, but obviously we can't show them in the minibuffer. :-)

Btw, I'm wondering about the special way you've treated `value` - is this really necessary? I can image we can both print the result and the image, although I guess that's a matter of personal preferences.

cider-repl.el
stardiviner: I agree with this. Should have an option to disable it to avoid early stage problems.

stardiviner:
About image too wide problem, I guess Emacs handle this. Org-mode has similar function to specify image with. I don't know how it is implemented. Even more, the image width can be smarter based on current Emacs window width.
Use code like this to get window width `(window-width (selected-window) t)`

bbatsov:
One more thing - does clearing output in the REPL work properly for images? This certainly needs to be tested.

zcaudate:
This is really cool. Is is possible to get it working with `cider-eval-last-sexp`/`cider-eval-last-sexp-to-repl`?

PR #2258, opened 2018-03-31
Add test selector support

benedekfazekas:
Test all and test ns understand leiningen style test selectors, eg
metadata markers on `deftest`s. No support for namespace level metada
markers for now.

If test selector is specified only the test(s) run that have that
selector in their metadata.

If all loaded tests or project tests are invoked with prefix, ask for
test selector in the minibuffer. A command is available to run ns
tests with a selector.

-----------------

Thanks!

manual][2] extremely useful.*


cider-test.el
```diff
@@ -611,7 +620,7 @@ This uses the Leiningen convention of appending '-test' to the namespace name."
 (declare-function cider-emit-interactive-eval-output "cider-interaction")
 (declare-function cider-emit-interactive-eval-err-output "cider-interaction")
 
-(defun cider-test-execute (ns &optional tests silent)
+(defun cider-test-execute (ns &optional tests silent test-selector)
```
bbatsov: You'll have to update all docstrings where you've added the new param.

doc/running_tests.md
```diff
@@ -13,12 +13,18 @@ test namespace inference logic (e.g. you have some tests in the implementation
 namespace that were defined with `clojure.test/with-test`)
 - in such cases you should use <kbd>C-u C-c C-t C-n</kbd>, which will simply run
 whatever tests are present in the currently visited/active namespace.
+You can also run <kbd>C-c C-t C-s</kbd> to run a subset of the tests defined in
```
bbatsov: "can also press to run" sounds better IMO.

I'd show some example of a test selector here, as I'm reasonably certain half the people have no idea what this is.

cider-test.el
```diff
@@ -677,17 +688,25 @@ If SILENT is non-nil, suppress all messages other then test results."
...
 
-(defun cider-test-run-project-tests ()
+(defun cider-test-run-project-tests (test-selector)
   "Run all tests defined in all project namespaces, loading these as needed."
-  (interactive)
-  (cider-test-execute :project))
+  (interactive "P")
+  (cider-test-execute :project nil nil test-selector))
+
+(defun cider-test-run-ns-tests-with-selector (suppress-inference)
```
bbatsov: I don't think we need a dedicated command for this. The prefix versions of the existing commands are enough IMO.

cider-test.el
```diff
@@ -677,17 +688,25 @@ If SILENT is non-nil, suppress all messages other then test results."
           (message "No prior failures to retest")))
     (message "No prior results to retest")))
 
-(defun cider-test-run-loaded-tests ()
+(defun cider-test-run-loaded-tests (test-selector)
```
bbatsov: I'd name the prefix arg `prompt-for-selector` or something like this, which is more descriptive. Otherwise I'd assume we have to pass a selector here.

cider-test.el
```diff
@@ -621,51 +630,53 @@ report is optionally displayed.  When test failures/errors occur, their sources
...
-                     (when (get-buffer cider-test-report-buffer)
-                       (with-current-buffer cider-test-report-buffer
-                         (let ((inhibit-read-only t))
-                           (erase-buffer)))
-                       (cider-test-render-report
-                        cider-test-report-buffer
-                        summary results))))))))
-      conn))
-   :clj))
+  (let ((selector (when test-selector (cider-read-from-minibuffer "Test selector (not as keyword): "))))
```
bbatsov: Not as keyword? :-) Maybe we can just say "As string".

cider-test.el
```diff
@@ -202,8 +208,11 @@ Add to this list to have CIDER recognize additional test defining macros."
         ["Rerun current test" cider-test-run-test]
         ["Rerun failed/erring tests" cider-test-rerun-failed-tests]
         ["Run all ns tests" cider-test-run-ns-tests]
+        ["Run all ns tests with selector" cider-test-run-ns-tests-with-selector]
         ["Run all loaded tests" cider-test-run-loaded-tests]
+        ["Run all loaded tests with selector" (apply-partially cider-test-run-loaded-tests 't)]
```
bbatsov: Use `'prompt-for-selector` here instead of `t`.

bbatsov:
Nice addition! I've left a bit of small feedback here and there.

benedekfazekas:
thanks for the review @bbatsov will update the PRs shortly

cider-test.el
benedekfazekas: I thought this feature made sense for running tests for an particular ns too. This can't be pulled off with a prefix argument as `cider-test-run-ns-tests` understands a prefix argument already. Any nicer way to do this or you think run-ns-tests does not need this feature?

bbatsov: Ah, I misread what this does. OK, I guess it makes some sense to have it.

benedekfazekas:
updated both PRs

bbatsov:
@benedekfazekas Looks ok to me, apart from the failing build:

```cider-repl.el:1412:1:Warning (check-declare): said
    ‘cider-test-run-loaded-tests’ was defined in
    cider-test-run-loaded-tests.el: file not found
cider-repl.el:1413:1:Warning (check-declare): said
    ‘cider-test-run-project-tests’ was defined in
    cider-test-run-project-tests.el: file not found
Checking cider-repl.el...2 problems found
Check-declare failed```

benedekfazekas:
fixed

bbatsov:
👍

benedekfazekas:
cheers

PR #2259, opened 2018-03-31
Adding cider-nrepl-port-search-paths defcustom

dantheobserver:
This defcustom is useful for cases when you want to connect to a local repl but
are using boot repl spawned from a location that is not a project folder.
The boot process does not display the start directory so we cannot search for
the nrepl-port file from the process. This defcustom allows you to list common
places where you would spawn your boot repl and add that to the search path when
looking for nrepl-port files to show in.

-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -473,7 +473,9 @@ and try to associate the created connection with this project automatically.
 ## 0.9.0 (2015-06-16)
 
 ### New features
-
+* New defcustom `cider-nrepl-port-search-paths`.
+  Used when `cider-connect` does a lookup of local ports. Mainly should be set when
+  using boot Repl server from some directory other than a specific project's.
 * [#1109](https://github.com/clojure-emacs/cider/issues/1109): New defcustom `cider-auto-mode`.
 On by default, when `nil` don't automatically enable `cider-mode` in all Clojure buffers.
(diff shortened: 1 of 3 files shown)


bbatsov:
> This defcustom is useful for cases when you want to connect to a local repl but
are using boot repl spawned from a location that is not a project folder.

Wouldn't you have same problem with lein as well? I'm also wondering how people are going to understand what this/make use of it.

dantheobserver:
@bbatsov the lein process has attached to it the directory it was spawned from.  The boot process is very bare.  I'm not familiar with a way to annotate a boot process but if that's possible perhaps that can be part of the instruction and make this obsolete.  I tried adding boot to the lookup but saw that the process itself is fairy bare.  

There is a way to do it, involving capturing the process PID and doing a pwdx <PID>, but I think that would be overkill.  I supposed this may be too.

bbatsov:
That's an interesting point. Frankly, I've never liked the approach of extracting info from `ps`, as obviously that's not exactly portable. Unfortunately I'm short on great alternatives as well. :-(

dantheobserver:
So I slept on it and agree, I think this option is really just for me as I applied it locally. I wanted to propose a non-invasive solution since I can't guarantee how boot or lein (or clj or whatever comes down the line) handles the .nrepl-port file if at all.  I looked into doing something with the boot repl task to output the port to the home folder and check for that by default, but didn't get far.  I'd like to have a one-size-fits-all solution, but I guess I need some more research to do so.

bbatsov:
@dantheobserver Perhaps boot can be updated to display similar info to the one displayed by lein? I assume that's nothing hard.

PR #2263, opened 2018-04-09
Fix `cider-repl-handle-shortcut` not handling plain lambdas

benedekfazekas:
The function expects values in `cider-repl-shortcuts` to be commands
(interactively callable functions). For some cases (currently
`browse-ns`, `test-all-with-selector` and
`test-project-with-selector`) this is not true that causes the `,`
command to fail. For these cases plain lambdas are
used to call certain otherwise interactive functions with predefined parameters.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1417,7 +1417,7 @@ constructs."
 (cider-repl-add-shortcut "clear-help-banner" #'cider-repl-clear-help-banner)
 (cider-repl-add-shortcut "ns" #'cider-repl-set-ns)
 (cider-repl-add-shortcut "toggle-pretty" #'cider-repl-toggle-pretty-printing)
-(cider-repl-add-shortcut "browse-ns" (lambda () (cider-browse-ns (cider-current-ns))))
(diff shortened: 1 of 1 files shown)


benedekfazekas:
yup all those are commands. but the things for them in `cider-repl-shortcuts` are not commands, they are just plain lambdas calling the commands. for example for `browse-ns` `(lambda () (cider-browse-ns (cider-current-ns)))`. sorry if I did not manage to explain this clearly above

benedekfazekas:
btw just try to call `browse-ns` from your repl using `,` in latest cider. you should see an error

bbatsov:
Hmm, shouldn't all the things we call from there be commands? At least this was the initial intention of those shortcuts - faster ways to run commands, not random functions.

benedekfazekas:
those lambdas are essentially running commands with preset params. like
partials. it makes sense this way I think. alternatively you could define
the lambdas interactive but that would look weird to me...

bbatsov:
I get your point, but you can make the same point about lambdas as keybindings in Emacs - they don't work unless they are interactive lambdas. It's not a big deal for me, but it seems more reasonable to stick to the established conventions.

bbatsov:
Thanks!

PR #2276, opened 2018-04-26
Wait for shadow-cljs client to connect after watch

OscarMarshall:
When jacking in to a ClojureScript REPL using Shadow CLJS, it is currently impossible to make it all the way to the CLJS REPL. The REPL init code looks like this:

```clojure
(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch :%s) (shadow/nrepl-select :%s))
```

`shadow/nrepl-select` currently always returns `[:no-client :app "Make sure your JS environment has loaded your compiled ClojureScript code."]` because there's no time to connect a client.

This pull request adds code that blocks until a client connects.

```
Arch Linux (Linux version: 4.16.3-1)
GNU Emacs 25.3.1
CIDER 0.17.0snapshot (package: 20180426.440), nREPL 0.2.13
Clojure 1.9.0, Java 1.8.0_172
```

-----------------

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -662,9 +662,9 @@ Generally you should not disable this unless you run into some faulty check."
 
 We have to prompt the user to select a build, that's why
 this is a command, not just a string."
-  (let ((form "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch :%s) (shadow/nrepl-select :%s))")
+  (let ((form "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch :%s) (while (not (shadow/repl-client-connected? (shadow/get-worker :%s))) (Thread/sleep 1000)) (shadow/nrepl-select :%s))")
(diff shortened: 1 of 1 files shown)


bbatsov:
> because there's no time to connect a client

Btw, what's a "client" in this context?

thheller:
Hmm. I added this because I wanted things to be less confusing when using the REPL. Typically people ran `nrepl-select` and then got errors when trying to eval something due to the Browser not having loaded their JS yet. So I thought I'd warn on `nrepl-select` instead. Which is bad when trying to do things like the above.

I will probably revert the change since the new warning is equally confusing and prevents other things from working properly.

> Btw, what's a "client" in this context?

Client refers to whichever runtime the intended target is. So for `:browser` builds that would be a Browser loading the generated JS. For `:node-script` that would be the `node` process and so on.

thheller:
Fixed in `shadow-cljs@2.3.10`. Sorry about the confusion.

PR #2290, opened 2018-05-12
Add more refloc regexps and improve relative file detection

vspinu:
Adding long promised tests on this occasion. Couldn't figure out the linter locally, but should be fine.

#2043

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1258,7 +1258,12 @@ regexes from `cider-locref-regexp-alist' to infer locations at point."
                       (or (cider-sync-request:ns-path var)
                           (nrepl-dict-get (cider-sync-request:info var) "file")))
                     ;; when not found, return the file detected by regexp
-                    (plist-get loc :file))))
+                    (when-let* ((file (plist-get loc :file)))
+                      (if (file-name-absolute-p file)
+                          file
+                        ;; when not absolute, expand within the current project
+                        (when-let* ((proj (clojure-project-dir)))
+                          (expand-file-name file proj)))))))
         (if file
             (cider--jump-to-loc-from-info (nrepl-dict "file" file "line" line) t)
           (error "No source location for %s" var)))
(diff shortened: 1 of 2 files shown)


bbatsov:
Nicely done! 👍

cider-repl.el
```diff
@@ -1198,7 +1198,9 @@ namespace to switch to."
   '((stdout-stacktrace "[ \t]\\(at \\([^$(]+\\).*(\\([^:()]+\\):\\([0-9]+\\))\\)" 1 2 3 4)
     (aviso-stacktrace  "^[ \t]*\\(\\([^$/ \t]+\\).*? +\\([^:]+\\): +\\([0-9]+\\)\\)" 1 2 3 4)
     (print-stacktrace  "\\[\\([^][$ \t]+\\).* +\\([^ \t]+\\) +\\([0-9]+\\)\\]" 0 1 2 3)
-    (timbre-log        "\\(TRACE\\|INFO\\|DEBUG\\|WARN\\|ERROR\\) +\\(\\[\\([^:]+\\):\\([0-9]+\\)\\]\\)" 2 3 nil 4))
+    (timbre-log        "\\(TRACE\\|INFO\\|DEBUG\\|WARN\\|ERROR\\) +\\(\\[\\([^:]+\\):\\([0-9]+\\)\\]\\)" 2 3 nil 4)
+    (cljs-message      "at line \\([0-9]+\\) +\\(.*\\)$" 0 nil 2 1)
```
bbatsov: Are those messages consistent, btw? I think @bhauman recently mentioned that their format is not exactly uniform, but I guess handling some beats handling none. :-)

vspinu: I don't know. I have seen "warning" messages so far and all seem to be consistent. But, I assumed there should be others, this is why `cljs-message` and not `cljs-warning`. We just have too keep an eye and update those when new patterns are spotted.

bbatsov: 👍

PR #2293, opened 2018-05-17
Fix eldoc on repl image

### LINKED ISSUE #2286: cider-eldoc try to display doc when point on weird symbols like on image content-type
When use cider display image content-type. I move point backward, then CIDER REPL can't evaluate sexp anymore.

Steps to reproduce:
Toggle nrepl log messaging.
Disable eldoc-mode in CIDER REPL buffer, then display image, then movie point backward around, on the image etc. All things fine, no weird symbol in nrepl log.
Enable eldoc-mode in CIDER REPL buffer, then move point backward around, on the image, then nrepl log displays those weird symbols. then input (+ 1 1) does not work any more. even though I disable eldoc-mode again.
Here is the nrepl log:
```
(-->
  id         "25"
  op         "eldoc"
  session    "07c655d4-444b-43c4-a695-43a9da90104e"
  time-stamp "2018-05-08 11:42:29.972220486"
  ns         "user"
  symbol     "\266J\206\306\251\347\355"
)
(-->
  id         "26"
  op         "eldoc"
  time-stamp "2018-05-08 11:42:35.941183374"
  ns         "user"
  symbol     "\266J\206\306\251\347\355"
)
(-->
  id           "27"
  op           "eval"
  time-stamp   "2018-05-08 11:42:41.580182208"
  code         "(print \"hi\")"
  column       7
  content-type "true"
  file         "*cider-repl localhost*"
  line         335
  ns           "user"
)
```

This weird symbol is from display image which eldoc try to get doc for them.

Guess might `cider-eldoc` need to filter on non-standard symbols?
With reference as background to original issue: https://github.com/clojure-emacs/cider/issues/2277

stardiviner:
Fix #2286 
-----------------

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -869,7 +869,7 @@ Handles an external-body TYPE by issuing a slurp request to fetch the content."
   (if-let* ((args        (cadr type))
             (access-type (nrepl-dict-get args "access-type")))
       (nrepl-send-request
-       (list "op" "slurp" "url" (nrepl-dict-get args "URL"))
+       (list "op" "slurp" "url" (nrepl-dict-get args access-type))
        (cider-repl-handler buffer)
        (cider-current-connection)))
   nil)
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -854,14 +854,14 @@ Returns an image instance with a margin per `cider-repl-image-margin'."
 Part of the default `cider-repl-content-type-handler-alist'."
   (cider-repl--display-image buffer
                              (cider-repl--image image 'jpeg t)
-                             show-prefix bol image))
+                             show-prefix bol " "))
 
 (defun cider-repl-handle-png (_type buffer image &optional show-prefix bol)
   "A handler for inserting a png IMAGE into a repl BUFFER.
 Part of the default `cider-repl-content-type-handler-alist'."
   (cider-repl--display-image buffer
                              (cider-repl--image image 'png t)
-                             show-prefix bol image))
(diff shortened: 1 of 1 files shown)


bbatsov:
@arrdem Please, take a look as well.

cider-repl.el
```diff
@@ -854,22 +854,22 @@ Returns an image instance with a margin per `cider-repl-image-margin'."
 Part of the default `cider-repl-content-type-handler-alist'."
   (cider-repl--display-image buffer
                              (cider-repl--image image 'jpeg t)
-                             show-prefix bol image))
+                             show-prefix bol " "))
```
bbatsov: What's the final param here? I'm too lazy to check the definition now.

stardiviner: Before change to string `" "`, the original parameter `image` (it is image data instead of string) is passed to `cider-repl--display-image` as parameter `string`, this param `string` is passed to elisp function `(insert-image image string)`. That why I get weird symbols https://github.com/clojure-emacs/cider/issues/2286#issuecomment-387662375

bbatsov: I see this string is " " by default, so I guess we can just omit it.

bbatsov: (unless somewhere we don't use a different value)

stardiviner: You're right, It can be omitted this optional argument.

arrdem: We don't

Issue #2295, opened 2018-05-19
Update docs to indicate cider-jack-in support for Clojure CLI


cbowdon:
I was (very pleasantly) suprised when `cider-jack-in` worked on a deps.edn project, since the current doc/up_and_running.md says that Lein, Boot or Gradle are required. Is clj support missing from the docs because it's incomplete? Otherwise would you like a PR for this?

bbatsov:
It's not incomplete, I just forgot to document it. :D I wouldn't mind a PR if you want to work on this. The magic code is most here https://github.com/clojure-emacs/cider-nrepl/blob/b0649ac8bcf1faa686ce9954ed4879622fdf58c3/src/cider_nrepl/main.clj

PR #2299, opened 2018-05-25
Indicate support for Clojure CLI tools in docs

cbowdon:
This is a quick contribution for #2295 . The only changes are to markdown files in `./doc` so I think most of the checks below are not applicable. I have not updated the changelog because looking through it I didn't see any entries just about documentation - hope that's ok.

-----------------


bbatsov:
Thanks!

PR #2304, opened 2018-05-31
typo in cider-connections function

fourtytoo:
oops


bbatsov:
That's more than a typo, that's a real bug! 👍

PR #2306, opened 2018-06-01
[Fix #2294] simplify stacktrace filter mechanism

### LINKED ISSUE #2294: Provide a better interface to set project-only filter for stacktrace mode
Currently we can't use positive filters on `cider-stacktrace-default-filters`. Options like `project` doesn't work here.
Also the documentation on `cider-stacktrace-default-filters` could be improved. I had to look more than I expected for the acceptable values for the variable.

dpsutton:
previously there were two separate lists of positive and negative
filters and complicated logic spread throughout of how these
interacted. Now there is a single list that you can toggle flags on or
off and their interaction is dictated inside of a single function
`cider-stacktrace-apply-filters`. Since their interaction is here we
no longer need to remember previous flags as they are just toggled on
and off.

Positive filters are 'all or 'project. These dictate that all frames
should be shown or only any frame from a project namespace should be
shown, despite other filters. Also, these two cannot exist at the same
time, so we need to check which is most "recent". Since if 'project is
selected, it hides all but project frames and if 'all is selected it
must show all. We do what was most recently selected in this case
then.

So the frame filtering is quite simply a list of tags you want to
filter by: '(java clj tooling repl dup project all). These are toggled
on and off as desired.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -89,7 +89,6 @@ cyclical data structures."
 
 (defvar-local cider-stacktrace-hidden-frame-count 0)
 (defvar-local cider-stacktrace-filters nil)
-(defvar-local cider-stacktrace-prior-filters nil)
 (defvar-local cider-stacktrace-cause-visibility nil)
 (defvar-local cider-stacktrace-positive-filters nil)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
I guess you also have to update the manual to reflect this change.

cider-stacktrace.el
```diff
@@ -470,41 +490,29 @@ When it reaches 3, it wraps to 0."
...
-      (progn
-        (setq-local cider-stacktrace-prior-filters cider-stacktrace-filters)
-        (setq-local cider-stacktrace-filters cider-stacktrace--all-negative-filters)
-        (setq-local cider-stacktrace-positive-filters '(project)))
-    (progn
-      (setq-local cider-stacktrace-filters cider-stacktrace-prior-filters)
-      (setq-local cider-stacktrace-positive-filters nil)))
-  (cider-stacktrace-apply-filters cider-stacktrace-filters
-                                  cider-stacktrace-positive-filters))
+  (setq cider-stacktrace-filters
```
bbatsov: The code here is mostly a duplication of the logic for `'all` so probably some small helper function should be extracted.

dpsutton:
updated and entry in the manual.

bbatsov:
Thanks!

PR #2307, opened 2018-06-05
Use a better error when a cljs repl form cannot be found

arichiardi:
Without this fix cider throws an error but it is not a very clear one.

-----------------

  - This fails with exit = 2 but I don't think it due to my change.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -755,14 +755,15 @@ you're working on."
 
 (defun cider-cljs-repl-form (repl-type)
   "Get the cljs REPL form for REPL-TYPE."
-  (let ((repl-form (cadr (seq-find
-                          (lambda (entry)
-                            (eq (car entry) repl-type))
-                          cider-cljs-repl-types))))
-    ;; repl-form can be either a string or a function producing a string
-    (if (symbolp repl-form)
-        (funcall repl-form)
-      repl-form)))
+  (if-let* ((repl-form (cadr (seq-find
+                              (lambda (entry)
+                                (eq (car entry) repl-type))
+                              cider-cljs-repl-types))))
+      ;; repl-form can be either a string or a function producing a string
+      (if (symbolp repl-form)
+          (funcall repl-form)
+        repl-form)
+    (user-error "No ClojureScript REPL type %s found.  Please make sure that `cider-cljs-repl-types' has an entry for it" repl-type)))
 
 (defun cider-verify-cljs-repl-requirements (repl-type)
   "Verify that the requirements for REPL-TYPE are met."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugs fixed
 
 * [#2286](https://github.com/clojure-emacs/cider/issues/2286): Fix eldoc issue with images in the REPL.
+* [#2307](https://github.com/clojure-emacs/cider/pull/2307): Use a better error when a cljs repl form cannot be found.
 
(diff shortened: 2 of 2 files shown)


bbatsov:
Hmm, I'm not sure that suppressing the error is a good idea. Probably it should just be made more meaningful using `user-error`. After your changes simply nothing's going to happen if people mistyped/misconfigured something, which is going to be pretty confusing.

arichiardi:
Uhm True, will try a better fix

cider.el
```diff
@@ -755,7 +755,7 @@ you're working on."
 
 (defun cider-cljs-repl-form (repl-type)
   "Get the cljs REPL form for REPL-TYPE."
-  (let ((repl-form (cadr (seq-find
+  (when-let ((repl-form (cadr (seq-find
```
bbatsov: I guess you can just use `if-let` here and show an `user-error` in the `else` branch. Something like `No ClojureScript REPL type x found. Please check ...`

arichiardi:
@bbatsov  done ;)

cider.el
```diff
@@ -755,14 +755,15 @@ you're working on."
...
-      repl-form)))
+  (if-let ((repl-form (cadr (seq-find
+                             (lambda (entry)
+                               (eq (car entry) repl-type))
+                             cider-cljs-repl-types))))
+      ;; repl-form can be either a string or a function producing a string
+      (if (symbolp repl-form)
+          (funcall repl-form)
+        repl-form)
+    (user-error "No ClojureScript REPL type x found.  Please make sure that `cider-cljs-repl-types` has an entry for `cljs-repl-type`")))
```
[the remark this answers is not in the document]
arichiardi: Ah got you! Let me google Elisp interpolations :D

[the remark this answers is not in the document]
arichiardi: oh I thought Elisp had a construct for interpolations :D

cider.el
```diff
@@ -755,14 +755,15 @@ you're working on."
...
   "Get the cljs REPL form for REPL-TYPE."
-  (let ((repl-form (cadr (seq-find
-                          (lambda (entry)
-                            (eq (car entry) repl-type))
-                          cider-cljs-repl-types))))
-    ;; repl-form can be either a string or a function producing a string
-    (if (symbolp repl-form)
-        (funcall repl-form)
-      repl-form)))
+  (if-let ((repl-form (cadr (seq-find
```
bbatsov: I see this in the build log `cider.el:756:30:Error: `(repl-form (cadr (seq-find (lambda (entry) (eq (car entry) repl-type)) cider-cljs-repl-types)))' is a malformed function`. If think by mistake someone made `if-let` work with just one set of parens and the form was immediately deprecated afterwards in Emacs 26.1. See how we're using this elsewhere in the code.

cider.el
```diff
@@ -755,14 +755,15 @@ you're working on."
...
+                             (lambda (entry)
+                               (eq (car entry) repl-type))
+                             cider-cljs-repl-types))))
+      ;; repl-form can be either a string or a function producing a string
+      (if (symbolp repl-form)
+          (funcall repl-form)
+        repl-form)
+    (user-error (format "No ClojureScript REPL type %s found.  Please make sure that `cider-cljs-repl-types` has an entry for it" repl-type))))
```
bbatsov: I think `user-error` does `format` implicitly inside. You can check the docstring.

bbatsov:
👍

Issue #2310, opened 2018-06-06
Add `cider-format-edn-last-sexp` for easy formatting


zcaudate:
Instead of highlighting the region and running `cider-format-edn-region`, this command will act like `C-x C-e` but instead, will prettify the last sexp.

PR #2311, opened 2018-06-06
Added cider-format-edn-last-sexp for [Feature #2310]

zcaudate:
This allows for easier formatting of data in the repl. It combines `cider-eval-last-sexp` and `cider-format-edn-region` to be able to format in-place.

-----------------

- [NA] You've added tests (if possible) to cover your change(s)

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bugs fixed
 
+* [#2310](https://github.com/clojure-emacs/cider/issues/2310): `cider-format-edn-last-sexp` will format the last sexp.
 * [#2294](https://github.com/clojure-emacs/cider/issues/2294): Fix setting default stacktrace filters.
 * [#2286](https://github.com/clojure-emacs/cider/issues/2286): Fix eldoc issue with images in the REPL.
 
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1981,6 +1981,11 @@ START and END represent the region's boundaries."
                           (lambda (edn)
                             (cider-sync-request:format-edn edn right-margin)))))
 
+(defun cider-format-edn-last-sexp ()
+  "Format the EDN data of the last sexp."
+  (interactive)
+  (apply 'cider-format-edn-region (cider-sexp-at-point 'bounds)))
+
 
 ;;; Interrupt evaluation
 
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -290,6 +290,10 @@ Configure `cider-cljs-repl-types' to change the ClojureScript REPL to use for yo
      ["Browse REPL input history" cider-repl-history]
      ["Browse classpath" cider-classpath]
      ["Browse classpath entry" cider-open-classpath-entry])
+    ("Format"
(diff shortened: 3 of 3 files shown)


cider-interaction.el
```diff
@@ -1981,6 +1981,17 @@ START and END represent the region's boundaries."
                           (lambda (edn)
                             (cider-sync-request:format-edn edn right-margin)))))
 
+(defun cider-format-edn-last-sexp ()
+  "Format the EDN data of the last sexp."
+  (interactive)
+  (save-excursion
+    (clojure-backward-logical-sexp 1)
+    (cider-format-edn-region (point)
```
bbatsov: Here you should simply use `(cider-sexp-at-point 'bounds)`. It's meant to be used with functions operating on regions.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 ## 0.17.0 (2018-05-07)
 
 ### New features
-
+* [#2310](https://github.com/clojure-emacs/cider/issues/2310) `cider-format-edn-last-sexp` will format the last sexp
```
bbatsov: You've added this to the wrong section of the changelog (0.17 was already released).

bbatsov: And this is not consistent with our format (missing `:` and a final `.`).

bbatsov:
New commands should be added to menus and the manual, otherwise no one will find them. :-)

cider-interaction.el
zcaudate: great. that's much easier

CHANGELOG.md
zcaudate: Fixed.

zcaudate:
I've added C-c M-c as the interactive hotkey and put a "Format" topic in the menu.

doc/interactive_programming.md
```diff
@@ -32,6 +32,7 @@ Here's a list of `cider-mode`'s keybindings:
 `cider-interrupt`                             |<kbd>C-c C-b</kbd>                   | Interrupt any pending evaluations.
 `cider-macroexpand-1`                         |<kbd>C-c C-m</kbd>                   | Invoke `macroexpand-1` on the form at point and display the result in a macroexpansion buffer.  If invoked with a prefix argument, `macroexpand` is used instead of `macroexpand-1`.
 `cider-macroexpand-all`                       |<kbd>C-c M-m</kbd>                   | Invoke `clojure.walk/macroexpand-all` on the form at point and display the result in a macroexpansion buffer.
+`cider-format-edn-last-sexp`                  |<kbd>C-c M-c</kbd>                   | Format the EDN data of the last sexp.
```
bbatsov: You forgot to actually bind this, but you also chose the same key we use in `clojure-mode` for `cider-connect`. :-) I think we should just put all the format related commands under the same keymap, so we can group them together and save some top-level keybindings. This can be done later, though.

zcaudate: I'll take this out. That'll be great if the `format` commands are grouped. I use maybe 3 hotkeys so I'll leave it to you.

bbatsov:
👍

PR #2313, opened 2018-06-06
Add cider-figwheel and figwheel-main REPL type

arichiardi:
This is more like a proposal, and something hacked quickly together. It mainly
adds `figwheel-main` as cljs REPL, now that it works. Note that
`cider/piggieback` needs releasing + bumping as well.

The `cider-figwheel.el` file will contain any custom configuration that Bruce
deems necessary. The `defcustom`s can of course be shared between "normal"
fighweel and figwheel-main.

I decided to create a file there, which might not necessary if we don't grow
the options at all. That `defcustom` and function can easily go into
`cider.el`.

-----------------

Thanks!

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -649,6 +649,12 @@ Generally you should not disable this unless you run into some faulty check."
   (unless (cider-library-present-p "figwheel-sidecar/figwheel-sidecar")
     (user-error "Figwheel-sidecar is not available.  Please check http://cider.readthedocs.io/en/latest/clojurescript")))
 
+(defun cider-check-figwheel-main-requirements ()
+  "Check whether we can start a Figwheel ClojureScript REPL."
+  (cider-verify-piggieback-is-present)
+  (unless (cider-library-present-p "bhauman/figwheel-main")
+    (user-error "Figwheel-main is not available.  Please check http://cider.readthedocs.io/en/latest/clojurescript")))
+
 (defun cider-check-weasel-requirements ()
   "Check whether we can start a Weasel ClojureScript REPL."
   (cider-verify-piggieback-is-present)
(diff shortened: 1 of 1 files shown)


cider-figwheel.el
```diff
@@ -0,0 +1,52 @@
...
+
+;;; Code:
+
+(defcustom cider-figwheel-main-start-options "dev"
+  "Defines the `figwheel.main/start' options.
+
+If this is `nil`, figwheel.main will be required but not started."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "0.19.0"))
```
bbatsov: When did we release 0.18? :-)

cider-figwheel.el
```diff
@@ -0,0 +1,52 @@
...
+
+;; This file is not part of GNU Emacs.
+
+;;; Commentary:
+
+;; Provides a registration functions for figwheel REPLs.
+
+;;; Code:
+
+(defcustom cider-figwheel-main-start-options "dev"
```
bbatsov: Might be better to just prompt for this and have some default for convenience.

bbatsov:
> cider/piggieback needs releasing + bumping as well.

I'll do this now. 

> The cider-figwheel.el file will contain any custom configuration that Bruce
deems necessary. The defcustoms can of course be shared between "normal"
fighweel and figwheel-main.

Seems odd to me to create a separate source file just for one config value and one function. Maybe down the road. The whole `cider.el` has to be reorganized, but this can't be done properly before we break up the damn `cider-interaction.el` (which I hope to do later this month).

cider-figwheel.el
arichiardi: Lol yeah I quickly put a number there :)

arichiardi:
Cool will change tomorrow

arichiardi:
Ok now `cider` requests the build id. I left the `defcustom` because `figwheel-main` can actually be configured with a full clojure map so a user can use that for doing it.

There is one small glitch that can be improved though, I don't know if in `cider` or `figwheel`:

`[Figwheel] Validating figwheel-main.edn` appears after some time so you have to wait a couple of seconds with two clj REPLs open and nothing happening. The cider spinner does not seem to signal this.

arichiardi:
Another problem is the the output of figwheel pollutes the REPL, I guess it is by design but probably it can be trimmed a bit.

2018-06-07-115456_813x933_scrot

bbatsov:
> Another problem is the the output of figwheel pollutes the REPL, I guess it is by design but probably it can be trimmed a bit.

I guess @bhauman can add some configuration for this.

PR #2314, opened 2018-06-08
Add cider-reload-ns and cider-ns-reload-all interactive command

arichiardi:
In order to send `(require 'ns :reload/:reload-all)` more easily.

-----------------

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -352,6 +352,8 @@ If invoked with a prefix ARG eval the expression after inserting it."
     ["Recursively load all files in directory" cider-load-all-files]
     ["Load all project files" cider-load-all-project-ns]
     ["Refresh loaded code" cider-ns-refresh]
+    ["Require and reload" cider-ns-reload]
+    ["Require and reload all" cider-ns-reload-all]
     ["Run project (-main function)" cider-run])
   "Menu for CIDER mode eval commands.")
 
@@ -410,6 +412,8 @@ If invoked with a prefix ARG eval the expression after inserting it."
 
 
 (declare-function cider-ns-refresh "cider-ns")
+(declare-function cider-ns-reload "cider-ns")
+(declare-function cider-ns-reload-all "cider-ns")
 (declare-function cider-browse-ns "cider-browse-ns")
 (declare-function cider-eval-ns-form "cider-eval")
 (declare-function cider-repl-set-ns "cider-repl")
@@ -427,6 +431,8 @@ If invoked with a prefix ARG eval the expression after inserting it."
     (define-key map (kbd "M-n") #'cider-repl-set-ns)
     (define-key map (kbd "r") #'cider-ns-refresh)
     (define-key map (kbd "M-r") #'cider-ns-refresh)
+    (define-key map (kbd "l") #'cider-ns-reload)
+    (define-key map (kbd "M-l") #'cider-ns-reload-all)
     map)
   "CIDER NS keymap.")
 
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -22,8 +22,38 @@
 
 ;;; Commentary:
 
-;; Smart code refresh functionality based on ideas
-;; fromhttp://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;; Smart code refresh functionality based on ideas from:
+;; http://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;;
+;; Note that refresh with clojure.tools.namespace.repl is a smarter way to
+;; reload code: the traditional way to reload Clojure code without restarting
+;; the JVM is (require ... :reload) or an editor/IDE feature that does the same
+;; thing.
+;;
+;; This has several problems:
+;;
+;; If you modify two namespaces which depend on each other, you must remember to
+;; reload them in the correct order to avoid compilation errors.
+;;
+;; If you remove definitions from a source file and then reload it, those
+;; definitions are still available in memory.  If other code depends on those
+;; definitions, it will continue to work but will break the next time you
+;; restart the JVM.
+;;
+;; If the reloaded namespace contains defmulti, you must also reload all of the
+;; associated defmethod expressions.
+;;
+;; If the reloaded namespace contains defprotocol, you must also reload any
+;; records or types implementing that protocol and replace any existing
+;; instances of those records/types with new instances.
+;;
+;; If the reloaded namespace contains macros, you must also reload any
+;; namespaces which use those macros.
+;;
+;; If the running program contains functions which close over values in the
+;; reloaded namespace, those closed-over values are not updated (This is common
+;; in web applications which construct the "handler stack" as a composition of
+;; functions.)
 
 ;;; Code:
 
@@ -144,6 +174,37 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
           (string-prefix-p project-root
                            (file-truename default-directory)
                            (eq system-type 'windows-nt))))))))
+
+;;;###autoload
+(defun cider-ns-reload (&optional prompt)
+  "Send a (require 'ns :reload) to the REPL.
+
+With an argument PROMPT, it prompts for a namespace name.  This is the
+Clojure out of the box reloading experience and does not rely on
+org.clojure/tools.namespace.  See Commentary of this file for a longer list
+of differences.  From the Clojure doc: \":reload forces loading of all the
+identified libs even if they are already loaded\"."
+  (interactive "P")
+  (let ((ns (if prompt
+                (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+              (clojure-find-ns))))
+    (cider-interactive-eval (format "(require '%s :reload)" ns))))
+
+;;;###autoload
+(defun cider-ns-reload-all (&optional prompt)
+  "Send a (require 'ns :reload-all) to the REPL.
+
+With an argument PROMPT, it prompts for a namespace name.  This is the
+Clojure out of the box reloading experience and does not rely on
+org.clojure/tools.namespace.  See Commentary of this file for a longer list
+of differences.  From the Clojure doc: \":reload-all implies :reload and
+also forces loading of all libs that the identified libs directly or
+indirectly load via require\"."
+  (interactive "P")
+  (let ((ns (if prompt
+                (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+              (clojure-find-ns))))
+    (cider-interactive-eval (format "(require '%s :reload-all)" ns))))
 
 ;;;###autoload
 (defun cider-ns-refresh (&optional mode)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#2373](https://github.com/clojure-emacs/cider/issues/2373): Make it possible to configure the welcome message displayed in scratch buffers via `cider-scratch-initial-message`.
 * Add the ability to jump to the profiler buffer using `cider-selector`.
 * [#1980](https://github.com/clojure-emacs/cider/issues/1980): Echo back missing namespace name on interactive eval (requires nREPL 0.4.3+).
+* [#2314](https://github.com/clojure-emacs/cider/pull/2314): Add `cider-ns-reload` and `cider-ns-reload-all` interactive commands.
 
 ### Bugs fixed
 
--- a/cider-util.el
+++ b/cider-util.el
@@ -772,7 +772,8 @@ through a stack of help buffers.  Variables `help-back-label' and
     "Exploring CIDER's menu-bar entries is a great way to discover features."
     "Keep in mind that some commands don't have a keybinding by default. Explore CIDER!"
     "Tweak `cider-repl-prompt-function' to customize your REPL prompt."
-    "Tweak `cider-eldoc-ns-function' to customize the way namespaces are displayed by eldoc.")
+    "Tweak `cider-eldoc-ns-function' to customize the way namespaces are displayed by eldoc."
+    "For no middleware, low-tech and reliable namespace reloading use <\\[cider-ns-reload]>.")
   "Some handy CIDER tips."
   )
 
--- a/doc/miscellaneous_features.md
+++ b/doc/miscellaneous_features.md
@@ -148,6 +148,11 @@ and
 * By default, all modified Clojure buffers are prompted to be saved. This
   behaviour can be customized using `cider-ns-save-files-on-refresh`.
 
+In case `cider-ns-refresh` does not work for you the
+`cider-ns-reload`|`cider-ns-reload-all` commands can be used instead: they will
+evaluate the Clojure's out of the box `(require ... :reload|:reload-all)` at
+the REPL.
+
 ## Tracing function execution
 
 You can trace the results produced by functions using <kbd>C-c M-t v</kbd>.  The


vspinu:
I think `cider-eval-ns-form` could do this instead of adding a new command.

arichiardi:
I would be fine with this, however however this command can also ask if you want to reload something explicitly. Very handy when you don't need to evaluate the whole `ns` form but only part of it.

We could add this as well to `cider-eval-ns-form` actually...so...well I am open to both approaches.

cider-interaction.el
```diff
@@ -1807,6 +1807,25 @@ refresh functions (defined in `cider-refresh-before-fn' and
           conn)))
      :clj 'any-mode)))
 
+(defun cider-reload (arg)
+  "Send a require :reload/:reload-all to the REPL.
+
+The prefix argument ARG can change the behavior of the command:
```
bbatsov: I'd be useful to explain here why would users want to prefer this over `cider-refresh`.

arichiardi:
@bbatsov what is you opinion on using `cider-eval-ns-form` for this?

cider-interaction.el
arichiardi: Adressed

cider-interaction.el
```diff
@@ -1840,6 +1840,33 @@ refresh functions (defined in `cider-refresh-before-fn' and
...
+
+  - (4)   \\[cider-reload]: prompts for a namespace name.
```
arichiardi: If there is a better way to show this, without havin a `C-u embedded in docstring` warning, I am eager to know.

cider-mode.el
```diff
@@ -273,6 +273,8 @@ Configure `cider-cljs-repl-types' to change the ClojureScript REPL to use for yo
      ["Clear latest output" cider-find-and-clear-repl-output]
      ["Clear all output" (cider-find-and-clear-repl-output t)
       :keys "\\[universal-argument] \\[cider-find-and-clear-repl-output]"]
+     ["Require and reload ns" (cider-reload t)
+      :keys "\\[universal-argument] \\[cider-reload]"]
```
arichiardi: Also need some advice on this one.

cider-interaction.el
```diff
@@ -1840,6 +1840,33 @@ refresh functions (defined in `cider-refresh-before-fn' and
...
+  - -     \\[cider-reload] : executes (require ... :reload-all).
+  - - (4) \\[cider-reload]: reloads all AND prompts.
+
+This the Clojure out of the box reloading experience and does not rely on
+org.clojure/tools.namespace.  From the Clojure doc:
+
+  :reload forces loading of all the identified libs even if they are already
+loaded
+  :reload-all implies :reload and also forces loading of all libs that the
+identified libs directly or indirectly load via require [...]"
```
bbatsov: This should definitely be mentioned, though - https://github.com/clojure/tools.namespace#reloading-code-motivation

cider-mode.el
```diff
@@ -273,6 +273,8 @@ Configure `cider-cljs-repl-types' to change the ClojureScript REPL to use for yo
      ["Clear latest output" cider-find-and-clear-repl-output]
      ["Clear all output" (cider-find-and-clear-repl-output t)
       :keys "\\[universal-argument] \\[cider-find-and-clear-repl-output]"]
+     ["Require and reload ns" (cider-reload t)
```
bbatsov: Why did you put this here?

arichiardi: I saw the other pattern, should it go somewhere else? Or take it away completely?

bbatsov: I meant that the local of this is kind of arbitrary.

vspinu:
BTW, I think `cider-reload` is not specific enough. It might be confused with `cider-restart` which does very different things.

cider-interaction.el
arichiardi: Actually maybe you are right and we should tune it for the most common cases - say I am working on ns A and I need to change it B: 

Flow 1)
- open ns B
- modify
- save
- reload the I am in - B

Flow 2)
- open B
- modify
- save
- go to A
- reload all in A

The ability of inputting a namespace stems from the fact that I am not a master of reload-all sometimes and I need to reload ad hoc a namespace that threw an error (was still stale), handy because I do need to open its buffer.

cider-interaction.el
arichiardi: Do you mean mention `tools.namespace` as alternative to reload?

arichiardi:
Clojure "reload" users should know the difference between the two I guess. There is not key binding for it at the moment so if you opt in probably you know the difference?

cider-mode.el
bbatsov: Local -> location. I meant that it should be somewhere around cider-refresh.

arichiardi: Oh ok cool I will move it there no problem

bbatsov:
Both of you make valid points - the current names can probably be improved, although generally most people should know what to do. Naming is hard! Maybe `cider-reload-ns/buffer` might be a better name. I don't have any great ideas right now.

arichiardi:
@bbatsov where should this go now? `cider-eval`? It is eval-ing something...

bbatsov:
I think it can go to `cider-common.el` if we plan to use this from repls, or to `cider-mode.el` otherwise.

arichiardi:
Ok cool, yes it is a repl command

arichiardi:
@bbatsov @vspinu I was thinking, what if the `reload` I have can be selected with a `defcustom`?

There are cases also when the `cider-refresh` would not work, one could be the absence of the middleware, the other is ClojureScript. Folks might want to disable it completely and use the `(require ... :reload)` instead.

arichiardi:
Well no, the functions have different arguments, completely different.

cider-mode.el
```diff
@@ -416,6 +418,7 @@ If invoked with a prefix ARG eval the expression after inserting it."
 (declare-function cider-toggle-trace-ns "cider-tracing")
 (declare-function cider-toggle-trace-var "cider-tracing")
 (declare-function cider-refresh "cider-refresh")
+(declare-function cider-reload-ns "cider-reload-ns")
```
bbatsov: That's not exactly correct. ;-)

cider-refresh.el
```diff
@@ -141,6 +171,37 @@ Its behavior is controlled by `cider-save-files-on-cider-refresh'."
                            (file-truename default-directory)
                            (eq system-type 'windows-nt))))))))
 
+;;;###autoload
+(defun cider-reload-ns (&optional arg)
```
bbatsov: Reload is different from refresh, so I wouldn't keep in the same source file.

arichiardi: Uhm why not? the two actually do the same thing. `cider-refresh` does it a bit better because of `clojure.tools.namespace`. I put the explanation in the commentary and it is evident from there that the end result is the same. That is why I opted for putting in here. If you feel very strong about moving, of course we will move it, please pick a file ;)

cider-mode.el
arichiardi: Oh got it

cider-refresh.el
vspinu: To be frank I am still not very clear about the difference. From the commentary which you added looks like refresh is uniformly better, right? Then why to bother with ns level reload? Does it have specific advantages besides small improvements in speed?

cider-refresh.el
```diff
@@ -22,8 +22,38 @@
 
 ;;; Commentary:
 
-;; Smart code refresh functionality based on ideas
-;; fromhttp://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;; Smart code refresh functionality based on ideas from:
+;; http://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;;
+;; Note that refresh with clojure.tools.namespace.repl is a smarter way to
+;; reload code.
```
vspinu: If I understand correctly, here you mean that refresh is "smarter" compared to "traditional reload" which you describe below? Wouldn't it be clearer to first describe "traditional reload" and then list all those features below as "improvements" of the refresh and not "problems" of reload?

I think this is a good write up and could be even placed in the manual.

arichiardi: I can do that - however I am not writing this myself but it is in the README of tools.namespace that Bohzidar linked. So kudos to the original writer.

cider-refresh.el
arichiardi: It is not speed the advantage, it is not using `tools.namespace` for something simple which is provided by Clojure natively. No middleware involved - less potential breakage

bbatsov: @arichiardi Noticed that everything in this file is about `cider-refresh`. Notice also that every function here has the `cider-refresh` prefix. `cider-reload-ns` simply looks out of place here. That's why I suggest `cider-common.el` or just putting this in `cider-mode.el` if you don't envision it to be used from the REPL much.

cider-refresh.el
```diff
@@ -141,6 +171,37 @@ Its behavior is controlled by `cider-save-files-on-cider-refresh'."
...
+  (interactive "P")
+  (let* ((reload-all-p (or (equal arg '-) (equal arg '(-4))))
+         (prompt-p (or (equal arg '(4)) (equal arg '(-4))))
+         (ns (if prompt-p
+                 (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+               (clojure-find-ns)))
+         (form (concat "(require '%s " (if (not reload-all-p)
+                                           ":reload)"
+                                         ":reload-all)"))))
+    (cider-interactive-eval (format form ns))))
```
bbatsov: Btw, with the default handler users won't really get any specific feedback the ns being reloaded. They'll just see some form got evaluated.

arichiardi: Uhm that is not good, there should be at least the spinner going. Which handler would you recommend?

cider-refresh.el
arichiardi: Oh ok yes that makes sense, unless we call it `cider-refresh-native-reload`

arichiardi: Or `cider-refresh-require-reload`

bbatsov: I actually have a better idea finally. We rename `cider-refresh.el` to `cider-ns.el` and we put there `cider-refresh` as `cider-ns-refresh` and `cider-reload-ns`, as `cider-ns-reload`. This will also make it a natural home for things like `cider-ns-eval` and `cider-ns-cleanup`, etc.

cider-refresh.el
bbatsov: Leave this for now. I'll think of some more versatile handler later. Basically we just need something like the current handler that should be aware if it was invoked in a REPL and then print its out there.

cider-refresh.el
arichiardi: Oh I like this will do it

arichiardi:
Oh ok didn't notice that you've already renamed. I will need to do a cleanup and rebase.

bbatsov:
👍

arichiardi:
Revisited this, splitting in two commands and assigning them to `C-c M-n l` and `C-c M-n M-l` respectively.

cider-ns.el
```diff
@@ -145,6 +175,46 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
...
+list of differences.  From the Clojure doc:
+
+  :reload forces loading of all the identified libs even if they are
+already loaded"
+  (interactive "P")
+  (let* ((prompt-p (or (equal arg '(4)) (equal arg '(-4))))
+         (ns (if prompt-p
+                 (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+               (clojure-find-ns)))
+         (form (concat "(require '%s :reload)")))
```
bbatsov: What's this `concat` here for? :-) I think you could have used `format` here directly to add the ns instead of doing it a line done.

cider-ns.el
```diff
@@ -145,6 +175,46 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
...
+list of differences.  From the Clojure doc:
+
+  :reload-all implies :reload and also forces loading of all libs that the
+identified libs directly or indirectly load via require"
+  (interactive "P")
+  (let* ((prompt-p (or (equal arg '(4)) (equal arg '(-4))))
+         (ns (if prompt-p
+                 (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+               (clojure-find-ns)))
+         (form (concat "(require '%s :reload-all)")))
```
bbatsov: Same here.

cider-ns.el
arichiardi: Yep sorry will fix that, leftover from the previous impl

cider-ns.el
```diff
@@ -145,6 +175,46 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
...
+With the universal prefix ARG, prompts for a namespace name and then
+
+This is the Clojure out of the box reloading experience and does not rely
+on org.clojure/tools.namespace.  See Commentary of this file for a longer
+list of differences.  From the Clojure doc:
+
+  :reload forces loading of all the identified libs even if they are
+already loaded"
+  (interactive "P")
+  (let* ((prompt-p (or (equal arg '(4)) (equal arg '(-4))))
```
vspinu: `arg` has no other function right? So you could name your argument `prompt-p` directly. Alos `-p` prefix is reserved for functions (predicates). We use `prompt` and `do-prompt` elsewhere.

cider-ns.el
```diff
@@ -22,8 +22,38 @@
 
 ;;; Commentary:
 
-;; Smart code refresh functionality based on ideas
-;; fromhttp://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;; Smart code refresh functionality based on ideas from:
+;; http://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded
+;;
+;; Note that refresh with clojure.tools.namespace.repl is a smarter way to
```
vspinu: Smarter compared to what? Now that this file contains loads of other ns functionality this comment should be more explicit regarding what it refers to.

cider-ns.el
```diff
@@ -145,6 +175,46 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
...
+                 (string-remove-prefix "'" (read-from-minibuffer "Namespace: " (clojure-find-ns)))
+               (clojure-find-ns)))
+         (form (concat "(require '%s :reload)")))
+    (cider-interactive-eval (format form ns))))
+
+;;;###autoload
+(defun cider-ns-reload-all (&optional arg)
+  "Send a (require current-ns :reload-all) to the REPL.
+
+With the universal prefix ARG, prompts for a namespace name and then
```
vspinu: Same here.

vspinu:
Are you sure about the keys? Usually base keys like `l` and `M-l` serve for the same command. In my experience when two similar commands are hooked on similar keys it's very easy to confuse them. But maybe it's not a problem in this case because one is a super-set of the other, right?

An alternative would be to have one command proposing `*all*` option along other namespaces when used with the prefix argument.  `*all*` could be the default such that `C-u C-c l RET` would get us the `:reload-all`.

cider-ns.el
bbatsov: Good catch. I guess this is a leftover from the original implementation that had a different behavior for single and double prefix argument. I really dislike double prefix args, so I asked @arichiardi to split this in two commands.

bbatsov:
I think it's fine in this this case, although probably we can have `M-l` and `l` for reload and `M-a` and `a` for reload-all. I also comtemplated on whether to comment on this in light of how most of CIDER behaves, but I think `l` and `M-l` would be ok as well, as nothing really bad can come out from mixing those up.

cider-ns.el
```diff
@@ -145,6 +175,46 @@ Its behavior is controlled by `cider-save-files-on-cider-ns-refresh'."
                            (file-truename default-directory)
                            (eq system-type 'windows-nt))))))))
 
+;;;###autoload
+(defun cider-ns-reload (&optional arg)
+  "Send a (require current-ns :reload) to the REPL.
+
+With the universal prefix ARG, prompts for a namespace name and then
```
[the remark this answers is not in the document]
arichiardi: Lol thanks, again did some poor job in splitting the original fn + comment yesterday

cider-ns.el
arichiardi: Cool will do

arichiardi:
Btw why do we reserve two keys per command? In this case it felt better to change the pattern a bit but I am open to any solution you guys find better

cider-ns.el
arichiardi: merged with the following line down there

bbatsov:
Common Emacs convention, as some people (like me) prefer not to lift their finger holding down some modifier key, and some don't. :-)

arichiardi:
Oh got it, I think now understand better your comment above about the behavior and yes, the worst you observe if you `:reload-all` instead of `:reload` is that it takes more time. Of course this method is not really optimal (that is why the `tools.namespace` method exists) so most of the times I personally do not rely on the exactness of it.

PR #2322, opened 2018-06-13
Add cider-insert-and-step-forward-p defcustom for cider-insert-in-repl

arichiardi:
For jumping to the next form after insertion of the current sexp. Default value
if nil.

-----------------

  -
- [byte-compilation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Byte-Compilation.html),
- [`checkdoc`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips.html),
- [check-declare](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declaring-Functions.html),
- packaging metadata, indentation, and trailing whitespace checks.

  - Do we have an entry for #2319 already?


arichiardi:
@bbatsov any objection to this super cool addition :smile: :angel:

bbatsov:
Seems really odd to change the behaviour of the function with a defcustom in this manner. I'd imagine some time you'd want one behavior, other time the alternative behaviour.

arichiardi:
Well don't know about that, I personally have it turned on all the time. Maybe this can go in as it is and then we see if folks require the other approach, the two approaches can live together I guess.

bbatsov:
I just think that when you modify commands in this manner the likelihood of someone changing the default (other than you) is going to be around 0. Let me think about this.

arichiardi:
Well @vspinu is already using the forward sexp so maybe it is already two people. Also note that this does not change yet the normal eval command.

bbatsov:
My main concern is that with this change you've made the behaviour is all-in in one direction or the other one. I really can't imagine why I'd always want my point to advance after evaluating something - e.g. if I'm experimenting with a definition I probably won't want this.

And I really think there should be at least some version eval-last-sexp and eval-defun that can behave similarly. I'd rather have commands like eval-last-sexp-and-go-to-next, etc, than a defcustom that impacts the flexibility of the the commands it touches.

Issue #2255, opened 2018-03-29
Support comment ignoring in `cider-insert-defun-in-repl`


halgari:
When writing code in a namespace it's not uncommon to keep a set of forms in a comment block for on-the-fly testing (example: https://github.com/clojure/spec.alpha/blob/master/src/test/clojure/clojure/test_clojure/spec.clj#L219-L221) In some other editors evaling a "top-level" form ignores comments and "top level" refers to the highest level form excluding the comment. 

## Expected behavior
```
(comment
(+ 1 2) ;; eval-in inside this form 

)
```
=> `3`

## Actual behavior
Loads the entire `(comment ... )` form in the repl

## Steps to reproduce the problem

Run `cider-insert-defun-in-repl` inside any comment block. 

### CIDER version information
20180131.913

halgari:
And it seems the Clojure guide just added a example video of the behavior I'd like to see: https://clojure.org/images/content/guides/repl/cursive-repl-demo.gif

https://clojure.org/guides/repl/enhancing_your_repl_workflow

bbatsov:
I guess now that #2323 is merged we can close this ticket.

PR #2323, opened 2018-06-13
Eval top level defuns inside comment forms

dpsutton:
Since this is a sensitive codepath (cider-defun-at-point), it is risky
to change its behavior due to introducing bugs or introducing
unexpected behavior. This _should_ only affect evaluation inside a
comment form but we want to be careful. Therefore there is a defcustom
`cider-eval-toplevel-inside-comment-form` while others test this. If
there is a bug, it is easy for users to turn this feature off completely.

The tests are pretty indicative of the desired behavior:

```lisp
      (cider-buffer-with-text
          "(comment
              (wrong)
              (wrong)
              (rig|ht)
              (wrong))"
          (expect (cider-defun-inside-comment-form) :to-equal "(right)"))

     (cider-buffer-with-text
         "(comment
              (wrong)
              (wrong)
              (right)     |
              (wrong))"
         (expect (cider-defun-inside-comment-form) :to-equal "(right)"))

      (cider-buffer-with-text
          "(comment
              (wrong)
              (wrong)
              (((((r|ight)))))
              (wrong))"
          (expect (cider-defun-inside-comment-form) :to-equal "(((((right)))))"))
```

This is a pretty hot codepath so it must be enabled for the time being with `cider-eval-toplevel-inside-comment-form` while we make sure it is both bug free and desired.

This combined with the new insert map `C-c C-j [ednr]` makes comment forms very usable.

@halgari asked for this in https://github.com/clojure-emacs/cider/issues/2255.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### New features
 
+* Allow evaling top level forms in a comment form rather than the entire comment form with `cider-eval-toplevel-inside-comment-form`.
 * Create keymap for inserting forms into the repl at `C-c C-j`.
 * Add new defcustom `cider-invert-insert-eval-p`: Set to cause insert-to-repl commands to eval the forms by default when inserted.
 * Add new defcustom `cider-switch-to-repl-after-insert-p`: Set to prevent cursor from going to the repl when inserting a form in the repl with the insert-to-repl commands.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### New features
 
+* Allow evaling top level forms in a comment form rather than the entire comment form with `cider-eval-toplevel-inside-comment-form`.
+* Create keymap for inserting forms into the repl at `C-c C-j`.
+* Add new defcustom `cider-invert-insert-eval-p`: Set to cause insert-to-repl commands to eval the forms by default when inserted.
+* Add new defcustom `cider-switch-to-repl-after-insert-p`: Set to prevent cursor from going to the repl when inserting a form in the repl with the insert-to-repl commands.
 * Inject piggieback automatically on `cider-jack-in-clojurescript`.
 
 ### Bugs fixed
@@ -17,10 +21,6 @@
 
 ### New features
 
-* Allow evaling top level forms in a comment form rather than the entire comment form with `cider-eval-toplevel-inside-comment-form`.
-* Create keymap for inserting forms into the repl at `C-c C-j`.
-* Add new defcustom `cider-invert-insert-eval-p`: Set to cause insert-to-repl commands to eval the forms by default when inserted.
-* Add new defcustom `cider-switch-to-repl-after-insert-p`: Set to prevent cursor from going to the repl when inserting a form in the repl with the insert-to-repl commands.
 * [#2248](https://github.com/clojure-emacs/cider/pull/2248): `cider-repl` can now display recognized images in the REPL buffer.
 * [#2172](https://github.com/clojure-emacs/cider/pull/2172): Render diffs for expected / actual test results.
 * [#2167](https://github.com/clojure-emacs/cider/pull/2167): Add new defcustom `cider-jdk-src-paths`. Configure it to connect stack trace links to Java source code.
--- a/cider-util.el
+++ b/cider-util.el
@@ -123,6 +123,21 @@ testing, give an easy way to turn this new behavior off."
   :type 'boolean
   :package-version '(cider . "0.18.0"))
 
+(defun cider-sexp-starts-until-position (position)
+  "Returns the starting points for forms before POSITION.
+Positions are in descending order to aide in finding the first starting
+position before the current position."
+  (save-excursion
+    (let (sexp-positions)
+      (condition-case nil
+          (while (< (point) position)
+            (clojure-forward-logical-sexp 1)
+            (clojure-backward-logical-sexp 1)
+            (push (point) sexp-positions)
+            (clojure-forward-logical-sexp 1))
+        (scan-error nil))
+      sexp-positions)))
+
 (defun cider-defun-inside-comment-form (&optional bounds)
   "Return the toplevel form inside a comment containing point.
 Assumes point is inside a (comment ....) form and will return the text of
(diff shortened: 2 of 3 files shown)


arichiardi:
Doesn't this work already?

I am using your recent `cider-insert-last-sexp-in-repl` in a `comment` form and I see the behavior in @halgari's gif.

arichiardi:
Oh I got it, you want to eval the right thing while inside and not at the end.

dpsutton:
@arichiardi it works if you eval the last sexp but that requires you to be at the end of each form. If you use the top-level-evaluation you get the whole `(coment ...)` form which always evals to nil. This does the obvious thing of recognizing when you're in a comment form and scoots over the top level forms _inside_ the comment form until it finds the one containing yours.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### New features
 
+* Allow evaling top level forms in a comment form rather than the entire comment form with `cider-eval-toplevel-inside-comment-form`.
```
arichiardi: This is the wrong place, see #2322 ;)

dpsutton: you're saying it should be in the bugfix part? If so I wasn't sure if this was a new feature or fixing a bug. I don't have a strong opinion either way

arichiardi: No sorry, basically you put it under `0.17.0`, it should be in `unreleased`

dpsutton: oh whoops

dpsutton: good call. i did the same for the last PR as well

arichiardi: Ok probably this should go in first then, then I'll rebase in #2322

dpsutton: wait and make sure its accepted first :)

bbatsov:
Overall the suggested changes look good to me, but they don't account for `#_`, which seems like a problem as many people prefer it over `comment`.

dpsutton:
I don't believe that eval last sexp or eval top level defun honor that
form. Also this is more for what Tim baldridge was talking about: a comment
form in a namespace for interactive programming with the repl. I've never
heard of someone using the reader macro for this case.

bbatsov:
Fair enough. Guess I should have thought more about this before raising this point. Seems reasonable to be putting multiple forms in a `comment`, not to use `#_` for each of them.

cider-util.el
```diff
@@ -96,17 +96,85 @@ If BUFFER is provided act on that buffer instead."
...
+(defun cider--text-or-limits (bounds start end)
+  "Returns the substring or the bounds of text.
+If BOUNDS is non-nil, returns the list (START END) of character
+positions.  Else returns the substring from START to END."
+  (funcall (if bounds #'list #'buffer-substring-no-properties)
+           start end))
+
+(defun cider-top-level-comment-p ()
+  "Return non-nil if point is in a comment form."
+  (save-excursion
```
bbatsov: This who function seems a bit too complex. What exactly is the reasoning here? Can't you just go find use some generic way to find the top-level form?

cider-util.el
```diff
@@ -96,17 +96,85 @@ If BUFFER is provided act on that buffer instead."
...
+  :group 'cider
+  :type 'boolean
+  :package-version '(cider . "0.18.0"))
+
+(defun cider-defun-inside-comment-form (&optional bounds)
+  "Return the toplevel form inside a comment containing point.
+Assumes point is inside a (comment ....) form and will return the text of
+that form or if BOUNDS, will return a list of the starting and ending
+position."
+  (save-excursion
```
bbatsov: This seems extremely complex, which leads me to believe we can probably simplify it somehow. 😆 What was your reasoning with this implementation?

cider-util.el
dpsutton: This seems about as generic as it gets to find the first form of a sexp. One subtlety you have to worry about is that the first form isn't necessarily nestled against the opening paren. ie, `(    comment ...)`. So there needs to be a reliable way to get to the beginning of that form. I could have scooted across whitespace but instead this goes forward (to the end) of the next sexp and then back to the beginning. This should handle the (rare) case that maybe a traditional comment like

```
( ;;; repl interaction forms
comment
   (launch-missles)
)
```
This shows that just consuming whitespace isn't enough. The easiest way that I know of to get to a meaningful form is to call `(clojure-forward-logical-sexp 1)` and then `(clojure-backward-logical-sexp 1)` to return to the start of it.

cider-util.el
dpsutton: The reasoning is very straightforward: find which form contains point. The complexities come from when point isn't inside of a form (what do we do? this initially moves point backwards across whitespace so it is hopefully "attached to" a form). Also, what happens when there are no forms `(comment)`, what happens if we can't identify which form contains it. Also, I wanted it to return the text of each sexp and not the trailing whitespace from the previous form in addition.

The result is that the function remembers the outer limits of the comment so that if it has no better idea it just evaluates the entire comment form and returns nil just as before.

I believe this and the previous function could be simpler if there were a `clojure-move-to-beginning-of-next-form` type function in clojure-mode. As it stands I'm using the previous trick of forward-logical-sexp, backwards logical-sexp to find the limits of each sexp, and then finding if the point (scooted backwards to be adjacent to non-whitespace) is in those bounds.

dpsutton:
@bbatsov I've simplified it a bit now with only a single change in behavior. Now it gets a list of the start points of the top level forms in the comment and looks for the first one that is before the current point.

I'm not sure what to do in the situation when point is outside of the comment form. An example where <> is point:

```clojure
(comment
    (def conn "connection-string")
    (def users (get-users conn)))    <>
```

Previously i just bailed and evaled the entire comment form. Now it evals that last form. I figure this might be more what people mean so they don't have to move the cursor on the last line. It also prevents having to do a special case since the start of the last form is before the current point. But I don't know what the right answer is there.

bbatsov:
I'm OK with the simplified version of the PR. We just need a green build and we're good to go.

dpsutton:
> travis-ci/install-cask.sh: 17: travis-ci/install-cask.sh: cask: not found
-e 
The command "cask_install_or_reset" failed 3 times.

I"ll have to retrigger it.

dpsutton:
@bbatsov all green

bbatsov:
Thanks!

PR #2324, opened 2018-06-13
New connection API and jack-in rewrite

vspinu:
Second attempt on #2069. A brief description of the new functionality follows.

1) __Jack-in/connect__ 
  
   - User level commands:
  
     Create new sessions:
     
      C-c M-j: cider-jack-in-clj
      C-c M-J: cider-jack-in-cljs
     
      C-c M-c: cider-connect-clj
      C-c M-C: cider-connect-cljs
     
     Add new REPLs to the current session:
     
      C-c M-s: cider-connect-sibling-clj
      C-c M-S: cider-connect-sibling-cljs
 
   - `cider-jack-in-clojurescript` no longer creates two repls, only the cljs repl
  
   - clj repl has no longer a special status of the "main" repl. All repls within a session share same server and are siblings of each other. You can create as many clj and cljs siblings as you want from any repl. 
    
   - Creation of the client is no longer tightly bounded with the nrepl-server startup. The dynamic communication mechanism between jack-in and nrepl-server filter has been replace by a simple `on-port-callback`.
    
2) __New Connection and Session Management API__
 
   - Connections (aka REPLs) are grouped in [sesman](https://github.com/vspinu/sesman) sessions. Sibling repls are added to the current session.
  
   - Cider connection commands (`cider-quit`, `cider-restart`, `cider-display-connection-info`) have been refactored to operate exclusively on the connection level.
    
   - Sesman commands operate on the whole session:
  
     sesman-map
   
   - Associations (links) between current context (buffer, directory, project) are governed by sesman and could be formed only on the session level. 
    
     In a nutshell, session can be linked to projects, directories and buffers. Buffer link have precedence over directory links, and directory have precedence over project links.
     
     When a sesman session is registered it's automatically linked with the lowest priority context (project, or directory if no project found). By default (configured with `sesman-1-to-1-links`) multiple sessions can be linked with a project or a directory, but only one session can be linked with a buffer. 
     
     Cider functionality (eval, completion, repl-switching etc) operate on linked sessions. When there are multiple linked sessions ambiguity is automatically resolved by the recency of the REPL buffers (configured with `sesman-disambiguate-by-relevance`). 
     
   - Show info on current links with `C-c C-s l`. Show info on current, linked or all sessions with `C-c C-s i`.
    
   - Micro-management of the server is not allowed (it's not useful and would complicate UI). All repls within a session share a server. Server can be either remote (`cider-connect`) or local (bootstraped within the emacs during `cider-jack-in-xyz`).
    
     In case of the local server, when the last connection is killed the server is automatically killed. `cider-restart` restarts the connection but not the server. `sesman-restart` restarts the server and all the connections.

This PR should still be considered WIP. Would be great if people could pull and test in the upcoming weeks. Conceptual, API and UI issues is the main focus right now. Minor issues with docs, tests etc will be fixed in the later stage once the dust has settled.

To try out (assuming you are loading cider from source):

```sh
   git remote add vspinu https://github.com/vspinu/cider.git
   git fetch vspinu
   git checkout vspinu/new-connection
   ## then restart emacs and play with `cider-jack-in-xyz`, `cider-restart`, 
   ## `cider-quit` and all sesman commands on `C-c C-s` map.
```

At least two issues I still plan to tackle here:

   - Restart of SSH tunneled connection has not been tested and probably doesn't work 
   - REPL buffer naming system is no longer adequate. It should be possible to include session name as part of the buffer name and add more flexibility into the customization of buffer name templates.

----- 

A tot of no longer necessary or questionable functionality has been removed. The goal is to start from scratch and add only what is really necessary. I am listing all the removed functions for the ease of lookup through the github interface.

Removed:

    cider--connection-host,
    cider--connection-port,
    cider--connection-project-dir,
    cider--connection-properties,
    cider--connection-type,
    cider--guess-cljs-connection, 
    cider--has-warned-about-bad-repl-type, 
    cider--in-connection-buffer-p, 
    cider--quit-connection,
    cider--restart-connection,
    cider-assoc-buffer-with-connection,
    cider-assoc-project-with-connection, 
    cider-change-buffers-designation,
    cider-clear-buffer-local-connection, 
    cider-close-nrepl-session,
    cider-connections (variable), 
    cider-current-connection (variable), 
    cider-current-messages-buffer,
    cider-current-repl-buffer,
    cider-default-connection, 
    cider-extract-designation-from-current-repl-buffer,
    cider-find-connection-buffer-for-project-directory, 
    cider-find-reusable-repl-buffer,
    cider-make-connection-default,
    cider-map-connections, 
    cider-other-connection,
    cider-project-connections, 
    cider-project-connections-types,
    cider-prompt-for-project-on-connect,
    cider-read-connection,
    cider-repl-buffers,
    cider-replicate-connection,
    cider-request-dispatch, 
    cider-rotate-default-connection,
    cider-toggle-buffer-connection, 
    cider-toggle-request-dispatch, 
    nrepl-connection-buffer-name-template,
    nrepl-create-client-buffer-function,
    nrepl-post-client-callback
    nrepl-prompt-to-kill-server-buffer-on-quit,
    nrepl-use-this-as-repl-buffer,

   Connection Browser Functionality:
   
    cider--connection-browser-buffer-name,
    cider--connection-ewoc,
    cider--connection-pp,
    cider--connections-close-connection,
    cider--connections-goto-connection,
    cider--connections-make-default,
    cider--connections-refresh,
    cider--connections-refresh-buffer,
    cider--ewoc-apply-at-point,
    cider--setup-connection-browser,
    cider--update-connections-display,
    cider-client-name-repl-type,
    cider-connection-browser,
    cider-connections-buffer-mode,
    cider-connections-buffer-mode-map
    cider-connections-close-connection,
    cider-connections-goto-connection,
    cider-connections-make-default,
    cider-display-connected-message,
    cider-project-name,
   
Renamed: 

    cider-current-session -> cider-nrepl-eval-session
    cider-current-tooling-session -> cider-nrepl-tooling-session
    cider-display-connection-info -> cider-describe-current-connection
    cider-create-sibling-cljs-repl -> cider-connect-sibling-cljs
    nrepl-connection-buffer-name -> nrepl-repl-buffer-name
    cider--close-connection-buffer -> cider--close-connection  

    ## repl <> connection overlap cleanup
    cider-connections -> cider-repls
    cider-current-connection -> cider-current-repl
    cider-map-connections -> cider-map-repls
    cider-connection-type-for-buffer -> cider-repl-type-for-buffer
    cider-repl-set-type -> cider-set-repl-type


dpsutton:
Could you share the steps we need to take to test this out?

1. `lein new sessions`
2. open a source file
3. ??
4. ??

Presumably just `cider-jack-in-clj` ?

Does everything currently work? 

And amazing work :)

vspinu:
The PR is self contained (`sesman.el` is temporarily part of this PR).  You should just pull my branch and try it out assuming you load cider from source. I have added the instructions.

bbatsov:
@vspinu In the interest of this not going the way it did last time I actually plan to just merge your PR and take it from there. The architecture is sound, we just need to polish things. Before there merge I'd like to ask you to just get the build to pass, everything else we can cleanup and fix later.

bbatsov:
> cider-replicate-connection,

Why is this removed? 

> A tot of no longer necessary or questionable functionality has been removed. 

It might have been nice if you added a line about everything on this list, as it's not really obvious which became redundant and which you consider questionable.

vspinu:
>>    cider-replicate-connection,

> Why is this removed?

Because it would be equivalent to creating a sibling with `C-c M-s/S`.  If a replica of a whole session is required then it should be part of sesman. I thought of adding such command but I am not entirely convinced of its value. Why just not jack-in in order to create a separate session?

 > It might have been nice if you added a line about everything on this list

Sure.

cider-connection.el
```diff
@@ -0,0 +1,395 @@
...
+(defun cider-connected-p ()
+  "Return t if CIDER is currently connected, nil otherwise."
+  (sesman-has-links-p 'CIDER))
+
+(defun cider-ensure-connected ()
+  "Ensure there is a linked CIDER session."
+  (let ((sesman-disambiguate-by-relevance t))
+    (sesman-ensure-linked-session 'CIDER)))
+
+(defun cider-connection-type-for-buffer (&optional buffer)
```
arichiardi: Given we are in refactoring mode, I have always thought of renaming this to something more appropriate and keep the "connection type" for something like "nRepl" or "socket". 

This one seems more like a repl type?

vspinu: Connection and repl are currently synonyms in the sense that they both refer to the repl buffer; historically that wasn't the case.

It might be a good time to reconsider and rename most of (all?) the  `..-connection-..` into `..-repl-..` on this ocassion, but I didn't do that because it would introduce a large amount of non-essential changes.

bbatsov: Yeah, that's fine by me. We can do it down the road. This still leaves open the question how to differentiate between a connection type (nREPL, socket, unrepl) and a REPL type (clj, cljs) whatever. It'd be nice if we came up with some good terminology for this - e.g. `repl-flavour` and `repl-connection-type` or whatever.

cider.el
```diff
@@ -918,122 +842,195 @@ own buffer."
...
+    (cider--jack-in prompt-project
+      (lambda (server-buffer)
+        (cider-connect-sibling-cljs server-buffer)))))
 
 ;;;###autoload
-(defun cider-jack-in-clojurescript (&optional prompt-project)
-  "Start an nREPL server and connect to it both Clojure and ClojureScript REPLs.
-If PROMPT-PROJECT is t, then prompt for the project for which to
-start the server."
+(defun cider-jack-in-cljcljs (&optional prompt-project soft-cljs-start)
```
bbatsov: I'd name this `combo` or something `cljcljs` looks like some type to me. :-)

bbatsov:
Just a few general notes on the jack-in changes:

* I think you've removed the option to be able to manually select the params to be passed to lein/boot/whatever on jack-in. That's quite useful so we should keep it.
* I wonder if we need dedicated create-sibling commands - my idea for this after we add sessions was to just prompt users for the session they want a connection to be added to if some session(s) already exist. The constant questions whether to have fewer smarter commands or many more specific ones. 
* I was also thinking of making commands like jack-in ask for the type of the REPL to start instead of having two of them. Same for `cider-connect`, although the cljs version might become redundant once the middleware type bug gets fixed.
* At some point I hope we'd have a single entry point `M-x cider` which will ask for the type of connection you'd like to create and delegate to `jack-in` or `connect`. 

Random feedback, mostly based on my observation that too many commands can be too confusing at some point.

cider-connection.el
vspinu: This makes sense to me. `cider-repl-type` is good for `cljs` and `clj` because REPL buffer name contains the type; it's a good association. `cider-connection-type` and `cider-server-type` would play well for nREPL, socket, unrepl etc. as those are specifics of the transport protocol which users should not care much about.

cider.el
vspinu: I am not a big fan of "combo" or "both" because it's simultaneously not specific enough (combo of what?, both of what?), and unnecessarily restrictive. If some day cider would support clr or stand alone clojure, how would you name those? cljcljs is good because it's consistent with standard postfixes -clj and -cljs. It can be -clj&cljs.

 The `&` separator is particularly useful convention because users can now create custom bootstraps like `jack-in-clj&node&feegweel` with ease.

vspinu:
>   I think you've removed the option to be able to manually select the params

I preserved that. On C-u both project directory and params are asked for. Just like before.

>    I wonder if we need dedicated create-sibling commands - 

I think that any type of interactive asking (even y-or-n) is less efficient than several commands with a clean separation of concerns.  With this PR we would have 3 keybinding M-j, M-c and M-s plus  the shifted versions for cljs. There is virtually zero cognitive burden in having an extra M-s key because all 3 bindings are mnemonic and work by similar pattern.
 
For me, pressing `C-c M-j` then `C-c M-S` for a clj&cljs session is way faster than answering some intermediate selection questions. Especially if you consider that till I finished pressing `C-c M-S` clj server bootstrap is already half way through!

Relatedly  I was never fond of "Do you want to reuse repl buffer xyz?" This functionality is no longer there. In fact it's not necessary because of the clean restart implementation. Even when an orphan (process-less or server-less) repl occurred, restarting a connection or a session should bring it back to life.

> my idea for this after we add sessions was to just prompt users for the session they want a connection to be added to if some session(s) already exist.

It full generality it would be a two step process 1) "You have a sessions foo, zoo, bar. Do you want to add a new sibling to one of them or create a new session"? 2) "pick session: foo, zoo, bar?".  Even when defaulting to the most relevant current session, one would need an y-or-n question.  

`y-or-n` might seem harmless but it creates bad memory habits. For example currently cider-quit asks two questions one for connection and one for server. Over years I got so used to typing `y y` that the quiting key essentially became `C-c C-q y y`. I never, ever pressed `n` on those questions. Does this sound familiar? 

 With this PR connections and sessions are exited and restarted without extra questions. If a user presses `C-c C-q` or `C-c C-s C-q` sesman assumes that he/she knows what he/she wants from the start. 

>  At some point I hope we'd have a single entry point M-x cider which will ask for the type of connection you'd like to create and delegate to jack-in or connect.

That entry point is there: `sesman-start` on `C-c C-s C-s`. It is precisely designed for such one-command-fits-all case. Currently it bootstraps clj session, checks for cljs requirement and starts cljs sibling if those requirements are met. But sure, it can be extended to do all possible scenarios as you envisage.

arichiardi:
I like the one command with no question approach as well, I see though many people do not, but this can always be customizable.

I will try this PR, the approach looks sound.

Let me submit an additional requirement as well. More and more often projects will have cljs only REPL requirements. For instance in my case - node server app - I only need the cljs REPL buffer. The clj in in the way and I just close it every time.

Maybe this PR solves this already?

vspinu:
> Maybe this PR solves this already?

Yes. `C-c C-J` creates one cljs repl only. You can add clj or cljs repl any time later with `C-c M-s/S`. In the new design `clj` repl is no more special than the `cljs` one.

cider-connection.el
vspinu: Just refactored:

```
    ## repl <> connection overlap cleanup
    cider-connections -> cider-repls
    cider-current-connection -> cider-current-repl
    cider-map-connections -> cider-map-repls
    cider-connection-type-for-buffer -> cider-repl-type-for-buffer
    cider-repl-set-type -> cider-set-repl-type
```

arichiardi:
Awesome I like the rename

bbatsov:
> I think that any type of interactive asking (even y-or-n) is less efficient than several commands with a clean separation of concerns. With this PR we would have 3 keybinding M-j, M-c and M-s plus the shifted versions for cljs. There is virtually zero cognitive burden in having an extra M-s key because all 3 bindings are mnemonic and work by similar pattern.

You'd be suprised how certain people think. Anyways, let's just go like this as it is for now.

> With this PR connections and sessions are exited and restarted without extra questions. If a user presses C-c C-q or C-c C-s C-q sesman assumes that he/she knows what he/she wants from the start.

I often types some keys by mistake, so I think that's a bad idea, but it's not a big deal. I'll just change this back later if needed. 

> That entry point is there: sesman-start on C-c C-s C-s. It is precisely designed for such one-command-fits-all case. Currently it bootstraps clj session, checks for cljs requirement and starts cljs sibling if those requirements are met. But sure, it can be extended to do all possible scenarios as you envisage.

Great! Still this will need to be wrapped in some `cider` command, as I'm assuming sesman can't really know what type of session it's supposed to start.

Anyways, things are looking good and we're just a green build away from merging this as far as I'm concerned.

vspinu:
> I often types some keys by mistake, so I think that's a bad idea, but it's not a big deal. I'll just change this back later if needed.

A valid point indeed. But C-c C-q is relatively hard to type; even so for C-c C-s C-q. I think it's worth the risk. In any event, if you plan to change this we need it to reconcile with sesman. It would be really akward if session level commands would differ from connection level in this regard.

> Still this will need to be wrapped in some cider command, 

It works the other way around, the interactive command `sesman-start` wraps generic `sesman-start-session`, a method for which is defined by CIDER.

> green build away from merging this as far as I'm concerned.

I am on it. It's quite still some work.

bbatsov:
> A valid point indeed. But C-c C-q is relatively hard to type; even so for C-c C-s C-q. I think it's worth the risk. In any event, if you plan to change this we need it to reconcile with sesman. It would be really akward if session level commands would differ from connection level in this regard.

Fair enough.

> It works the other way around, the interactive command sesman-start wraps generic sesman-start-session, a method for which is defined by CIDER.

Ah, got it. Guess I should have read the source first. :-)

> I am on it. It's still quite some work.

👍

dpsutton:
Suggestion: we should make a release right before this gets merged and let people know that melpa snapshot will be considered experimental while this gets hammered out. maybe an announcement in slack and also a post on reddit. Let everyone know we would appreciate feedback and bug reports but that the snapshot will be more experimental than usual.

Lots of people use the snapshot at work and a heads up about this invasive change would be welcome I think.

cider-interaction.el
```diff
@@ -1811,34 +1772,33 @@ refresh functions (defined in `cider-refresh-before-fn' and
...
-                   `("pprint-fn" ,(cider--pprint-fn)))
-                 (when (and (not inhibit-refresh-fns) cider-refresh-before-fn)
-                   `("before" ,cider-refresh-before-fn))
-                 (when (and (not inhibit-refresh-fns) cider-refresh-after-fn)
-                   `("after" ,cider-refresh-after-fn)))
-          (lambda (response)
-            (cider-refresh--handle-response response log-buffer))
-          conn)))
-     :clj 'any-mode)))
+    (cider-map-repls :clj
```
dpsutton: This seems strange to me. If you are in a cljs buffer and invoke `cider-refresh`, it would do nothing to your cljs repl while silently reloading your clj buffers. Shouldn't this be `:clj-strict`?

cider-mode.el
```diff
@@ -107,32 +100,25 @@ that of the namespace in the Clojure source buffer."
...
                                  (unless a-repl
-                                   (setq a-repl b))
-                                 (equal type (cider-connection-type-for-buffer b))))
+                                   (setq a-repl buf))
+                                 (equal type (cider-repl-type-for-buffer buf))))
                              (buffer-list))))
-    (if-let* ((repl (or the-repl a-repl)))
-        (cider--switch-to-repl-buffer repl set-namespace)
-      (user-error "No REPL found"))))
+    (let ((repl (or the-repl a-repl)))
```
dpsutton: previously this used `buffer-list` to get the list of buffers in order of their last usage I believe. I thought that sesman had a way to prioritize this now without scanning all open buffers which can be quite large (`sesman-more-relevant-p`). Shouldn't this find all connections of the type for the buffer and reduce over them with `sesman-more-relevant-p` ? and if non are found of the type then reduce over all linked sessions with `sesman-more-relevant-p` ?

cider-client.el
```diff
@@ -701,14 +163,13 @@ If ABORT-ON-INPUT is non-nil, the function will return nil
...
-                           (or connection (cider-current-connection))
+                           (or connection (cider-current-repl))
                            abort-on-input))
 
-(defun cider-nrepl-send-unhandled-request (request)
-  "Send REQUEST to the nREPL server and ignore any responses.
-Immediately mark the REQUEST as done.
-Return the id of the sent message."
-  (let* ((conn (cider-current-connection))
+(defun cider-nrepl-send-unhandled-request (request &optional connection)
```
bbatsov: You've renamed the functions, but you forgot about the params and local variables. It's odd if we refer to something as a repl in one place and a connection at another place.

cider-interaction.el
vspinu: I just swapped the new mapper and didn't question/check the implementation. The old :clj + 'any-mode should be equivalent to the new :clj  and old :clj is equivalent to :clj-strict. The latter throws an error if in wrong buffer.

cider-mode.el
vspinu: Buffers don't hold last-usage timestamp so `sesman-more-relevant-p` has to use buffer-list anyhow.  This is unlikely to be an issue in practice because sorting short-circuit on the most recent repl which is commonly located on the top of the list.

cider-client.el
vspinu: Hmm. I think the intent is not to rename all occurrences of connection. Semantics of the "connection" is more appealing in many situations. This is the case in this particular example and all other request/response handling functions. I am sending request through the "connection" not a "repl". I renamed only those functions which explicitely deal with the repl object.

It might be a good idea to keep `cider-current-connection` as an alias for `cider-current-repl` and use it in places when "connection" is more appropriate.

sesman.el
```diff
@@ -0,0 +1,722 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Commentary:
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Code:
+
+(require 'project)
+(require 'mule-util)
```
bbatsov: What do you need this for?

sesman.el
```diff
@@ -0,0 +1,722 @@
...
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Commentary:
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Code:
+
+(require 'project)
```
bbatsov: I don't think we've implemented any integration with project in clojure-mode/CIDER, so I'm not sure why this is needed.

cider-client.el
bbatsov: Yeah, I'm mostly concerned about code like `(let ((conn (cider-current-repl))...` and passing repls for conn params. I've grown used to this mess, but for newcomers all of this can be quite confusing.

sesman.el
vspinu: I don't. It's removed in the upstream sesman. I have already moved sesman.el out of cider as I need to prepare it for MELPA and it's quite a long requests back-log over there at the moment.

sesman.el
vspinu: It's needed by sesman to create links between sessions and projects. You might not like it, but that's the built-in generic way available from emacs25 to detect projects roots. More and more modes support it and I think  clojure-mode should too.

cider-client.el
vspinu: When we say "connection" we mean "process" and when we say "REPL" we mean the buffer. The source of the trouble is that all relevant to the connection values are stored as buffer local variables in the repl buffer. Thus, for convenience we pass buffers as connections. If we would use process plist to store connection parameters then we could pass the processes as "connections" and buffers as "repls". In fact we do pass the server process around but still use buffer-local variables to track its state. 

I would suggest that we implement this change after this PR. Working with process plists is a superior strategy anyhow.  Incidentally, I was half way through into implementing this idea more than two years ago, but then stupidly lost my code in a crash and have never picked it up again.

bbatsov: Yeah, that sounds like a good strategy.

sesman.el
bbatsov: Yeah, yeah - I'm very well aware what it is, but as everyone knows I have a different idea about how this should be done. :-)

I was mostly surprised you opted to use something we don't really support in `clojure-mode.el`, meaning it will just fallback to vcs roots.

sesman.el
bbatsov: Btw, would mind hosting the upstream version under `clojure-emacs`? As I told you before - a third party package I can't fix myself directly if needed is a big liability, especially if it's at the heart of CIDER.

I've good a good relationship with the MELPA maintainers, so hopefully we'll be able to speed along the process. Anyways, I'm guessing the initial merge is going to use this version of sesman.el committed here.

sesman.el
bbatsov: Hmm, I just grepped in Emacs's source code and I see none of the built-in modes are using `project.el`. :-)

vspinu: What alternative do you have in mind?  IMO, yet another module specific hack along `sesman-project-root-function` is inferior to what I have done.  

For a module which cares only about project root, `project.el` is an obvious choice. From an outsider prospective, if both clojre-mode and projectile would register their functions with `project-find-function` it would be a great step towards uniformization.

AFAICS the two alternatives to project.el are `projectile`, which is too heavy for the purpose at hand, and `project-root` which hasn't been updated on MELPA since 2011 and doesn't seem to be used much.

sesman.el
bbatsov: Yeah, making a co-owner of the project would be enough. I proposed `clojure-emacs` mostly because it gives projects bigger visibility and there are many people in the org, but I'm fine either way.

bbatsov:
Guess we're ready for mass testing and further polish. :-)

vspinu:
Ha. You were fast. I didn't even have time to test it myself after the green light :)  Let's dive then! 

Would be good to open related issues with a common tag (or 0.19 milestone). I will try to address things as we go.

vspinu:
I already see that there are a few obviously broken things. You were warned, don't be surprised if things don't work as expected.

bbatsov:
> Would be good to open related issues with a common tag (or 0.19 milestone). I will try to address things as we go.

Will do!

> Ha. You were fast. I didn't even have time to test it myself after the green light :) Let's dive then!

Well, now we got you more testers. :-) I definitely didn't want us to repeat the mistake of the past where we let the PR linger for way too long. Plus this one was already huge and I'd rather us continue with the improvements in a more incremental manner. 

> I already see that there are a few obviously broken things. You were warned, don't be surprised if things don't work as expected.

I warned people on twitter and slack about the big changes, hopefully only the adventurous people will update before the end of the week.

arichiardi:
I am going to put this here, `make build` has got some warning on latest master:

```
In end of data:
cider-repl.el:1733:1:Warning: the following functions are not known to be
    defined: cider-jack-in-clojure, cider-jack-in-clojurescript,
    cider-connect-clojure, cider-connect-clojurescript,
    cider-connect-sibling-clojure, cider-connect-sibling-clojurescript
```

vspinu:
Thanks. I forgot to declare those. Will fix in the next PR.

PR #2338, opened 2018-06-18
Pass cider-default-cljs-repl value to server's on-port-callback

### LINKED ISSUE #2336: New connection manager reads the wrong cider-default-cljs-repl
## Expected behavior

When I call `C-c M-J` the `cider-connect-sibling-cljs` should read the `cider-default-cljs-repl` from the code buffer's locals.

## Actual behavior
It always reads from the REPL buffer.

Have a `.dir-locals.el` like this in your folder:

```elisp
((nil . ((cider-default-cljs-repl . figwheel-main)))
```

`C-c M-J` in a code buffer. You will always be prompted no matter what is the value of `cider-default-cljs-repl`.

### CIDER version information
`latest master`

The problems seems in  [here](https://github.com/clojure-emacs/cider/blob/master/cider.el#L947), the variable should be read from the code buffer but it is read from the REPL instead.

vspinu:
Fixes #2336

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -899,10 +899,12 @@ Prompt for the project and nREPL server command when DO-PROMPT is non-nil."
   (interactive "P")
   (let ((cider-jack-in-dependencies (append cider-jack-in-dependencies cider-jack-in-cljs-dependencies))
         (cider-jack-in-lein-plugins (append cider-jack-in-lein-plugins cider-jack-in-cljs-lein-plugins))
-        (cider-jack-in-nrepl-middlewares (append cider-jack-in-nrepl-middlewares cider-jack-in-cljs-nrepl-middlewares)))
+        (cider-jack-in-nrepl-middlewares (append cider-jack-in-nrepl-middlewares cider-jack-in-cljs-nrepl-middlewares))
+        (cljs-repl-type cider-default-cljs-repl))
     (cider--jack-in do-prompt
       (lambda (server-buffer)
-        (cider-connect-sibling-cljs server-buffer)))))
+        (let ((cider-default-cljs-repl cljs-repl-type))
+          (cider-connect-sibling-cljs server-buffer))))))
 
 ;;;###autoload
 (defun cider-jack-in-clj&cljs (&optional do-prompt soft-cljs-start)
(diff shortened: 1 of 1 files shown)


cider.el
```diff
@@ -899,10 +899,12 @@ Prompt for the project and nREPL server command when DO-PROMPT is non-nil."
   (interactive "P")
   (let ((cider-jack-in-dependencies (append cider-jack-in-dependencies cider-jack-in-cljs-dependencies))
         (cider-jack-in-lein-plugins (append cider-jack-in-lein-plugins cider-jack-in-cljs-lein-plugins))
-        (cider-jack-in-nrepl-middlewares (append cider-jack-in-nrepl-middlewares cider-jack-in-cljs-nrepl-middlewares)))
+        (cider-jack-in-nrepl-middlewares (append cider-jack-in-nrepl-middlewares cider-jack-in-cljs-nrepl-middlewares))
+        (cljs-repl-type cider-default-cljs-repl))
     (cider--jack-in do-prompt
       (lambda (server-buffer)
-        (cider-connect-sibling-cljs server-buffer)))))
+        (let ((cider-default-cljs-repl cljs-repl-type))
```
bbatsov: I'm not a big fan of dynamic binding, as it makes the code harder to read. Maybe we should pass this as a param?

cider.el
```diff
@@ -944,8 +948,7 @@ Figwheel, etc), unless you've set `cider-default-cljs-repl'.  OTHER-REPL
 can also be a server buffer, in which case a new session with a REPL for
 that server is created."
   (interactive (list (cider-current-repl)))
-  (let ((cljs-repl-type (or cider-default-cljs-repl
-                            (cider-select-cljs-repl)))
+  (let ((cljs-repl-type (or cider-default-cljs-repl (cider-select-cljs-repl)))
```
bbatsov: Or we can obtain this differently here.

cider.el
vspinu: I think we should do it anyhow. `cider-connect-siblings-cljs` accepts other-repl as first and the only argument. This is useless for interactive use. So I would say we make two argument with the first one being either `cljs-repl-type` (or maybe more generic params) which would ask on C-u for repl type. Then even with dir-local `default-cljs-repl`  set one can spark a different repl type with `C-u C-c M-s`.

bbatsov: OK

Issue #2328, opened 2018-06-17
define cider-eval-last-sexp-at-point


zcaudate:
I'm hoping to have a functionality to be used for evaluating `->` and `->>` macros where:

--------

given the form:

```
(-> (+ 1 1)
    (+ 1)
    (+ 1)
    (+ 1))
```

it is possible to have evalution at subsequent steps, ie:

```
(-> (+ 1 1)| ; => 2
    (+ 1)
    (+ 1)
    (+ 1))
```
```
(-> (+ 1 1)
    (+ 1)| ;=> 3
    (+ 1)
    (+ 1))
```
```
(-> (+ 1 1)
    (+ 1)
    (+ 1)| ;=> 4
    (+ 1))
```
```
(-> (+ 1 1)
    (+ 1)
    (+ 1)
    (+ 1)|) ;=> 5
```
-------------------

The interaction would be to `move` the parenthesis forward to the point of the cursor, evaluate and then move the parenthesis back to the original position.

I tried but was a bit stuck. It'd be super useful for me if someone can implement it.

bbatsov:
Funny enough something like this already exists. See https://github.com/clojure-emacs/cider/commit/2db31b0069caf4331100207948ca2bcbbf217dc5

What I've implemented works only for top-level forms, but obviously we can use the same approach for inner forms as well.

zcaudate:
Oh that's awesome. Let me have a play with it.

zcaudate:
is this the function?

```
(defun cider-read-and-eval-defun-at-point ()
  "Insert the toplevel form at point in the minibuffer and output its result.
The point is placed next to the function name in the minibuffer to allow
passing arguments."
  (interactive)
  (let* ((fn-name (cadr (split-string (cider-defun-at-point))))
         (form (format "(%s)" fn-name)))
    (cider-read-and-eval (cons form (length form)))))
```

So is it possible to change the behaviour by replacing `cider-defun-at-point` with something like `cider-current-form`?

bbatsov:
No, the function is `cider-eval-defun-to-point`. Seems you want to create a variant of it - `cider-eval-sexp-to-point`.

zcaudate:
I have this function, and it works

```
(defun cider-eval-sexp-to-point ()
  (interactive)
  (let* ((begin (save-excursion (up-list) (backward-list) (point)))
         (code  (buffer-substring-no-properties begin (point)))
         (code  (concat code '(41))))
[...]

PR #2339, opened 2018-06-19
Added `cider-eval-sexp-to-point` for [Feature #2328]

zcaudate:

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/cider-interaction.el
+++ b/cider-interaction.el
@@ -1440,6 +1440,24 @@ It constructs an expression to eval in the following manner:
          (code (concat code (cider--calculate-closing-delimiters))))
     (cider-interactive-eval code)))
 
+(defun cider-eval-sexp-to-point (&optional  output-to-current-buffer)
+  "Evaluate the current sexp form up to point.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current buffer.
+It constructs an expression to eval in the following manner:
+
+- It finds the code between the point and the start of the sexp expression;
+- It balances this bit of code by closing the expression;
+- It evaluates the resulting code using `cider-interactive-eval'."
+  (interactive "P")
+  (let* ((beg-of-sexp (save-excursion (up-list) (backward-list) (point)))
+         (beg-delimiter (save-excursion (up-list) (backward-list) (char-after)))
+         (beg-set?  (save-excursion (up-list) (backward-list) (char-before)))
+         (code (buffer-substring-no-properties beg-of-sexp (point)))
+         (code (if (= beg-set? ?#) (concat (list beg-set?) code) code))
+         (code (concat code (list (cider--matching-delimiter beg-delimiter)))))
+    (cider-interactive-eval code
+                            (when output-to-current-buffer (cider-eval-print-handler)))))
+
 (defun cider-pprint-eval-defun-at-point (&optional output-to-current-buffer)
   "Evaluate the \"top-level\" form at point and pprint its value.
 If invoked with OUTPUT-TO-CURRENT-BUFFER, insert as comment in the current
@@ -1491,6 +1509,7 @@ passing arguments."
     (define-key map (kbd "r") #'cider-eval-region)
     (define-key map (kbd "n") #'cider-eval-ns-form)
     (define-key map (kbd "v") #'cider-eval-sexp-at-point)
+    (define-key map (kbd "o") #'cider-eval-sexp-to-point)
     (define-key map (kbd ".") #'cider-read-and-eval-defun-at-point)
     (define-key map (kbd "z") #'cider-eval-defun-to-point)
     (define-key map (kbd "c") #'cider-eval-last-sexp-in-context)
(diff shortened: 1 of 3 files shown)


cider-interaction.el
```diff
@@ -1440,6 +1440,22 @@ It constructs an expression to eval in the following manner:
...
+
+- It find the code between the point and the start of the sexp expression;
+- It balances this bit of code by closing the expression;
+- It evaluates the resulting code using `cider-interactive-eval'
+
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current buffer."
+  (interactive "P")
+  (let* ((beg-of-sexp (save-excursion (up-list) (backward-list) (point)))
+         (code (buffer-substring-no-properties beg-of-sexp (point)))
+         (code (concat code '(41))))
```
bbatsov: Hmm, seems to me you hardcode a single closing paren here, which might be incorrect depending on the sexp, right?

bbatsov: Here's a simple problem with the current solution - the current sexp might not be a list. Imagine you call this function in some vector or hash literal for instance. Why didn't you use the matching delimiter functionality which already exists?

zcaudate:
I've added additional code to support sets as well.

bbatsov:
The build is failing due to some problems with the docstring. See my inline comments.

cider-interaction.el
```diff
@@ -1440,6 +1440,23 @@ It constructs an expression to eval in the following manner:
          (code (concat code (cider--calculate-closing-delimiters))))
     (cider-interactive-eval code)))
 
+(defun cider-eval-sexp-to-point (&optional  output-to-current-buffer)
```
bbatsov: Why did you add this param in the first place? It makes no sense to me.

cider-interaction.el
```diff
@@ -1440,6 +1440,23 @@ It constructs an expression to eval in the following manner:
...
+  "Evaluate the current sexp form up to point.
+It constructs an expression to eval in the following manner:
+
+- It find the code between the point and the start of the sexp expression;
+- It balances this bit of code by closing the expression;
+- It evaluates the resulting code using `cider-interactive-eval', working with C-u."
```
bbatsov: Related to the first comment - no idea why you did this like that. You can simply write in `C-u` in a docstring, btw, and you have to refer explicitly to all params. I'd just remove the param.

cider-interaction.el
```diff
@@ -1440,6 +1440,23 @@ It constructs an expression to eval in the following manner:
...
+  (interactive "P")
+  (let* ((beg-of-sexp (save-excursion (up-list) (backward-list) (point)))
+         (beg-delimiter (save-excursion (up-list) (backward-list) (char-after)))
+         (beg-set?  (save-excursion (up-list) (backward-list) (char-before)))
+         (code (buffer-substring-no-properties beg-of-sexp (point)))
+         (code (if (= beg-set? ?#) (concat (list beg-set?) code) code))
+         (code (concat code (list (cider--matching-delimiter beg-delimiter)))))
+    (cider-interactive-eval code
```
bbatsov: This should be just a simple `cider-interactive-eval` to be consistent with other commands we have.

zcaudate: outputting is really important (for me anyways) It's for inline editing:

out

cider-interaction.el
zcaudate: I didn't do the merge right. will resubmit

cider-interaction.el
bbatsov: I understand you're point but we need to be uniform in our command interface. There are dedicated commands for evaling something to comment, which is what you seem to want (.e.g. `cider-pprint-eval-last-sexp-to-comment` ). It seems to me we just need two command. The basic command that evals a sexp up to point and another command that wraps the first one and provides the pprint-to-comment functionality.

zcaudate: well... I copied this function:

```
(defun cider-eval-last-sexp (&optional output-to-current-buffer)
  "Evaluate the expression preceding point.
If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current buffer."
  (interactive "P")
  (cider-interactive-eval nil
                          (when output-to-current-buffer (cider-eval-print-handler))
                          (cider-last-sexp 'bounds)))
```

the great thing is that C-x C-e outputs as a message whilst C-u C-x C-e outputs to the buffer.

bbatsov: Oh, that's a great point. Turns out it's me who made a mistake by forgetting to include this in eval-defun-to-point. :-)

cider-mode.el
```diff
@@ -233,6 +234,7 @@ Configure `cider-cljs-repl-types' to change the ClojureScript REPL to use for yo
     ["Insert ns form in REPL" cider-insert-ns-form-in-repl]
     ["Eval top-level sexp to comment" cider-eval-defun-to-comment]
     ["Eval top-level sexp and pretty-print to comment" cider-pprint-eval-defun-to-comment]
+    ["Eval top-level sexp to point" cider-eval-defun-to-point]
```
bbatsov: I already this command in the menu (a bit higher up).

cider-interaction.el
```diff
@@ -1440,6 +1440,23 @@ It constructs an expression to eval in the following manner:
          (code (concat code (cider--calculate-closing-delimiters))))
     (cider-interactive-eval code)))
 
+(defun cider-eval-sexp-to-point (&optional  output-to-current-buffer)
+  "Evaluate the current sexp form up to point.
+It constructs an expression to eval in the following manner:
+
+- It finds the code between the point and the start of the sexp expression;
+- It balances this bit of code by closing the expression;
+- It evaluates the resulting code using `cider-interactive-eval', if invoked with OUTPUT-TO-CURRENT-BUFFER, insert as comment in the current buffer."
```
bbatsov: I'd structure the docstring as in here - https://github.com/clojure-emacs/cider/commit/28bdc1f2897d351f8a1761e4cf121a1dd353a667

This avoid the very long line.

PR #2343, opened 2018-06-21
Fix unbalanced parentheses

dirceu:
This was broken on 28bdc1f2897d351f8a1761e4cf121a1dd353a667.


bbatsov:
Ops, sorry about that! And thanks for fixing it!

PR #2345, opened 2018-06-22
Optionally pass repl-type to cider-insert-in-repl

arichiardi:
Necessary in case you want to override the autodetection, which at the moment
is flimsy.

-----------------

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -200,12 +200,14 @@ and eval and the prefix is required to prevent evaluation."
   :group 'cider
   :package-version '(cider . "0.18.0"))
 
-(defun cider-insert-in-repl (form eval)
+(defun cider-insert-in-repl (form eval &optional repl-type)
   "Insert FORM in the REPL buffer and switch to it.
-If EVAL is non-nil the form will also be evaluated."
+If EVAL is non-nil the form will also be evaluated.
+
+Optionally pass the REPL-TYPE on which you want to act."
   (while (string-match "\\`[ \t\n\r]+\\|[ \t\n\r]+\\'" form)
     (setq form (replace-match "" t t form)))
-  (with-current-buffer (cider-current-repl)
+  (with-current-buffer (cider-current-repl repl-type)
     (goto-char (point-max))
     (let ((beg (point)))
       (insert form)


bbatsov:
I don't get the point of this change, so some explanations would be welcome. You think the connection dispatch is not picking the right REPL or what?

arichiardi:
@bbatsov yes when I try to invoke the function it seems that the REPL type is not identified correctly (with `shadow`), so I need to pass `"clj"` in. This is also good for completeness.

arichiardi:
Rebased and fixed the linter.

bbatsov:
> @bbatsov yes when I try to invoke the function it seems that the REPL type is not identified correctly (with shadow), so I need to pass "clj" in. This is also good for completeness.

The current repl type should be clear from the file you're visiting, so I don't think this is a good idea. Shouldn't you use `cider-set-repl-type` instead to fix the incorrect repl type?

arichiardi:
I am using the function `cider-insert-in-repl` to send my own string from the code buffer to the REPL, but the function contains `current-repl-buffer`, which fails for `shadow`. This is why I need to kind of override...Maybe the new `defcustom` there might help actually...

arichiardi:
@bbatsov I copied that function over to my local code. I still feel this should go in to solve the current mismatch between the visited buffer and the underlying REPL. The scenario is this, which I know is handled in another way in `cider`, but I am going to write here again anyways:

 * open terminal
 * launch clojure REPL
 * connect to it with `cider-connect`
 * try to send an upgrade form from a cljs buffer *fail* - no "cljs" buffer found

Probably @vspinu has a better way to handle this now.

arichiardi:
this is what I am using for insertion now:

```
(with-current-buffer (cider-current-repl "clj")
      (goto-char (point-max))
      (let ((beg (point)))
        (insert form)
        (indent-region beg (point)))
      (cider-repl-return))
```

arichiardi:
Alternative to manual insertion:

```
(cider-register-cljs-repl-type
   'shadow-select
   #'ar-emacs-cljs-shadow-select-init-form
   #'cider-check-shadow-cljs-requirements)
```

arichiardi:
No, too soon:

`user-error: No cljs REPLs found.  Have you linked a session?`

Really need help on this one.

arichiardi:
Oh right, so now I need to detect the REPL - switch on auto detection again ;)

arichiardi:
I think I found an alternative but you are right it should be fixed properly. Closing.

PR #2348, opened 2018-06-24
Fix formatting error in interactive_programming.md

Blevs:
Fix a formatting error in interactive_programming.md by adding a missing '`' after cider-describe-current-connection.


bbatsov:
👍

PR #2355, opened 2018-06-28
Connection api fixes

### LINKED ISSUE #2340: cider-quit does not quit the active session
## Expected behavior
After a successful `cider-jack-in-cljs`, <kbd>C-c C-q</kbd> should quit the active CIDER session.

## Actual behavior
<kbd>C-c C-q</kbd> does nothing, the CLJS REPL is still present (and, in my case, working with Figwheel).

- create a new project with `lein new`
- add `[org.clojure/clojure "1.9.0"]` and `[org.clojure/clojurescript "1.9.946"]` as dependencies.
- create a new CLJS file
- hit <kbd>C-c M-J</kbd> or <kbd>M-x cider-jack-in-cljs</kbd>
- hit <kbd>C-c C-q</kbd> or <kbd>M-x cider-quit</kbd>

### CIDER version information
```
;; CIDER 0.18.0snapshot (package: 20180617.2338), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_171
```

### System details
```
GNU Emacs 27.0.50 (build 1, x86_64-debian-linux-gnu, GTK+ Version 3.22.30)
 of 2018-06-19

Repository revision: 3057260b476a5c8efde523d0e5e72202b96ed9d1
Configured using:
--host=x86_64-debian-linux-gnu

Emacs uptime: 15 minutes, 50 seconds
Colour theme: sanityinc-tomorrow-night
Operating system: Ubuntu 18.04 LTS
Window system: x11
Desktop environment: GNOME Shell 3.28.1
```

### LINKED ISSUE #2342: Add warnings about things that were removed
I've noticed that @vspinu add many warnings for the things that were removed (I noticed just one). Probably it'd nice to add a few of them with pointers to whatever replaced the removed stuff (otherwise end users and extension writers would struggle with some aspects of the upgrade). 

Plus, people should know which configuration items to drop from their init.el's (e.g. `cider-connection-dispatch`).

### LINKED ISSUE #2349: C-c M-d `(wrong-type-argument stringp nil)`
I receive this error when running `C-c M-d` in a clj or cljs repl if the last visited repl is of the opposite type.
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  cider--connection-info(nil)
  cider-describe-current-connection()
  funcall-interactively(cider-describe-current-connection)
  call-interactively(cider-describe-current-connection nil nil)
  command-execute(cider-describe-current-connection)
```

There are also error when trying to evaluate code by any means.

### CIDER version information

```
;; CIDER 0.18.0snapshot, nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_172
```

(Commit `527466cb1dc73014b411a54a1c6c3061f68b7cfd`)

### Emacs version

25.3.1

### Operating system

NixOS

### LINKED ISSUE #2350: Fix build warnings after new connection API changes
Just a reminder for myself of whoever can do it first.

## Expected behavior
No warnings.

## Actual behavior
```
In end of data:
cider-connection.el:589:1:Warning: the function ‘cider-mode’ is not known to
    be defined.

...

In end of data:
cider-repl.el:1742:1:Warning: the following functions are not known to be
    defined: cider-jack-in-clojure, cider-jack-in-clojurescript,
    cider-connect-clojure, cider-connect-clojurescript,
    cider-connect-sibling-clojure, cider-connect-sibling-clojurescript
```
make build

### CIDER version information

```
;; CIDER 0.18.0snapshot latest master
```

vspinu:
A range of simple comits which should fix #2350, fix #2349, fix #2342 and fix #2340.

I am strugling with cask install locally. For unclear reason it pulls an one week old sesman. Both MELPA and MELPA-stable have up to date sesman. Any ideas?


vspinu:
Same happens on travis. Old sesman is pulled there as well.

vspinu:
Looks like it's a melpa issue actually (melpa/melpa#5580).  I was mislead by the github badge which was up to date.

cider-repl.el
```diff
@@ -1027,10 +1027,12 @@ text property `cider-old-input'."
...
   ;; FIXME: implement cycling as session can hold more than two REPLs
-  (if-let* ((this-repl (cider-current-repl))
-            (other-repls (seq-remove (lambda (r) (eq r this-repl)) (cider-repls))))
-      (switch-to-buffer (car other-repls))
-    (message "There's no other REPL for the current project")))
+  (let* ((this-repl (cider-current-repl nil t))
+         (other-repl (car (seq-remove (lambda (r) (eq r this-repl)) (cider-repls nil t)))))
+    (if other-repl
+        (switch-to-buffer other-repl)
+      (user-error "No other REPL in current session (%s)"
```
bbatsov: I think it'd be much better to use the type of the other REPL in this message, instead of the vague word "other". E.g. "No ClojureScript REPL...".

cider-connection.el
```diff
@@ -517,32 +516,42 @@ function with the repl buffer set as current."
 
 ;;; Current/other REPLs
 
-(defun cider-current-repl (&optional type)
+(defun cider-current-repl (&optional type ensure)
```
bbatsov: I'm puzzled by this change. Can you even have a REPL without a session?

cider-connection.el
```diff
@@ -337,7 +336,7 @@ Don't restart the server or other connections within the same session.  Use
 (defun cider-describe-current-connection ()
   "Display information about the current connection."
   (interactive)
-  (message "%s" (cider--connection-info (cider-current-repl))))
+  (message "%s" (cider--connection-info (cider-current-repl nil t))))
```
bbatsov: I prefer passing symbols like `'ensure-session` instead of `t` for such boolean params. That makes the code much more readable.

cider-connection.el
bbatsov: P.S. I'd also name this param `ensure-session`.

cider.el
```diff
@@ -1177,7 +1177,7 @@ choose."
 
 ;; TODO: Implement a check for command presence over tramp
 (defun cider--resolve-command (command)
-  "Find COMMAND on `exec-path' if possible, or return nil.
+  "Find COMMAND in list variable ‘exec-path’ if available, or return nil.
```
bbatsov: You'll have to explain to me this change. Seems pretty weird to me.

bbatsov:
So, we'll have to wait until the proper build is generated, right?

cider-repl.el
vspinu: This function is a bit more generic now. It will work with sessions containing two cljs connections for example. So the other here literally means "other repl besides the current one". 

This function should be fixed to cycle through all of the repls in case there are more than two. Will fix that in due time.

cider-connection.el
vspinu: Ensure means that it will throw an error if there is no repl. And it means that it will throw both when there is no session and when there is no repl of type TYPE within a session. We cannot have sessions with no repls, but we can have sessions consisting of, for example, clj repls only.

cider-connection.el
vspinu: Sure, will change. But I am a bit divided on this one.  For frequently used functions like `cider-current-repl` everyone will know what 2nd parameter means, so adding `'ensure` just eats space for nothing.

cider-connection.el
```diff
@@ -587,6 +581,64 @@ Error is signaled if no REPL buffer of specified type exists."
         (user-error "No %s REPLs found.  Have you linked a session?" type))
       (mapcar function repls))))
 
+
+;; Aliases and Deprecation after #2324
+
+;; REPLs double as connections in CIDER, so it's useful to be able to refer to
+;; them as connections in certain contexts.
+(defalias 'cider-current-connection #'cider-current-repl)
```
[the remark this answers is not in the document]
vspinu: These 4 for a unit of related functionality with the inline comment above relating to all of them. Plus all of these functions are defined right there an inch above.

cider-connection.el
```diff
@@ -587,6 +581,64 @@ Error is signaled if no REPL buffer of specified type exists."
...
+
+(define-obsolete-function-alias 'cider-current-repl-buffer 'cider-current-repl "0.18")
+(define-obsolete-function-alias 'cider-repl-buffers 'cider-repls "0.18")
+(define-obsolete-function-alias 'cider-current-session 'cider-nrepl-eval-session "0.18")
+(define-obsolete-function-alias 'cider-current-tooling-session 'cider-nrepl-tooling-session "0.18")
+(define-obsolete-function-alias 'cider-display-connection-info 'cider-describe-current-connection "0.18")
+(define-obsolete-function-alias 'nrepl-connection-buffer-name 'nrepl-repl-buffer-name "0.18")
+(define-obsolete-function-alias 'cider-repl-set-type 'cider-set-repl-type "0.18")
+
+(make-obsolete 'cider-assoc-buffer-with-connection 'sesman-link-with-buffer "0.18")
```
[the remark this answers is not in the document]
vspinu: I am not sure what you mean here, but `make-obsolete` is used here for all functions which don't have one-to-one mapping into new functionality.

cider.el
vspinu: It is, but checkdoc insists to disambiguate `exec-path` with `variable` or `function` in front. Pretty silly but I don't see how this could be done otherwise without disabling the check or removing `exec-path`. "list variable `exec-path` is a bit more meaningful than `in variable exec-path`.  An alternative would be to just write `in exec path`.

vspinu:
Yes. Once MELPA is on I will re-base with your suggestions.

vspinu:
MELPA is back. I have addressed your comments except for the location of connection aliases. It's the most natural grouping in my mind, but feel free to move around as you see fit. They won't be aliases much longer though, if follow with the connection <=> process refactoring.

Tests fail due to a bug in clojure-mode. Should be fine once fixed.

bbatsov:
> They won't be aliases much longer though, if follow with the connection <=> process refactoring.

Yeah, that'd be best. Btw, now that sesman is standalone, please don't forget to make me a co-owner of the project.

cider-connection.el
bbatsov: Btw, why is the type a string instead of a symbol? That seems a bit unlispy to me.

vspinu: :) that's because `cider-repl-type` returns a string, which in turn (my guess) is because the state handler returns a type as a string. So I just don't want to convert things while passing around. 

It might be good idea to change the type to a keyword throughout. It's a keyword in `cider-map-connections` argument.

PR #2356, opened 2018-06-29
classpath libs support on Windows

kevindragon:
handle classpath on windows with directory seperator `\` when call cider-classpath-libs.


bbatsov:
Thanks!

PR #2360, opened 2018-06-29
[Fix #2317] Allow cancelling stdin prompt

### LINKED ISSUE #2317: (slurp *in*) can't be interrupted
This is just creating a ticket for https://stackoverflow.com/questions/50493580/cider-for-emacs-stdin

It seems that slurping from the stdin right now can't be interrupted, which should certainly be possible.

TimoFreiberg:
* Add handler for the `quit` signal which sends an empty message to nREPL
* Add C-c C-c binding for `abort-recursive-edit` (which sends a `quit` signal)

I did not add any tests so far. The only thing I think makes sense to test is the quit signal handler that returns nil. I will add a test if you want me to :)

I did not add an entry to the user documentation because there is no Stdin minibuffer section yet and I feel it's not necessary to add an entry only to document that `C-c C-c` cancels the input.

-----------------

- [not needed, I think?] You've updated the [user manual][4] (if adding/changing user-visible functionality)

Thanks!

manual][2] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
+* [#2317](https://github.com/clojure-emacs/cider/issues/2317): The stdin prompt can now be cancelled.
 * [#2328](https://github.com/clojure-emacs/cider/issues/2328): Added `cider-eval-sexp-to-point`.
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks good! Thank you!

TimoFreiberg:
Glad I could help!

PR #2364, opened 2018-06-30
[Fix #2341] Update connection management section in the manual

### LINKED ISSUE #2341: Update the section of the manual regarding connection management
All of this http://www.cider.mx/en/latest/managing_connections/ has to be rewritten in light of the recent changes by @vspinu 

Some menu entries and other references to jack-in/connect commands should also be updated before the next release.

vspinu:
First version of the connection management docs. 

I moved the section up close to the "Using the REPL" because it seems related.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -88,7 +88,7 @@ CIDER packs plenty of features. Here are some of them (in no particular order):
 * [Scratchpad](miscellaneous_features.md#using-a-scratchpad)
 * [Minibuffer code evaluation](miscellaneous_features.md#evaluating-clojure-code-in-the-minibuffer)
 * Integration with [company-mode][]
-* [Support for working with multiple simultaneous nREPL connections](managing_connections.md)
+* [Support for working with multiple REPLs](managing_connections.md)
 
 ![CIDER Screenshot](images/cider-overview.png)
 
--- a/doc/managing_connections.md
+++ b/doc/managing_connections.md
@@ -1,27 +1,93 @@
-## Overview
+Because connections map one-to-one to REPL buffers, for the purpose of this
+section we use "REPL" and "connection" interchangeably.
 
-You can connect to multiple nREPL servers using <kbd>M-x</kbd> `cider-jack-in`
-(or `cider-connect`) multiple times.  To close the current nREPL connection, use
-<kbd>C-c C-q</kbd> (`cider-quit`). You can close all connections with
-<kbd>C-u C-c C-q</kbd>.
+## Sessions
 
-CIDER maintains a list of nREPL connections and a single 'default'
-connection. When you execute CIDER commands in a Clojure editing buffer such as
-to compile a namespace, these commands are executed against a specific
-connection.
+CIDER maintains a grouped view of opened connections through [Sesman]
+sessions. Each session is a collection of connections which share same nREPL
+server.
 
-You can display the current nREPL connection using <kbd>C-c M-d</kbd>
-and rotate the default connection using <kbd>C-c M-r</kbd>.
+Start new sessions with
 
-## Switch to connection buffer
+   - <kbd>C-c M-j</kbd> `cider-jack-in-clj`
+   - <kbd>C-c M-J</kbd> `cider-jack-in-cljs`
+   - <kbd>C-c M-c</kbd> `cider-connect-clj`
+   - <kbd>C-c M-C</kbd> `cider-connect-cljs`
+   - <kbd>M-x</kbd> `cider-jack-in-clj&cljs`
 
-The REPL buffers double as connection buffers.
+Add new REPLs to the current session with
 
-To switch to the relevant REPL buffer based on the Clojure namespace
-in the current Clojure buffer, use: <kbd>C-c C-z</kbd>. You can then
-use the same key combination to switch back to the Clojure buffer you
-came from.
+   - <kbd>C-c M-s</kbd> `cider-connect-sibling-clj`
+   - <kbd>C-c M-S</kbd> `cider-connect-sibling-cljs`
 
-The single prefix <kbd>C-u C-c C-z</kbd>, will switch you to the
-relevant REPL buffer and set the namespace in that buffer based on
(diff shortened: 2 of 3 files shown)


doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
-You can display the current nREPL connection using <kbd>C-c M-d</kbd>
-and rotate the default connection using <kbd>C-c M-r</kbd>.
+Start new sessions with
 
-## Switch to connection buffer
+   - <kbd>C-c M-j</kbd> `cider-jack-in-clj`
+   - <kbd>C-c M-J</kbd> `cider-jack-in-cljs`
+   - <kbd>C-c M-c</kbd> `cider-connect-clj`
+   - <kbd>C-c M-C</kbd> `cider-connect-cljs`
+   - <kbd>M-x</kbd> `cider-jack-in-clj&cljs`
```
bbatsov: I think you forgot something here.

doc/index.md
```diff
@@ -88,7 +88,7 @@ CIDER packs plenty of features. Here are some of them (in no particular order):
 * [Scratchpad](miscellaneous_features.md#using-a-scratchpad)
 * [Minibuffer code evaluation](miscellaneous_features.md#evaluating-clojure-code-in-the-minibuffer)
 * Integration with [company-mode][]
-* [Support for working with multiple simultaneous nREPL connections](managing_connections.md)
+* [Support for working with multiple simultaneous REPLs](managing_connections.md)
```
bbatsov: I guess `simultaneous` is redundant now.

doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
 
-The REPL buffers double as connection buffers.
+Add new REPLs to the current session with
 
-To switch to the relevant REPL buffer based on the Clojure namespace
-in the current Clojure buffer, use: <kbd>C-c C-z</kbd>. You can then
-use the same key combination to switch back to the Clojure buffer you
-came from.
+   - <kbd>C-c M-s</kbd> `cider-connect-sibling-clj`
+   - <kbd>C-c M-S</kbd> `cider-connect-sibling-cljs`
```
bbatsov: Btw, is there currently some visual feedback for users that they created a new session/adding something to an existing session?

doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
+
+  - <kbd>C-c C-s b</kbd> `sesman-link-with-buffer`
+  - <kbd>C-c C-s d</kbd> `sesman-link-with-directory`
+  - <kbd>C-c C-s p</kbd> `sesman-link-with-project`
+  - <kbd>C-c C-s u</kbd> `sesman-unlink`
+
+## Current Session
+
+All CIDER commands (evaluation, completion, switching to REPL etc.) operate on
+the current REPL within the current session. Current session is the most
```
bbatsov: The current

doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
+"relevant" session among all linked session. Session relevance is decided by the
+specificity of the linked contexts and recency of the REPL buffers.
```
bbatsov: I guess you can briefly mention what the possible contexts here.

doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
+
+Retrieve info on the current session with <kbd>C-c C-s i</kbd>
+(`sesman-show-session-info`). With a single prefix <kbd>C-u</kbd> it shows info
+on all linked sessions and with double prefix <kbd>C-u C-u</kbd> on all CIDER
+sessions. Show current links with <kbd>C-c C-s l</kbd> (`sesman-show-links`).
+
+
+## Current REPL
+
+Current REPL is the most relevant REPL from the current session. REPL relevance
```
bbatsov: The current

doc/managing_connections.md
```diff
@@ -1,27 +1,92 @@
...
+is determined by the type of current buffer. For example if current buffer is a
```
bbatsov: of the current/if the current

doc/managing_connections.md
vspinu: No, that command is not bound to anything, it's on M-x.

bbatsov: Hmm, it's a bit confusing you mentioned the other commands with a different notation then - I just assumed your intent was to do keybinding / command-name. 

Anyways, not a big deal.

doc/managing_connections.md
vspinu: New REPL is created. In principle buffer mode line could include name of current session and number of session within but I am afraid that might be more confusing because session names are same as project names. In any case the session naming and REPL buffer names should be reworked, so I am not paying much attention to it at the moment.

Or do you have something else in mind?

bbatsov: > New REPL is created.

Well, the sibling commands are going to create new REPLs as well, so I can't see how a user would notice the creation a new session.

> In any case the session naming and REPL buffer names should be reworked, so I am not paying much attention to it at the moment.

Yeah, that's what I mean basically - the session info should be prominent, so people would:

* understand what's their current session
* understand when they've created new sessions

That's just a comment for a future improvement, not something to be done in the scope of this PR. I think that most likely we'll have to replace the project name with the session name or something like this.

doc/managing_connections.md
vspinu: I don't think it's needed. It's a generic concept being defined in that sentence.  One doesn't use "the" in definitions. You say "Cat is an animal", but not "The cat is an animal".

doc/managing_connections.md
vspinu: I mentioned them in the preceding paragraph and I am explaining that very sentence in the following two paragraphs. The sentence is heavy already, so I would not load it further.

doc/managing_connections.md
vspinu: Here it's needed indeed.

vspinu:
Done!

PR #2365, opened 2018-06-30
Correctly format shadow command when it is "npx shadow-cljs"

dpsutton:
Previously called `(cider--resolve-command (car (split-string
cider-shadow-cljs-command)))` on "npx shadow-cljs" would return "npx"
and drop the shadow-cljs. This would lead to the error message that
server was not a valid option as the command was "npx [cider-stuff]
server" instead of "npx shadow-cljs [cider-stuff] server"

Old version would create commands like
```bash
/home/dan/.nvm/versions/node/v8.9.2/bin/npx -d org.clojure/tools.nrepl:0.2.13 -d cider/piggieback:0.3.5 -d cider/cider-nrepl:0.18.0-SNAPSHOT server
```

which omits the all-important `shadow-cljs` binary to run with npx

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -327,7 +327,9 @@ Throws an error if PROJECT-TYPE is unknown.  Known types are
     ("clojure-cli" (cider--resolve-command cider-clojure-cli-command))
     ;; here we have to account for the possibility that the command is either
     ;; "npx shadow-cljs" or just "shadow-cljs"
-    ("shadow-cljs" (cider--resolve-command (car (split-string cider-shadow-cljs-command))))
+    ("shadow-cljs" (let ((parts (split-string cider-shadow-cljs-command)))
+                     (when-let* ((command (cider--resolve-command (car parts))))
+                       (mapconcat #'identity (cons command (cdr parts)) " "))))
     ("gradle" (cider--resolve-command cider-gradle-command))
     (_ (user-error "Unsupported project type `%s'" project-type))))
 


cider.el
```diff
@@ -327,7 +327,9 @@ Throws an error if PROJECT-TYPE is unknown.  Known types are
     ("clojure-cli" (cider--resolve-command cider-clojure-cli-command))
     ;; here we have to account for the possibility that the command is either
     ;; "npx shadow-cljs" or just "shadow-cljs"
-    ("shadow-cljs" (cider--resolve-command (car (split-string cider-shadow-cljs-command))))
+    ("shadow-cljs" (let ((parts (split-string cider-shadow-cljs-command)))
```
bbatsov: The actual problem is different. We don't really want to used the resolved commands. I thought I had changed this, now I'm puzzled how this went back to what it was before. (I mean that we check for `npx`, but then run `cider-shadow-cljs-command`).

dpsutton: oh i thought the resolved command was aware of different hosts and such. these functions all return the `shell-quoted-argument` of the command.

I don't understand what npx does then. And checking for npx doesn't guarantee that shadow-cljs is installed right? Should we check that shadow-cljs is installed and then just run it with npx shadow-cljs if that's what the user's shadow command is?

arichiardi: I think your strategy is correct @dpsutton - one way to detect if `shadow` is installed would be:

```
npm ls --depth=0 | grep shadow-cljs
```

but it could be installed globally as well so this should also be attempted:

```
npm ls --global --depth=0 | grep shadow-cljs
```

The `npx` always find the executable - does not matter if global or local - if I remember correctly.

dpsutton: sounds good.  I was learning some about shadow-cljs (amazing software) and ran into this issue and remember people having a similar issue in the past. I was confused because you left a comment saying you were dealing with this case but the return value did not deal with it. Getting lost in the merge makes sense. :)

bbatsov: Btw, did you test your changes with a single entry command (e.g. just `shadow-cljs`)? I guess we should add some tests for the two cases. I recall in the beginning I had a similar idea in mind and planned to put this directly in `cider--resolve-command`.

dpsutton: yes i did. works equally well with `npx shadow-cljs` and `shadow-cljs`. I can add two simple tests this evening if you would like that.

bbatsov: Yeah, let's have them mostly to future proof things.

Issue #2352, opened 2018-06-28
cider-load-file broken


abo-abo:
Results in:

    error in process filter: user-error: No clj REPLs found.  Have you linked a session?

bbatsov:
Are you using this in a different project? Not the one in which you jacked-in initially perhaps?

The message is very confusing and we have to improve it, but basically this happens when there's no session linked to the project you're currently visiting. Before 0.18 we didn't check the project in any of the eval commands and they'd just get dispatched to whatever was considered the current REPL, but this behaviour is now stricter (so people won't end up evaluating code in the wrong REPL). 

//cc @vspinu

abo-abo:
> Are you using this in a different project? Not the one in which you jacked-in initially perhaps?

No, using it from a jacked-in project. Simply removing the checks makes it work:

```elisp
(defun lispy-cider-load-file (filename)
  (let ((ns-form  (cider-ns-form)))
    (cider-map-repls :auto
      (lambda (connection)
        (when ns-form
          (cider-repl--cache-ns-form ns-form connection))
        (cider-request:load-file (cider--file-string filename)
                                 (funcall cider-to-nrepl-filename-function
                                          (cider--server-filename filename))
                                 (file-name-nondirectory filename)
                                 connection)))))
```

vspinu:
Looks like it's either an older version of CIDER where sesman wasn't correctly matching current directory to the project path, or it's a variant of #2351 where VC sub-directories interfere with the clojure projects.

abo-abo:
I'd like to call for some robustness here. There's no reason why I should get "no session associated" message when in 100% of my use cases I only have a single CIDER session.

I understand that it's useful to accommodate workflows with multiple sessions. But it should not be done at the expense of edge cases where the simple case of a single REPL stops working.

vspinu:
I intentionally left the configuration of how associations work in single-session workflow for later. I think we first have to be crystal clear about the semantics of session management in multi-session context and then add a custom variable which would allow a lax session matching. 

But I am not convinced about the "robustness" argument though. I am incline to think that the lax option should be off by default. It's too easy to overlook that you no-longer have 2 sessions running and send wrong stuff, kill or interrupt the (now) only REPL.

This might not feel like a big deal for clojure users, but in my other use case regarding analytics in R, when I have some sessions running for hours I really don't want to send anything wrong or kill the last session by mistake.

bbatsov:
[...]

PR #2366, opened 2018-06-30
Preserve current context in cider-load-file and cider-load-buffer

vspinu:
The motivation is in this [comment](https://github.com/clojure-emacs/cider/issues/2352#issuecomment-401570948).

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1024,45 +1024,47 @@ passing arguments."
 
 (defun cider-load-buffer (&optional buffer)
   "Load (eval) BUFFER's file in nREPL.
-If no buffer is provided the command acts on the current buffer.
-
-If the buffer is for a cljc file, and both a Clojure and
-ClojureScript REPL exists for the project, it is evaluated in both REPLs."
+If no buffer is provided the command acts on the current buffer.  If the
+buffer is for a cljc file, and both a Clojure and ClojureScript REPL exists
+for the project, it is evaluated in both REPLs."
   (interactive)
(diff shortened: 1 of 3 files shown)


cider-eval.el
```diff
@@ -1024,54 +1024,55 @@ passing arguments."
 
 (defun cider-load-buffer (&optional buffer)
   "Load (eval) BUFFER's file in nREPL.
-If no buffer is provided the command acts on the current buffer.
-
-If the buffer is for a cljc file, and both a Clojure and
-ClojureScript REPL exists for the project, it is evaluated in both REPLs."
+If no buffer is provided the command acts on the current buffer.  If the
```
bbatsov: Btw, a note for change we have to do - as @dpsutton mentioned this behavior for cljc files should be configurable. Some people don't want to load/eval stuff in both type of repls. Maybe we need a follow up ticket, so won't forget it. I postponed changing this for after you're done with the big PR.

vspinu: Agreed. This part will require a bit of thinking. But I don't see how a simple custom could do a good job here. First, one can always create separate sessions for clj and cljs interaction. Second, adding and removing various species of sibling REPls to a session is now trivial. Third, such a custom would not generalize well to more complex scenario (e.g. multiple cljs REPLs).

I think what could work well is an interactive toggling or cycling command that would temporary narrow interaction with a session to one REPL in that session.  But let's wait and see if people really need such an extra layer of connection micro-management, or existing tools could be just enough.

PR #2367, opened 2018-07-01
Require cider-find from cider.el

dpsutton:
Nothing loaded this code from cider previously leading to
cider-find-var being undefined. Probably other things broken as well.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -79,6 +79,7 @@
 (require 'cider-common)
 (require 'cider-compat)
 (require 'cider-debug)
+(require 'cider-find)
 
 (require 'tramp-sh)
 (require 'subr-x)


cider.el
```diff
@@ -79,6 +79,7 @@
 (require 'cider-common)
 (require 'cider-compat)
 (require 'cider-debug)
+(require 'cider-find)
```
bbatsov: That's not an oversight - commands from `cider-find.el` are supposed to be autoloaded when needed to avoid requiring too much code when CIDER is loaded. Generally most non-essential command should be auto-loaded. Perhaps I forgot to add the auto-load or you've installed cider without compiling the auto-loads?

arichiardi:
It looks like `cider-find` is not in `cider-autoloads.el`

arichiardi:
Also I have noticed `cider-refresh.el` is not in there.

cider.el
dpsutton: Not sure. I am running from git. I wasn't aware there were steps beyond requiring it.

arichiardi: I do `make build` when using it from `git`.

bbatsov:
Hmmm, those are supposed to be generated automatically when the package is built. @tarsius is it possible we hit some bug in MELPA? How does it decide with files to extract the autoloads from?

cider.el
bbatsov: From git the autoloads won't be generated automatically. Maybe we'll ned to add some autoloads file for people like you. They are supposed to be generated automatically by package.el using the metadata in each source file.

tarsius:
The autoloads file is generated when the package is being installed. So you will have to look into `package-generate-autoloads` and `update-directory-autoloads`.

bbatsov:
Turns out I forgot mark this command as autoloaded. I guess we're good overall, although people using cider from git will still have issues.

arichiardi:
Trying this `(autoload 'cider "cider.el")`

but...when I press my key binding it still says: `command-execute: Wrong type argument: commandp, cider-find-var`

bbatsov:
It's `(autoload 'command-name "file-where-it's-defined")`. You need one entry for each command you want to autoload (require) the file in question. If you add `(autoload 'cider-find-var "cider-find-var.el")` that should work.

bbatsov:
More info - https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html

arichiardi:
Oh this works! I wish there was a way to do what MELPA is doing but for local packages not installed.

arichiardi:
Thank you a lot @bbatsov

arichiardi:
For @dpsutton I found in the above link that you can do `update-directory-autoloads` and save them in an arbitrary `loaddefs.el` file.

Probably we could add a `Makefile` task for doing that?

dpsutton:
Thanks for the clarification. I thought this was a simple oversight in moving cider find around. Learn new emacs every day :)

PR #2371, opened 2018-07-01
Add little autoload explanation to Hacking on Cider doc page

arichiardi:
From what I gathered and tried locally this should work, let me know if I need
to fix things.

-----------------

### DIFF
--- a/doc/hacking_on_cider.md
+++ b/doc/hacking_on_cider.md
@@ -12,16 +12,20 @@ simply clone the code from GitHub and use it. In general - avoid editing the
 code of an installed package.
 
 Alternatively you can simply load CIDER in your Emacs straight from its source
-repo:
+repo (you'll have to manually install all the packages CIDER depends on
+in advance).
+
+Additionally you will have to generate and require the autoloads: this can be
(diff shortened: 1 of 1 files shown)


bbatsov:
Good idea! Make we should also add a simple make task to generate this file, so people could generate it in an easier fashion.

doc/hacking_on_cider.md
```diff
@@ -12,17 +12,21 @@ simply clone the code from GitHub and use it. In general - avoid editing the
 code of an installed package.
 
 Alternatively you can simply load CIDER in your Emacs straight from its source
-repo:
+repo (you'll have to manually install all the packages CIDER depends on
+in advance).
+
+Additionally you will have to generate and require the autoloads: this can be
+done interactively within Emacs with `update-directory-autoloads`. The function
+will prompt for saving a file, let's call it `cider-loaddefs.el` that you will
```
bbatsov: Btw, why don't we suggest the name `cider-autoloads.el`? `loaddefs` doesn't mean much to most people. Even to me. :D

arichiardi: I thought we had something similar to that in our code base. Cannot check now, can change it for sure.

PR #2376, opened 2018-07-13
Implement uniform jack-in and connect parameter handling

vspinu:
In a nutshell
  - all connect and jack-in commands now accept PARAMS plist argument
  - interactive updating of parameters happens by pipelining specialized functions

This entirely eliminated dynamic parameter dispatch and brings all connect and jack-in function on the same footing regarding parameter handling. It's an open system, in the sense that adding new interactive parameters in the future won't break the UI.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -811,6 +811,22 @@ Return REPL-TYPE if requirements are met."
       (funcall fun))
     repl-type))
 
+(defun cider--check-cljs (&optional cljs-type no-error)
+  "Verify that all cljs requirements are met for CLJS-TYPE connection.
+Return REPL-TYPE of requirement are met, and throw an ‘user-error’ otherwise.
+When NO-ERROR is non-nil, don't throw an error, issue a message and return
+nil."
+  (if no-error
+      (condition-case ex
+          (progn
+            (cider-verify-clojurescript-is-present)
+            (cider-verify-cljs-repl-requirements cljs-type))
+        (error
+         (message "Invalid CLJS dependency: %S" ex)
+         nil))
+    (cider-verify-clojurescript-is-present)
+    (cider-verify-cljs-repl-requirements cljs-type)))
+
 (defun cider--offer-to-open-app-in-browser (server-buffer)
   "Look for a server address in SERVER-BUFFER and offer to open it."
   (when (buffer-live-p server-buffer)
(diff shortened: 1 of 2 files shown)


vspinu:
Regarding the release in a week time. I think it's a bit too early. I am still aware of at least two issues which should be fixed (connection/buffer naming and ssh tunneling). 

The connection refactoring is a few days of work and is likely to break a few things internally. So there is no way we could manage that feat by next week. I could commit to make that change within a week or two if you decide to postpone the release 5-6 weeks from now.

bbatsov:
> Regarding the release in a week time. I think it's a bit too early. I am still aware of at least two issues which should be fixed (connection/buffer naming and ssh tunneling).

Yeah, I was thinking pretty much the same yesterday. Too much work remains to be done, so I guess I was too optimistic. I haven't even finished updating the manual to reflect all the changes. 

> The connection refactoring is a few days of work and is likely to break a few things internally. So there is no way we could manage that feat by next week. I could commit to make that change within a week or two if you decide to postpone the release 5-6 weeks from now.

Deal.

cider.el
```diff
@@ -811,88 +811,38 @@ Return REPL-TYPE if requirements are met."
...
-                             (y-or-n-p "Are you sure you want to run `cider-jack-in' without a Clojure project? ")))
-                (let* ((cmd (format "%s %s" command-resolved params)))
-                  (nrepl-start-server-process project-dir cmd on-port-callback)))
-            (user-error "`cider-jack-in' is not allowed without a Clojure project")))
-      (user-error "The %s executable isn't on your `exec-path'" command))))
-
-(defun cider--check-cljs (&optional repl-type no-error)
-  "Verify that all cljs requirements are met for cljs REPL-TYPE.
+(defun cider--check-cljs (&optional cljs-type no-error)
+  "Verify that all cljs requirements are met for CLJS-TYPE connection.
```
bbatsov: It's a cljs repl type, not a cljs type. That name is pretty confusing. cljs type would be hosted or self-hosted (for instance).

cider.el
```diff
@@ -1193,12 +1260,12 @@ Use `cider-ps-running-nrepls-command' and `cider-ps-running-nrepl-path-regexp-li
                       (car candidate)))
                   build-files))))
 
-(defun cider-project-type ()
-  "Determine the type of the current project.
+(defun cider-project-type (&optional project-dir)
+  "Determine the type of the project in PROJECT-DIR.
```
bbatsov: or the current project.

cider.el
```diff
@@ -1179,9 +1246,9 @@ Use `cider-ps-running-nrepls-command' and `cider-ps-running-nrepl-path-regexp-li
           (setq paths (cons (match-string 1) paths)))))
     (seq-uniq paths)))
 
-(defun cider--identify-buildtools-present ()
-  "Identify build systems present by their build files."
-  (let* ((default-directory (clojure-project-dir (cider-current-dir)))
+(defun cider--identify-buildtools-present (project-dir)
```
bbatsov: Shouldn't this be made optional? Otherwise you'd have to pass nil for the default behaviour.

bbatsov:
One downside of the proposed change is that functions signatures become more opaque this way. Without reading the docstring you can't know what can/should be in `params`. It's a pity Elisp doesn't have keyword arguments (not counting `cl-lib.el`, of course).

bbatsov:
Btw, I also can't find how are the prefix args handled now - I see they are an entry in the plist, but I don't see where's the check of the `current-prefix-arg` (or whatever else you're doing).

cider.el
vspinu: cljs-repl-type is not ideal either. It's a property of the connection, more like "cljs-runtime-type". I resorted to cljs-type for brevity but will change to cljs-repl-type if no other ideas creep in.

cider.el
vspinu: Need not be "current" if PROJECT-DIR is in current path. Or what do you mean by "current" here?

cider.el
vspinu: Right.

vspinu:
> One downside of the proposed change is that functions signatures become more opaque this way. 

Indeed. One would need to go to the doc string to figure out how to use the function, but it's a small price to pay.  How often would one use those commands in programs in the first place?

> I also can't find how are the prefix args handled now -

I am just testing  for `'(4)`  in every interactive updater and leaving the marker `:do-prompt` in the plist if PARAMS is `'(4)`. Each interactive plist updater `cider--update-xyz` then checks for the marker when needed.

vspinu:
> so I guess I was too optimistic

To be frank I haven't even tested this food on a real clojure workflow yet. All this time was busy with non-clojure stuff.

bbatsov:
> To be frank I haven't even tested this food on a real clojure workflow yet. All this time was busy with non-clojure stuff.

Same here.

> Indeed. One would need to go to the doc string to figure out how to use the function, but it's a small price to pay. How often would one use those commands in programs in the first place?

Not often for sure, but still - it's most likely that if something needs to be changed there I'll have to sift through the code. :-) Anyways, I agree that the uniformity is definitely desirable and I'll stomach this.

cider.el
bbatsov: I just see that in the absence of the optional param this operates on the current project (which is not obvious from the current docstring).

PR #2379, opened 2018-07-19
Modify cider-boot-parameters to match the new boot task

arichiardi:
This follows changes in `cider-nrepl` introducing the new `nrepl-server` boot
task.

-----------------

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -129,7 +129,7 @@ version from the CIDER package or library.")
   :package-version '(cider . "0.14.0"))
 
 (defcustom cider-boot-parameters
-  "repl -s -H :: wait"
+  "nrepl-server -b :: wait"
   "Params passed to boot to start an nREPL server via `cider-jack-in'."
   :type 'string
   :group 'cider


arichiardi:
I have tested this with a Clojure file and no problem. There is still a problem with the cljs REPL, which is probably going to need a change in `boot-cljs-repl`.

bbatsov:
Nice! 

> I have tested this with a Clojure file and no problem. There is still a problem with the cljs REPL, which is probably going to need a change in boot-cljs-repl.

@arichiardi I guess you'll take care of those, right? What are those problems, btw?

arichiardi:
A couple of deps are outdated, see https://github.com/adzerk-oss/boot-cljs-repl/issues/60.

About taking care of those, it is very low in my priority atm because I have switched to `shadow-cljs` and `figwheel-main`. I will definitely be able to help there though or do it myself at some point.

The only thing I am afraid of is that I am not very familiar with the changes to piggieback, so probably I need to spend some time for learning things there: `boot-cljs-repl` seems to do something very similar.

bbatsov:
> About taking care of those, it is very low in my priority atm because I have switched to shadow-cljs and figwheel-main. I will definitely be able to help there though or do it myself at some point.

Yeah, I get this. Seems this boot-cljs-repl is not getting any love these days, but I doubt many people are using it. 

> The only thing I am afraid of is that I am not very familiar with the changes to piggieback, so probably I need to spend some time for learning things there: boot-cljs-repl seems to do something very similar.

Isn't it just using piggieback?

arichiardi:
> Isn't it just using piggieback?

https://github.com/adzerk-oss/boot-cljs-repl/blob/master/src/adzerk/boot_cljs_repl.clj#L18-L21

It seems so, so maybe it is an easy one after all, will try smth out before talking nonsense again :smile:

bbatsov:
Oh, it's using Weasel as well... The changes required really are trivial, but if there isn't anyone to merge them, that's a problem...

arichiardi:
So this one can go in now?

bbatsov:
Yeah, we'll merge this soon. The only holdup right now is `refactor-nrepl`, but I'm not that concerned about it.

bbatsov:
Let's get wild! ;-)

PR #2384, opened 2018-07-23
[Fix #2291] Fix `cider-use-tooltips` custom variable to work as expected

### LINKED ISSUE #2291: cider-stacktrace doesn't seem to honor cider-use-tooltips
Hi there!

I have `cider-use-tooltips` set to nil, but I still can see `"View source at this location"` tooltips.

Relevant code - https://github.com/clojure-emacs/cider/blob/ff3bfa2642cfcd4eb5f2756280ea057dc69f8aff/cider-stacktrace.el#L698

Thanks - Victor

greenfork:
I have added a `(when cider-use-tooltips ...)` condition to every `'help-echo` property of `insert-text-button` in `cider-stacktrace-el`. This way all tooltips in stacktrace buffer will be displayed only when `cider-use-tooltips` is set to `t`.

For some reason I get "cider.el:73:1:Error: Cannot open load file: No such file or directory, cider-client" when I run `make lint` and this is the only error.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -602,6 +602,10 @@ prompt and whether to use a new window.  Similar to `cider-find-var'."
 
 ;; Rendering
 
+(defun cider-stacktrace-tooltip (tooltip)
+  "Return TOOLTIP if `cider-use-tooltips' is set to true, nil otherwise."
+  (when cider-use-tooltips tooltip))
+
 (defun cider-stacktrace-emit-indented (text &optional indent fill fontify)
   "Insert TEXT, and optionally FILL and FONTIFY as clojure the entire block.
 INDENT is a string to insert before each line.  When INDENT is nil, first
@@ -638,8 +642,9 @@ others."
                           'filter (cadr filter)
                           'follow-link t
                           'action 'cider-stacktrace-filter
-                          'help-echo (format "Toggle %s stack frames"
-                                             (car filter)))
+                          'help-echo (cider-stacktrace-tooltip
+                                       (format "Toggle %s stack frames"
+                                               (car filter))))
       (insert " "))
     (insert "\n")
     (insert "  Hide: ")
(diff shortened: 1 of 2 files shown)


cider-stacktrace.el
```diff
@@ -726,7 +731,8 @@ This associates text properties to enable filtering and source navigation."
                             'file file 'line line 'column column 'follow-link t
                             'action (lambda (_button)
                                       (cider-jump-to (cider-find-file file)
-                                                     (cons line column))))
+                                                     (cons line column)))
+                            'help-echo (when cider-use-tooltips ()))
```
bbatsov: I think you forgot to add some text here.

cider-stacktrace.el
```diff
@@ -638,8 +638,9 @@ others."
                           'filter (cadr filter)
                           'follow-link t
                           'action 'cider-stacktrace-filter
-                          'help-echo (format "Toggle %s stack frames"
-                                             (car filter)))
+                          'help-echo (when cider-use-tooltips
```
bbatsov: Why not introduce some `cider-tooltip` helper instead that does the check internally. Probably it should take a string as its only param. I think that would cleanup up the code.

cider-stacktrace.el
greenfork: It was empty, so I just added this line for consistency. Should I compose a new tooltip?

cider-stacktrace.el
greenfork: Sure, good idea!

cider-stacktrace.el
bbatsov: I guess that'd be best.

greenfork:
I have wrapped all the behavior in `cider-stacktrace-tooltip` and added a proper tooltip for the.. last chunk.

bbatsov:
Thanks!

P.S. The build was failing due to wrong indentation, but I addressed this myself.

PR #2385, opened 2018-07-24
accept plain symbol as figwheel-main build, normalizing to keyword

euccastro:
When using cider-jack-in-cljs and choosing figwheel-main as the REPL type, you are prompted to select a figwheel build (although in principle you could enter a map of options there). If you provide the bare build name (i.e., `dev` instead of `:dev`), you get an exception and the operation fails.  This patch normalizes build names so they are passed to `figwheel/start` as keywords, but it won't modify anything that looks like a keyword, map, or s-expr.

Known edge case: if you prepend commas (,) to your figwheel-main options then a colon will be prepended to them, causing a `RuntimeException: Invalid Token: :`.  I don't expect this to ever happen in practice, since there's no point in prepending commas anyway.  This PR should be robust to other leading or trailing whitespace, including trailing commas (whitespace liberally added in tests.)

-----------------

- [ ] The commits are consistent with our [contribution guidelines](../blob/master/CONTRIBUTING.md) **I couldn't find this doc**
- [X] All code passes the linter (`make lint`) **cider.el does, which is where I put my additions.  there are some other failures but I'm remiss to touch unrelated code in my first PR**
- [ ] You've updated the [user manual](../blog/master/doc) (if adding/changing user-visible functionality) **not necessary since old behavior should keep working**.

Thanks!

manual][1] extremely useful.*

[1]: _https://cider.readthedocs.io/en/latest/hacking_on_cider/_

### DIFF
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -207,6 +207,38 @@
               :and-return-value '())
       (expect (cider-project-type) :to-equal cider-default-repl-command))))
 
+(describe "cider-figwheel-main-init-form"
+  ;; whitespace checks sprinkled amongst other tests
+  (describe "from options"
+    (it "leaves keywords from options alone"
+        (setq-local cider-figwheel-main-default-options ":dev ")
+        (expect (cider-figwheel-main-init-form) :to-equal "(do (require 'figwheel.main) (figwheel.main/start :dev))"))
+    (it "leaves maps from options alone"
+        (setq-local cider-figwheel-main-default-options " {:a 1 :b 2}")
+        (expect (cider-figwheel-main-init-form) :to-equal "(do (require 'figwheel.main) (figwheel.main/start {:a 1 :b 2}))"))
+    (it "leaves s-expr from options alone"
+        (setq-local cider-figwheel-main-default-options " (hashmap :a 1 :b 2)")
+        (expect (cider-figwheel-main-init-form) :to-equal "(do (require 'figwheel.main) (figwheel.main/start (hashmap :a 1 :b 2)))"))
+    (it "adds colon when necessary from options"
+        (setq-local cider-figwheel-main-default-options " dev")
+        (expect (cider-figwheel-main-init-form) :to-equal "(do (require 'figwheel.main) (figwheel.main/start :dev))")))
+
+  (describe "from minibuffer"
+    (before-each
+     (setq-local cider-figwheel-main-default-options nil))
+    (it "leaves keywords entered in minibuffer alone"
+        (spy-on 'read-from-minibuffer :and-return-value " :prod ")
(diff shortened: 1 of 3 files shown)


test/cider-tests.el
```diff
@@ -207,6 +207,38 @@
               :and-return-value '())
       (expect (cider-project-type) :to-equal cider-default-repl-command))))
 
+(describe "cider-figwheel-main-init-form"
+  ;; whitespace checks sprinkled amongst other tests
+  (describe "from options"
+    (it "leaves keywords alone"
+        (setq-local cider-figwheel-main-default-options ":dev ")
```
bbatsov: I'd simply use `let` here instead of `setq-local`.

bbatsov: The indentation after `it` doesn't look right. Probably you didn't eval `buttercup`, so Emacs doesn't know how to indent its macros.

euccastro: Thanks!  Fixed.  This is literally my first emacs-lisp code barring basic config var setting, so I was just parroting what I saw elsewhere in this file.

euccastro: Yeah, I don't know how to do that in spacemacs, and this file is not checked by `make lint`.  For the time being, I just fixed the indentation manually to match other `it` forms in this file.

euccastro:
Thanks for the review, @bbatsov!

Yeah, I can send a PR for that later today.  Maybe everything related to piggieback could be removed or deemphasized, if it is really not needed as it says there.

I haven't checked whether that is the case.  If it is, then [this check](https://github.com/clojure-emacs/cider/blob/master/cider.el#L665) should be removed, right?

arichiardi:
Great work! @bhauman will be happy to know that you can pass edn there directly :)

bbatsov:
> Yeah, I can send a PR for that later today. 

Yeah, that'd be nice. 

> Maybe everything related to piggieback could be removed or deemphasized, if it is really not needed as it says there.

It's not needed only if piggieback is auto-injected. I guess we should just mention this. We started injecting piggieback only in `master`, that's why the docs are not completely up-to-date. 

> I haven't checked whether that is the case. If it is, then this check should be removed, right?

Why? I'm pretty sure if `piggieback`'s not on the classpath figwheel will not work with CIDER. :-)

euccastro:
> @bhauman will be happy to know that you can pass edn there directly :)

That was the case already; I just tried not to break that. :)

> Why? I'm pretty sure if piggieback's not on the classpath figwheel will not work with CIDER. :-)

I didn't understand this bit:

> It's not needed only if piggieback is auto-injected.

That makes sense now, thanks!

I took longer than expected at work today so I won't get into the manual PR until tomorrow.

bbatsov:
👍

Issue #2387, opened 2018-07-25
empty first host on the list when using `cider-connect-clj` to nrepl on localhost


tsulej:
## Expected behavior

When connecting to locally run nREPL with `cider-connect-clj` I expect to have `localhost` at first position in `host:` prompt like it was in cider 0.17.
I want to just hit enter twice to connect.

## Actual behavior

Empty first position in proposed host. 
This is caused by assumption that `buffer-file-name` contains host name when is a string (`cider-current-host`). Which is not true in this case.

## Steps to reproduce the problem

* `lein repl` in console
* open any project file in freshly run emacs
* `M-x cider-connect-clj`

Prompt looks like: `Host: { | 192... }`

### CIDER version information
```
;; Connected to nREPL server - nrepl://localhost:54920
;; CIDER 0.18.0snapshot (package: 20180701.1420), nREPL 0.2.12
;; Clojure 1.9.0, Java 1.8.0_66
```

### Lein/Boot version

Leiningen 2.8.1

### Emacs version

GNU Emacs 26.1 (build 1, x86_64-unknown-cygwin, GTK+ Version 3.22.28) of 2018-05-28

### Operating system

Windows 10

PR #2388, opened 2018-07-25
#2387 fix - localhost is returned when connecting to local nREPL

genmeblog:
Fix of #2387 

- [byte-compilation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Byte-Compilation.html) - compiled manually with `byte-compile-file`

I run tests but `cider-load-file` fails on my environment (cider). Test fails also on master without my fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@
 
 ### Bugs fixed
 
-* [#2387](https://github.com/clojure-emacs/cider/issues/2387): `localhost` as first host returned when connecting to local nREPL
 * [#2317](https://github.com/clojure-emacs/cider/issues/2317): The stdin prompt can now be cancelled.
 * [#2328](https://github.com/clojure-emacs/cider/issues/2328): Added `cider-eval-sexp-to-point`.
 * [#2310](https://github.com/clojure-emacs/cider/issues/2310): `cider-format-edn-last-sexp` will format the last sexp.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 ### Bugs fixed
 
+* [#2387](https://github.com/clojure-emacs/cider/issues/2387): `localhost` as first host returned when connecting to local nREPL
```
vspinu: If this was introduced in 0.18, then it shouldn't be in the news.

genmeblog: looks like introduced in 0.18. Understand, to be removed, right?

genmeblog:
One check timed out. Please rerun.

vspinu:
There is no need to call `file-remote-p` twice there. Let me take care of this as part of #2391. There is another tiny simplification which could be done in there. 

In any case, I don't think it's a regression in CIDER. I guess you are using ido-ubiquitous and either ido or ido-ubiquitous doesn't handle well nils in the candidates which wasn't the case before.

genmeblog:
Ok, no problem, this can be a part of #2391. Thx!

Yes, I use ido. I don't have a problem with empty value. I have a problem, that in my case 'localhost' is not displayed at first position in host proposition list (it's at the end). And I have to type in or select with the arrows. I just want to hit enter twice to connect.

genmeblog:
I've just checked your changes. Much better.
Closing this PR.

PR #2390, opened 2018-07-26
Fixed tramp usage for emacs 26.1 and up

shlomiv:
Tramp with emacs version 26.1 and up is using  `cl-defstruct` instead of vanilla `vec`

This solves issue #2377

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1200,6 +1200,12 @@ Return a list of the form (HOST PORT), where PORT can be nil."
     ;; remove the label
     (if (= 3 (length host)) (cdr host) host)))
 
+(defun ->tramp-file-name (vec)
+  "Tramp version starting 26.1 is using a cl-defstruct rather than vanilla vec"
+  (if (version< emacs-version "26.1") vec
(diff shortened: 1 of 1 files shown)


shlomiv:
Not sure what to do about the error `the function ‘make-tramp-file-name’ is not known to be defined.` It passes in version 26.1 obviously, but whats the right thing to do with older versions? its already guarded by a version check.

vspinu:
Wrap in `with-no-warnings`.

shlomiv:
@vspinu Thanks 👍

cider.el
```diff
@@ -1200,6 +1200,13 @@ Return a list of the form (HOST PORT), where PORT can be nil."
     ;; remove the label
     (if (= 3 (length host)) (cdr host) host)))
 
+(defun ->tramp-file-name (vec)
```
xiongtx: Function should start w/ package name, i.e. `cider-tramp-file-name`.

shlomiv: done, thanks

PR #2391, opened 2018-07-27
Initialize sesman earlier in REPL buffers and add support for sesman-project

vspinu:
Fixes vspinu/sesman#6 and adds support for sesman-project. After discovering a few more issues with project.el I am no longer relying on it.


bbatsov:
Seems you broke one sesman-related test.

vspinu:
Should be good to go.

PR #2392, opened 2018-07-28
New session and REPL buffer naming system

vspinu:
Make session names arbitrarily customisable and make REPL buffer (and satellites) use session name as part of their name. 

By default session names look like this `project/dir:localhost:1234` and buffer names like `*project/dir:localhost:1234(cljs:node)*`. See the docstring of `cider-format-connection-params` for what is available.  

I believe the only potentially contentious change is that I made `nrepl-buffer-name-template` not to start with `cider-repl`. The new names are rather long and `cider-repl` prefix is not particularly useful, especially that the mode-line already displays [REPL(cljs)]. I am not very sure about this change though, nor I am able to find a shorter prefix than `cider-repl `.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * **(Breaking)** Move `cider-ns-refresh`, previously on `C-c C-x`, on `C-c M-n (M-)r` in the `cider-ns-map`.
 * **(Breaking)** Bump the minimum required Emacs version to 25.1.
 * **(Breaking)** Drop support for Java 7 and Clojure(Script) 1.7.
+* **(Breaking)** Use session name as part of CIDER buffers names (REPL, server, messages), and obsolete `nrepl-buffer-name-separator` and `nrepl-buffer-name-show-port`. See `cider-session-name-template` and `cider-format-connection-params` for how to customize CIDER buffer names. 
 * Rename `cider-eval-defun-to-point` to `cider-eval-defun-up-to-point`.
 * Add support for printing to the current buffer to `cider-eval-defun-up-to-point`.
 * Remove `cider-ping` command.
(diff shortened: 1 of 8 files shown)


bbatsov:
> By default session names look like this project/dir:localhost:1234 and buffer names like *project/dir:localhost:1234(cljs:node)*. See the docstring of cider-format-connection-params for what is available.

I think the cljs repl type should be in the REPL[...] in the interest of saving space. It's not something you'd normally use when selecting a buffer anyways. 

> The new names are rather long and cider-repl prefix is not particularly useful, especially that the mode-line already displays [REPL(cljs)]. I am not very sure about this change though, nor I am able to find a shorter prefix than cider-repl

I'd love to see shorter names, but I used this prefix to search in ido and ivy about the currently running REPL buffers, and I assume others do as well. Unfortunately REPL management commands don't always work out great and from time to time you have to hunt those buffers manually. :-) I also think it's just common sense to have the word "REPL" in the name of a REPL buffer.

vspinu:
> it's not something you'd normally use when selecting a buffer anyways.

I think this is no longer true. If my session contains one nodejs and one figwheel repl I do want to see "node" and "figwheel" in the buffer name (for selection purposes). Otherwise you would get the lame <2> postfix for one of them. 

> I'd love to see shorter names, but I used this prefix to search in ido and ivy about the currently running 

I did the same but it's not particularly useful to diferentiate cljs and clj repls. Nowadays I search for `j)` and `s)`  (or `s*`) to get to clj or cljs repls respectively.

> I also think it's just common sense to have the word "REPL" in the name of a REPL buffer.

Ok, let's then keep the long name by default and let users customize accordingly. There are some abbreviation options available in the formats (shorter project name - %j, no "localhost" - %H).

bbatsov:
👍

PR #2393, opened 2018-07-30
[Fix #2381] Start REPLs in the original jack-in buffer

### LINKED ISSUE #2381: `cider-default-cljs-repl` still shows 'select cljs repl type' popup
After doing some investigation, `cider-default-cljs-repl` is `nil` inside of `cider--update-cljs-type`. I'm not sure where it's getting reset though.

Cider 0.18.0 snapshot (Latest commit `4b7aea3`)

vspinu:


bbatsov:
I think that's not going to work when you decide to do jack-in with the option to select a project prior to jacking in.

vspinu:
Indeed. Even worse, all connection commands are affected by this. A fix will probably destroy the nice pipelining for parameter construction currently in place :/  I need to give it a bit of thought.

vspinu:
This turned to be quite tricky. First of all `cider-default-cljs-repl-type` is affected only when connection functions are used in programs because on `C-u` all parameters are asked for, including cljs-repl-type.

Even more importantly either way (hacking dir-locals or using current buffer) the behavior will be against expectation of some users. Some users might be setting variables in `clojure-mode-hook`, others in `dir-locals`. So a thorough solutions has to take into account both, variables in the current buffer and variables hacked in in the `:project-dir` parameter.

I think what I have now implemented solves the problem reliably. First, when no project-dir is supplied as a parameter, all repls are created withing the `current-buffer`. When project-dir is supplied (either on `C-u` or programaticaly) a temporary buffer is set-up, all local variables from the current buffer are cloned into this temporary buffer and dir locals are hacked from the `project-dir`. Finally repls are created with this temporary buffer set as the current. 

The above strategy should also provide a more reliable experience during the restarts, because the restarts are likely to be executed from different contexts than the original jack-in/connect.

bbatsov:
> I think what I have now implemented solves the problem reliably. First, when no project-dir is supplied as a parameter, all repls are created withing the current-buffer. When project-dir is supplied (either on C-u or programaticaly) a temporary buffer is set-up, all local variables from the current buffer are cloned into this temporary buffer and dir locals are hacked from the project-dir. Finally repls are created with this temporary buffer set as the current.

I opened the issue today with the intention to propose exactly this approach! 👍 Nicely done!

PR #2394, opened 2018-07-31
Improve session and REPL buffer naming

vspinu:
- better abbreviation of directory names
 - re-create context buffer on every connect instead of `kill-all-local-variables` which could fail

-----------------


bbatsov:
👍

PR #2395, opened 2018-08-01
A range of connection fixes and support for sesman browser

vspinu:
- A range of fixes for various connection problems which I discovered while working on connection browser. 
 - Add support for the new full featured sesman browser.  

Sesman simplified the "info" UI, btw. There are now only two keys `C-c C-s i` for info on linked sessions and `C-c C-s C-w` the the browser. 

Tests will fail till sesman is propagated to MELPA.

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -335,7 +335,7 @@ REPL defaults to the current REPL."
 
 (defun cider-restart (&optional repl)
   "Restart CIDER connection associated with REPL.
-REPL defaults to the current REPL. Don't restart the server or other
+REPL defaults to the current REPL.  Don't restart the server or other
 connections within the same session.  Use `sesman-restart' to restart the
 entire session."
   (interactive)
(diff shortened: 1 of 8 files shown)


cider-connection.el
```diff
@@ -553,6 +553,8 @@ Assume that the current buffer is a REPL."
       (setq cider-repl-type type)
       (setq mode-name (format "REPL[%s]" type))
       (let ((params (cider--gather-connect-params)))
+        ;; in order to re-use this buffer's name on restarts, rename temporally
```
bbatsov: temporarily

cider-connection.el
```diff
@@ -553,6 +553,8 @@ Assume that the current buffer is a REPL."
...
+        (rename-buffer (generate-new-buffer-name "*dummy-cider-repl-buffer*"))
```
bbatsov: Won't it be better if `nrepl-repl-buffer-name` handled this based on `params`? Right now this code looks pretty weird.

vspinu: I don't know how that could be done. The issue is that if the current buffer is named "*repl something*" and the new buffer name is also "*repl something*" then `nrepl-repl-buffer-name` will generate `*repl something*<2>`. So each time you restart the name will toggle <2> back and forth. Do you see what I mean? 

I could improve on the inline comment but I think this is the right fix because `nrepl-repl-buffer-name` doesn't know for which buffer I need the new name, so it cannot make the decision whether to duplicate the name or not.

vspinu:
Hmm. Melpa fetcher has been [down](https://github.com/melpa/melpa/issues/5662) for several days. Will have to wait on this.

cider-connection.el
bbatsov: I got what the problem is immediately, I was just wondering if this `nrepl-repl-buffer-name` can't simply check if it's invoked in a REPL buffer or that the current command is restart and act accordingly. On the other hand that some more coupling there, so probably the current appoach is OK. The comment should probably be extended, though.

vspinu: Indeed. Making `nrepl-repl-buffer-name` "aware" of the current buffer name doesn't seem like a good solution. I thought of having a restart marker as the buffer local, but it's complex and it's no impossible that "same-name" situation will occur outside of the restarts in the future.

I have extended the comment.

bbatsov:
Seems some other byte compilation error is causing the failure right now.

vspinu:
Fixed.

bbatsov:
👍

PR #2397, opened 2018-08-02
Add shadow-select CLJS REPL type

arichiardi:

-----------------

Thanks!

manual][1] extremely useful.*

This kind of CLJS REPL is very helpful for cases when you are already watching
one or more shadow builds in shadow either after having launched the server in
a terminal or from within Cider.
This patch also refactors option normalization out of figwheel-main so that it
can be reused by every other function.


arichiardi:
Oh good idea I would just avoid to prompt twice though, so maybe some other `defcustom` could be useful

PR #2416, opened 2018-08-14
 [Fix #2415] Filter out killed buffers in `cider-repls` 

### LINKED ISSUE #2415: Opening .cljc throws an exception if `(sessman-current-session 'CIDER)` contains a dead buffer
## Expected behavior

Opening `.cljc` does not trigger an exception.

## Actual behavior

Opening `.cljc` throws an exception:

```
File mode specification error: (error Selecting deleted buffer)
cider-resolve-ns-symbols: Selecting deleted buffer
```

Open a REPL in a project, then kill the REPL buffer. Now try to open a `.cljc`

### CIDER version information

checkout `12674b3de377d6ae6f67ac155caea12f5de3c743`

### Emacs version

27.0.50

### Operating system

macOS 10.13

### Diagnosis

`(sessman-current-session 'CIDER)` returns a killed buffer. For `.clj` and `.cljs` these buffers get filtered out in `cider-repls`, but not for `.cljc`.

dottedmag:
If a REPL buffer is killed manually it will leak out of `cider-repls` for .cljc
files, so make sure `cider-repls` only return live REPL buffers.

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -658,6 +658,13 @@ session."
                       type (car (sesman-current-session 'CIDER)))
         repl))))
 
+(defun cider--match-repl-type (type buffer)
+  "Return non-nil if TYPE matches BUFFER's REPL type."
+  (let ((buffer-repl-type (cider-repl-type buffer)))
+    (cond ((null buffer-repl-type) nil)
+          ((or (null type) (equal type "multi")) t)
+          (t (string= type buffer-repl-type)))))
+
 (defun cider-repls (&optional type ensure)
   "Return cider REPLs of TYPE from the current session.
 If TYPE is nil or \"multi\", return all repls.  If ENSURE is non-nil, throw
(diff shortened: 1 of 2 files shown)


cider-connection.el
```diff
@@ -658,18 +658,22 @@ session."
                       type (car (sesman-current-session 'CIDER)))
         repl))))
 
+(defun cider--match-repl-type (type buffer)
+  "Return t if BUFFER's REPL type matches TYPE."
```
arichiardi: My 2c -> "Return non-nil if TYPE matches BUFFER's REPL.`

vspinu:
Solutions looks good for this particular issue but I think the problem might pop somewhere else. Cleaning of the buffers should occur in process sentinel I guess, or in sesman proper.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 * [#2307](https://github.com/clojure-emacs/cider/pull/2307): Use a better error when a cljs repl form cannot be found.
 * Fix the broken test selector functionality.
 * [#2291](https://github.com/clojure-emacs/cider/issues/2291): `cider-use-tooltips` custom variable works as expected.
+* [#2415](https://github.com/clojure-emacs/cider/issues/2291): Opening .cljc throws an exception if `(sessman-current-session 'CIDER)` contains a dead buffer
```
bbatsov: This should be removed as the bug was introduced in 0.18, meaning that's not a change that would affect users of stable CIDER releases.

dottedmag: Done

bbatsov:
Thanks!

PR #2429, opened 2018-08-29
Connection fixes

vspinu:
A range of connection related fixes. One commit per issue. 

I will fix the last issue regarding auto-association to projects and jars in another day or two.

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -92,6 +92,21 @@ PARAMS is a plist containing :host, :port, :server and other parameters for
   "Ensure there is a linked CIDER session."
   (sesman-ensure-session 'CIDER))
 
+(defun cider--session-server (session)
+  "Return server buffer for SESSION or nil if there is no server."
+  (seq-some (lambda (r)
+              (buffer-local-value 'nrepl-server-buffer r))
+            (cdr session)))
+
+(defun cider--gather-session-params (session)
+  "Gather all params for a SESSION."
+  (let (params)
+    (dolist (repl (cdr session))
+      (setq params (cider--gather-connect-params params repl)))
+    (when-let* ((server (cider--session-server session)))
+      (setq params (cider--gather-connect-params params server)))
+    params))
+
 (defun cider--gather-connect-params (&optional params proc-buffer)
   "Gather all relevant connection parameters into PARAMS plist.
 PROC-BUFFER is either server or client buffer, defaults to current buffer."
@@ -443,9 +458,7 @@ Fallback on `cider' command."
 (cl-defmethod sesman-restart-session ((_system (eql CIDER)) session)
   (let* ((ses-name (car session))
          (repls (cdr session))
-         (srv-buf (seq-some (lambda (r)
-                              (buffer-local-value 'nrepl-server-buffer r))
-                            repls)))
+         (srv-buf (cider--session-server session)))
     (if srv-buf
         ;; session with a server
         (let ((s-params (cider--gather-connect-params nil srv-buf)))
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -627,7 +627,7 @@ function with the repl buffer set as current."
             cider-session-name ses-name
             nrepl-project-dir (plist-get params :project-dir)
             ;; REPLs start with clj and then "upgrade" to a different type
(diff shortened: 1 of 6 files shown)


cider.el
```diff
@@ -1208,6 +1215,29 @@ non-nil, don't start if ClojureScript requirements are not met."
...
+                              (let ((ses-params (cider--gather-session-params ses)))
+                                (and (equal proj-dir (plist-get ses-params :project-dir))
+                                     (or (null port)
+                                         (equal port (plist-get ses-params :port)))
+                                     (or (null host)
+                                         (equal host (plist-get ses-params :host))))))
+                            sessions)))
+
+    (when session
+      (unless (y-or-n-p "A session with the same parameters exists.  You can connect a sibling instead.  Proceed? ")
```
bbatsov: Maybe you can also print `session` as part of the message?

cider.el
```diff
@@ -1208,6 +1215,29 @@ non-nil, don't start if ClojureScript requirements are not met."
                               cider-offer-to-open-cljs-app-in-browser)
                      (cider--offer-to-open-app-in-browser nrepl-server-buffer)))))))
 
+(defun cider--check-same-kind-session (params)
```
bbatsov: I'm not sure if `kind` is the right word for this. Maybe `check-for-existing` or `check-for-similar`?

cider-connection.el
```diff
@@ -639,6 +639,17 @@ function with the repl buffer set as current."
 
 ;;; Current/other REPLs
 
+(defun cider--no-repls-user-error (type)
+  "Throw \"No REPL\" user error customized for TYPE."
+  (let ((type (cond
```
bbatsov: Probably `pcase` would simplify this code. Or `cl-case`.

cider.el
vspinu: Could be, just a name though. Putting the full session is an overkill; too much stuff.

bbatsov:
Overall the changes look good to me. Probably some simple test for the first and the last fix would be handy to make sure we don't accidentally break this down the road. 

I'm not sure I like the term `pending-cljs`, but I can't suggest something better of the top of my head. I'd like some term that suggests better that some REPL upgrade is in progress. I'm also wondering how are going to recover REPLs that for some reason don't managed to complete their "upgrade".

cider.el
vspinu: `check-for-existing` sound right to me.

cider.el
bbatsov: 👍

cider.el
bbatsov: 👍

vspinu:
> I'm not sure I like the term pending-cljs,

Yeh. I actually did some digging on this one with synonyms et. al. But nothing better came up. I started with `uninitialized-cljs` but that's way too long for the modeline. 

> I'm also wondering how are going to recover REPLs that for some reason don't managed to complete their "upgrade".

We don't need to. If upgrade failed the repl returns to clj automatically on the first state message.

cider-connection.el
vspinu: :) I don't see how it could be even simpler than that, it's a bare-bones conditional already.

vspinu:
I have addressed the comments and added a test for the last commit. Writing tests for the first issue is quite pointless. One would need to overwrite every single function in `cider-connect-sibling-cljs`, thus tightly coupling the test with the implementation.

cider-connection.el
```diff
@@ -92,6 +92,21 @@ PARAMS is a plist containing :host, :port, :server and other parameters for
...
+  "Return server buffer for SESSION or nil if there is no server."
+  (seq-some (lambda (r)
+              (buffer-local-value 'nrepl-server-buffer r))
+            (cdr session)))
+
+(defun cider--gather-session-params (session)
+  "Gather all params for a SESSION."
+  (let (params)
+    (dolist (repl (cdr session))
+      (setq params (cider--gather-connect-params params repl)))
```
dpsutton: am i misreading this? This is a dolist and `setq` without accumulating previous values? Should this be `add-to-list` or `(setq params (cons params ...`?

dpsutton: i missed `params` being passed into that function. ignore me :)

PR #2430, opened 2018-08-30
Open archive entries within AVFS directory if available

vspinu:

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -290,7 +290,7 @@ or an entry within a zip/jar archive.  If AVFS (archive virtual file
 system; see online docs) is mounted the archive entry is opened inside the
 AVFS directory, otherwise the entry is archived into a temporary read-only
 buffer.  If URL doesn't contain a scheme prefix and is an absolute path, it
-is treated as such. Finally, if URL is relative, it is expanded within each
+is treated as such.  Finally, if URL is relative, it is expanded within each
 of the open Clojure buffers till an existing file ending with URL has been
 found."
   (require 'arc-mode)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 ## master (unreleased)
 
+### New features
+
 * [#2430](https://github.com/clojure-emacs/cider/issues/2375) `cider-find-var` opens archive files inside [AVFS](http://avf.sourceforge.net) folders if AVFS is detected.
+
+### Bug fixes
+
+* Fix jack-in from inside of remote buffers
+
 
 ## 0.18.0 (2018-09-02)
 
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -501,7 +501,11 @@ key-values depending on the connection type."
                    (message "[nREPL] Falling back to SSH tunneled connection ...")
                    (nrepl--ssh-tunnel-connect host port))
               ;; fallback is either not enabled or it failed as well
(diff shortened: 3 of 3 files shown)


cider-common.el
```diff
@@ -286,10 +286,12 @@ If no local or remote file exists, return nil."
 (defun cider-find-file (url)
   "Return a buffer visiting the file URL if it exists, or nil otherwise.
 If URL has a scheme prefix, it must represent a fully-qualified file path
-or an entry within a zip/jar archive.  If URL doesn't contain a scheme
-prefix and is an absolute path, it is treated as such.  Finally, if URL is
-relative, it is expanded within each of the open Clojure buffers till an
-existing file ending with URL has been found."
+or an entry within a zip/jar archive.  If AVFS is present and mounted the
+archive entry is opened inside the archive AVFS directory, otherwise the
+entry is archived into a temporary read-only buffer.  If URL doesn't
```
bbatsov: archived or unarchived?

cider-common.el
```diff
@@ -303,36 +305,44 @@ existing file ending with URL has been found."
...
-                   (setq-local buffer-read-only t)
-                   (set-buffer-modified-p nil)
-                   (set-auto-mode)
-                   (current-buffer))))))
+                     (name  (format "%s:%s" path entry))
+                     (avfs (format "%s%s#uzip/%s"
+                                   (expand-file-name (or (getenv "AVFSBASE")  "~/.avfs/"))
+                                   path entry)))
+           (cond
+            ;; use avfs
```
bbatsov: I have no idea what `avfs` is, so I guess some explanations here might be useful. :-) Sounds like some filesystem, but I've certainly never heard fo this.

vspinu: I assumed it's quite well know. It maps archives to browsable directories. There is a dired extension [here](https://github.com/Fuco1/dired-hacks#dired-avfs) and [sunrise commander](https://www.emacswiki.org/emacs/Sunrise_Commander) activates it by default. 

Simple `mountavfs` in `.bash_profile` makes it work.

cider-common.el
vspinu: "archive's AVFS directory" would probably be the most precise. I will add some more pointers.

bbatsov: You should probably also add something this in the manual. Otherwise almost no one is going to figure this out.

cider-common.el
bbatsov: The cond is getting a bit big, so maybe we can extract 2-3 helper methods from it.

cider-common.el
```diff
@@ -303,36 +305,44 @@ existing file ending with URL has been found."
...
-                 (with-current-buffer (generate-new-buffer
-                                       (file-name-nondirectory entry))
-                   (archive-zip-extract path entry)
-                   (set-visited-file-name name)
-                   (setq-local default-directory (file-name-directory path))
-                   (setq-local buffer-read-only t)
-                   (set-buffer-modified-p nil)
-                   (set-auto-mode)
-                   (current-buffer))))))
+                     (name  (format "%s:%s" path entry))
```
bbatsov: There's an extra space after `name`.

cider-common.el
vspinu: I have simplified the inner cond.

vspinu:
Added a doc tip and fixed a jack-in issue on remotes. Host was always set to localhost in server sentinel.

PR #2454, opened 2018-09-16
Fix erratic inspector behavior when multiple REPLs are connected

alexander-yakushev:
Whenever you have more than one connected REPL, and you use the inspector in one of them, the initial value is displayed correctly, but the subsequent actions (push, pop, change page) show something completely off. 

Turns out, the inspector buffer does not know the correct REPL where it was initiated from, so `(cider-current-repl)` can return a different REPL where the inspector middleware, obviously, doesn't match.

I've been suffering from this problem for quite a while, but only now got annoyed enough to tackle it. I don't know if this is the best solution; but since the current inspector does not allow multiple concurrent inspections for different REPLs, this should work.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -134,6 +134,11 @@ This is used as an alternative to the built-in `last-command'.  Whenever we
 invoke any command through \\[execute-extended-command] and its variants,
 the value of `last-command' is not set to the command it invokes.")
 
+(defvar cider-inspector--current-repl nil
+  "Contains the reference to the REPL where inspector was last invoked from.
+This is needed for internal inspector buffer operations (push,
+pop) to execute against the correct REPL session.")
+
 ;; Operations
 ;;;###autoload
 (defun cider-inspect-expr (expr ns)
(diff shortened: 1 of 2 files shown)


bbatsov:
Seems like a reasonable approach to me. Another would be to link the inspector buffer to the current sesman session when it's created I guess. That should ensure `cider-current-repl` behaves consistently IMO.

@vspinu Any thoughts from you?

cider-inspector.el
```diff
@@ -134,6 +134,11 @@ This is used as an alternative to the built-in `last-command'.  Whenever we
 invoke any command through \\[execute-extended-command] and its variants,
 the value of `last-command' is not set to the command it invokes.")
 
+(defvar cider-inspector--current-repl nil
```
bbatsov: I think this should should probably be a buffer-local variable for the inspector buffers.

bbatsov:
Yeah, that's a good point as well. I've added one small inline remark. You should also mention this bugfix in the changelog.

cider-inspector.el
alexander-yakushev: I might be wrong, but the inspector buffer is not retained between actions. At least, when I try to change a buffer-local variable, it is set back to `nil` once I do push/pop/whatever.

vspinu:
This looks like the right approach to me as well. If I understand correctly inspector is really attached to a single connection and not a session. 

There is also a possibility that the REPL is killed in the meanwhile and I don't think there is a provision to kill the inspector with it. A more generic `cider--current-repl` might work better in the long run; I bet there must be other similar cases. Then we could clean such repl-dependent buffers when the repl is killed.

cider-inspector.el
bbatsov: Oh, well, this part of the codebase is pretty bad and needs a general cleanup. :-) I keep forgetting it's "special" compared to the rest of the code. No need for additional changes now, but generally we should reusing the buffer most of the time instead of re-creating it all the time.

alexander-yakushev:
Thank you!

PR #2467, opened 2018-09-30
[Fix #2466] Make generic cider ops use any available nrepl connection

### LINKED ISSUE #2466: Formatting code in .cljs buffers doesn't work without cljs repl
## Expected behavior
When formatting code with cider-format-* commands, but only a clj repl active, formatting code works.

## Actual behavior
`Lisp error: (wrong-type-argument stringp nil)` when formatting code in a cljs buffer

1) cider-connect-clj
2) cider-format-defun in a .cljs buffer

### CIDER version information

```
;; CIDER 0.18.1snapshot (package: 20180929.818), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_102
```

kommen:
This makes formatting code always use the clj repl type, which makes the `cider-format-*` commands work in cljs buffers even if there is no cljs repl active.

Fixes #2466 

[This is my first contribution to cider, discussion & feedback very welcome]

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -153,7 +153,7 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
                                     \"par1\" ... ).
 If CONNECTION is provided dispatch to that connection instead of
 the current connection.  Return the id of the sent message."
-  (nrepl-send-request request callback (or connection (cider-current-repl))))
+  (nrepl-send-request request callback (or connection (cider-current-repl "any"))))
 
 (defun cider-nrepl-send-sync-request (request &optional connection abort-on-input)
   "Send REQUEST to the nREPL server synchronously using CONNECTION.
@@ -163,13 +163,13 @@ If ABORT-ON-INPUT is non-nil, the function will return nil
 at the first sign of user input, so as not to hang the
 interface."
   (nrepl-send-sync-request request
-                           (or connection (cider-current-repl))
+                           (or connection (cider-current-repl "any"))
                            abort-on-input))
 
 (defun cider-nrepl-send-unhandled-request (request &optional connection)
   "Send REQUEST to the nREPL CONNECTION and ignore any responses.
 Immediately mark the REQUEST as done.  Return the id of the sent message."
-  (let* ((conn (or connection (cider-current-repl)))
+  (let* ((conn (or connection (cider-current-repl "any")))
          (id (nrepl-send-request request #'ignore conn)))
     (with-current-buffer conn
       (nrepl--mark-id-completed id))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 * Fix jack-in from inside of remote buffers.
 * [#2454](https://github.com/clojure-emacs/cider/pull/2454): Fix erratic inspector behavior when multiple REPLs are connected
+* [#2467](https://github.com/clojure-emacs/cider/pull/2467): Make generic cider ops use any available nrepl connection
 
 ## 0.18.0 (2018-09-02)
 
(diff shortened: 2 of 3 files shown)


bbatsov:
> [This is my first contribution to cider, discussion & feedback very welcome]

Congratulations! Thanks for the nicely executed PR! Hopefully it's going to be the first of many - we sure need more people to help out!

cider-client.el
```diff
@@ -533,7 +533,7 @@ The result entries are relative to the classpath."
   "Perform nREPL \"format-code\" op with CODE."
   (thread-first `("op" "format-code"
                   "code" ,code)
-    (cider-nrepl-send-sync-request)
+    (cider-nrepl-send-sync-request (cider-current-repl "clj"))
```
bbatsov: Looking at the other code in the file I think of two things:

* what happens if you only have a cljs connection? Seems to me the formatting should still work, as it's not really clj or cljs specific
* I see we're utilizing the same pattern for other ops as well, which means they would likely exhibit the same problem as the formatting issue

kommen: > what happens if you only have a cljs connection? Seems to me the formatting should still work, as it's not really clj or cljs specific

True. I didn't consider having just a cljs repl to be a thing.

So I guess `"multi"` would be the right thing to pass to `cider-current-repl`?

> I see we're utilizing the same pattern for other ops as well, which means they would likely exhibit the same problem as the formatting issue

Perhaps, yes. I just investigate since I didn't run into any of those other issues yet.

kommen: Just tested using `(cider-current-repl "multi")` and it does make it work with either an cljs or an clj repl active.

kommen: re other ops: additionally to change other ops, utility functions like `cider-nrepl-op-supported-p` would also have to be changed.

kommen:
@bbatsov would you prefer closing this PR in favor of a more thorough one addressing this in a general fashion for all ops?

bbatsov:
We'll certainly need a generic solution for this, but I'll have to think a bit about what makes the most sense. 

I think historically we sent all middleware evaluations to the Clojure REPL, as before 0.18 it wasn't possible to have just a ClojureScript REPL. @vspinu What do you think about this?

vspinu:
Sorry for late reply (I am on vacation atm). I think any connection would do for the purpose of format and other generic ops. Hence "multi" seems like the right approach. Unfortunately in the context of `cider-current-repl` "multi" is a misnomer, maybe we could have "any" as a synonym here.

bbatsov:
> Hence "multi" seems like the right approach. Unfortunately in the context of cider-current-repl "multi" is a misnomer, maybe we could have "any" as a synonym here.

I like this!

kommen:
@bbatsov @vspinu I've pushed an attempt to as you suggested. As [eval uses it's connection selection](https://github.com/clojure-emacs/cider/blob/e071b027c1dafea8599db8c2b039f11d9eacf994/cider-client.el#L178-L191), it seemed safe to me to just adjust the `cider-nrepl-send-sync-request`, `cider-nrepl-send-request` and `cider-nrepl-send-unhandled-request` functions.

What do you think?

cider-client.el
```diff
@@ -153,7 +153,7 @@ REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
                                     \"par1\" ... ).
 If CONNECTION is provided dispatch to that connection instead of
 the current connection.  Return the id of the sent message."
-  (nrepl-send-request request callback (or connection (cider-current-repl))))
+  (nrepl-send-request request callback (or connection (cider-current-repl "any"))))
```
bbatsov: Seems to me we should assume "any" by default.

bbatsov: Actually forget about this.

bbatsov:
Looks good to me overall. Just add some entry in the changelog for this.

kommen:
@bbatsov thanks, I've added the changelog entry

PR #2471, opened 2018-10-01
Fix typos in clj nrepl command documentation

AdamFrey:
The cider nrepl command in the manual had two missing characters

-----------------

Thanks!


bbatsov:
Thanks!

PR #2478, opened 2018-10-10
Ensure and mention Cljs Figwheel Main options

arichiardi:
This patch checks and ensures we send back a sensible error in case of missing
Figwheel Main options and makes sure we mention them and link to the Scripting
API page.

-----------------


arichiardi:
I will close this one, because the option map position right there is not working, I misread the doc: Figwheel accepts an option map *and* a build id in the multi-arity `figwheel.main.api/start`. Sorry for the noise.

PR #2483, opened 2018-10-11
Implement friendly sessions

### LINKED ISSUE #2408: Jumping to a library definition opens a file not bound to any CIDER session
I have a workflow when I <kbd>M-.</kbd> into a function defined by a dependency. This opens a code buffer from the JAR, which I then can edit and evaluate forms to make some quick runtime-only hacks to the library. Now, with the connection handling rewrite, this changed – a buffer opened by <kbd>M-.</kbd> is not bound to a CIDER session (the message in the minibuffer is `No linked CIDER sessions` when I try to evaluate anything).

The question is whether it would be sensible to propagate the current CIDER connection to any buffers opened by `Go to definition`.

Meanwhile, I'm fine manually linking the buffer to the session for this particular usecase.

```
;; CIDER 0.18.0snapshot (package: 20180808.1857), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_102
```

### LINKED ISSUE #2446: Checkouts no longer working
## Expected behavior

When switching to a namespace for a code file that's in checkouts, the REPL should go there. Same for evaluating a form in that code file.

## Actual behavior

Cider says "no linked CIDER sessions"

Create project with checkouts. Open file from one of the checkouts folders.

### CIDER version information

```
;; CIDER 0.18.0 (Saigon), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_181
```

### Lein/Boot version

Leiningen 2.7.1

### Emacs version

25.3.50.2

### Operating system

Ubuntu 16.04

### LINKED ISSUE #2464: There have been so many reports from users that want to eval code from random context into whatever their most recent REPL is that we should probably add an option to allow this.
There's at least 1 person complaining about the changes we've made every day. :-) 

For that particular issue I agree with your second suggestion.

_Originally posted by @bbatsov in https://github.com/clojure-emacs/cider/issues/2446#issuecomment-420600951_

vspinu:
Implement "friendly" sessions - sessions linked to dependent projects. Fix #2446, close #2464 and re-fix #2408 in a more elegant way.

The best way I could find to implement this is by searching through the classpath for the buffer file path. I have no estimation of how fast this is and whether it could cause performance issues. Remains to be seen. I am caching the classpath on the first use and haven't seen any issues in my experiments.

Regarding #2464, I will add at some point an option to sesman for picking first available session in non-linked contexts, but let's first see if this is really necessary for the CIDER's use case. 

-----------------

- [x] All code passes the linter (`make lint`) which is based on [`elisp-lint`]


arrdem:
Nice! Note that there may be some unfortunate intricacies with (non)absolute paths in the classpath here, and buffers whose file names aren't absolute. Spoken as someone who uses a lot of symlinks ;)

cider.el
```diff
@@ -1235,7 +1235,7 @@ canceled the action, signal quit."
                                          (equal port (plist-get ses-params :port)))
                                      (or (null host)
                                          (equal host (plist-get ses-params :host))))))
-                            (sesman-linked-sessions 'CIDER '(project)))))
+                            (sesman-current-sessions 'CIDER '(project)))))
```
bbatsov: Btw, shouldn't this be just `sesman-sessions` or something? Both `linked` and `current` seem misleading to me.

bbatsov:
> Implement "friendly" sessions - sessions linked to dependent projects

And jars I guess.

cider.el
vspinu: There is already `sesman-sessions` which returns ALL sessions for a system. `linked` and `current` have slightly different semantics. Linked should be very clear given that linking is such a central concept in sesman; "current" conveys a general idea of "active" sessions in the current context. It basically means linked + friendly sessions.

cider-connection.el
```diff
@@ -417,6 +417,24 @@ REPL defaults to the current REPL."
...
+      (let* ((classpath (or (process-get proc :cached-classpath)
+                            (let ((cp (with-current-buffer repl
+                                        (cider-sync-request:classpath))))
+                              (process-put proc :cached-classpath cp)
+                              cp)))
+             (file (file-truename file))
+             (file-len (length file)))
+        (seq-find (lambda (path)
+                    (when (>= file-len (length path))
+                      (string= path (substring file 0 (length path)))))
```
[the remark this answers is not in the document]
vspinu: Ha. I didn't know about string-prefix-p. Will do.

doc/managing_connections.md
```diff
@@ -50,11 +50,27 @@ Sessions can be linked to contexts (projects, directories and buffers)
   - <kbd>C-c C-s p</kbd> `sesman-link-with-project`
   - <kbd>C-c C-s u</kbd> `sesman-unlink`
 
+## Friendly Sessions
```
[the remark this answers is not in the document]
vspinu: I thought of "external" after similar concept in project.el but I don't like it much. I think "friendly" is pretty suggestive and generic. There might be other notions of "friendliness" than dependency. If you have other ideas I am all ears.

vspinu:
> And jars I guess.

Yes.

cider.el
bbatsov: Yeah, I get this, but I feel that `sesman-sessions` should just return all sessions by default and narrow this down depending on optional params passed to it. Seems to me we've used `relevant` instead of `current` in CIDER, but I generally feel that in the present of the params it's pretty clear what sessions we're talking about without adding more functions to the API.

vspinu: I agree in general and in case of `sesman-sessions` this  might actually work better. There is also `sesman-links` and `sesman-current-links`. Plus `sesman-current-session` which is the `(car (sesman-current-sessions))`.  Let me see if I could simplify the api without making too many tradeoffs.

vspinu: I gave it a bit of thought and I am afraid the `sesman-current-sessions` and `sesman-current-links` will have to stay. The reason for the "current" instead of "relevant" is because those are sessions active in the _current_ context and the `sesman-current-session` is  the most relevant session for the _current_ context. The latter is also how CIDER calls its most relevant repl. The reason for not subsuming `sesman-current-session` into `sesman-sessions` is because I would need to subsume `sesman-current-links` (which is a bit trickier) and the CIDER code would have to look like `(sesman-sessions 'CIDER '(linked friendly) 'sort)` which is rather annoying as all readers will have to learn about all those internals, while `sesman-current-sessions` is kinda self-explanatory. 

It makes sense to subsume `sesman-friendly-sessions` and `sesman-linked-sessions` though. Those are somewhat internal. Looking into that right now.

vspinu:
Done!

PR #2484, opened 2018-10-12
Fix shortcut for REPL buffers when type is "any"

kommen:
This fixes a regression from #2467 where in clojure repl buffers the cider eldoc feature stopped working.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -728,6 +728,7 @@ are synonyms.  If ENSURE is non-nil, throw an error if either there is
 no linked session or there is no REPL of TYPE within the current session."
   (if (and (derived-mode-p 'cider-repl-mode)
            (or (null type)
+               (string= "any" type)
                (string= cider-repl-type type)))
       ;; shortcut when in REPL buffer
       (current-buffer)
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -644,13 +644,13 @@ PARAMS is a plist as received by `cider-repl-create'."
     (let* ((proj-dir (plist-get params :project-dir))
            (host (plist-get params :host))
            (port (plist-get params :port))
-           (cljsp (string-match-p "cljs" (symbol-name (plist-get params :repl-type))))
+           (cljsp (member (plist-get params :repl-type) '(cljs pending-cljs)))
            (scored-repls
             (delq nil
                   (mapcar (lambda (b)
                             (let ((bparams (cider--gather-connect-params nil b)))
-                              (when (eq cljsp (string-match-p "cljs"
-                                                              (symbol-name (plist-get bparams :repl-type))))
+                              (when (eq cljsp (member (plist-get bparams :repl-type)
+                                                      '(cljs pending-cljs)))
                                 (cons (buffer-name b)
                                       (+
                                        (if (equal proj-dir (plist-get bparams :project-dir)) 8 0)
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -712,7 +712,7 @@ function with the repl buffer set as current."
 (defun cider--no-repls-user-error (type)
   "Throw \"No REPL\" user error customized for TYPE."
   (let ((type (cond
-               ((equal type "multi")
+               ((or (equal type "multi") (equal type "any"))
                 "clj or cljs")
                ((listp type)
                 (mapconcat #'identity type " or "))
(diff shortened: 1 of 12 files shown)


kommen:
Test failures seem to be due to an upstream change: https://github.com/vspinu/sesman/issues/10

cider-connection.el
```diff
@@ -702,7 +702,8 @@ are synonyms.  If ENSURE is non-nil, throw an error if either there is
 no linked session or there is no REPL of TYPE within the current session."
   (if (and (derived-mode-p 'cider-repl-mode)
            (or (null type)
-               (string= cider-repl-type type)))
+               (string= cider-repl-type type)
+               (string= "any" type)))
```
bbatsov: Hmm, as `any` and `multi` can't be synonyms here maybe we should treat them differently by default? It also seems to me that this check should be after the `null` check, so the code reads better. 

On a related note - can you please change these types to be symbols instead of strings everywhere. It's very weird to pass around types as strings in Elisp and I guess this was just some oversight of ours when REPL types were added. For backward compatibility you can just coerce type arguments to symbol in methods like this one (there's a function `cider-maybe-symbol` or something like this.

bbatsov: Also - if we're using symbols we can use the more effecient `eq` checks instead of `string=` and `equal`.

kommen: > Hmm, as any and multi can't be synonyms here maybe we should treat them differently by default?

How do you men differently?

I can take care of the rest.

vspinu: The fix doesn't look quite right to me. This function becomes unreadable with all those if-else and cons. 

Any objection against reverting #2467 and just adding ` (when (equal type "any") (setq type "multi"))` as the first line in the function?

kommen: > Any objection against reverting #2467 and just adding (when (equal type "any") (setq type "multi")) as the first line in the function?

How would this address the original issue #2467 fixed?

bbatsov: I think it'd better to use something like `(member type '(any multi))` instead of converting `any` to `multi`. As a said - there are cases where semantically they won't be exactly the same, so I'd rather keep them separate.

vspinu: This would need to be done at least twice in this function and at least once in the utility function. Plus there is already a re-assignment to "multi" in the `let` part.  I If the semantics of this function ever diverges one can modify it later, for now the simplicity of the code is more important. None of us noticed the issue with the original PR. 

@kommen
 sorry, by reverting I meant reverting the "any" implementation for this function only, not the whole patch

bbatsov:
@kommen Ping :-)

kommen:
@bbatsov I've addressed two of your comments.

Regarding changing the "clj", "cljs", "any" and "multi" repl types from strings to symbols:
I don't think I fully grasp the implications of changing it everywhere, e.g here: https://github.com/clojure-emacs/cider/blob/d1eef83124759ee4efa5e37b8406dcb5d71f8c98/cider.el#L1011

You think it is fine to change it here as well?

bbatsov:
Yep. In the end there should be not strings referring to repl types, only symbols.

cider-connection.el
```diff
@@ -632,12 +632,13 @@ PARAMS is a plist as received by `cider-repl-create'."
...
            (host (plist-get params :host))
            (port (plist-get params :port))
-           (cljsp (string-match-p "cljs" (plist-get params :repl-type)))
+           (cljsp (string-match-p "cljs" (symbol-name (plist-get params :repl-type))))
            (scored-repls
             (delq nil
                   (mapcar (lambda (b)
                             (let ((bparams (cider--gather-connect-params nil b)))
-                              (when (eq cljsp (string-match-p "cljs" (plist-get bparams :repl-type)))
+                              (when (eq cljsp (string-match-p "cljs"
```
kommen: Anybody knows what `cljsp` is supposed to be about? I assume it is to consider `pending-cljs` repls to be reused by `cljs` repls? But not sure.

kommen:
@bbatsov I've pushed the current state of my branch. Things still needed:

- [ ] figure out which functions need to be made backwards compatible by accepting strings and turn them into symbols with `cider-maybe-intern`
- [ ] ensure https://github.com/clojure-emacs/cider/pull/2484/commits/fec7691951db9df1e01b9f28278a87f75b274418#r228822378 is doing the right thing
- [ ] investigate ci failure
- [ ] update doc strings
- [ ] add changelog

Not sure I will have time this week to address these though.

cider-connection.el
vspinu: It's non-nil when it's either cljs or pending-cljs.

kommen:
This is ready for review.

bbatsov:
Sounds like a plan to me!

bbatsov:
Is the PR ready to go?

kommen:
using it the branch for day to day stuff i found some more issues stilll. i’d rather test it a bit more.

kommen:
but i don’t know of any remaining issues right now

bbatsov:
Well, if you work it day-to-day I guess it doesn't break anything major. :-) Anyways, I'm fine with you testing it longer if you prefer that.

kommen:
@bbatsov I've squashed the commits to combine them logically. 

As [pauld on the slack](https://clojurians.slack.com/archives/C0617A8PQ/p1542549831353600) also tried the branch and confirmed it fixed an issue for him, I think this should be good to merge.

bbatsov:
Thanks!

arichiardi:
Awesome job folks I have just bumped into this and wanted to say thanks!

PR #2493, opened 2018-10-18
Quote middleware parameter when invoking clj

plexus:
Add quotes around the --middleware flag

Before: `--middleware ["cider.nrepl/nrepl-middleware"]`
After:  `--middleware '["cider.nrepl/nrepl-middleware"]'`

This prevents issues with shells like zsh that treat square brackets special.


bbatsov:
👍

PR #2500, opened 2018-10-30
Fix face names in docstring of cider-font-lock-dynamically variable

mallt:
This PR fixes the names of the faces that are used for macros and vars in the docstring of the `cider-font-lock-dynamically` variable.


bbatsov:
Thanks!

Issue #1544, opened 2016-02-04
Improve SSH behavior in cider-connect


cap10morgan:
As per the discussion in #1142, we should improve how and when CIDER decides it needs to try an SSH connection to a remote host in `cider-connect`.

The preferred option seems to be to do 2 things:
1. Add a custom variable for SSH files to sniff for ports, like `'(~/.ssh/known_hosts ~/.ssh/config)`. It's value would default to `'(~/.ssh/config)`.
2. Have a user level logical `cider-sniff-for-ports-when-connected-only` and use `(file-remote-p file nil t)` to check if there is already an emacs connection open outside of cider to that server.

This should serve to make `cider-connect`'s behavior less surprising for users. Currently it tries to SSH to any hostname in `~/.ssh/known_hosts` which is often incorrect behavior and is difficult to debug in those cases.

Malabarba:
If someone is going to tackle this, I'd suggest looking into tramp too. Tramp is Emacs' default interface to ssh (and other protocols), and many aspects of Emacs (besides just visting files) are already integrated with it, like `dired`, `M-x shell`, and `eww`.

It might be overkill for us to offer this integration, but it might also be that tramp has some functions/variables we can reuse for deciding what ports to sniff.

bbatsov:
Perhaps using tramp is the way to... I'm not sure...

1. should be pretty easy to do, but number 2. might turn out to be pretty complex.

Frozenlock:
Is there a way to avoid this behavior?

I usually open an SSH connection with port forwarding if I want a secure interaction with a remote REPL.

Now I'm in a situation where **I don't want** SSH and yet `cider-connect` is trying on its own to use it.

Malabarba:
@frozenlock maybe I misunderstood, but it looks like you're experiencing a genuine bug, while the current issued is really just a slightly related feature request. If that's the case, Could you file a separate issue describing how to reproduce your problem?

Frozenlock:
Emacs crashing is indeed a bug unrelated to this issue.

However, I'd still prefer to have a way to disable the automatic SSH fallback, which is pertinent for this issue.

Malabarba:
I see. Thanks for explaining. 👍

aiba:
I also have a bit of trouble getting cider to do a direct connection when I want it.

[...]

PR #2506, opened 2018-11-02
add defcustom cider-use-ssh-to-infer-remote-ports

aiba:
This is just an idea to address #1544.  If the cider maintainers want it, I can update this PR to also include documentation of the new defcustom.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 ### Changes
 
 * [#2484](https://github.com/clojure-emacs/cider/pull/2484): REPL types are now symbols instead of strings.
+* [#1544](https://github.com/clojure-emacs/cider/issues/1544): Add a new defcustom `cider-infer-remote-nrepl-ports` to control whether we use tramp/ssh to infer remote ports.  Now defaulting to `nil` (previously it always tried to infer).
 
 ## 0.18.0 (2018-09-02)
 
--- a/cider.el
+++ b/cider.el
@@ -1327,6 +1327,12 @@ Tramp version starting 26.1 is using a `cl-defstruct' rather than vanilla VEC."
       (make-tramp-file-name :method (elt vec 0)
                             :host   (elt vec 2)))))
 
+(defcustom cider-infer-remote-nrepl-ports nil
+  "When true, cider will use ssh to try to infer nREPL ports on remote hosts."
+  :type 'boolean
+  :safe #'booleanp
+  :package-version '(cider . "0.19.0"))
+
 (defun cider--infer-ports (host ssh-hosts)
   "Infer nREPL ports on HOST.
 Return a list of elements of the form (directory port).  SSH-HOSTS is a list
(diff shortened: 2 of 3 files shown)


bbatsov:
I don't use tramp, so I don't really care about this much. With me it's perfectly fine to just delete everything tramp-related, as it has been nothing but a pain in the ass for me. At this point I hate tramp (both because of CIDER and Projectile). :-) 

On the other hand a lot of people seem to be using it for some weird reason... In principle I'm fine with the suggested change.

cider.el
```diff
@@ -1338,14 +1343,15 @@ of remote SSH hosts."
...
-      (let ((vec (vector "sshx" nil host "" nil))
-            ;; change dir: user might want to connect to a different remote
-            (dir (when (file-remote-p default-directory)
-                   (with-parsed-tramp-file-name default-directory cur
-                     (when (string= cur-host host) default-directory)))))
-        (tramp-maybe-open-connection (cider--tramp-file-name vec))
-        (with-current-buffer (tramp-get-connection-buffer (cider--tramp-file-name vec))
-          (cider-locate-running-nrepl-ports dir))))))
+      (when cider-use-ssh-to-infer-remote-nrepl-ports
+        (let ((vec (vector "sshx" nil host "" nil))
```
bbatsov: I wonder if we shouldn't move the remote logic to a separate function.

cider.el
```diff
@@ -1327,6 +1327,11 @@ Tramp version starting 26.1 is using a `cl-defstruct' rather than vanilla VEC."
       (make-tramp-file-name :method (elt vec 0)
                             :host   (elt vec 2)))))
 
+(defcustom cider-use-ssh-to-infer-remote-nrepl-ports t
```
bbatsov: Probably we can shorten this simply to `cider-infer-remote-nrepl-ports`. 
The defcustom should be moved to the top of the file and annotated accordingly.

aiba:
> With me it's perfectly fine to just delete everything tramp-related, as it has been nothing but a pain in the ass for me.

Yeah, same with me.  My workflow is that I always edit files locally.  When I connect to a remote repl, it's always a direct connection and I use git to sync files, rather than editing files remotely with tramp.

But I'm sympathetic to the workflow of editing files remotely with tramp.

Another strategy would be to have a master var, `cider-enable-ssh-integration` or `cider-enable-tramp-integration`, and then hide everything behind this var so non-tramp users (like us) can disable all those code paths with one var.  Would you prefer that strategy?

cider.el
aiba: I tried that, but it didn't seem like a big win, especially since this function is already entangled with an `ssh-hosts` arg.

aiba:
>  I think it's fine for now to have two variables, both of which should be nil by default and should be well documented

OK, i'll clean this up and add docs for it.

aiba:
OK, I think this PR is ready for another look.  I considered adding documentation to `doc/configuration.md` but I'm not sure it's worth it given how infrequently we expect people to turn this variable on, and that we might want to phase it out in the future.

bbatsov:
I think it might actually be nice to add some section on connecting over ssh, but that's not that big of a deal.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 ### Changes
 
 * [#2482](https://github.com/clojure-emacs/cider/issues/2482): Don't bind nREPL server started by `cider-jack-in` to `::` (use `localhost` instead).
+* [#1544](https://github.com/clojure-emacs/cider/issues/1544): Add a new defcustom `cider-infer-remote-nrepl-ports` to control whether we use tramp/ssh to infer remote ports.  Now defaulting to `nil` (previously it alwasy tried to infer).
```
bbatsov: alwasy -> alwasy

cider.el
```diff
@@ -1327,6 +1327,11 @@ Tramp version starting 26.1 is using a `cl-defstruct' rather than vanilla VEC."
       (make-tramp-file-name :method (elt vec 0)
                             :host   (elt vec 2)))))
 
+(defcustom cider-infer-remote-nrepl-ports nil
+  "When true, cider will use ssh to try to infer nREPL ports on remote hosts."
+  :type 'boolean
+  :safe #'booleanp)
```
bbatsov: You should also add a `:package-version` property (0.19.0).

bbatsov:
Btw, it seems to me that the two ssh options should probably be mentioned in http://www.cider.mx/en/latest/up_and_running/, as this makes more sense than the generic configuration section.

aiba:
> Btw, it seems to me that the two ssh options should probably be mentioned in http://www.cider.mx/en/latest/up_and_running/, as this makes more sense than the generic configuration section.

I can take a stab at docs for this.  To me it makes sense to have a section somewhere (perhaps up-and-running, or perhaps a new page) called "Connecting over SSH".  And I'm not sure I understand all of cider's behavior wrt ssh connections well enough to write this section, but I'll give it a try.

aiba:
I think this is ready for another look.

PR #2510, opened 2018-11-04
[Fix #2508] Include keyword and namespace jumping in cider-find-dwim

### LINKED ISSUE #2508: Cider-find-dwim should jump to keyword
Hello folks!

This is not a problem per se but something I find a bit not straightforward to understand.
Basically when I call `cider-find-dwim` on a keyword, I expect it to use `cider-find-keyword`.

I have to keep this distinction in my head at them moment while the `-dwim` command suggests it should handle whatever I mean :smile: 

Not a blocker, not a thing even maybe :smile:

arichiardi:
They were not previously included so it did not felt dwim enough.

-----------------

### DIFF
--- a/cider-find.el
+++ b/cider-find.el
@@ -84,22 +84,28 @@ CALLBACK upon failure to invoke prompt if not prompted previously.
 Show results in a different window if OTHER-WINDOW is true."
   (if-let* ((info (cider-var-info symbol-file)))
       (cider--jump-to-loc-from-info info other-window)
-    (progn
-      (cider-ensure-op-supported "resource")
-      (if-let* ((resource (cider-sync-request:resource symbol-file))
-                (buffer (cider-find-file resource)))
-          (cider-jump-to buffer 0 other-window)
-        (if (cider--prompt-for-symbol-p current-prefix-arg)
-            (error "Resource or var %s not resolved" symbol-file)
-          (let ((current-prefix-arg (if current-prefix-arg nil '(4))))
-            (call-interactively callback)))))))
+    ;; jumps and returns point or does not jump and returns nil
+    (unless (cider--find-keyword symbol-file other-window 'noerror)
+      ;; jumps and returns point or does not jump and returns nil
+      (unless (cider--find-ns symbol-file other-window 'noerror)
+        (progn
+          (cider-ensure-op-supported "resource")
+          (if-let* ((resource (cider-sync-request:resource symbol-file))
+                    (buffer (cider-find-file resource)))
+              (cider-jump-to buffer 0 other-window)
+            (if (cider--prompt-for-symbol-p current-prefix-arg)
+                (error "Resource or var %s not resolved" symbol-file)
+              (let ((current-prefix-arg (if current-prefix-arg nil '(4))))
+                (call-interactively callback)))))))))
 
 (defun cider--find-dwim-interactive (prompt)
   "Get interactive arguments for jump-to functions using PROMPT as needed."
-  (if (cider--prompt-for-symbol-p current-prefix-arg)
-      (list
-       (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
-    (list (or (thing-at-point 'filename) ""))))  ; No prompt.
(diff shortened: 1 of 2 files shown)


cider-find.el
```diff
@@ -84,22 +84,28 @@ CALLBACK upon failure to invoke prompt if not prompted previously.
...
-    (progn
-      (cider-ensure-op-supported "resource")
-      (if-let* ((resource (cider-sync-request:resource symbol-file))
-                (buffer (cider-find-file resource)))
-          (cider-jump-to buffer 0 other-window)
-        (if (cider--prompt-for-symbol-p current-prefix-arg)
-            (error "Resource or var %s not resolved" symbol-file)
-          (let ((current-prefix-arg (if current-prefix-arg nil '(4))))
-            (call-interactively callback)))))))
+    ;; jumps and returns point or does not jump and returns nil
```
bbatsov: I'm not very found of this approach of nested conditionals and optional `no-error` params. It seems to me it'd be better if we just had some pipeline of functions which try to resolve to some file and we invoke the find-file with results of whatever function returns first. Maybe the dwim functions should just be some list of functions and then the user can also customize those?

bbatsov:
Glad to see you tackling this one, @arichiardi!

cider-find.el
arichiardi: I see the point of customizing but I see the dwim command as some sort of magic that handles everything for me. Also I did not want it to be too clever - it is basically just side effects and jumps - I would have needed way more work for making it pretty.

About the `'noerror` stuff, I too disliked it but again it's been done this way so that I would not refactor all these side effectful functions...

arichiardi: Ok will do my best if you think that.

bbatsov: Thanks!

arichiardi: Had a look at this again today, it feels like too much of a monster to tackle. It is not just the jumps, the code also has a lot of `completing-read` all over.

I have to say this is unfortunately the kind of piece of code that I don't know if I want to touch for risking of breaking/having to fix things.

It is a daunting task that I don't know if I have strength for facing now. And I do not have Excalibur with me either :crossed_swords:  

Any suggestion on how to proceed?

bbatsov: The should be just a single jump. All internal functions should just try to return something you can jump to, they should be evaluated in order until one of them returns something and then you jump. No internal function should be have completing reads - those make sense only for the wrapper commands.

arichiardi:
So @bbatsov ideally we should have something like this to pass to `cider--jump-to-loc-from-info`.

```elisp
;; this should probably go where cider-var-info lives.
(defun cider-all-info (symbol)
  "Retrieve the info data structure for SYMBOL-FILE.

Return nil if and only if nothing can be retrieved."
  (or (cider-var-info symbol-file)
      (cider-keyword-info symbol-file)
      (cider-resource-info symbol-file)))

(defun cider--find-dwim (symbol-file callback &optional other-window)
  "Find the SYMBOL-FILE at point.
CALLBACK upon failure to invoke prompt if not prompted previously.
Show results in a different window if OTHER-WINDOW is true."
  (when-let ((info (or (cider-all-info symbol-file))))
    (cider--jump-to-loc-from-info info other-window)))
```

The `cider-keyword-info` is the one that is really missing. And it probably would also solve - if I understand how this all work - the issue #2523.

arichiardi:
The only other thing that I am still investigating is what the `resource` call to nrepl return and what is its use case. It seems like it would not return something compatible with `cider--jump-to-loc-from-info`.

arichiardi:
Actually even `(cider-sync-request:ns-path ns)` is kind of different. Returning something else.

arichiardi:
I am thinking that if the `info` map had a key to indicate which kind of info we are returning `:namespace`, `:var`, etc, ... this functionality would be way leaner.

bbatsov:
Yeah, probably some unification of the API would be a good idea. Much of this functionality was developed in a piece-meal manner and we didn't think much about how all the pieces would fit together.

arichiardi:
Oh no, this didn't get in after all?

PR #2512, opened 2018-11-06
[inspector] Use visual-line-mode instead of truncate-lines

alexander-yakushev:
I'm not really sure what `truncate-lines` has been achieving here. I removed it, but I can bring it back if it is still useful.

Before:

screen shot 2018-11-07 at 00 28 40

After:

screen shot 2018-11-07 at 00 28 57


bbatsov:
Fine by me! 👍

PR #2516, opened 2018-11-09
Improve friendly session detection

vspinu:
Currently detection of  sessions in dependency projects (friendly sessions in sesman terminology) fails in two cases - AVFS paths and files in the root of the project. The latter is because classpath doesn't include the path to the root. This patch fixes both issues. 

-----------------


bbatsov:
👍

PR #2517, opened 2018-11-09
[Fix #2514] Don't auto-jump to warnings even when they are sent through *err*

### LINKED ISSUE #2514: Restrict auto jump to reflection warnings in the current project only
Currently, when `*warn-on-reflection*` is true and `cider-atuo-jump-to-error` is non-nil, each time some external code with reflections (notably cider-nrepl itself) is loaded the user is faced with a pop-up buffer in that foreign project.   

I think a meaningful solution would be to auto-jump only to locations in the current project only.

vspinu:
It's hard to think of anything more reliable than this. Either on nrepl or emacs side there should be a regexp search for "warning" keyword. 

-----------------


bbatsov:
I think that's reasonable fix, but I it also seems strange that the reflection warnings are marked as errors. Probably we should fix this as well.

vspinu:
They are printed to the *err* stream in the clojure compiler. So either on cider-nrepl side one has to regexp and redirect to warn response, or do what this PR does.

PR #2522, opened 2018-11-17
Improve the Makefile autoload and compilation sections

arichiardi:
Inspired by the immense magit, we now have a similar way to build the
autoloads.el file and additionally we combine both autoload and compilation
steps under the build target. With this change, manual installation and tools
like borg can just call make and have everything ready for consumption.

-----------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,15 @@ export EMACS ?= emacs
 EMACSFLAGS = -L .
 CASK = cask
 VERSION = $(shell git describe --tags --abbrev=0 | sed 's/^v//')
-PACKAGE_NAME = cider-$(VERSION)
+PKG = cider
 
-ELS = $(wildcard *.el)
-LINTELS = $(filter-out cider-autoloads.el,$(ELS))
+ELS_ALL = $(wildcard *.el)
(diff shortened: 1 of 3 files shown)


arichiardi:
I need to tweak the readme a bit still, I found a better way to load autoloader, please do not merge

Makefile
```diff
@@ -2,13 +2,16 @@ export EMACS ?= emacs
...
 CASK = cask
 VERSION = $(shell git describe --tags --abbrev=0 | sed 's/^v//')
-PACKAGE_NAME = cider-$(VERSION)
+PKG = cider
 
-ELS = $(wildcard *.el)
-LINTELS = $(filter-out cider-autoloads.el,$(ELS))
+ELS_ALL = $(wildcard *.el)
+ELS = $(filter-out $(PKG)-autoloads.el,$(ELS_ALL))
+LINTELS = $(filter-out $(PKG)-autoloads.el,$(ELS_ALL))
```
bbatsov: This seems to be the same as ELS.

arichiardi: Right thank you I will modify it

arichiardi:
Pingy ping :smile:

PR #2524, opened 2018-11-19
Re-set the current buffer after display-buffer

glittershark:
If the user has `:select nil` in their popup rules for the Cider REPL
buffer, the call to `display-buffer` to display the repl buffer while
initializing it will reset the `current-buffer` to the buffer the user
ran `jack-in` from, which breaks any successive calls to
`cider-current-repl`. Resetting the current buffer back to the original
buffer after `display-buffer` is resilient against this config change,
without breaking anything else in the case the config is absent.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
The patch seems reasonable to me. Thanks!

PR #2526, opened 2018-11-22
[Fix #2474] Fix end-of-output detection

### LINKED ISSUE #2474: Background threads print to the wrong position
## Expected behavior

When "background" threads print to `*out*`, the output should appear at the end of the buffer, like it does normally.  Here's how the bare repl behaves:

```
Clojure 1.8.0
user=> (do
        (println "before")
        (doto (Thread.
               (fn []
                 (Thread/sleep (rand 500))
                 (println "in thread")))
          (.start)
          (.join))
        (println "after"))
before
in thread
after
```

## Actual behavior

"Background" threads seem to save the position in the buffer when the thread is started and put their `*out*` starting there:

```
user>
in thread (do
        (println "before")
        (doto (Thread.
               (fn []
                 (Thread/sleep (rand 500))
                 (println "in thread")))
          (.start)
          (.join))
        (println "after"))
before
after
```
Note that I only pasted the `(do ...)` form into the repl, directly after the `user>` prompt.  The output of the thread that was started is printed at a new line, and the input form is moved after that output.

Paste into a cider repl:
```
(do
        (println "before")
        (doto (Thread.
               (fn []
                 (Thread/sleep (rand 500))
                 (println "in thread")))
          (.start)
          (.join))
        (println "after"))
```

OSX, Clojure 1.8

### CIDER version information

```
;; CIDER 0.16.0 (Riga), nREPL 0.2.13
;; Clojure 1.8.0, Java 1.8.0_101
```

### Lein/Boot version

```
Leiningen 2.8.1 on Java 1.8.0_101 Java HotSpot(TM) 64-Bit Server VM
```

### Emacs version

```
GNU Emacs 25.3.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911)) of 2017-09-12
```

### Operating system

```
OSX 10.13.6
```

vspinu:
The original issue comes down to the incorrect end-of-output detection (rationale of why #2474 should be fixed on emacs side and not on nrepl side is [here](https://github.com/nrepl/nrepl/issues/62#issuecomment-441015237)). 

I am renaming `cider-repl--end-of-line-before-input-start` into `cider-repl--end-of-output` because the former is less suggestive of what that function is supposed to do.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -677,8 +677,7 @@ If BOL is non-nil insert at the beginning of line.  Run
 (defun cider-repl--emit-interactive-output (string face)
   "Emit STRING as interactive output using FACE."
   (with-current-buffer (cider-current-repl)
-    (let ((pos (cider-repl--end-of-line-before-input-start))
-          (string (replace-regexp-in-string "\n\\'" "" string)))
+    (let ((pos (cider-repl--end-of-output)))
       (cider-repl--emit-output-at-pos (current-buffer) string face pos t))))
 
 (defun cider-repl-emit-interactive-stdout (string)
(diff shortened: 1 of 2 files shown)


cider-repl.el
```diff
@@ -1055,10 +1054,15 @@ See also the related commands `cider-repl-clear-output' and
     (recenter t))
   (run-hooks 'cider-repl-clear-buffer-hook))
 
-(defun cider-repl--end-of-line-before-input-start ()
+(defun cider-repl--end-of-output ()
```
bbatsov: I guess you should also update the docstring. I don't like the term output very much as it's unclear whether this refer to a value or string output, so this should probably be clarified in the docstring. Apart from this the changes look good.

bbatsov:
> closes nrepl/nrepl#86, closes nrepl/nrepl#62

I disagree with this. While this fixes the CIDER issue in the REPL we still have to fix the missing ids and output. Ideally at some point output handing will be fixed in nREPL and the need for the middleware creating the current id mess would disappear.

cider-repl.el
vspinu: Fixed.

vspinu:
Ok. I have removed the "closes". That discussion is indeed only loosely connected to this fix.

bbatsov:
Thanks!

kommen:
Awesome, thanks to both of you!

PR #2529, opened 2018-11-23
Add a cider-repl-set-ns-as command

fhalde:
**Allows you to provide an alias for the ns**

-----------------

- [] You've added tests (if possible) to cover your change(s)

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1145,6 +1145,27 @@ command will prompt for the name of the namespace to switch to."
       (cider-nrepl-request:eval (format "(in-ns '%s)" ns)
                                 (cider-repl-switch-ns-handler connection)))))
 
+(defun cider-repl-set-ns-as (ns as)
+  "Load the namespace in the REPL buffer with an alias.
+If called from a cljc buffer act on both the Clojure and ClojureScript REPL
+if there are more than one REPL present.  If invoked in a REPL buffer the
+command will prompt for the name of the namespace to switch to."
+  (interactive (list (if (or (derived-mode-p 'cider-repl-mode)
+                             (null (cider-ns-form)))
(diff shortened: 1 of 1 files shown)


fhalde:
the CI is failing because of some error in the CI configuration it seems

```
The command "cask_install_or_reset" failed 3 times.
The command "sh travis-ci/install-cask.sh" failed and exited with 1 during .
```

bbatsov:
What's the problem this command would solve?

fhalde:
with `cider-repl-set-ns`, the REPL namespace is switched .. more often i'd want the buffer to be loaded in the REPL without changing the existing namespace of the REPL.

this basically does a `(require '[a.b.c] :as something)` and you can straightaway start using it as `something/somefn` e.t.c

@bbatsov

bbatsov:
I'm not sure that a command for this will help much, though. Seems to me that writing the require manually is more or less the same effort. I'll think about this.

benedekfazekas:
would the user need to type the ns and the alias or does this provide some defaults (sorry only read the comment thread here, not the code)

bbatsov:
Now you have to type it, but if we provide some defaults that might make the feature a bit more useful.

benedekfazekas:
agreed

fhalde:
will work on the default .. thinking of using the first letter of ever word in the namespace
i.e. `hello.world` will become `hw` or `www.users.handlers` will become `wuh`

let me know your thoughts

benedekfazekas:
i would go for the last word rather. but question of taste/style so your approach is fine too

fhalde:
@alexander-yakushev my goal was to require it in the current REPL's namespace & yes the name `cider-repl-set-ns-as` does look misleading if I think about it now

PR #2532, opened 2018-11-28
Don't stall due to dynamic source location discovery on long lines in repl

vspinu:
I am observing long hangs when I am hovering on long lines in repl. While the root cause is somewhere down in emacs redisplay or regexp lookup, the cause on the CIDER side is in the dynamic source location sniffing. 

In this PR I am disallowing source loc discovery on lines longer than 300 characters (should be more than enough)  and improving source locs regexp.


cider-repl.el
```diff
@@ -1200,8 +1197,11 @@ found.  Returned keys are :type, :highlight, :var, :file, :line, where
 :highlight is a cons of positions, :var and :file are strings or nil, :line
 is a number.  See `cider-locref-regexp-alist' for how to specify regexes
 for locref look up."
-  (seq-some (lambda (rl) (cider--locref-at-point-1 rl pos))
-            cider-locref-regexp-alist))
+  (save-excursion
+    (goto-char (or pos (point)))
+    (when (< (- (point-at-eol) (point-at-bol)) 300)
```
bbatsov: Probably we should explain that magic number. Or maybe make it a variable.

bbatsov:
Apart from the failing build and my small remark, the changes look good.

test/cider-repl-tests.el
```diff
@@ -204,10 +204,29 @@ PROPERTY shoudl be a symbol of either 'text, 'ansi-context or
...
-  (it "works with reflection warnings"
+  (it "works with warnings"
     (with-temp-buffer
       (insert "\nReflection warning, cider/nrepl/middleware/slurp.clj:103:16 - reference to field getInputStream can't be resolved.")
       (move-to-column 20)
       (expect (cider-locref-at-point)
               :to-equal
-              '(:type reflection :highlight (22 . 61) :var nil :file "cider/nrepl/middleware/slurp.clj" :line 103)))))
+              '(:type warning :highlight (22 . 61) :var nil :file "cider/nrepl/middleware/slurp.clj" :line 103))))
+  (it "works with warnings"
```
bbatsov: I see you've got two blocks with the same description - an oversight maybe?

bbatsov:
@vspinu ping :-)

vspinu:
Sorry. I got detached for a while and didn't notice your comments. Both issues fixed. I think adding a variable for line length is an overkill. I would think that paths longer than 300 characters are very unlikely.  I have added a comprehensive comment.

PR #2542, opened 2018-12-16
[Fix #2105] Fix no comment syntax defined message when loading buffer after running failing test

### LINKED ISSUE #2105: "No comment syntax is defined.  Use: " in minibuffer when compiling after test/error
(Thanks, as ever, for CIDER. It's the best and you're the best)

## Expected behavior
File compiles fine, no buffers move or close, minibuffer is not focused, workflow uninterrupted.

## Actual behavior
`*cider-error*` buffer is buried, minibuffer is focused and says `No comment syntax is defined.  Use: `, file may or may not be compiled.

1) Run a test (`cider-test-run-test`) that has an error (throws)
    - `*cider-test-report*` buffer appears (as expected)
2) Select the `error:` link at the bottom of the buffer
    - `*cider-error*` buffer appears w/ stack-trace in it (as expected)
3) Select the line in your code that caused the error
    - File with errant code it in appears (as expected)
4) Make a change to the code - attempting to fix the error
5) `C-x C-s` `C-c C-k` to save the file and compile the new (hopefully fixed) code
    - :boom: `cider-error` buffer weirdly is buried, minibuffer is focused and says `No comment syntax is defined.  Use: ` ...

GNU Emacs 24.5.1 on Linux gibson 4.8.0-59-generic #64-Ubuntu SMP Thu Jun 29 19:38:34 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

### CIDER version information

```
;; CIDER 0.15.1 (London), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_111
```

### Lein/Boot version

```
$ lein --version
Leiningen 2.7.2-SNAPSHOT on Java 1.8.0_111 Java HotSpot(TM) 64-Bit Server VM
```

mallt:
This PR fixes the no comment syntax defined message when loading a buffer after running a failing test, by adding a `save-excursion` to make sure the loaded buffer does not lose focus.

-----------------

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -168,7 +168,8 @@ When invoked with a prefix ARG the command doesn't prompt for confirmation."
 (defun cider--quit-error-window ()
   "Buries the `cider-error-buffer' and quits its containing window."
   (when-let* ((error-win (get-buffer-window cider-error-buffer)))
-    (quit-window nil error-win)))
+    (save-excursion
+      (quit-window nil error-win))))
 
 
 ;;; Dealing with compilation (evaluation) errors and warnings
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -1055,9 +1055,10 @@ for the project, it is evaluated in both REPLs."
                    (or (eq cider-save-file-on-load t)
                        (y-or-n-p (format "Save file %s? " buffer-file-name))))
           (save-buffer))
-        (remove-overlays nil nil 'cider-temporary t)
-        (cider--clear-compilation-highlights)
-        (cider--quit-error-window)
+        (save-excursion
```
bbatsov: Isn't it enough to wrap this just `cider--quit-error-window` for this? I also wonder if we shouldn't have the `save-excursion` in it instead.

mallt: I think you're right, thanks! During testing it seemed the removal of the overlays also caused switching of the focus but I can't seem to reproduce that now... I will give it some further testing and update the PR.

bbatsov:
Thanks!

PR #2546, opened 2018-12-21
Fix "save before refresh" functionality and use `files.el` for path comparisons/operations

cichli:
I tried setting `cider-ns-save-files-on-refresh` to both `t` and `'prompt` but found neither had any effect. It turns out that `clojure-project-dir` returns a non-canonical path (e.g. `~/clojure/...`), whereas the predicate passed to `save-some-buffers` was using `(file-truename default-directory)` which _does_ return canonical paths (e.g. `/Users/griffithsm/clojure/...`).

I then made the further observation that `tools.namespace` actually reloads all modified Clojure files on the entire classpath - not just those inside the current project!

To solve this I have modified `cider-ns-refresh--save-project-buffers` to:

* Check modified file-visiting buffers against _all_ classpath directories, not just the current project directory.
* Use proper file/directory path handling functions from `files.el` rather than `string-prefix-p`. This means we get rid of the Windows-specific workaround for case-insensitivity there too.

Similarly, the use of string comparisons in `cider-expected-ns` meant we had to be sure that its input was a canonical path. Again by using functions from `files.el` we can avoid any platform-related headaches or similar.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 ### Changes
+* [#2546](https://github.com/clojure-emacs/cider/pull/2546): New defcustom `cider-ns-save-files-on-refresh-modes` to control for which buffers `cider-ns-refresh` should save before refreshing.
 
 ### Bug fixes
 * Fix values for `cider-preferred-build-tool` variable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 ### New features
 
 ### Changes
-* Fix values for `cider-preferred-build-tool` variable.
-* Fix value and safe property for `cider-allow-jack-in-without-project` variable.
 
 ### Bug fixes
+* Fix values for `cider-preferred-build-tool` variable.
+* Fix value and safe property for `cider-allow-jack-in-without-project` variable.
+* `cider-ns-save-files-on-refresh` will now save any modified buffers visiting files on the classpath, rather than just in the current project.
 
 ## 0.20.0 (2019-01-14)
 
@@ -40,6 +41,7 @@
 
 ### Bug fixes
 
+* `cider-expected-ns` no longer requires an absolute path as its argument, and now internally handles paths canonically and consistently.
 * [#2474](https://github.com/clojure-emacs/cider/issues/2474): Fix incorrect detection of output and out-of-order printing.
 * [#2514](https://github.com/clojure-emacs/cider/issues/2514): Don't auto-jump to warnings when `cider-auto-jump-to-error` is set to 'errors-only.
 * [#2453](https://github.com/clojure-emacs/cider/issues/2453): Make it possible to debug deftype methods by direct insertion of #dbg and #break readers into the deftype methods.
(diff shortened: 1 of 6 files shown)


test/cider-client-tests.el
```diff
@@ -109,10 +109,15 @@
             :to-throw 'user-error)))
 
 (describe "cider-expected-ns"
-  (before-all
+  (before-each
```
cichli: Without this change, the `cider-load-file` test started failing - I'm afraid I'm not familiar enough with buttercup to understand why. It's definitely related to the `(spy-on 'file-directory-p ...)` line below.

cider-ns.el
```diff
@@ -162,18 +162,18 @@ namespace-qualified function of zero arity."
...
       (save-some-buffers
-       (eq cider-save-files-on-cider-ns-refresh t)
+       (not (eq cider-ns-save-files-on-refresh 'prompt))
        (lambda ()
-         (and
-          (derived-mode-p 'clojure-mode)
-          (string-prefix-p project-root
-                           (file-truename default-directory)
-                           (eq system-type 'windows-nt))))))))
+         (and (derived-mode-p 'clojure-mode)
```
cichli: Many programs might read other files on the classpath (e.g. config or static data) when refreshing (either as a side effect of namespaces loading, or in the `after` fn). It would definitely be handy to have any such files be automatically saved too, which makes me want to relax or even remove the `(derived-mode-p 'clojure-mode)` check. What I'm not yet sure about is if/when such behaviour would be undesirable or even dangerous. So I will think about it some more for now.

bbatsov:
Long time, no see! It's good to see you back! :-)

cider-ns.el
```diff
@@ -162,18 +162,18 @@ namespace-qualified function of zero arity."
       (cider--render-stacktrace-causes error))))
 
 (defun cider-ns-refresh--save-project-buffers ()
-  "Ensure modified project buffers are saved before certain operations.
+  "Ensure any relevant modified buffers are saved before refreshing.
```
bbatsov: The name of the function remains confusing.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Bug fixes
 
+* `cider-ns-save-files-on-refresh` will now save any modified buffers visiting files on the classpath, rather than just in the current project.
```
bbatsov: Probably it's worth expanding on this in the manual as well.

bbatsov:
@cichli ping :-)

cider-ns.el
cichli: Renamed to `cider-ns-refresh--save-modified-buffers`.

cichli:
I have added a new defcustom to control for which modes the saving should happen – defaults to `'(clojure-mode)`.

CHANGELOG.md
cichli: Updated the relevant part of the manual with this and the new defcustom.

CHANGELOG.md
```diff
@@ -57,6 +59,7 @@
 
 ### Changes
 
+* [#2546](https://github.com/clojure-emacs/cider/pull/2546): New defcustom `cider-ns-save-files-on-refresh-modes` to control for which buffers `cider-ns-refresh` should save before refreshing.
```
bbatsov: I think you forgot to move this.

cichli: Whoops :-)

bbatsov:
Great work!

PR #2551, opened 2018-12-24
Fix typo in navigating_stacktraces.md

lischenko:
cider-stack-trade-fill-column -> 
    cider-stacktrace-fill-column

-----------------

- [x] The commits are consistent with our [contribution guidelines](../blob/master/CONTRIBUTING.md) - **Broken link**, but the change should not be violating any sensible guidelines
- [ ] You've added tests (if possible) to cover your change(s) - **N/A, doc change**
- [ ] All tests are passing (`make test`) - **N/A, doc change**
- [ ] All code passes the linter (`make lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes  - **N/A, doc change**
- [ ] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality) - **N/A, doc change**
- [ ] You've updated the [user manual](../blog/master/doc) (if adding/changing user-visible functionality) - **N/A, doc change**

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/navigating_stacktraces.md
+++ b/doc/navigating_stacktraces.md
@@ -87,7 +87,7 @@ if they are both present.
 
 Finally, CIDER can wrap error messages when they are displayed in a
 buffer to help improve their readability. CIDER uses
-`cider-stack-trade-fill-column` for this, which can take on three
+`cider-stacktrace-fill-column` for this, which can take on three
 types of values:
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #2557, opened 2019-01-04
Support Kotlin DSL for Gradle Projects.

dpassen:
With [Gradle 5.0](https://docs.gradle.org/5.0/release-notes.html), the Kotlin DSL left beta status. Projects making use of this DSL have build scripts named `build.gradle.kts`

It is necessary to recognize these files as Gradle projects to allow for running REPLS, tests, etc. from Gradle projects making use of this alternative build scripts.


bbatsov:
Thanks!

PR #1983, opened 2017-04-29
Add cider-load-all-files and clojure-files-in-directory functions

fromheten:
Useful when the nrepl host is on another machine, and you wish to "force" the
state of your local file system onto that remote nrepl all at once. 

For this use case `cider-load-all-project-ns` will not do as it runs on the nrepl server machine (unless I've misunderstood). 

-----------------

Before submitting the PR make sure the following things have been done (and denote this
by checking the relevant checkboxes):

- [x] The commits are consistent with our [contribution guidelines][1]
- [N/A] You've added tests (if possible) to cover your change(s)
- [x] All tests are passing (`make test`)
- [x] The new code is not generating bytecode or `M-x checkdoc` warnings
- [x] You've updated the changelog (if adding/changing user-visible functionality)
- [N/A] You've updated the readme (if adding/changing user-visible functionality)
- [N/A] You've updated the refcard (if you made changes to the commands listed there)

PR #2558, opened 2019-01-05
Load all files (clj, cljc, & cljs) on `cider-load-all-files`.

rduplain:
#1983 adds `cider-load-all-files`; the current pull request touches it up to include cljc files in addition to clj files.

There's a larger discussion on ClojureScript repls, especially since loading cljc will eval in all active cljs repls. But cljc is needed in order to load "all files" in a Clojure repl. Before the current pull request, `cider-load-all-files` is incomplete for clj repls. If this proposed change does not have consensus, then let's discuss options to `cider-load-all-files` or variations for `clj`/`clj+cljc`/`cljs+cljc`.

-----------------

- [N/A] You've added tests (if possible) to cover your change(s).

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Changes
 
 * [#2496](https://github.com/clojure-emacs/cider/issues/2496): Replace CIDER's pprint implementation with nREPL 0.5's built-in pprint support.
-* [#2558](https://github.com/clojure-emacs/cider/pull/2558): Load clj & cljc files on `cider-load-all-files`.
+* [#2558](https://github.com/clojure-emacs/cider/pull/2558): Load clj, cljc, & cljs (if cljs repl available) files on `cider-load-all-files` (`C-c C-M-l`). Previously, this only loaded clj files.
 
 ## 0.19.0 (2019-01-01)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changes
 
 * [#2496](https://github.com/clojure-emacs/cider/issues/2496): Replace CIDER's pprint implementation with nREPL 0.5's built-in pprint support.
+* [#2558](https://github.com/clojure-emacs/cider/pull/2558): Load clj & cljc files on `cider-load-all-files`.
 
 ## 0.19.0 (2019-01-01)
 
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -798,7 +798,10 @@ session."
                    (:auto (if (eq cur-type 'multi)
                               '(clj cljs)
                             cur-type))))
-           (repls (cider-repls type 'ensure)))
+           (ensure (cl-case which
+                     (:auto nil)
+                     (t 'ensure)))
+           (repls (cider-repls type ensure)))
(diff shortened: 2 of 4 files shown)


rduplain:
Thanks for the fast reply. My primary concern was that `cider-load-all-files` has a bug, as below, but let's find a complete solution.


#### topic: `cider-load-all-files`

#1983 states the intent behind `cider-load-all-files`:

> Useful when the nrepl host is on another machine, and you wish to "force" the state of your local file system onto that remote nrepl all at once.

In order to accomplish this (on a clj repl), CIDER needs to load cljc files in addition to clj files. Before this pull request, `cider-load-all-files` does not work as documented, which is a bug.

For context in this discussion, `cider-load-all-files` is bound to `C-c C-M-l`.


#### topic: active cljs repls

> There's a larger discussion on ClojureScript repls, especially since loading cljc will eval in all active cljs repls.

The initial commit (338146e1c8ddca1a57cf6ba376c01c8c3f49f2b2) on this pull request is a fix for the original intent of `cider-load-all-files`. With this patch, `cider-load-all-files` will "just work" on clj repls in CIDER. But if CIDER is loading into a cljs repl, then the "all-files" part of the function name is a lie (hence the update to the doc in this pull request, to name clj/cljc explicitly), because `cider-load-all-files` makes no attempt to load cljs files.


#### conclusion: expand `cider-load-all-files` to "do the right thing"

The doc for `cider-load-file` says:

> The heavy lifting is done by `cider-load-buffer`.

Indeed. The `cider-load-buffer` function is iterating all active repls and loading the buffer into each. The necessary APIs are already in place to load clj, cljc, and cljs as part of "all-files" just by updating the regular expression:

```elisp
(mapcar #'cider-load-file
          (directory-files-recursively directory "\\.clj[cs]?$")))
```

The only missing piece in doing this: a user error is reported for every file which does not have a matching repl, cljs files when there is no active cljs repl. That's a simple change in `cider-map-repls`.

New commit incoming ...

rduplain:
I tested 75a683fc4a3ad0b50f2bf63522c1d4bf55000b26 manually with `cider-jack-in` and `cider-jack-in-clj&cljs`.
`cider-load-all-files` does the following:

* loads clj/cljc files in the clj repl created with `cider-jack-in`; it finds cljs files, too, and without any error or repl effect.
* simultaneously loads clj/cljc files in the clj repl and cljc/cljs files in the cljs repl (from `cider-jack-in-clj&cljs`, respectively).

Two very minor and harmless effects:

* When both a clj and a cljs repl are available, the cljs ns names will autocomplete in the clj repl, but of course without any definitions within those cljs-only namespaces. Just the names autocomplete. The cljs ns names do not autocomplete in the clj-only repl case.
* The `*Messages*` buffer lists "Loading path/to/file.cljs..." even when there's only a clj repl.

Ready.

bbatsov:
Thanks for working on this! I really appreciate it! 🙇

> Two very minor and harmless effects:

Probably we should take some time to address those, but I agree that's not a big deal.

rduplain:
Thanks, and I really appreciate that you maintain many great Clojure tools!

PR #2560, opened 2019-01-07
Detect the repl type in cider-sync-request:info based on the current buffer

ak-coram:
`cider-sync-request:info` is using `cider-nrepl-send-sync-request` without a connection argument, which leads to `(cider-current-repl 'any)` being used to determine the current connection. This means that when M-. (`cider-find-var`) is used in a cljs buffer it still sometimes tries to (unsuccessfully) find vars via the clj session of the project. I've added an explicit connection argument to the `cider-nrepl-send-sync-request` call in `cider-sync-request:info` using `(cider-current-repl)` (which determines the repl type based on the current buffer). This resolved the issue for me.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -458,7 +458,8 @@ CONTEXT represents a completion context for compliment."
                                     "ns" ,(cider-current-ns)
                                     "symbol" ,str
                                     "context" ,context)
-                      (cider-nrepl-send-sync-request nil 'abort-on-input))))
+                      (cider-nrepl-send-sync-request (cider-current-repl)
+                                                     'abort-on-input))))
     (nrepl-dict-get dict "completions")))
 
 (defun cider-sync-request:complete-flush-caches ()
@@ -486,7 +487,8 @@ CONTEXT represents a completion context for compliment."
                                      ,@(when symbol `("symbol" ,symbol))
                                      ,@(when class `("class" ,class))
                                      ,@(when member `("member" ,member)))
-                       (cider-nrepl-send-sync-request nil 'abort-on-input))))
+                       (cider-nrepl-send-sync-request (cider-current-repl)
+                                                      'abort-on-input))))
     (if (member "no-eldoc" (nrepl-dict-get eldoc "status"))
         nil
       eldoc)))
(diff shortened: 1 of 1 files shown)


ak-coram:
I've noticed that `cider-sync-request:eldoc` has the same problem because `nil` is passed as the connection. I can add another commit for that as well if you'd like.

ak-coram:
Probably the same also goes for `cider-sync-request:complete` and possibly several other functions, but I'm not really sure about the ones related to namespaces (and the ones for specs).

bbatsov:
I recall we discussed something like this with @kommen recently, but I'm fuzzy on the details and whether we actually decided on anything. I agree that picking some random connection is not the best way to go, and what you suggest seems reasonable. 

> I've noticed that cider-sync-request:eldoc has the same problem because nil is passed as the connection. I can add another commit for that as well if you'd like.

Yeah, that's weird. In such cases nREPL will create some transient session for the duration of the request.

kommen:
This is the PR where we discussed this: https://github.com/clojure-emacs/cider/pull/2467

I think we we came to the solution we have right now because of this comment:

> Looking at the other code in the file I think of two things:
> what happens if you only have a cljs connection? Seems to me the formatting should still work, as it's not really clj or cljs specific

https://github.com/clojure-emacs/cider/pull/2467#discussion_r221466368

ak-coram:
Instead of having to figure this out for each request type, maybe having `cider-current-repl` prefer repls of type `(cider-repl-type-for-buffer)` when called with `'any` would be the easiest solution. It could still pick the most recent one available, but if there are any with the current buffer's repl type then pick the most recent one of those. I'll gladly update this PR if you prefer this solution.

bbatsov:
To summarize again - we need some more specific routing only for ops that have ClojureScript versions. Currently I think those are only "eval" (which should have it done right by now), "complete" and "info".

bbatsov:
Also - rebase on top of `master` to fix the build error.

ak-coram:
I've rebased on top of master and added another commit for the
"complete" and "eldoc" ops (as eldoc isn't working for me reliably
without this change either).

kommen:
Thank you @ak-coram!

PR #2568, opened 2019-01-18
Fix values for `cider-preferred-build-tool` and `cider-allow-jack-in-without-project`

pjstadig:
The values for `cider-preferred-build-tool` were quoted, which resulted in the
variable getting the value `(quote lein)`.  This caused an error in
`cider-maybe-intern` when checking the value for stringness.

`cider-allow-jack-in-without-project` had a similarly quoted value, and a safe
property that checked for strings, even though the value should be a symbol.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### New features
 
 ### Changes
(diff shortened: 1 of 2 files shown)


cichli:
Thanks!

PR #2579, opened 2019-02-11
Update to nREPL 0.6.0 / cider-nrepl 0.21.0

cichli:
Please see the commit messages for more detail, but find a summary of the
changes below:

* The printing config is now controlled by three options: `cider-print-fn`,
  `cider-print-options`, and `cider-print-quota`. Now-redundant printing options
  have been made obsolete.

* There are now functions for constructing partial requests for both the
  pretty-print and non-pretty-print case – we need to ensure we force
  `clojure.core/pr` printing for all the non-`pprint` commands.

* `cider-print-options` can now be any map representation that `map.el` accepts
  (seeing as we support Emacs 25+ now). We use `map-merge` to combine the
  partial requests that specify the printing config. 

  * `map.el` has been updated in ELPA to support plists. Once it has a new
    release we can depend on, we could change `nrepl-send-request` to accept any
    map representation as its `request` argument. There are probably many other
    places where we use ad-hoc representations of maps that we should change to
    use `map.el`.
&nbsp;

* Some general enhancements to the REPL experience, mostly focused on performance:

  * Simplify insertion of images so we don't need to manually fix up any markers
    or the point.

  * New option `cider-repl-init-code` replaces `cider-repl-print-length` and
    `cider-repl-print-level`. This offers a more general way to configurably
    `set!` any dynamic vars or do anything else needed to prep the REPL
    environment.

  * Remove redundant `cider-scroll-on-output` option (and its associated custom
    recentering logic) and document how to restore the old behaviour by setting
    `scroll-conservatively`.

  * I spent some time profiling printing of some large results at the REPL and
    found a couple of easy wins around excessive object allocation (we spend a
    surprising amount of time in GC when the REPL is printing):

    * Reuse the same buffer in `nrepl-bdecode` rather than creating a new one
      each time.

    * Combine calls to `replace-regexp-in-string`.

  * Ensure we always insert output _before_ the REPL prompt.

    
* Also fixed a bug causing `cider-pprint-eval-last-sexp-to-comment` and
  `cider-pprint-eval-defun-to-comment` to not insert anything.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -22,7 +22,6 @@
                               (cider--jack-in . 1)
                               (cider--make-result-overlay . 1)
                               ;; need better solution for indenting cl-flet bindings
-                              (multiline-comment-handler . defun) ;; cl-flet
                               (insert-label . defun)              ;; cl-flet
                               (insert-align-label . defun)        ;; cl-flet
                               (insert-rect . defun)               ;; cl-flet
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,20 @@
 
 ### New features
 
-* The `cider-test-run-*` and `cider-ns-refresh-*` commands are now interruptible with the `cider-interrupt` command.
+* The `cider-test-run-*` and `cider-ns-refresh-*` commands are now interruptible by the `cider-interrupt` command.
+* Many commands now stream printed results back to the client incrementally – meaning it's now possible to, for example, interrupt evaluations while their result is being rendered.
+* New option: `cider-repl-init-code`. This is a list of strings containing Clojure code to evaluate when the REPL starts (with bindings for any `set!`-able vars in place). Replaces `cider-print-length` and `cider-print-level`, which are now obsolete.
+* New option: `cider-print-quota`. This is a hard limit on the number of bytes that will be returned by any printing operation. This defaults to one megabyte and can be set to `nil` if no limit is desired.
 
 ### Changes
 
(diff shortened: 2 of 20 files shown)


bbatsov:
Wow! That’s one truly epic PR! Fantastic work! 🙇 

I don’t see any obviously problems, so you can go ahead with merging this, so we can see if the snapshot will report any problems.

cichli:
@bbatsov I have asked @dpsutton to review and test this also, but I can't add them as reviewer since they're not (yet?) a member of the clojure-emacs org.

bbatsov:
I’ve just invited him to the org.

bbatsov:
I hadn’t noticed the release of Lein 2.9.0 - it’s pretty cool it uses nREPL 0.6 by default! 

In general I agree with your assessment - if we are certain that the core functionality works than it’s fine to merge this, wait until the end of the week for user feedback and cut a stable release then. I’m reasonably certain most users won’t noticed the printing changes, so we should probably do some promotional work at some point - blog posts, etc. I certainly intend to write something about nREPL 0.6 when I’m back home. 

We should probably add some upgrade notes somewhere (e.g. a plain file or something in the manual), as users are often complaining that it’s not clear what are they supposed to do when upgrading CIDER. But we can also not do it, as the current strategy seems to work, despite its flaws. 😃

Macroz:
Thanks for the updates. I bumped into `2.9.0` problems yesterday.

arichiardi:
Will test this out as well from master, thanks!

PR #2583, opened 2019-02-13
Fixes cider--find-var-other-window

agzam:
as docstring for cider-find-var states:

> A prefix of ‘-‘ or a double prefix argument causes the results to be displayed
  in a different window.

that appears to be broken and it's no longer possible to find things in other window.


cichli:
Thanks 😄

PR #2585, opened 2019-02-13
Add a line to docs telling people how to join the slack discussion

rnikander:
This is very simple, since it's my first PR. Just adding a line to the docs.


dpsutton:
@rnikander I asked around on slack and I think the best url to use is http://clojurians.net/  . Could you update the PR to point here instead?

Thanks so much for updating docs. If there's anything else you see with fresh eyes I'm sure countless others would appreciate it.

rnikander:
Okay, just did that.

dpsutton:
Thanks for the sharp eye. Keeping the documentation clear is hard work and PRs are _very_ much appreciated.

PR #2586, opened 2019-02-14
Edit jack in command with prefix

dpsutton:
The current prefix sends `:do-prompt` through the jack in
process. This allows the user to specify: the project directory, the
cljs repl type, the params to the jack in command, and the
endpoint. However, the most common of these are slight edits to the
jack in command to add lein profiles or deps aliases.

This makes a single prefix edit only the jack in command in its
entirety (not the previous command line args only version) rather than
require action on all of the parameters.

For instance, notice here the editing of the jack in command to add the Eastwood profile for cider-nrepl:

screen shot 2019-02-14 at 12 24 03 am

compared to the previous behavior which lacks clarity about what you are allowed to edit and where:

screen shot 2019-02-14 at 12 32 14 am

I like seeing the whole jack in string as I know where the aliases and profiles go. The current way is rather ambiguous to my eye. Also its annoying to have more questions than what I most commonly want to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* Allow editing of jack in command with prefix or when `cider-edit-jack-in-command` is truthy.
 * Stop cursor moving when initialising the CIDER REPL, when `cider-repl-pop-to-buffer-on-connect` is nil. This fixes a bug introduced by [commit e0aca78b](https://github.com/clojure-emacs/cider/commit/e0aca78ba56425e50ea895c5adc7c0331cee0b19).
 
 * [#2593](https://github.com/clojure-emacs/cider/issues/2593): The REPL's initial namespace is now set correctly if configured in another tool (e.g. Leiningen's `:init-ns`).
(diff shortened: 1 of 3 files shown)


bbatsov:
I’m on the fence about the proposed change, but it makes some sense to me. Unfortunately I can’t speculate how exactly other people are using the jack-in commands and how often they have to edit something. Let’s get more feedback here.

If we go forward with this the manual will also need to be updated.

bhauman:
I like the proposed behavior since it requires a prefix argument. I often need to add args to my clojure launch command.

stardiviner:
It make some sence for me too. But not in most case. I will give a star.

stardiviner:
It make some sence for me too. But not in most case. I will give a star.

dpsutton:
pending a successful style check this should be good to go @cichli

cichli:
Thanks again for this! Pushed as 74974b4cbfb20ea9a50484607a890a1aebf1ab31 and 88328955435a4d5d8961a56fc96b2a6a40fd113c. There was another changelog conflict which I resolved.

bbatsov:
Btw, I don’t think that adding the defcustom is a good idea, as it seems like a config that almost no one will probably need. Not to mention it’s confusing what kind of edit it allows you exactly.

bbatsov:
And on a stylistic level - for consistency it might be better to have something like `prompt-command` and `prompt-all`. The current names confuse me, and I assume they’ll confuse others as well.

PR #2590, opened 2019-02-16
Scroll buffer after using cider-insert-X-in-repl

dpsutton:
As it stood, buffer would not scroll, even if you included a
`(goto-char (point-max))` call at the end. Using solution from
https://stackoverflow.com/questions/13530025/emacs-scroll-automatically-when-inserting-text

Have to watch out if the buffer is visible or not. If it is, we can
scroll it with `with-selected-window`. Otherwise, the best we can do
is the current behavior. Note `get-buffer-window` returns nil if
something is not visible which is why we need to branch on it.

## Currently

Using the `cider-insert-[last-sexp|defun|region]-inrepl` commands (C-c M-j [e,d,r]) were very convenient but would not scroll the buffer, limiting their usability.

before

Notice that the repl buffer is not scrolled. Simply adding a `(goto-char (point-max))` in `cider-insert-in-repl` was not sufficient due. (see stack overflow post from commit message above).

## After

By using `with-selected-window` we can manipulate the view of that buffer. So now we end up with this much more sensical view when using these commands:

after

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -223,6 +223,17 @@ With a prefix argument, prompt for function to run instead of -main."
   :group 'cider
   :package-version '(cider . "0.18.0"))
 
+(defcustom cider-switch-to-repl-on-insert-p t
+  "Whether to switch to the repl when inserting a form into the repl."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.21.0"))
+
+(define-obsolete-variable-alias
+  'cider-switch-to-repl-after-insert-p
+  'cider-switch-to-repl-on-insert-p
+  "0.21.0")
+
 (defcustom cider-invert-insert-eval-p nil
   "Whether to invert the behavior of evaling.
 Default behavior when inserting is to NOT eval the form and only eval with
@@ -238,7 +249,7 @@ If EVAL is non-nil the form will also be evaluated.  Use
 `cider-invert-insert-eval-p' to invert this behavior."
   (while (string-match "\\`[ \t\n\r]+\\|[ \t\n\r]+\\'" form)
     (setq form (replace-match "" t t form)))
-  (when cider-switch-to-repl-after-insert-p
+  (when cider-switch-to-repl-on-insert-p
     (cider-switch-to-repl-buffer))
   (let ((repl (cider-current-repl)))
     (with-selected-window (or (get-buffer-window repl)
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -121,7 +121,7 @@ Clojure buffer."
 (defun cider-switch-to-last-clojure-buffer ()
   "Switch to the last Clojure buffer.
 The default keybinding for this command is
-the same as `cider-switch-to-repl-buffer',
+the same as variable `cider-switch-to-repl-buffer',
 so that it is very convenient to jump between a
 Clojure buffer and the REPL buffer."
   (interactive)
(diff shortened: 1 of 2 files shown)


cichli:
We could switch to the REPL _before_ inserting to make sure the buffer is visible, but `cider-switch-to-repl-after-insert-p` would have to be renamed.

arichiardi:
I am fine with the rename as seeing the output is more important ;)))

Thanks for opening this btw!

dpsutton:
I've marked obsolete and created a new defcustom

```lisp
(defcustom cider-switch-to-repl-on-insert-p t
  "Whether to switch to the repl when inserting a form into the repl."
  :type 'boolean
  :group 'cider
  :package-version '(cider . "0.21.0"))

(define-obsolete-variable-alias
  'cider-switch-to-repl-after-insert-p
  'cider-switch-to-repl-on-insert-p
  "0.21.0")
```

I figure we can just say "on-insert" rather than before or after so hopefully we don't go through this again if it makes sense to focus in the buffer before/after insertion in the future.

Let me know if I didn't obsolete it correctly.

arichiardi:
Oh it seems the problem could just be that it is not rebased on `master`.

dpsutton:
@arichiardi do you have the correct settings? By default, CIDER doesn't eval the inserted forms automatically. use

```clojure
(setq cider-invert-insert-eval-p t)

**Comment by arichiardi** [comment]
Ah lol maybe not :)

**Comment by dpsutton** [comment]
also added fontlocking

with-fontlocking

File: cider-mode.el [hunk 1]
```diff
@@ -223,6 +223,17 @@ With a prefix argument, prompt for function to run instead of -main."
   :group 'cider
   :package-version '(cider . "0.18.0"))
 
+(defcustom cider-switch-to-repl-on-insert-p t
```
bbatsov: Defcustoms should not have predicate names IMO. I’ve never seen such naming for them.

cider-mode.el
```diff
@@ -223,6 +223,17 @@ With a prefix argument, prompt for function to run instead of -main."
...
+  "Whether to switch to the repl when inserting a form into the repl."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.21.0"))
+
+(define-obsolete-variable-alias
```
bbatsov: I don’t see the var that was made obsolete anywhere, so I guess that’s not needed.

dpsutton: the var is right above the new one.
```lisp
(defcustom cider-switch-to-repl-after-insert-p t
  "Whether to switch to the repl after inserting a form into the repl."
  :type 'boolean
  :group 'cider
  :package-version '(cider . "0.18.0"))

(defcustom cider-switch-to-repl-on-insert-p t
  "Whether to switch to the repl when inserting a form into the repl."
  :type 'boolean
  :group 'cider
  :package-version '(cider . "0.21.0"))

(define-obsolete-variable-alias
  'cider-switch-to-repl-after-insert-p
  'cider-switch-to-repl-on-insert-p
  "0.21.0")
```

The original had a `-p` so I mimic'ed it. I'll remove that from the new one.

arichiardi:
This looks very good to me :smile: Eager to have it in :smile: 

Thanks a lot @dpsutton!

cichli:
Thanks for this! Sorry it's taken a while for me to get around to reviewing it.

arichiardi:
Awesome thank you!

Issue #2577, opened 2019-02-10
Can't get stacktrace for exception in test report


kommen:
## Expected behavior
Clicking an error in a test report should show the stacktrace for that error.

## Actual behavior

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  nrepl-send-request(("op" "test-stacktrace" "ns" "com.nextjournal.graph-test" "var" "normalize-test" "index" 1 "pprint-fn" "cider.nrepl.pprint/pprint" "print-options" (dict "length" 50 "level" 50)) (closure ((causes) (index . 1) (var . "normalize-test") (ns . "com.nextjournal.graph-test") t) (response) (let ((class (nrepl-dict-get response "class")) (status (nrepl-dict-get response "status"))) (cond (class (setq causes (cons response causes))) (status (if causes (progn (cider-stacktrace-render ... ...))))))) nil)
  cider-nrepl-send-request(("op" "test-stacktrace" "ns" "com.nextjournal.graph-test" "var" "normalize-test" "index" 1 "pprint-fn" "cider.nrepl.pprint/pprint" "print-options" (dict "length" 50 "level" 50)) (closure ((causes) (index . 1) (var . "normalize-test") (ns . "com.nextjournal.graph-test") t) (response) (let ((class (nrepl-dict-get response "class")) (status (nrepl-dict-get response "status"))) (cond (class (setq causes (cons response causes))) (status (if causes (progn (cider-stacktrace-render ... ...))))))) nil)
  (let (causes) (cider-nrepl-send-request (nconc (list "op" "test-stacktrace" "ns" ns "var" var "index" index) (if (cider--pprint-fn) (progn (list "pprint-fn" (cider--pprint-fn)))) (if cider-stacktrace-print-options (progn (list "print-options" cider-stacktrace-print-options)))) (function (lambda (response) (let ((class (nrepl-dict-get response "class")) (status (nrepl-dict-get response "status"))) (cond (class (setq causes ...)) (status (if causes ...)))))) (cider-current-repl)))
  cider-test-stacktrace-for("com.nextjournal.graph-test" "normalize-test" 1)
  cider-test-stacktrace()
  (lambda (_button) (cider-test-stacktrace))(#<marker (moves after insertion) at 255 in *cider-test-report*>)
  button-activate(#<marker (moves after insertion) at 255 in *cider-test-report*> t)
  push-button(255 t)
[...]

PR #2591, opened 2019-02-17
Ensure connections for better error messages and avoid exceptions

kommen:
Follow up to the discussion with @vspinu in #2577:

If for any reason `cider-current-repl` doesn't find a repl, this ensures a proper user friendly error is displayed instead of a type error.

In the #2577 case, this would report an `No linked CIDER sessions` instead of `(wrong-type-argument stringp nil)`

I tried to adopt all code locations where `cider-current-repl` was used in an `(or connection (cider-current-repl))` pattern, where the result was then passed to a function which did not have any fallback if the connection param was nil and thus caused the `wrong-type-argument` error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 * Fixed a bug causing REPL output to be inserted after the prompt.
 * Fixed a bug causing `cider-pprint-eval-last-sexp-to-comment` and `cider-pprint-eval-defun-to-comment` to not insert anything.
 * `cider-find-var` now correctly uses a new window when passed a prefix of `-` or a double prefix argument.
+* [#2577](https://github.com/clojure-emacs/cider/issues/2577): Ensure user friendly error messages if a repl connection is expected but none was found in certain situations. 
 
 ## 0.20.0 (2019-01-14)
 
--- a/cider-client.el
+++ b/cider-client.el
@@ -375,7 +375,7 @@ The library is a string of the format \"group-id/artifact-id\"."
   "Interrupt any pending evaluations."
   (interactive)
   ;; FIXME: does this work correctly in cljc files?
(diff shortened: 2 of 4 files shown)


dpsutton:
Are there any callsites that are fine with nil connection returned?

kommen:
I checked every call site I touched and as far as I understood each one would cause an error with a nil connection.

dpsutton:
I'm wondering if `ensure` should be a parameter then if every call site invokes it with that option. Feel free to ignore if this is bike shedding though.

kommen:
> if every call site invokes it with that option
Sorry, maybe my comment about "every call site" wasn't clear: I only checked the call sites with the `(or connection (cider-current-repl))` pattern. There are still a lot of `cider-current-repl` calls which don't get `ensure` passed.

vspinu:
Shouldn't at least all `-eval` and -`sync` calls be 'ensured?

bbatsov:
I’m fine with the proposed changes, but ideally we need to be consistent about our error handling, so probably this requires a deeper pass of the code base indeed. (Same for things like commands requiring connections in general, commands that don’t work with a ClojureScript connection, and so on).

kommen:
> but ideally we need to be consistent about our error handling

I agree, that's why I tried to find as many call sites as possible which could exhibit the issue. But as @vspinu mentioned, there are still some more or less obvious calls which can be `'ensure`ed. I'll see to find some time to go over the calls again, can't promise when I can make it happen though.

kommen:
@cichli done. I found a few more call sites where `'ensure` is appropriate, pushed them in 2033a04 for easier review.

cichli:
Fixed in 690b1343cf15cb0e8a0c58c6af0645341d354f8d. There was another changelog conflict which I resolved. Thanks for this!

PR #2594, opened 2019-02-21
support figwheel-connected sibling cljs repl

bendlas:
## Use Case
- Start figwheel via `lein figwheel` and have it configured to start an nrepl
- Connect to nrepl via `C-c M-c`
- Start a sibling cljs repl of `figwheel-running` type, to get into a figwheel-connected browser

Right now, this has to be done with a custom init form, which is annoying

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -763,6 +763,8 @@ The supplied string will be wrapped in a do form if needed."
     (figwheel "(do (require 'figwheel-sidecar.repl-api) (figwheel-sidecar.repl-api/start-figwheel!) (figwheel-sidecar.repl-api/cljs-repl))"
               cider-check-figwheel-requirements)
     (figwheel-main cider-figwheel-main-init-form cider-check-figwheel-main-requirements)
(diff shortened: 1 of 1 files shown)


bbatsov:
So, that’s the same as what we have right now, but with the difference it doesn’t start figwheel, right? Seems to me the naming is a bit confusing.

bendlas:
Historically, the `-sidecar` moniker has been associated with starting figwheel in your process and dropping down to it via `cljs-repl`, for me.

I agree that there may be better names for it. What about `figwheel-piggieback` or `figwheel-running`?

bbatsov:
Or maybe `figwheel-connect`? I’ll have to think about this a bit. @cichli Any thoughts?

bendlas:
`figwheel-connected` implies the precondition of an active nrepl connection. I think I'll go with suggesting that.

cichli:
Thanks for this! 🙌

PR #2598, opened 2019-02-24
Stop cursor moving when initialising the CIDER REPL

simon-katz:
SORRY — I'm trying to work out how to put this into DRAFT status. I haven't been through the checklist.

Fixes this bug: When cider-repl-pop-to-buffer-on-connect is nil the cursor jumps
when initialising the CIDER REPL. The cursor jumps in whatever buffer is
current.

This commit fixes commit e0aca78b, whose commit message says:

    Make sure REPL window point is at point-max at startup
    Needed when cider-repl-pop-to-buffer-on-connect is 'display-only

This commit applies the new functionality only when
cider-repl-pop-to-buffer-on-connect is display-only.

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -304,7 +304,8 @@ client process connection."
   (with-current-buffer buffer
     (cider-repl--insert-banner)
     (cider-repl--insert-prompt cider-buffer-ns)
-    (set-window-point (get-buffer-window) (point-max)))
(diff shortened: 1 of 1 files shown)


simon-katz:
Closing this for now. I will do it again, properly.

PR #2599, opened 2019-02-24
Stop cursor moving when initialising the CIDER REPL

simon-katz:
Fixes this bug: When cider-repl-pop-to-buffer-on-connect is nil the cursor jumps
when initialising the CIDER REPL. The cursor jumps in whatever buffer is
current.

This commit fixes [commit e0aca78b](https://github.com/clojure-emacs/cider/commit/e0aca78ba56425e50ea895c5adc7c0331cee0b19), whose commit message says:

    Make sure REPL window point is at point-max at startup
    Needed when cider-repl-pop-to-buffer-on-connect is 'display-only

This commit applies the new functionality only when
cider-repl-pop-to-buffer-on-connect is display-only.

The following are not done, but I think they're not applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bug fixes
 
+* Stop cursor moving when initialising the CIDER REPL, when `cider-repl-pop-to-buffer-on-connect` is nil. This fixes a bug introduced by [commit e0aca78b](https://github.com/clojure-emacs/cider/commit/e0aca78ba56425e50ea895c5adc7c0331cee0b19).
+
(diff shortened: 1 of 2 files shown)


manuel-uberti:
Thanks for this, it's a rather annoying thing. :+1:

cichli:
Thanks for the PR 😄!

What do you think about using `get-buffer-window` rather than checking the value of the option? Maybe we shouldn't ever assume that the selected window is the correct one, even if the option is `t`.

```elisp
    (when-let ((window (get-buffer-window)))
      (set-window-point window (point-max)))
```

simon-katz:
@cichli 
> What do you think about using `get-buffer-window` rather than checking the value of the option? Maybe we shouldn't ever assume that the selected window is the correct one.

I'm not sure — I don't understand the initialisation code very well.

If someone tells me it's the right thing to do, I'm happy to make the change.

cichli:
Thanks for this! 🙌

PR #2603, opened 2019-03-05
Switch to CircleCI

danielcompton:
Opening early for feedback. One thing that is tricky is that CircleCI doesn't have the concept of allowable failures like Travis does. At the moment Emacs master is failing on `cider-jack-in-clj&cljs` (timing out?), but 25 and 26 are running successfully: https://circleci.com/workflow-run/dff69ef0-ec14-4c24-8f19-457e4168af37.

Still needed:

- [x] Update hacking_on_cider doc to show how to do the same thing with CircleCI
- [x] Remove/convert the Travis Docker scripts to CircleCI scripts

-----------------

***

See also https://github.com/clojure-emacs/clomacs/pull/16, https://github.com/clojure-emacs/clojure-mode/pull/514, and https://github.com/clojure-emacs/inf-clojure/pull/159.


.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
+version: 2.1
+
+# Default actions to perform on each Emacs version
+default: &default-steps
+  steps:
+    - checkout
```
bbatsov: Do we need Lein at all?

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+    - run: apt-get update && apt-get install make leiningen=2.8.1-6 -y
+    - run: make elpa
+    - run: emacs --version
+    - run: make test
```
bbatsov: Maybe we should have separate steps for the testing and linting, so that we would get nicer reports at the end?

danielcompton: The issue with this is that the majority of CI time is spent getting the image booted and into the right state to run the tests. The lint and test [phases](https://circleci.com/gh/danielcompton/cider/8) only take 6s and 3s. Given this is a free service, I wanted to try and minimise resource usage where practicable. If we wanted better errors from CircleCI we could look at producing JUnit XML reports which they can render.

.circleci/config.yml
danielcompton: `cider-jack-in-clj&cljs` requires `lein` to be around: See the test failure at https://circleci.com/gh/danielcompton/cider/11 when removing Leiningen from the build.

danielcompton:
I've migrated the docs for running locally, I think this is now ready to be reviewed.

.circleci/config.yml
bbatsov: I'll take a look at the test. We don't do any real jack-ins anywhere in the tests, so probably something just needs to be stubbed.

.circleci/config.yml
bbatsov: Fair enough. My only issue is that the linting produces a lot of (hard to read) output and this makes it hard to find the tests at least for me.

bbatsov:
Well, perhaps we can just remove this build or fix the failure there? I don't see any other options. :-)

bbatsov:
Actually checking the report you've linked it seems to me that output is not bad even now. I'm still a bit confused when navigating Circle's UI, as I was pretty used to Travis's. 

I'll see what I can do about that failing test. Weirdly I don't see anything in the report and I also don't see anything special in the test.

bbatsov:
Thanks for tackling this, @danielcompton! Much appreciated!

PR #2607, opened 2019-03-19
Use markers for evaluating to comments

yuhan0:
Fixes a slightly annoying bug with comments appearing in unintended locations
when calling (expensive) async `cider-eval-***-to-comment` commands on a form,
then making subsequent edits while the evaluation is pending.

--------------------------

Thanks!

manual][1] extremely useful.*


bbatsov:
The change seems reasonable to me. Please, document it in the changelog.

yuhan0:
Alright, just edited the changelog. 
I skipped the other 2 checkboxes because it's not a user-facing change to include in the manual, and would be quite difficult to write tests for this case.

bbatsov:
Thanks!

PR #2609, opened 2019-03-21
Add current ns to `spec-list` and `spec-form` requests

tatut:
Add current ns to `spec-list` and `spec-form` requests so the server can use that information to resolve namespace aliases when providing completions and spec information.

See: https://github.com/clojure-emacs/orchard/pull/42

-----------------


bbatsov:
Thanks!

PR #2615, opened 2019-03-26
[debugger] Fix inspector behavior when multiple REPLs are connected

alexander-yakushev:
Identical to #2454, this ensures that the inspector in the debugger mode works with the correct REPL and doesn't jump around.


bbatsov:
Thanks!

PR #2620, opened 2019-03-31
Remove C-g binding in test map to make C-g work

wasamasa:
To quote (info "(elisp) Key Binding Conventions"):

    • Similarly, don’t bind a key sequence ending in ‘C-g’, since that is
      commonly used to cancel a key sequence.

-----------------

No tests added because, well, this was a convenience shortcut. I'd like to update the user manual, but it doesn't even cover the `C-c ,` map and instead documents the obsolete `C-c C-t` map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * `cider-print-options` is now supported by the `pr` option for `cider-print-fn`. The options will now be also used by interactive eval commands that do not use pretty-printing.
 * `spec-list` and `spec-form` requests send the current namespace for alias resolution.
+* `C-c , C-g` and `C-c C-t C-g` cancel the key chord instead of rerunning the last test. The respective command has been moved to `C-c , C-a`, `C-c , a`, `C-c C-t C-a` and `C-c C-t a`.
 
 ### Bug fixes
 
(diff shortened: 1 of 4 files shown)


bbatsov:
You bring up a good point, but we can't just remove this. We also have to decide what to use instead.

Malabarba:
Seems this command is already bound to `C-c , g` (besides `C-c , C-g`), which makes more and is easier to type. So I think this is a good change as-is.

wasamasa:
Correct, there's `C-c , g`. I've assumed that this should settle things.

Another thing I've noticed is that the `C-c C-t` map is still a thing and does the same mistake. Would you like me to fix the key binding and documentation for it as well?

Malabarba:
Yes, that makes sense.

wasamasa:
Alright, both are changed now.

bbatsov:
> Seems this command is already bound to C-c , g (besides C-c , C-g), which makes more sense and is easier to type. So I think this is a good change as-is.

But this breaks the uniformity of the commands. It's super weird to have some commands that don't have a `C-something` counterpart. I'd rather change the `g` with something else than make a special case out of it.

wasamasa:
Well, personally I believe that to be a far more invasive change (to get rid of g/C-g in favor of <letter>/C-<letter>). If you really prefer it, please name what you consider better.

bbatsov:
Consistency is important. We can't have a single keybinding that doesn't fit our entire pattern of keybindings. I guess we can use `C-e` or `C-a` instead. Generally everything free on the left side of the keyboard works for me.

wasamasa:
OK, I've switched to `C-a` and adjusted the changelog entry appropriately.

bbatsov:
Thanks!

PR #2623, opened 2019-04-09
[Fix #2495] Add option to require ns before setting it in the repl

### LINKED ISSUE #2495: Do a require before cider-repl-set-ns
Hi folks!

I have been bitten by some problem that is not related directly but triggered some thoughts on `cider-repl-set-ns`. At the moment it sends a `in-ns`, which means that if the code has not required the ns before, in Clojure it will create an empty namespace - which is never what you want to do - and in ClojureScript, depending on the REPL type, might generate weird error messages like `[:did-not-find...your namespace...]` in shadow-cljs.

Would it be too breaking for folks to send something like `(require 'namespace) (in-ns 'namespace)` instead?

Or have another command for doing that (that I would use all the time :smile:) ?

Thanks!

paulojean:
Allow CIDER to require the namespace when calling `cider-repl-set-ns`.

-----------------

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -143,6 +143,12 @@ will be used."
   :group 'cider-repl
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-repl-require-ns-on-set nil
+  "Controls wether to require the ns before setting it in the repl."
+  :type 'boolean
+  :group 'cider-repl
+  :package-version '(cider . "0.21.0"))
+
 (defcustom cider-repl-result-prefix ""
   "The prefix displayed in the REPL before a result value.
 By default there's no prefix, but you can specify something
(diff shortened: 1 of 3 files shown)


cider-repl.el
```diff
@@ -1083,7 +1089,9 @@ command will prompt for the name of the namespace to switch to."
     (user-error "No namespace selected"))
   (cider-map-repls :auto
     (lambda (connection)
-      (cider-nrepl-request:eval (format "(in-ns '%s)" ns)
+      (cider-nrepl-request:eval (if cider-repl-require-ns-on-set
+                                    (format "(require '%s) (in-ns '%s)" ns ns)
```
bbatsov: Probably you should wrap those two forms in a `do`.

PR #2627, opened 2019-04-21
Remove 'ensure in initial REPL when switching to other

stacksideflow:
Reason of this PR (very common if you are working with both `cljs` and `clj`):

1. I'm in `cljs` file, just connected to my `clj` REPL
2. Now I'd like to go to my `clj` REPL buffer, maybe convert it to `cljs` REPL (e.g. with `cljs-repl`) - which is common
3. Oops, you can't do that - use `SPC b b` to find either your `clj` REPL or `clj` file

-> Now you can cycle freely

-----------------

- [x] All tests are passing (make test)
- [x] All code passes the linter (make lint) which is based on elisp-lint and includes byte-compilation, checkdoc, check-declare, packaging metadata, indentation, and trailing whitespace checks.
- [x] You've updated the user manual (if adding/changing user-visible functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 * `cider-print-options` is now supported by the `pr` option for `cider-print-fn`. The options will now be also used by interactive eval commands that do not use pretty-printing.
 * `spec-list` and `spec-form` requests send the current namespace for alias resolution.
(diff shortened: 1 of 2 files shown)


bbatsov:
I'll have to think about this, as it changes a bit the semantics of the function.

PR #2636, opened 2019-05-19
[DOCS] Fix autocompletion for `map-indexed`

pfeodrippe:
Fix autocompletion for `map-indexed` at docs, I'm testing with `compliment` and it accepts, for this case, only `mi`, not `mp`.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Thanks!

PR #2643, opened 2019-06-03
Drop cider.tasks/nrepl-server usage

arichiardi:
Now that boot supports the new nRepl, we do need a custom task anymore so we
are going to use the built-in one.

-----------------


bbatsov:
Thanks!

PR #2646, opened 2019-06-08
Add minibuffer completion for shadow-cljs builds

yuhan0:
This PR adds `completing-read` minibuffer completions for shadow-cljs builds, by looking for and parsing a `shadow-cljs.edn` file in the project root. Default build options of `node-repl` and `browser-repl` are also included.

Inspired by the equivalent functionality in Calva: https://github.com/BetterThanTomorrow/calva/blob/2571d802c0be7c82230db8583caf1443a3c93244/calva/shadow.ts

(Is it necessary to add tests or update the docs for this?)

-----------------

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -709,13 +709,27 @@ Generally you should not disable this unless you run into some faulty check."
   :safe (lambda (s) (or (null s) (stringp s)))
   :package-version '(cider . "0.18.0"))
 
+(defun cider--shadow-parse-builds ()
+  "Extract build names from the shadow-cljs.edn config file in the project root.
+The default options of `browser-repl' and `node-repl' are also included."
+  (let* ((shadow-edn
+          (concat (clojure-project-dir (cider-current-dir)) "shadow-cljs.edn"))
+         (edn-hash (when (file-exists-p shadow-edn)
+                     (with-temp-buffer
+                       (insert-file-contents shadow-edn)
+                       (car (parseedn-read)))))
+         (builds-hash (when edn-hash (gethash :builds edn-hash)))
+         (builds (when builds-hash (hash-table-keys builds-hash))))
+    (append builds '(browser-repl node-repl))))
+
 (defun cider-shadow-select-cljs-init-form ()
   "Generate the init form for a shadow-cljs select-only REPL.
 We have to prompt the user to select a build, that's why this is a command,
 not just a string."
   (let ((form "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/nrepl-select %s))")
         (options (or cider-shadow-default-options
-                     (read-from-minibuffer "Select shadow-cljs build (e.g. dev): "))))
+                     (completing-read "Select shadow-cljs build: "
+                                      (cider--shadow-parse-builds)))))
     (format form (cider-normalize-cljs-init-options options))))
 
 (defun cider-shadow-cljs-init-form ()
@@ -724,7 +738,8 @@ We have to prompt the user to select a build, that's why
 this is a command, not just a string."
   (let* ((form "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch %s) (shadow/nrepl-select %s))")
          (options (or cider-shadow-default-options
-                      (read-from-minibuffer "Select shadow-cljs build (e.g. dev): ")))
+                      (completing-read "Select shadow-cljs build: "
+                                       (cider--shadow-parse-builds))))
          (build (cider-normalize-cljs-init-options options)))
     (format form build build)))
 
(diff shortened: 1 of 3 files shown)


cider.el
```diff
@@ -709,13 +709,27 @@ Generally you should not disable this unless you run into some faulty check."
   :safe (lambda (s) (or (null s) (stringp s)))
   :package-version '(cider . "0.18.0"))
 
+(defun cider--shadow-parse-builds ()
+  "Extract build names from the shadow-cljs.edn config file in the project root.
+The default options of `browser-repl' and `node-repl' are also included."
+  (let* ((shadow-edn
+          (concat (clojure-project-dir (cider-current-dir)) "shadow-cljs.edn"))
```
bbatsov: What will happen if run this outside a project dir? Btw, passing some dir here is not necessary.

cider.el
```diff
@@ -709,13 +709,27 @@ Generally you should not disable this unless you run into some faulty check."
   :safe (lambda (s) (or (null s) (stringp s)))
   :package-version '(cider . "0.18.0"))
 
+(defun cider--shadow-parse-builds ()
```
bbatsov: Ideally the reading of the file and its parsing, should be split, so you can add a couple of tests for the parsing function.

bbatsov:
> (Is it necessary to add tests or update the docs for this?)

Tests for sure. I'm not sure if there's any docs that need to be updated.

cider.el
yuhan0: Thanks for the review! Would you prefer that the split take place before or after the edn is parsed?

ie. (where `get-builds` is the new function)
```clojure
(let ((edn-file ...)
      (hash (with-temp-buffer
              (insert-file-contents edn-file)
              (car (parseedn-read)))))
  (get-builds hash))  ; <--- pure function from hash -> list of builds

;; vs


(let ((edn-file ...))
  (with-temp-buffer
    (insert-file-contents edn-file)
    (get-builds))) ; <------ expects to be in the edn buffer
```

cider.el
yuhan0: There are nil checks at each step, so if the edn file doesn't exist / isn't parsed properly / doesn't have the required structure, it simply returns the default options `(browser-repl node-repl)` for completing-read, which I thought would be more reasonable than throwing an error or returning the empty list.

cider.el
bbatsov: I'd make `get-builds` a pure function. Sorry for the slow response!

bbatsov:
@yuhan0 ping :-)

yuhan0:
Ah, I must have missed your previous comment - just pushed a couple of commits, I hope the tests are sufficient (not sure if the parsing function should be wrapped in `ignore-errors` in case the edn file is ill-formed). Let me know if you'd llke the commits to be squashed or any other changes :)

test/cider-tests.el
```diff
@@ -251,6 +251,23 @@
     (it "prepends colon to plain names"
       (expect (cider-normalize-cljs-init-options "dev") :to-equal ":dev"))))
 
+(describe "cider--shadow-parse-builds"
+  (before-each
+    (fset 'cider--with-edn (lambda (str)
```
bbatsov: I think it's going to be simpler to just have this a helper that you call on whatever you pass to `cider--shadow-parse-builds`. This makes it easier to understand which is the actual function under the test.

bbatsov:
> (not sure if the parsing function should be wrapped in ignore-errors in case the edn file is ill-formed)

Will shadow start if the edn file is ill-formed? Maybe it'd be best to just display an user-friendly error that something's wrong with the file in this case.

test/cider-tests.el
yuhan0: Okay, I made it a top-level defun - haven't used Buttercup before and somehow thought the `before-each .. fset` elsewhere in the file made a test-local wrapper

yuhan0:
shadow-cljs won't start if the file is syntactically ill-formed, so eg. jacking in throws the error:

```
error in process sentinel: Could not start nREPL server: ===== ERROR =================
Failed to read config file: /Users/yuhan/scratch/shadow/shadow-cljs.edn
[line 31, col 1] Unexpected EOF while reading item 6 of map, starting at line 2 and column 1.
=============================
```

before these build-selection functions have a chance to be called.

test/cider-tests.el
bbatsov: @yuhan0 My point was a different one - normally in the tests you call directly the functions under test. That's why I was thinking you can do something like `(cider--shadow-parse-builds (parse-edn-str ...))`. Or just set the parsed edn as vars in before blocks. Something like this. Pretty common setup for an RSpec-like test framework (like buttercup).

yuhan0: Got it, refactored to use `parseedn-read-str` inline and force-pushed.

bbatsov:
Nicely done! Thanks! 🙇

Issue #2654, opened 2019-06-22
sesman docs vague


jimrthy:
**Is your feature request related to a problem? Please describe.**

I'm working on one project. I wanted to update a library dependency using the same REPL.

I spent about half an hour reading the "Managing Connections" doc over and over, and then the sesman source code trying to figure it out.

I kept trying to run either `sesman-link-with-project` or `sesman-link-with-directory` from the REPL buffer. I kept getting errors that it couldn't find the project/directory I was trying to specify.

I finally figured it out by switching to the specific buffer I wanted to update and just running `sesman-link-with-buffer` because I figured that working with one buffer at a time was better than nothing.

After I saw how that worked, it was pretty obvious.

**Describe the solution you'd like**

This seems like a documentation tweak. I'll submit a PR to at least start the conversation about how to make this more clear.

**Describe alternatives you've considered**

It would be nice to also update the sesman docs so other dependent projects get the same benefit.

bbatsov:
Yeah, I agree there's a lot of room for improvement there. sesman-related questions are probably the number 1 support questions we're getting, so we should improve the docs. //cc @vspinu

vspinu:
Looks like the source of your confusion was that you tried to link from the repl buffer and not from the project/dir/buffer. As you said, it's obvious once you know it: go to the project/dir/buffer and link with the session.  One always associates a session to the the (current) context (project/dir/buffer etc), so you have to `C-c C-s l` and friends from the current context. 

From the browser (`C-c C-s C-w`) you can also do the other way around: link session at point with a project/dir/buffer.

I think a step by step working example would definitely help here. Sesman docs are raw and probably outdated. CIDER's are pretty ok. I plan for another bout of sesman related work in July. Will revamp the docs then.

PR #2655, opened 2019-06-22
Add a note about how to link sessions

jimrthy:
Added a clarification about where to run the sesman-link commands.

(That clarification may very well have lots of nuances and variations I don't understand. I'm really just trying to document a method that worked for me).

This is a proposed fix for https://github.com/clojure-emacs/cider/issues/2654

### DIFF
--- a/doc/modules/ROOT/pages/usage/managing_connections.adoc
+++ b/doc/modules/ROOT/pages/usage/managing_connections.adoc
@@ -43,6 +43,9 @@ To quit or restart individual connections use cider commands
 
 Sessions can be linked to contexts (projects, directories and buffers)
 
+Run these commands from within a buffer that's inside the buffer/directory/project
+you wish to link.
+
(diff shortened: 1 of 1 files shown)


doc/modules/ROOT/pages/usage/managing_connections.adoc
```diff
@@ -43,6 +43,9 @@ To quit or restart individual connections use cider commands
 
 Sessions can be linked to contexts (projects, directories and buffers)
 
+Run these commands from within a buffer that's inside the buffer/directory/project
```
bbatsov: Buffer within a buffer sounds a bit weird. :) 

I'd put this after the keybindings with a "NOTE:" prefix, so it'd be rendered as an admonition in the manual.

bbatsov:
I've made a few improvements to this page, which hopefully solve the problem for you.

Issue #2656, opened 2019-06-23
Change command line quoting when calling clojure cli


cark:
**Is your feature request related to a problem? Please describe.**

Cider does not start a deps.edn project under windows (using C-c M-j). 

This is due to the lack of a clojure command, which in turn is due to the clojure team deciding to use a powershell command to run clojure. There is no file to run.

Even outside cider, the crux of the problem is that, under windows, we have to deal with no less than 3 competing escaping rules for the double quotes in the clojure command line (the shell we're using, powershell and java).

There are some tools to replace the command line provided by the clojure team. But even those have still to deal with 2 separate escaping rules. 

Here is one of these : https://github.com/frericksm/clj-windows

I propose to change the way cider builds the command it uses to start clojure. By using double quote and backslash-double quote instead of single-quote, double-quote, we can make it compatible with windows in addition to unix. 

I've done this in my local copy of cider and tested it under windows and ubuntu. 
It works and i'll send a pull request for review.

Please note that this does not solve the windows problem. It only makes it more palatable. It is still necessary to use the command line tool linked above rather than the one provided by the clojure team. It is my belief that even if/when the clojure team changes its command line tool, we'll still need to have this change anyways.

**Describe alternatives you've considered**

I've been spending this day and several more hours considering any way to do it, tried so many tricks... Nothing worked but changing cider and using the command line tool listed above.

There is an alternative that wouldn't require an external tool : Directly executing the clojure launch command with powershell would be brittle and might require real parsing as we're back at dealing with 3 competing escaping rules. But cider could create a temp powershell file *.ps1 then execute powershell with this file as a parameter.

The change is very small, PR to follow.

cark:
This is not an issue to me anymore, I've written a [binary wrapper](https://github.com/cark/clojure-win-cli-wrap) to the powershell command line, and also a complete [clojure powershell scripts replacement](https://github.com/cark/clojure-cli-portable/blob/v0.0.2/src/clojure.nim). Both are sporting command line compatibility with the Posix clojure command line.

I'll leave the issue open because these are in no way official, but solutions exist outside of a cider change.  Admins, feel free to close it, as well as the PR.

scottdw:
I have a solution which uses the `powershell -encodedCommand' option. If the user sets `cider-clojure-cli-command' to `"powershell"' the clojure command and arguments are base64 encoded and sent to the powershell executable. This does make the printed jack-in command rather opaque but it should be compatible with whatever shell the user is running.

[...]

PR #2657, opened 2019-06-23
changed quoting to double quote outside, escaped double quote inside …

cark:
PR Supporting #2656 
Changed quoting for command line invocation of clojure.
Tested working with windows 10 and ubuntu.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -157,7 +157,7 @@ it using rlwrap."
   :package-version '(cider . "0.17.0"))
 
 (defcustom cider-clojure-cli-parameters
-  "-m nrepl.cmdline --middleware '%s'"
+  "-m nrepl.cmdline --middleware \"%s\""
   "Params passed to clojure to start an nREPL server via `cider-jack-in'.
 This is evaluated using `format', with the first argument being the Clojure
 vector of middleware variables as a string."
@@ -362,7 +362,7 @@ Throws an error if PROJECT-TYPE is unknown."
                           (concat
                            "["
                            (mapconcat
-                            (apply-partially #'format "\"%s\"")
+                            (apply-partially #'format "\\\"%s\\\"")
                             (cider-jack-in-normalized-nrepl-middlewares)
                             ", ")
                            "]")))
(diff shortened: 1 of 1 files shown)


Malabarba:
All of this manual escaping feels scary. Won't `shell-quote-argument` work for this? It's supposed to support all systems Emacs runs on (Windows included).

cark:
In this case, we're dealing with 3 different quoting/escaping rules on windows, all conflicting and fighting with each other. I can't imagine shell-quote-argument making sense on any of that. I explained the rationale for this pr in #2656

Malabarba:
What worries me is that I think the change you made might break the command on linux and Mac. At least, it reads like you're changing a `'` for a `"`, and that has several consequences in bash.

cark:
I tested this on ubuntu, i don't have a mac though

bbatsov:
The behaviour should be the same, as the shell's what matters not the OS. It'd be nice if a few people tested the change before we merge this PR, though. Right now I'm traveling and I'm short on time for this.

nate:
This is definitely the wrong way to do escaping in sh and bash. Things within double quotes get evaluated for variable substitution. Single quotes is the right way to go, and it should change `'` within the argument to `'\''` like so:

```
$ echo 'hi$varg'\''uys'
hi$varg'uys
```

As opposed to with double quotes where this happens:

```
$ echo "hi$varg\"uys"
hi"uys
```

nate:
It sounds like platform-specific quoting is needed, where you could just call `(quote "hi$varg'uys")` and get back `'hi$varg'\''uys'` already wrapped in quotes and the internal single quotes already escaped. For windows it could wrap it in different, appropriate quoting and escape quotes within it.

bbatsov:
See also https://github.com/clojure-emacs/cider/pull/2685 I think it's pretty elegant way to solve this quoting mess.

cark:
That's actually how i did it with my binary wrapper, it works

bbatsov:
Guess we can close this one.

PR #2658, opened 2019-06-23
Correctly exclude namespaces from apropos

dpsutton:
Confluence of bugs in orchard, cider-nrepl, and CIDER. CIDER was using
the wrong key to send excluded namespaces. cider-nrepl was not putting
this key in the right position for the orchard query. And orchard was
not correctly eliding cider.nrepl inlined deps.

- orchard: https://github.com/clojure-emacs/orchard/pull/59/
- nrepl: https://github.com/clojure-emacs/cider-nrepl/pull/618

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #2665, opened 2019-06-30
Run lint tests in emacs 26

dpsutton:
I _think_ this clears up the lint tests. Perhaps indent-region got better with the emacs upgrade. Now to tackle the tests running on emacs master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,8 @@ jobs:
 
   test-lint:
     docker:
-      - image: silex/emacs:25-dev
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -42,7 +42,8 @@ jobs:
 
   test-lint:
     docker:
-      - image: silex/emacs:25-dev
+      - image: silex/emacs:26-dev
+      - entrypoint: bash
```
bbatsov: Why is this needed?

PR #2668, opened 2019-07-02
[Fix #2659] handle `#shadow/env` tag when getting shadow-cljs builds

### LINKED ISSUE #2659: No reader for tag #shadow/env in (inst uuid)
## Expected behavior
after `sider-connect-sibling-cljs` and selecting `shadow` I should be able to select shadow-cljs build
## Actual behavior
```
error in process filter: user-error: No reader for tag #shadow/env in (inst uuid)
error in process filter: No reader for tag #shadow/env in (inst uuid)
```
use https://shadow-cljs.github.io/docs/UsersGuide.html#shadow-env in `shadow-cljs.edn`

### CIDER version information
```
;; CIDER 0.22.0snapshot (package: 20190623.1003), nREPL 0.6.0
;; Clojure 1.10.1, Java 11.0.3
```
### Emacs version
```
GNU Emacs 26.2 (build 1, x86_64-suse-linux-gnu, GTK+ Version 3.22.30)
```
### Operating system
Linux, openSUSE Leap 15.1

rpkarlsson:
Since builds are unlikely to be set from env variables just pass the
`#shadow/env` tag values to `identity`, preventing things from crashing. 
An alternative could be to read the actual env value using `getenv`.
Perhaps that would be better even though setting a build from a env variable seems highly unlikely?

Had some trouble writing tests for this as `cider--shadow-get-builds` seems to be untested and I can't figure out how to mock the edn file.
Perhaps adding a test for this isn't worth the effort and that's the reason it currently is untested?

Let me know if anything needs an improvement.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 * [#2607](https://github.com/clojure-emacs/cider/pull/2607): Use markers for specifying insertion point for `cider-eval-*-to-comment`commands. This fixes a bug where editing the buffer during a pending evaluation resulted in comments appearing in unintended locations.
 * [#2308](https://github.com/clojure-emacs/cider/issues/2308): Don't rely on the classpath in `cider-library-present-p`. Now it does a `require` instead to check if some library is present or not.
 * [#2541](https://github.com/clojure-emacs/cider/issues/2541): Hook properly CIDER's code completion machinery.
+* [#2659](https://github.com/clojure-emacs/cider/issues/2659): Handle `#shadow/env` reader tags in `cider--shadow-get-builds`.
 
 ## 0.21.0 (2019-02-19)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
> Had some trouble writing tests for this as cider--shadow-get-builds seems to be untested and I can't figure out how to mock the edn file.

Mocking the file is not hard, as you can stub `insert-file-contents`, but I'm not sure if that's needed. First I didn't know we had to deal with some reader tags, but in light of this some test might be a good idea. I guess we have to teach `parseedn` about those.

bbatsov:
//cc @yuhan0

rpkarlsson:
Added some tests to `cider--shadow-get-builds`.

bbatsov:
Great work! Thanks!

PR #2669, opened 2019-07-04
Fix tests

dpsutton:
- [ ] All code passes the linter (`make lint`) which is based on [`elisp-lint`]

Some changes and now tests pass locally on linux with emacs built from master and in the Cloud™. Fun bug that checking whether the filesystem is case sensitive goes into an infinite loop if file doesn't exist and doesn't begin with a slash. Great commit message on that one. I want to submit a bug report to emacs but I'm not sure how to get a minimal example.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,6 @@ jobs:
   test-lint:
     docker:
       - image: silex/emacs:26-dev
-      - entrypoint: bash
     steps:
       - setup
       - lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,8 @@ jobs:
 
   test-lint:
     docker:
-      - image: silex/emacs:25-dev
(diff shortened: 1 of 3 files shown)


dpsutton:
Oh, running the linter in emacs 26 makes it just work. I guess there's some improvement to known when you're in a string and not to indent. The error on emacs 25 is it attempts to indent a string of clojure code according to emacs lisp rules.

The linter also fails under emacs 27 because
> ‘seq-contains’ is an obsolete function (as of 27.1); use ‘seq-contains-p’ instead.

Not sure if `seq-contains-p` exists in 25 and 26 so we can wait on that one.

dpsutton:
emacs [bug report](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=36502)

bbatsov:
Thanks for tracking this down and fixing it! 🙇

PR #2670, opened 2019-07-05
Remove lein from circle images

dpsutton:
part of jack in is to determine which build tool and to resolve the
command. when not present it signals an error. Lein is not actually
used in the test other than just being present on the path

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,6 @@ commands:
   setup:
     steps:
       - checkout
-      - run: apt-get update && apt-get install make
       - run: make elpa
       - run: emacs --version
   test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ commands:
   setup:
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


dpsutton:
Tests are all green in circle on my fork.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ commands:
   setup:
     steps:
       - checkout
-      - run: apt-get update && apt-get install make leiningen=2.8.1-6 -y
+      - run: apt-get update && apt-get install make
```
bbatsov: Btw, I wonder if we need `apt-get update` here. I'm pretty sure any version of `make` will do.

dpsutton:
Good call @bbatsov . Apt removed and tests passing

I think the apt-get stuff took 11 seconds with lein and java. Then 4 seconds for the update and now we have shaved that off. Not huge savings but its nice to know we aren't grabbing and installing a jdk to test our emacs package :)

bbatsov:
Indeed! Thanks for tackling this, @dpsutton! 🙇

PR #2671, opened 2019-07-05
Remove ellipses ("...") from startup command log

dpsutton:
so many people (myself included) assumed this meant the command
displayed was truncated rather than "pending" or ongoing.


bbatsov:
Yeah, probably we should do this. The ellipses are a remnant of the days where the command was not displayed at all. I've been wondering now and then whether to hide it by default and show it only for debug purposes, as it's a bit intimidating and not very useful most of the time.

dpsutton:
what if we showed this in the repl banner?

bbatsov:
That's an option, although the command is so long that it won't fit comfortably in the REPL. Maybe we can just dump it in `*Messages*` and whoever needs it can find it there?

bbatsov:
(or in the `*nrepl-server*` buffer)

bbatsov:
Well, I guess we can sort this out down the road.

Issue #2682, opened 2019-07-30
cider-defun-at-point doesn't work correctly with #_ prefixed forms


alexander-yakushev:
Partially related to #2375.

`cider-defun-at-point` goes completely nuts when invoked from forms prefixed by `#_` (or any other data reader for that matter). Here's an example (█ stands for cursor):

```clj
#_(foo█)

#_(bar)
```

`(cider-defun-at-point)` here returns `"#_"`! Basically, it returns any form that follows the #_-prefixed form we are in; and that *might* be the desired behavior – but certainly not returning `#_`.

bbatsov:
Likely the underlying problem is somewhere in `clojure-mode`. We'll figure that out.

alexander-yakushev:
Keep it aliiiiiiive 😄!

alexander-yakushev:
Bump, bad bot :).

dpsutton:
Good bump! I had no idea about this and this is my fault.

dpsutton:
@alexander-yakushev 

After looking at this I don't think that I actually broke this but its a bit scary. I messed with this arcane machinery once to handle the top level stuff inside of `(comment ...)` forms. I think there are some assumptions around `beginning-of-defun` that are considered invariants and start breaking things if they aren't followed. From the docstring:

> If search is successful, return t; point ends up at the beginning of the line where the search succeeded.  Otherwise, return nil.

...

> Regardless of the values of ‘defun-prompt-regexp’ and
‘beginning-of-defun-function’, point always moves to the
beginning of the line whenever the search is successful.

Messing with this stuff can cause havoc from what i remember. One possible workaround is to use

```clj
#_
(top level form)
```

rather than `#_ (top level form)`.

However, you can try this patch to clojure-mode and see if it solves your woes:


```
diff --git a/clojure-mode.el b/clojure-mode.el
index 91f1410..bf983c5 100644
--- a/clojure-mode.el
+++ b/clojure-mode.el
@@ -2069,7 +2069,8 @@ operators can be used.  Given a positive N it will do it that
 many times."
   (let ((beginning-of-defun-function nil))
     (if (and clojure-toplevel-inside-comment-form
-             (clojure-top-level-form-p "comment"))
+             (or (clojure-top-level-form-p "comment")
+                 (clojure-top-level-form-p "#_")))
         (condition-case nil
             (save-match-data
               (let ((original-position (point))
```

[...]

Issue #2375, opened 2018-07-06
`cider-eval-toplevel-inside-comment-form` breaks autocomplete inside `comment` form


dpsutton:
Example:

```clojure
(defn thing [])

(comment
  (th|)
  )
```

if point is at `|` you would expect the autocompletion to suggest `thing`. And it does if `cider-eval-toplevel-inside-comment-form` is nil. But if this is truthy you are greeted with the following message:

> Company: backend company-capf error "Args out of range: "(th)", 0, 12" with args (candidates th)

The problem is in `cider-completion-get-context-at-point` which gets the context for the completion which includes in relevant part:

```lisp
      (let* ((pref-end (point))
             (pref-start (cider-completion-symbol-start-pos))
             (context (cider-defun-at-point))  ;; aware of comment forms
             (_ (beginning-of-defun))          ;; unconditionally goes to top level
             (expr-start (point)))
        (concat (when pref-start (substring context 0 (- pref-start expr-start)))
                "__prefix__"
                (substring context (- pref-end expr-start))))
```

The problem is that it uses `cider-defun-at-point` and then `beginning-of-defun`. The former is aware that top level may be one level deep inside of a `comment` form, the later goes to the top level open parens no questions asked. This breaks assumptions about the substring it will get.

A simple fix is to only use "beginning of defun" aware functions:

```lisp
      (let* ((pref-end (point))
             (pref-start (cider-completion-symbol-start-pos))
             (context-bounds (cider-defun-at-point t))    ;; get beginning and ending char positions
             (context (apply #'buffer-substring-no-properties context-bounds))
             (expr-start (car context-bounds)))
        (concat (when pref-start (substring context 0 (- pref-start expr-start)))
                "__prefix__"
                (substring context (- pref-end expr-start))))
```

But if we wish to keep the top-level-defun inside of comment forms feature, we need to eradicate all usages of `beginning-of-defun` in the codebase since it is not aware that top level has a bit more logic to it. I count 11 non-comment usages of this. It should be mechanical but I didn't realize this feature had this price originally.

I think it is fine and we should change `(beginning-of-defun)` to some type of `cider-beginning-of-defun` but I just wanted to make sure this was acceptable beforehand.

bbatsov:
That's a bit more nuanced than this:

* `beginning-of-defun` is also used in `clojure-mode`, so a custom replacement should likely live there
* changing this might break some third-party modes relying on `beginning-of-defun`
* is `end-of-defun` affected as well

[...]

PR #2683, opened 2019-07-30
[completion] Protect get-context function from errors

alexander-yakushev:
Mostly because of #2682, is also related to #2375.

This is useful in any case since it's better to provide fewer completions (because we couldn't obtain the context) than returning an error to the user and not showing any completions at all. Even if #2682 is somehow fixed, there could be other errors that we don't want to break on.

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -132,7 +132,11 @@ form, with symbol at point replaced by __prefix__."
                           ;; `ending-of-defun' work incorrectly in the REPL
                           ;; buffer, so context extraction fails there.
                           (derived-mode-p 'clojure-mode))
-                     (or (cider-completion-get-context-at-point)
+                     ;; We use ignore-errors here since grabbing the context
+                     ;; might fail because of unbalanced parens, or other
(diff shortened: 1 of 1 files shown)


cider-completion.el
```diff
@@ -116,14 +116,16 @@ form, with symbol at point replaced by __prefix__."
...
     (save-excursion
-      (let* ((pref-end (point))
-             (pref-start (cider-completion-symbol-start-pos))
-             (context (cider-defun-at-point))
-             (_ (beginning-of-defun))
-             (expr-start (point)))
-        (concat (when pref-start (substring context 0 (- pref-start expr-start)))
-                "__prefix__"
-                (substring context (- pref-end expr-start)))))))
+      (condition-case _
```
bbatsov: This can be replaced by `ignore-errors`. I'd also add some explanations why we're doing in the code, as suppressing errors is rarely a good idea.

alexander-yakushev:
How about this one?

bbatsov:
👍

PR #2685, opened 2019-08-06
[Fix #2656] Base64 encode command when using PowerShell

### LINKED ISSUE #2656: Change command line quoting when calling clojure cli

scottdw:
To avoid multiple layers of escaping when using tools.deps with
PowerShell, base64 encode the clojure command and arguments when
calling jack-in with `cider-clojure-cli-command` as `"powershell"`.

Initial stab at using PowerShell's encodedCommand to use tools.deps. On the plus side it works and you can connect using just PowerShell and tools.deps, downside is the launch command is ugly. eg:

```
[nREPL] Starting server via "c:/WINDOWS/System32/WindowsPowerShell/v1.0/powershell.exe" -encodedCommand YwBsAG8AagB1AHIAZQAgAC0AUwBkAGUAcABzACAAJwB7ADoAZABlAHAAcwAgAHsAbgByAGUAcABsACAAewA6AG0AdgBuAC8AdgBlAHIAcwBpAG8AbgAgACIAIgAwAC4ANgAuADAAIgAiAH0AIAByAGUAZgBhAGMAdABvAHIALQBuAHIAZQBwAGwAIAB7ADoAbQB2AG4ALwB2AGUAcgBzAGkAbwBuACAAIgAiADIALgA0AC4AMAAiACIAfQAgAGMAaQBkAGUAcgAvAGMAaQBkAGUAcgAtAG4AcgBlAHAAbAAgAHsAOgBtAHYAbgAvAHYAZQByAHMAaQBvAG4AIAAiACIAMAAuADIAMgAuADAALQBiAGUAdABhADgAIgAiAH0AfQB9ACcAIAAtAG0AIABuAHIAZQBwAGwALgBjAG0AZABsAGkAbgBlACAALQAtAG0AaQBkAGQAbABlAHcAYQByAGUAIAAnAFsAIgAiAHIAZQBmAGEAYwB0AG8AcgAtAG4AcgBlAHAAbAAuAG0AaQBkAGQAbABlAHcAYQByAGUALwB3AHIAYQBwAC0AcgBlAGYAYQBjAHQAbwByACIAIgAsACAAIgAiAGMAaQBkAGUAcgAuAG4AcgBlAHAAbAAvAGMAaQBkAGUAcgAtAG0AaQBkAGQAbABlAHcAYQByAGUAIgAiAF0AJwA=
```
``` elisp
(insert (decode-coding-string (base64-decode-string "YwBsAG8AagB1AHIAZQAgAC0AUwBkAGUAcABzACAAJwB7ADoAZABlAHAAcwAgAHsAbgByAGUAcABsACAAewA6AG0AdgBuAC8AdgBlAHIAcwBpAG8AbgAgACIAIgAwAC4ANgAuADAAIgAiAH0AIAByAGUAZgBhAGMAdABvAHIALQBuAHIAZQBwAGwAIAB7ADoAbQB2AG4ALwB2AGUAcgBzAGkAbwBuACAAIgAiADIALgA0AC4AMAAiACIAfQAgAGMAaQBkAGUAcgAvAGMAaQBkAGUAcgAtAG4AcgBlAHAAbAAgAHsAOgBtAHYAbgAvAHYAZQByAHMAaQBvAG4AIAAiACIAMAAuADIAMgAuADAALQBiAGUAdABhADgAIgAiAH0AfQB9ACcAIAAtAG0AIABuAHIAZQBwAGwALgBjAG0AZABsAGkAbgBlACAALQAtAG0AaQBkAGQAbABlAHcAYQByAGUAIAAnAFsAIgAiAHIAZQBmAGEAYwB0AG8AcgAtAG4AcgBlAHAAbAAuAG0AaQBkAGQAbABlAHcAYQByAGUALwB3AHIAYQBwAC0AcgBlAGYAYQBjAHQAbwByACIAIgAsACAAIgAiAGMAaQBkAGUAcgAuAG4AcgBlAHAAbAAvAGMAaQBkAGUAcgAtAG0AaQBkAGQAbABlAHcAYQByAGUAIgAiAF0AJwA=") 'utf-16le))
clojure -Sdeps '{:deps {nrepl {:mvn/version ""0.6.0""} refactor-nrepl {:mvn/version ""2.4.0""} cider/cider-nrepl {:mvn/version ""0.22.0-beta8""}}}' -m nrepl.cmdline --middleware '[""refactor-nrepl.middleware/wrap-refactor"", ""cider.nrepl/cider-middleware""]'nil
```

I can add tests and manual updates if it is decided this is the correct approach and they are necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2656](https://github.com/clojure-emacs/cider/issues/2656): Base64 encode clojure command and arguments on jack-in when `cider-clojure-cli-command` is `"powershell"` to avoid escaping issues. If no `clojure` command is found on windows `cider-clojure-cli-command` defaults to `"powershell"`.
 * Allow editing of jack in command with prefix or when `cider-edit-jack-in-command` is truthy.
 * New defcustom `cider-repl-require-ns-on-set`: Set it to make cider require the namespace before setting it, when calling `cider-repl-set-ns`.
 * [#2611](https://github.com/clojure-emacs/cider/issues/2611): Add `eval`-based classpath lookup fallback. It's used when cider-nrepl is not present.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New features
 
-* [#2656](https://github.com/clojure-emacs/cider/issues/2656): Base64 encode clojure command and arguments on jack-in when `cider-clojure-cli-command` is `"powershell"` to avoid escaping issues. If no `clojure` command is found on windows `cider-clojure-cli-command` defaults to `"powershell"`.
+* [#2656](https://github.com/clojure-emacs/cider/issues/2656): Base64 encode clojure command and arguments on jack-in when `cider-clojure-cli-command` is `"powershell"` to avoid escaping issues. If no `clojure` command is found on Windows `cider-clojure-cli-command` defaults to `"powershell"`.
 * Allow editing of jack in command with prefix or when `cider-edit-jack-in-command` is truthy.
 * New defcustom `cider-repl-require-ns-on-set`: Set it to make cider require the namespace before setting it, when calling `cider-repl-set-ns`.
 * [#2611](https://github.com/clojure-emacs/cider/issues/2611): Add `eval`-based classpath lookup fallback. It's used when cider-nrepl is not present.
--- a/cider.el
+++ b/cider.el
@@ -139,7 +139,7 @@ version from the CIDER package or library.")
   :package-version '(cider . "0.9.0"))
 
 (defcustom cider-clojure-cli-command
-  (if (and (= system-type 'windows-nt)
+  (if (and (eq system-type 'windows-nt)
            (null (executable-find "clojure")))
       "powershell"
     "clojure")
--- a/cider.el
+++ b/cider.el
@@ -139,10 +139,14 @@ version from the CIDER package or library.")
   :package-version '(cider . "0.9.0"))
 
 (defcustom cider-clojure-cli-command
-  "clojure"
+  (if (and (= system-type 'windows-nt)
+           (null (executable-find "clojure")))
+      "powershell"
+    "clojure")
   "The command used to execute clojure with tools.deps (requires Clojure 1.9+).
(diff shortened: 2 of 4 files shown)


bbatsov:
I like this idea! Go ahead with adding some documentation and tests for it.

scottdw:
Tests and manual updates added. If everything is okay I can rebase and squash the commits.

bbatsov:
Yeah, it looks good to to me. Btw, I wonder if we shouldn't set the command to powershell by default on Windows. From what I gathered there aren't any good alternatives, right?

scottdw:
I don't believe any of the other options have become overwhelmingly popular or blessed by the clojure core team. But in case the user is using an alternate option we also can check that no `clojure` executable is found before setting the default.

scottdw:
Bah, didn't mean to make that a separate commit. I'll have to either clean it up or create a new one tomorrow.

Issue #2472, opened 2018-10-02
'help-echo text-properties make messages less persistent


vemv:
Hi there,

I don't use the functionality provided by `cider--help-echo`. Still, 'help-echo text properties are unconditionally added, which makes invocations to `(message...)` less persistent: the emitted message will disappear after hovering Clojure code with the mouse.

I propose that if `cider-use-tooltips` is customized to nil, this LOC https://github.com/clojure-emacs/cider/blob/e071b027c1dafea8599db8c2b039f11d9eacf994/cider-mode.el#L987 (and possibly other similar ones) isn't executed.

WDYT?

Thanks - Victor

bbatsov:
Sounds reasonable to me.

vemv:
Not stale

(can open a PR when I find the time)

vemv:
Time for some elispin'

PR #2688, opened 2019-08-10
Observe `cider-use-tooltips` so it can prevent filling the echo area

vemv:
Fixes clojure-emacs/cider#2472 using the agreed strategy.

I have tested out the changes in my day-to-day Emacs usage, and they work fine.


bbatsov:
Thanks!

PR #2689, opened 2019-08-10
Optionally specify callback to cider load buffer

magnars:
- `cider-load-buffer` (normally bound to C-c C-k) does not pass a callback to `cider-request:load-file`. With this change, a user can now optionally specify this callback. 

- when no callback is passed to `cider-request:load-file`, it will fall back on `cider-load-file-handler`, which does not specify a done-handler. With this change, a user can now optionally specify a done-handler.

These two changes together would allow users to perform some task after loading the current buffer, by doing something like this:

```elisp
(cider-load-buffer nil (cider-load-file-handler nil (lambda (&rest _) (message "I'm done!"))))
```

In my case that would be running tests.

-----------------

- [-] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1079,7 +1079,7 @@ passing arguments."
       (widen)
       (substring-no-properties (buffer-string)))))
 
-(defun cider-load-buffer (&optional buffer)
+(defun cider-load-buffer (&optional buffer callback)
   "Load (eval) BUFFER's file in nREPL.
 If no buffer is provided the command acts on the current buffer.  If the
 buffer is for a cljc file, and both a Clojure and ClojureScript REPL exists
@@ -1114,7 +1114,8 @@ for the project, it is evaluated in both REPLs."
                                        (funcall cider-to-nrepl-filename-function
                                                 (cider--server-filename filename))
                                        (file-name-nondirectory filename)
-                                       repl)))
+                                       repl
+                                       callback)))
           (message "Loading %s..." filename))))))
 
 (defun cider-load-file (filename)
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -491,7 +491,7 @@ or it can be a list with (START END) of the evaluated region."
                                    (cider-handle-compilation-errors err eval-buffer))
                                  '())))
 
-(defun cider-load-file-handler (&optional buffer)
+(defun cider-load-file-handler (&optional buffer done-handler)
   "Make a load file handler for BUFFER."
   (let ((eval-buffer (current-buffer)))
     (nrepl-make-response-handler (or buffer eval-buffer)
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice idea! Let me think a bit about this, as there's another approach to do the same thing - namely trigger some hook functions in the done handler, which might be more user-friendly.

bbatsov:
Let me also ping @Malabarba here, as back in the day he had a different approach to this https://endlessparentheses.com/test-driven-development-in-cider-and-emacs.html

bbatsov:
Hmm, actually I see we already have the file loaded hook, but it's being triggered in another handler (the value handler). I'm not sure if this has any practical differences, so I guess you can help me understand why using the done callback is preferable for this.

magnars:
I'm happy to use the `cider-file-loaded-hook` for my use case. I just didn't notice it there, since I was thinking in callbacks. :) I still see some use cases for this change, in that hooks are global. When passing in the callback you can affect only this invocation, without having to juggle un-registering hooks afterwards.

magnars:
As a side note, the trick @Malabarba uses is the one I'm trying to avoid here - since I had to resort to a `(reset)` in my save-handler, which doesn't scale well with project size.

bbatsov:
Btw, I've noticed one small problem with our existing hook - starting with nREPL 0.6 you can actually get multiple `value` responses (chunks of the same result). While this is unlikely to affect ns loading, it might also be better to move the triggering of the hook to the `done` handler as well.

> When passing in the callback you can affect only this invocation, without having to juggle un-registering hooks afterwards.

Yeah, that's a perfectly valid argument.

cider-eval.el
```diff
@@ -501,7 +502,7 @@ or it can be a list with (START END) of the evaluated region."
                                  (lambda (_buffer err)
                                    (cider-emit-interactive-eval-err-output err)
                                    (cider-handle-compilation-errors err eval-buffer))
-                                 '()
+                                 (or done-handler '())
```
Malabarba: since `'()` is the same as `nil`, this whole line is the same as just `done-handler`

Malabarba:
I  think it's reasonable for the function to receive a callback argument, since it's already implemented in the lower functions. I think both the hook and the local callback will probably have its uses in different situations.

bbatsov:
@Malabarba Fine by me. Let's get this baby merged! Btw, do you think we should move the hook in the done handler as well?

@magnars Please, add a changelog entry for the PR.

magnars:
Done, as well as freshly rebased onto master.

bbatsov:
Thanks!

PR #2694, opened 2019-08-26
Fix two typos in misc_features.adoc

jhacksworth:
Typos in the keybindings for the prefix and double-prefix arguments to `cider-ns-refresh`


bbatsov:
Thanks!

PR #2700, opened 2019-08-31
Fix docstring for cider-mode-menu

dsedivec:
Currently, at least on macOS, the text "Menu for CIDER mode." appears as the final item under the "CIDER" menu, with no effect when clicked.  I believe this string was meant to be the docstring for `cider-mode-menu`, and this commit makes it so.

-----------------


bbatsov:
Thanks!

Issue #2597, opened 2019-02-22
figwheel-main not available


cfraz89:
Since updating cider, it seems to no longer able to detect and start fighwheel-main in my project. This setup was previously working with a version of cider from late last year.

Environment:
;; Connected to nREPL server - nrepl://localhost:36971
;; CIDER 0.21.0snapshot (package: 20190216.2255), nREPL 0.6.0
;; Clojure 1.9.0, Java 1.8.0_201

Messages:
```
[nREPL] Starting server via /usr/local/bin/clojure -Sdeps '{:deps {nrepl {:mvn/version "0.6.0"} cider/piggieback {:mvn/version "0.4.0"} cider/cider-nrepl {:mvn/version "0.21.1"}}}' -m nrepl.cmdline --middleware '["cider.nrepl/cider-middleware", "cider.piggieback/wrap-cljs-repl"]'...
[nREPL] server started on 34675
[nREPL] Establishing direct connection to localhost:34675 ...
[nREPL] Direct connection to localhost:34675 established
[nREPL] Establishing direct connection to localhost:34675 ...
[nREPL] Direct connection to localhost:34675 established
error in process filter: user-error: Figwheel-main is not available.  Please check https://docs.cider.mx/en/latest/clojurescript
error in process filter: Figwheel-main is not available.  Please check https://docs.cider.mx/en/latest/clojurescript
```

deps.edn
```
{:deps {bidi {:mvn/version "2.1.4"}
        cljsjs/create-react-class {:mvn/version "15.6.3-0"}
        cljsjs/csv {:mvn/version "1.1.1-0"}
        cljsjs/react {:mvn/version "16.0.0-0"}
        cljsjs/react-dom {:mvn/version "16.0.0-0"}
        com.andrewmcveigh/cljs-time {:mvn/version "0.5.2"}
        org.clojure/clojure {:mvn/version "1.9.0"}
        org.clojure/clojurescript {:mvn/version "1.10.339"}
        org.clojure/core.async {:mvn/version "0.4.474"}
        org.omcljs/om {:mvn/version "1.0.0-beta4"}
        prismatic/dommy {:mvn/version "1.1.0"}
        venantius/accountant {:mvn/version "0.2.4"}}
 :aliases {:cljs {:main-opts ["-m" "cljs.main"]}
           :fig {:extra-deps {com.bhauman/figwheel-main {:mvn/version "0.2.0"}
[...]

PR #2698, opened 2019-09-04
"RuntimeException Invalid token: :  " with cider-jack-in-cljs in figwheel-main template

krathdog:
## Expected behavior

connect to figwheel repl.

## Actual behavior

RuntimeException Invalid token: :  clojure.lang.Util.runtimeException (Util.java:221)

First I tried to get emacs-cider + figwheel-main + clojure-cli working with the flappy bird demo repo:

clone https://github.com/bhauman/flappy-bird-demo-new.git
run cider-jack-in-cljs
select clojure-cli figwheel-main :dev

But I'm getting this error before the figwheel repl:

RuntimeException Invalid token: :  clojure.lang.Util.runtimeException (Util.java:221)

I switched to a more basic repo to debug it.
I used clj-new tool and figwheel-main template:

clj -A:new figwheel-main clj-new-figwheel-main-template.core -- --reagent

my deps.edn has:

{:new {:extra-deps {seancorfield/clj-new {:mvn/version "0.7.8"}} :main-opts ["-m" "clj-new.create"]}}

Verified it works at command-line (gives me figwheel repl and opens browser on page):

clojure -A:fig:build

Now open the cljs file in emacs to test with cider.
Try to start figwheel repl:

run cider-jack-in-cljs
select clojure-cli figwheel-main :dev

Got this error:

error in process filter: user-error: Figwheel-main is not available.  Please check https://docs.cider.mx/en/latest/clojurescript
error in process filter: Figwheel-main is not available.  Please check https://docs.cider.mx/en/latest/clojurescript

Fixed it by adding this entry to .dir-local.el (see https://github.com/clojure-emacs/cider/issues/2597):

((clojurescript-mode . ((cider-clojure-cli-global-options . "-R:fig"))))

This is in the clj-new-figwheel-main deps.edn to support that:

 :aliases {:fig {:extra-deps
                  {com.bhauman/rebel-readline-cljs {:mvn/version "0.1.4"}
                   com.bhauman/figwheel-main {:mvn/version "0.2.3"}}

Now getting this same error again as with the flappy-bird-demo:

RuntimeException Invalid token: :  clojure.lang.Util.runtimeException (Util.java:221)

Try to start clojure repl instead and then manually start figwheel repl in it.
First update cider to automatically inject dependencies and middleware for clojure repl:

(cider-add-to-alist 'cider-jack-in-dependencies "cider/piggieback" "0.4.1")
(add-to-list 'cider-jack-in-nrepl-middlewares "cider.piggieback/wrap-cljs-repl")

Now they look like this:

cider-jack-in-dependencies -> (("cider/piggieback" "0.4.1") ("nrepl" "0.6.0"))
cider-jack-in-nrepl-middlewares -> ("cider.piggieback/wrap-cljs-repl" ("refactor-nrepl.middleware/wrap-refactor" :predicate cljr--inject-middleware-p) "cider.nrepl/cider-middleware")

run cider-jack-in-clj

Verified it now works when manually starting figwheel-main:

user> (require 'figwheel.main.api)
2019-08-29 13:25:16.831:INFO::nRepl-session-22f4e246-5964-44b0-af74-8c866f9a7869: Logging initialized @35929ms to org.eclipse.jetty.util.log.StdErrLog
;; => nil
user> (figwheel.main.api/start :dev)
[Figwheel] Validating figwheel-main.edn
...

The emacs testing above was all done with doom emacs. But I also did a quick install
of prelude emacs and ran the same test with jack-in-cider-cljs on the clj-new-figwheel-main
repo and got the same error Invalid token: exception.

I am new to clojure and emacs and cider (and this is my first github bug) so let me know
if there is anything I missed doing right or anything I can do to help resolve it. Thanks!

### CIDER version information

;; CIDER 0.22.0snapshot, nREPL 0.6.0
;; Clojure 1.9.0, Java 1.8.0_71

### Emacs version

"GNU Emacs 26.2 (build 1, x86_64-apple-darwin18.7.0, Carbon Version 158 AppKit 1671.6)
 of 2019-08-18"

Doom v2.0.9 (Emacs v26.2)

~/.emacs.d.prelude (master % u=)$ git lg1 -1
* c951f8e - (2 months ago) Update PDF cheatsheet with new Rust bindings - apiraino (HEAD -> master, origin/master, origin/HEAD)

### Operating system

macOS Mojave 10.14.6


krathdog:
I created this PR from emacs with magit/forge, which I'm new to, so I'm not sure how to add the checklist that I've seen in some other PRs.

I did install cask and ran 'make test' and 'make lint' successfully.

I thought about adding a test but didn't have any good ideas for how to do it using the buttercup facilities.

I didn't see any docs too touch based on this but let me know if I missed something.

@dpsutton Regarding your comment:

"The fallback to cider-figwheel-main-default-options never happens since the
read-from-minibuffer always returns an empty string"

I don't believe this is the case since cider-figwheel-main-default-options is
the first test in the 'or' so it would always be checked and if non-nil then
read-from-minibuffer wouldn't happen.

dpsutton:
haven't looked at your code yet but you're right, its not a fallback, the minibuffer input is the fallback. I was mistaken not you :)

krathdog:
Previously I tested using eval-defun and it was working as expected, but as I'm testing it now it stopped working. I'm not sure if its the fix itself or something in my environment and how I'm testing it, I'm investigating.

cider.el
```diff
@@ -767,7 +767,8 @@ Figwheel for details."
   (let ((form "(do (require 'figwheel.main) (figwheel.main/start %s))")
         (options (string-trim
                   (or cider-figwheel-main-default-options
-                      (read-from-minibuffer "Select figwheel-main build (e.g. :dev): ")))))
+                      (read-from-minibuffer "Enter figwheel-main build-id (must match file named <build-id>.cljs.edn. Default is 'dev'): "
```
bbatsov: Such a big prompt seems pretty weird to me. I'd rather just check in the next step whether the input was empty and whether it matches an existing build file.

krathdog: I had same hesitation about the prompt length - I can remove the advisement about the filename. 

If you prefer it to additionally check for the existence of the build file, I can look into how to do that. In which case it'd verify the build file whether input was empty or not, in case it was mistyped, etc.

As I mentioned, this fix is actually not correct. In order for the read-from-minibuffer to use the default value, the 'read' field must be non-nil. I'm testing fix now and will update when ready.

krathdog:
Ok, so I determined I just went through the same sequence of realizations as the OP on this thread:

https://groups.google.com/forum/#!topic/gnu.emacs.help/sZzprGr_aKI

Basically, I can't use the read-from-minibuffer default-value arg as hoped.

Instead I'll do what cider-repl-history-read-regexp does -- handle the empty string input separately.

krathdog:
Not sure if there is anything else for me to do on this, so just wanted to note that the fix is done on my end, I'll wait for any review comments. I did unit testing but I'll do another round of cask test and lint after its ready to go. Also, I can add any useful tests and update documentation.

dpsutton:
Looks great but I've got one concern and one minor thing:
1. minor: use `string-blank-p` rather than comparing to an empty string
2. I like the idea of checking if the build file exists but it seems like this implementation doesn't look at the project directory but uses the current directory of the buffer. This means this would fail my most common usage which is to jack in from a source buffer not at the project root.

Couple ways to solve this:
- make sure you check project root for the build file
- look into checking the project root for the *.cljs.edn build files to present a list to the user
- abandon the check for the build file existence.

I've not often hit a the issue of selecting an incorrect build but when I do figwheel has a helpful error message:
> ExceptionInfo File dev.cljs.edn does not exist  clojure.core/ex-info (core.clj:4739)

Also, since this check happens once the jvm has started up, the user-error from CIDER occurs at the same time as the error message from figwheel so we are adding no benefit to the user but introducing a codepath that can introduce a false negative: signalling a bad build when the build is actually valid.

I'd probably back this out to just ensure we aren't sending blank strings to the init form and we can wait for more reports of confusion or have a better way to validate the figwheel build.

bbatsov:
I think a better approach for all of this would be to just check for the names of the builds in the root dir of the project (you can easily get it with `clojure-project-dir`) and then present those as options in the prompt. Showing all valid options beats making the users have to remember those. We already do something similar for shadow-cljs, so I'd just replicate this logic.

bbatsov:
See https://github.com/clojure-emacs/cider/pull/2646/files

krathdog:
Thanks for the reviews, makes sense! I'll circle back with another delta soon.

bbatsov:
Thanks!

krathdog:
I looked at the cider--shadow-get-builds impl but I think the figwheel-main version needed to be different and actually simpler. This seems to work so I'm providing it for feedback before I look at updating tests or docs.

bbatsov:
The changes look good to me. Might be a good idea to add a couple of simple unit tests as well.

bbatsov:
Just one thing - I wonder what should be the behaviour if we find no builds. Should we just show some error saying that something's missing?

krathdog:
If cider does not find any build files, I think we should warn user and allow
them to abort.

One unfortunate aspect of this is that in this state if you try to run cider-quit
it will say: "user-error: No cljs REPLs in current session" presumably because
the started repl is not linked to the cljs file until the cljs-repl is started.

If the user chooses to continue anyways, either by hitting enter with no build
name, or entering a build name that doesn't exist, we can proceed and let
figwheel report the error, which as @dpsutton noted it does print a useful
error message.

In the case where user hits enter with no build name, I need to make sure not
to send figwheel the ":" that currently happens.

Let me know if this sounds right or not and I'll try to come up changes.

bbatsov:
> If cider does not find any build files, I think we should warn user and allow
them to abort.

Sounds good to me.

> One unfortunate aspect of this is that in this state if you try to run cider-quit
it will say: "user-error: No cljs REPLs in current session" presumably because
the started repl is not linked to the cljs file until the cljs-repl is started.

You'll have to explain to me further. Which REPL would this be? After all we wouldn't have created anything, right?

> If the user chooses to continue anyways, either by hitting enter with no build
name, or entering a build name that doesn't exist, we can proceed and let
figwheel report the error, which as @dpsutton noted it does print a useful
error message.

Or just throw some user error and not do anything.

krathdog:
> > If cider does not find any build files, I think we should warn user and allow
> > them to abort.
> 
> Sounds good to me.

Ok, though I didn't end up doing this exactly. Instead, if there are no build files I
just exited. The reasoning was that allowing user to continue would definitely result
in an error from Figwheel anyways, so it didn't seem worth the extra code to provide
that option.

FYI, I rebased to the master branch and force-pushed to this branch. Hope that's not
a problem.

> > If the user chooses to continue anyways, either by hitting enter with no build
> > name, or entering a build name that doesn't exist, we can proceed and let
> > figwheel report the error, which as @dpsutton noted it does print a useful
> > error message.
> 
> Or just throw some user error and not do anything.

From my testing, the user has to somewhat go out of their way to enter an empty
build name, since they have to Ctrl-P out of the completion choices to select the
prompt and then from there hit enter while still blank. So I didn't add extra code
to catch that possibility.

> 
> > One unfortunate aspect of this is that in this state if you try to run cider-quit
> > it will say: "user-error: No cljs REPLs in current session" presumably because
> > the started repl is not linked to the cljs file until the cljs-repl is started.
> 
> You'll have to explain to me further. Which REPL would this be? After all we wouldn't have created anything, right?
> 

One unfortunate aspect of this is that in this state if you try to run cider-quit
it will say: "user-error: No cljs REPLs in current session" presumably because
the started repl is not linked to the cljs file until the cljs-repl is started.

You'll have to explain to me further. Which REPL would this be? After all we wouldn't have created anything, right?

If you run cider-jack-in-cljs but figwheel-main initialization doesn't finish
(e.g., missing build file) then you'll have a connected repl buffer but without
figwheel. If you try to run cider-quit from the .cljs buffer you'll get this
error:

user-error: No cljs REPLs in current session "figwheel/clj-new-figwheel-main.core:localhost:54335"

If you switch to the repl buffer and run cider-quit you'll get:

cider-ns-refresh: Reloading (figwheel.main.css-reload cljs.stacktrace figwheel.repl figwheel.main.async-result figwheel.main.testing figwheel.core figwheel.main)

If you just try to kill the repl process by killing the buffer you might get
an error as well (could be specific to my doom emacs config):

Error killing transient buffer: (error Process nrepl-connection is not a subprocess)

To kill the buffer and its repl processes I have to open the buffer one more
time and kill it again.

Its cider-jack-in-cljs calling nrepl-start-server-process that results in the
repl buffer opening with a prompt and associated repl server and connection
processes.

But its in cider--connected-handler call to cider-repl-init that the
initialization form generated by cider-figwheel-main-init-form and stored in the
:repl-init-function of params passed to it that figwheel-main is started.
So this second step can fail after the initial repl is successfully started.

It seems that if cider-repl-init fails during a cider-jack-in-cljs, that the
repl buffer's cider-repl-type remains as 'clj' instead of changing to 'cljs'.
And cider-current-repl returns nil because cider-repl-type-for-buffer returns
'cljs' which causes cider-repls to return nil for that type.

```emacs-lisp
(defun cider-current-repl (&optional type ensure)
...
      (let* ((type (or type (cider-repl-type-for-buffer)))
             (repls (cider-repls type ensure))
             (repl (if (<= (length repls) 1)
                       (car repls)
```

Well that explains why cider-quit called from the .cljs buffer doesn't close
the repl it started. I didn't look into why it fails when called from the repl
buffer.

I didn't find where cider-repl-init changes the cider-repl-type from
'clj' to 'cljs'.

I also don't see where the figwheel.main.api/cljs-repl or equivalent call is
made.

bbatsov:
It seems to me that the issues related to quitting are completely unrelated to the problem at hand, so I guess we can tackle them in some dedicated ticket.

Just add a changelog entry, test and docs for your change and we can wrap it up.

bbatsov:
@krathdog ping :-)

krathdog:
hi! sorry for delay, am looking at this right now, should have an update today. but one issue i had that i haven't figured out, maybe you have a clue. when i try to run the tests from cask i get this:

```Some dependencies were not available: spinner, queue```

i've been getting that same message every time i try for last couple weeks so seems like its not issue with connectivity or uptime on the archive servers but i don't know much about it.

instead, i'm trying to run the tests manually in emacs.

since cider-figwheel-main-init-form will throw an error if cider--figwheel-main-get-builds returns no builds, I thought I could add a test for this behavior but I don't see a way to verify it with buttercup nor an example of any other tests verifying thrown errors.

when i started looking at the figwheel-main tests, i noticed under the (describe "from minibuffer"...)
tests that they pass various argument types that figwheel.main/start will accept. so apparently while
the existing prompt asks the user to enter their build name, it actually accepts more than that and cider tests for it. so perhaps the prompt should be changed to reflect that? e.g., instead of asking
to select the build from the list of found builds, we could ask to either select a build or provide an argument supported by figwheel.main/start, however that could be phrased. or i could just ignore that cider has these tests for this capability that the user is not apprised of, just let me know, i'm just observing what i found.

i thought about whether i could add any new tests that would be useful but didn't come up with any. one possibility was spying on directory-files and have it return some set of filenames to verify that the list of builds captured by the regex were correct. another possibility was spying on the completing-read and returning a particular build name and verifying that the init-form was properly formatted. but i felt that both of these were testing trivial functionality so didn't add them. if you think there is some other test that would be useful to add let me know. otherwise i've just fixed the existing tests to work.

dpsutton:
I've asked the emacs mailing list about the elpa situation here: https://lists.gnu.org/archive/html/emacs-devel/2019-10/msg00063.html

I wouldn't jump through any big hoops to get this tested. It seems like a simple enough workflow and mocking up build files for a test seems a bit much.

dpsutton:
Re the unable to download packages: helpful information from a former maintainer

https://lists.gnu.org/archive/html/emacs-devel/2019-10/msg00064.html

https://www.reddit.com/r/emacs/comments/bn6k1y/updating_gnu_elpa_keys/

the gist is that the gpg key has been updated. If you're on the latest 26.3 or master it will work fine otherwise you need to update keys. Info in the reddit thread but i'm still digesting it myself so I won't put instructions here.

krathdog:
> Re the unable to download packages: helpful information from a former maintainer
> 
> https://lists.gnu.org/archive/html/emacs-devel/2019-10/msg00064.html
> 
> https://www.reddit.com/r/emacs/comments/bn6k1y/updating_gnu_elpa_keys/
> 
> the gist is that the gpg key has been updated. If you're on the latest 26.3 or master it will work fine otherwise you need to update keys. Info in the reddit thread but i'm still digesting it myself so I won't put instructions here.

Thanks much for following up on that issue, I'll check up on those threads now.

I just pushed some updates to the tests and docs if you get a chance to look. I'm not sure how to build the docs to test them. If you look at the changes and think I should build them before checking in then let me know if there are instructions.

Lastly, in general, should I rebase and collapse the deltas and provide a proper commit message or is that something you do when merging.

bbatsov:
> I just pushed some updates to the tests and docs if you get a chance to look. I'm not sure how to build the docs to test them. If you look at the changes and think I should build them before checking in then let me know if there are instructions.

I'll just rebuild it myself. The markup seems correct to me. The instructions are in the documentation, btw. :D 

> Lastly, in general, should I rebase and collapse the deltas and provide a proper commit message or is that something you do when merging.

I'll take care of this. Thanks for your contribution!

bbatsov:
@krathdog Btw, it seems some figwheel related tests are failing on master right now.

krathdog:
> @krathdog Btw, it seems some figwheel related tests are failing on master right now.

How do I look at the failing tests? I will try to reproduce by updating my elpa keys to get cask running again. Once I can reproduce I'll submit a fix.

bbatsov:
@dpsutton fixed the problem in #2725.

Issue #2707, opened 2019-09-09
cider-selector with a non clojure buffer/repl selected produces error: "Wrong type argument: stringp, nil"


ahungry:
I did some fast debugging last week on this, and it seems to be with how it's calling sesman to get the active sessions.

I think CIDER should work around this, as the behavior appears well defined at the sesman level (given an active buffer, give linked sessions).

However, the Common Lisp variant of this with SLIME and slime-selector works regardless of what buffer the user is currently in (it has a jump to last visited lisp file, and a jump to last visited REPL feature).

I think this would make the CIDER selector much more useful, if a single hotkey could jump back to the last visited clojure file or clojure repl if the current file returns no matches.

bbatsov:
I guess we can close this one, as it was fixed by #2711.

PR #2711, opened 2019-09-14
Bugfix/fix cider selector problem

ahungry:
This fixes : 
https://github.com/clojure-emacs/cider/issues/2707
-----------------

- [ x ] The commits are consistent with our [contribution guidelines](../.github/CONTRIBUTING.md)
- [ - ] You've added tests (if possible) to cover your change(s)
- [ x ] All tests are passing (`make test`)
- [ x ] All code passes the linter (`make lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes
- [ x ] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- [ x ] You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-selector.el
+++ b/cider-selector.el
@@ -49,20 +49,23 @@ DESCRIPTION is a one-line description of what the key selects.")
 Not meant to be set by users.  It's used internally
 by `cider-selector'.")
 
-(defun cider-selector--recently-visited-buffer (mode)
+(defun cider-selector--recently-visited-buffer (mode &optional consider-visible-p)
   "Return the most recently visited buffer, deriving its `major-mode' from MODE.
-Only considers buffers that are not already visible."
+CONSIDER-VISIBLE-P will allow handling of visible windows as well.
+First pass only considers buffers that are not already visible.
+Second pass will attempt one of visible ones for scenarios where the window
+is visible, but not focused."
   (cl-loop for buffer in (buffer-list)
            when (and (with-current-buffer buffer
                        (derived-mode-p mode))
                      ;; names starting with space are considered hidden by Emacs
                      (not (string-match-p "^ " (buffer-name buffer)))
-                     ;; Why not? If we have a window split and want to use cider-selector
-                     ;; to jump between the latest clojure / REPL?
-                     ;; (null (get-buffer-window buffer 'visible))
-                     )
+                     (or consider-visible-p
+                         (null (get-buffer-window buffer 'visible))))
            return buffer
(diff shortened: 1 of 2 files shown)


cider-selector.el
```diff
@@ -139,7 +145,8 @@ is chosen.  The returned buffer is selected with
 
 (def-cider-selector-method ?r
   "Current REPL buffer."
-  (cider-current-repl))
+  (or (cider-current-repl)
```
bbatsov: The docstring probably needs to be adjusted, as now this does something different.

bbatsov:
Might also be a good idea to expand a bit the documentation for `cider-selector` in the manual.

cider-selector.el
```diff
@@ -49,17 +49,23 @@ DESCRIPTION is a one-line description of what the key selects.")
 Not meant to be set by users.  It's used internally
 by `cider-selector'.")
 
-(defun cider-selector--recently-visited-buffer (mode)
+(defun cider-selector--recently-visited-buffer (mode &optional consider-visible-p)
   "Return the most recently visited buffer, deriving its `major-mode' from MODE.
-Only considers buffers that are not already visible."
+CONSIDER-VISIBLE-P will allow handling of visible windows as well.
+First pass only considers buffers that are not already visible.
```
bbatsov: I think we can probably simplify the logic and consider all buffers right away. I've mostly copied this code from SLIME back in the day and I didn't think much about it. Treating differently hidden/visible buffers seems weird to me.

ahungry: I had thought about that, and on one of the commits actually implemented as such, but that had a side-effect of removing the ability to use cider-selector to flip-flop between the latest 2 buffers of the same type (to change from a clojure-mode buffer to the last visited but not shown clojure-mode buffer - not using the second pass would change this operation into a noop essentially).

cider-selector.el
ahungry: updated

cider-selector.el
bbatsov: Got it. Well, I guess it really depends on what do we consider the purpose of this command to be - for me it was always to jump to a source buffer from some special buffer and vice versa. Of course, there's also the question of how people are actually using something and what do they consider useful. :D 

That's why I was thinking we should extend the documentation of `cider-selector` in general so people would know how it's supposed to be used.

PR #2716, opened 2019-09-20
Display evaluation result overlays at end of sexp

yuhan0:
Rename argument from point -> where to match cider--make-result-overlay and
update docstring
-----------------

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -59,16 +59,26 @@ font-locking it."
 
 (defcustom cider-use-overlays 'both
   "Whether to display evaluation results with overlays.
-If t, use overlays.  If nil, display on the echo area.  If both, display on
-both places.
+If t, use overlays determined by `cider-result-overlay-position'.
+If nil, display on the echo area.
+If both, display on both places.
 
 Only applies to evaluation commands.  To configure the debugger overlays,
 see `cider-debug-use-overlays'."
-  :type '(choice (const :tag "End of line" t)
-                 (const :tag "Bottom of screen" nil)
+  :type '(choice (const :tag "Display using overlays" t)
+                 (const :tag "Display in echo area" nil)
                  (const :tag "Both" both))
   :group 'cider
   :package-version '(cider . "0.10.0"))
+
+(defcustom cider-result-overlay-position 'at-eol
+  "Where to display result overlays for inline evaluation and the debugger.
+If 'at-eol, display at the end of the line.
+If 'at-point, display at the end of the respective sexp."
+  :group 'cider
+  :type ''(choice (const :tag "End of line" at-eol)
+                  (const :tag "End of sexp" at-point))
+  :package-version '(cider . "0.22.1"))
 
 (defcustom cider-eval-result-prefix "=> "
   "The prefix displayed in the minibuffer before a result value."
@@ -156,10 +166,9 @@ END is the position where the sexp ends, and defaults to point."
                                             (format (concat " " cider-eval-result-prefix "%s "))
                                             (prepend-face 'cider-result-overlay-face)
                                             &allow-other-keys)
-  "Place an overlay displaying VALUE at the end of line.
+  "Place an overlay displaying VALUE at the position determined by WHERE.
 VALUE is used as the overlay's after-string property, meaning it is
-displayed at the end of the overlay.  The overlay itself is placed from
-beginning to end of current line.
+displayed at the end of the overlay.
 Return nil if the overlay was not placed or if it might not be visible, and
 return the overlay otherwise.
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Can you add some screenshots before and after?

yuhan0:
Before:
image

After:
image

yuhan0:
I realise a simpler change would be modifying the semantics of `cider--make-result-overlay` to treat a single marker as placing the overlay there instead of the EOL. 

That function itself is rather confusing- the docstrings, comments and code don't match up. 

```
  Place an overlay displaying VALUE at the end of line.
The overlay itself is placed from
beginning to end of current line.
```
...
```
        ;; Make sure the overlay is actually at the end of the sexp.
```
...
```
        (let* (beg (if (consp where)
                        (car where)
                      (save-excursion
                        (clojure-backward-logical-sexp 1)
                        (point))))
               (end (if (consp where)
                        (cdr where)
                      (line-end-position)))

**Comment by yuhan0** [comment]
And a related change would be in `cider--debug-display-result-overlay` which would display intermediate results after the respective forms. 
Making the above semantic change to `cider--make-result-overlay` automatically solves this as well.

**Comment by bbatsov** [comment]
I think I just remembered why we did it the way it was - the overlay moving the code to the right seemed kind of weird and we felt that some people would not like it. I'm not sure how other editors are handling inline evaluations for comparison. Obviously we can always make this configurable.

**Comment by bbatsov** [comment]
//cc @Malabarba

**Comment by yuhan0** [comment]
For what it's worth, VS Code's Calva displays results end-of-sexp:
image

whereas Atom's Proto-repl displays it end-of-line:
image


One small advantage of Calva's approach is that you can see several intermediate results on the same line
image
(I don't think Cider currently supports multiple overlays)

**Comment by yuhan0** [comment]
Another ergonomic issue: long lines cause the overlay to be truncated or pushed off screen completely
image
and you end up having to shift things around or use `cider-inspect-last-result`. 
That's especially troublesome in the debugger where editing the buffer causes you to lose the state, and there's no easy way to invoke the inspector on the current value.

This seems more natural:
image

**Comment by bbatsov** [comment]
> (I don't think Cider currently supports multiple overlays)

No, it doesn't. Your points are good, but knowing how users typically react to this type of changes (usually some of them don't like them) I'd really prefer for us to make this a configuration option instead of just changing the default behaviour. 

Thanks for looking into the behaviour of other editors as well!

**Comment by yuhan0** [comment]
Would you like it to be a separate boolean variable like `cider-display-overlays-at-eol` or expand the options to ﻿`cider-use-overlays`?

**Comment by bbatsov** [comment]
A separate one variable, but not boolean - `cider-overlays-position` (or something like this) with values `'at-point` and `'at-eol`. We might come up with other positions down the road (e.g. above/under) and I want to keep our options open.

**Comment by yuhan0** [comment]
Ok, I named it `cider-result-overlay-position` because `overlays` might be ambiguous.

Should the default remain 'at-eol ? I personally think 'at-point is a more sensible default with the option for the user to revert to the previous behaviour.

**Comment by bbatsov** [comment]
Let's keep the default. I might survey the users to check if we can change it. We should mention this configurable behaviour somewhere in the docs as well.

**Comment by Malabarba** [comment]
I would prefer we keep the default too. 
Your report of overlays getting pushed off the side of the window intrigues me though. I could swear we had some code in there that prints the overlay in a newline (below the current line) when that happens.

File: cider-overlays.el [hunk 1]
```diff
@@ -193,14 +201,19 @@ overlay."
...
         (skip-chars-backward "\r\n[:blank:]")
-        (let* ((beg (if (consp where)
+        (let* ((eolp (eq cider-result-overlay-position 'at-eol))
+               (beg (if (consp where)
                         (car where)
-                      (save-excursion
-                        (clojure-backward-logical-sexp 1)
-                        (point))))
+                      (if eolp
+                          (line-beginning-position)
```
Malabarba: Here you're changing the overlay behaviour in the current default.

cider-overlays.el
```diff
@@ -59,17 +59,27 @@ font-locking it."
...
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-result-overlay-position 'at-eol
+  "Where to display result overlays for inline evaluation and the debugger.
+If 'at-eol, display at the end of the line.
+If 'at-point, display at the end of the respective logical sexp."
+  :group 'cider
+  :type ''(choice (const :tag "End of line" at-eol)
+                  (const :tag "End of logical sexp" at-point))
```
Malabarba: We can just call this "End of sexp". I think "logical sexp" is more of an internal implementation detail

cider-overlays.el
yuhan0: Thanks for the review! I was going off the docstring which said "apply the overlay over
  the entire line at that place", but doesn't match up with the implementation. 
I can change it back if that's not appropriate?

bbatsov: I think you should change the docstring instead.

cider-overlays.el
bbatsov: Indeed.

yuhan0:
Ok, made the requested changes and added documentation

bbatsov:
Great work! Thank you! 🙇

cider-overlays.el
```diff
@@ -59,17 +59,27 @@ font-locking it."
...
                  (const :tag "Both" both))
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-result-overlay-position 'at-eol
+  "Where to display result overlays for inline evaluation and the debugger.
+If 'at-eol, display at the end of the line.
+If 'at-point, display at the end of the respective sexp."
+  :group 'cider
+  :type ''(choice (const :tag "End of line" at-eol)
```
Malabarba: Sorry for forgetting to finish the review. And thanks for the great work.

Just one thing, I think this line has an extra `'`

PR #2722, opened 2019-10-03
Expired elpa key

dickmao:
The elpa key expiration has wrought havoc among elisp repos that actually conduct proper testing, proving yet again that no good deed goes unpunished.

The silex images should pre-bake the new key into the various images.  Until then, this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ commands:
   setup:
     steps:
       - checkout
+      - run: make elpa-key
       - run: make elpa
       - run: emacs --version
   test:
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ elpa-key:
 	  exit 1 ; \
 	)
 	-mkdir -p $(shell $(CASK) package-directory)
-	rsync -azSH $(HOME)/.emacs.d/elpa/gnupg $(shell $(CASK) package-directory)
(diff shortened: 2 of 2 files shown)


dpsutton:
@dickmao thank you so much. this would have been so annoying to figure out without your help.

dpsutton:
tests pass on my personal repo. thanks so much!

https://circleci.com/gh/dpsutton/cider/tree/gpg-key-fix

bbatsov:
@dickmao Thanks! //cc @silex

Silex:
@bbatsov @dickmao: thanks for the notification! 

If I understand correctly, I should do https://github.com/clojure-emacs/cider/blob/master/Makefile#L26-L38 in all my images? only some of them?

Some pointers would be useful, I didn't quite follow what happend I only know I had to personally switch to `https` and `26.3` in my config for it to work again, it didn't occur to me that I could fix this in the images themselves.

Silex:
Ok, so far my understanding is that I need to do that for all images below 26.3

bbatsov:
@Silex Yeah, exactly.

dickmao:
The cask part likely won't apply at the silex level, so perhaps just
https://github.com/clojure-emacs/cider/blob/master/Makefile#L26-L36

It's not clear to me how to bake the elpa key such that `cask install` works with zero hassle.  It may not be possible without modifying cask.

Silex:
Please comment on https://github.com/Silex/docker-emacs/pull/52 about if I need to do something else or if there are mistakes.

PR #2725, opened 2019-10-06
Fix tests for figwheel-main init form

dpsutton:
fixing tests for figwheel main init form.

I want to point out one thing that I don't like in the solution for this though: it will error if CIDER can't detect figwheel builds. This happens after jvm startup so it doesn't save any time it just adds a point where we could falsely fail a valid figwheel startup. If there isn't a valid build figwheel will provide a nice error message almost instantaneously after this code is run. I don't see a reason to error here.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -764,9 +764,10 @@ Figwheel for details."
 
 (defun cider--figwheel-main-get-builds ()
   "Extract build names from the <build-id>.cljs.edn config files in the project root."
-  (let ((builds (directory-files (clojure-project-dir) nil ".*\\.cljs\\.edn")))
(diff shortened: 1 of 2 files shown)


bbatsov:
I think the main point of this feature is spare people from having to remember their builds. This seems like a time saver to me. I'm open to prompting people for some manual input if the auto-detection fails, though.

dpsutton:
@bbatsov I like the autocomplete feature. I dislike the user-error when no builds are detected.

bbatsov:
Well, feel free to replace this with my suggestion to just ask for some build if it can't detect any. I just really don't know how likely it's for the current logic to fail in the wild.

PR #2726, opened 2019-10-07
[Fix #2614] Update `cider-compilation-regexp` to match error message in 1.10

### LINKED ISSUE #2614: Red squiggly in source buffer is gone with Clojure 1.10
## Expected behavior
Whenever I do `C-c C-c` with a form that has an error, I get a red squiggly in my source buffer indicating where the error is. 
## Actual behavior
I’m not getting the red squiggly, but I get a rather friendly message in the `*cider-repl..*` buffer telling me that I have an error somewhere in my source file 
Create a project with Clojure 1.10, jack-in, create a form with an error in it, evaluate it with `C-c C-c`. You will not have a red squiggly under the form with error, but you will get a message in the `*cider-repl ...*` buffer

Any recent version of Cider, Clojure 1.10.

## Additional information 
The messages I get in the repl are: for 1.10
```
Syntax error compiling at (core.clj:10:3).
      Unable to resolve symbol: somthing in this context
```
Wheras for 1.9.0 it’s
```
CompilerException java.lang.RuntimeException: Unable to resolve symbol: somthing in this context, compiling:(/Users/erik/Documents/github.com/refactor-nrepl-ns-kw-bug/src/foo/core.clj:10:3)
``` 

So a good starting point would be to look at  this regex here:
https://github.com/clojure-emacs/cider/blob/master/cider-eval.el#L300

slipset:
This is a draft PR to start the discussion of wether we should support Clojure versions prior to 1.10, and if so, how that support should be implemented.


dpsutton:
Is this a proposal to drop support for error messages prior to 1.10?

slipset:
No, not at all, but I need some guidance on how to support both sets of error messages if that is desired.

slipset:
https://no.surveymonkey.com/results/SM-S9JVNXNQV/ shows that as of this years Clojure survey, there are more 1.9 users than 1.10 users, so I guess we need to support error handling for both

bbatsov:
I had totally forgotten about this issue. Thanks for reminding me! I agree that we should support older Clojure versions as well. Probably short term producing a smarter regexp is the way to go, unfortunately that's kind of painful with Emacs's crazy quoting in regexps. I guess we can use the `rx` macro like in this PR https://github.com/clojure-emacs/clojure-mode/pull/545/files

A better solution would probably be to extra the error info from `*e` and just parse the map in Emacs. That way we won't have to deal with regular expressions at all, but it will require a bit more work.

cider-eval.el
```diff
@@ -303,7 +303,8 @@ It delegates the actual error content to the eval or op handler."
     (cider-default-err-eval-handler)))
 
 (defvar cider-compilation-regexp
-  '("\\(?:.*\\(warning, \\)\\|.*?\\(, compiling\\):(\\)\\(.*?\\):\\([[:digit:]]+\\)\\(?::\\([[:digit:]]+\\)\\)?\\(\\(?: - \\(.*\\)\\)\\|)\\)" 3 4 5 (1))
+  '("Syntax error compiling at (\\(.*?\\):\\([[:digit:]]+\\):\\([[:digit:]]+\\))"
```
bbatsov: Btw, the regexp in its current format is ignoring warnings.

slipset: Crap. The only warning I can think of is reflection-warnings, and of course it follows the original pattern :/
```
user> (set! *warn-on-reflection* true)
;; => true
user> (def lol "3")
;; => #'user/lol
user> (Integer. lol)
Reflection warning, *cider-repl ardoq.com/ardoq-api:localhost:60882(clj)*:48550:7 - call to java.lang.Integer ctor can't be resolved.
```
So a mix of the regexes it will be then :(

slipset: Also, the regex is flawed, as Clojure reports differently for different errors:
```
Syntax error (IllegalStateException) compiling at (test/ardoq/security_test.clj:1:1).
Alias json already exists in namespace ardoq.security-test, aliasing clojure.data.json
```

slipset: So 
```
'("Syntax error .*? compiling at (\\(.*?\\):\\([[:digit:]]+\\):\\([[:digit:]]+\\))"
				 1 2 3)
``` 
catches that as well.

bbatsov: Btw, also keep in mind that the resulting regexp needs to have 4 components (the final one being the type - error or warning), as that's what the function doing the matching expects. 

As for warnings - there are more, but I think they have more or less the same format http://insideclojure.org/2014/12/15/warn-on-boxed/

bbatsov:
Btw, I was planning to cut a CIDER release today, but I can wait a bit if you plan to update this regexp today/tomorrow.

slipset:
I don't think you should wait for this. 
I'm running it now locally (with `.dir-locals.el`, thanks for that hint :))` and figuring out more and more cases I've missed :/

bbatsov:
Fine by me.

slipset:
@bbatsov I think this should cover both 1.9 and 1.10 (and be readable and extendable)

bbatsov:
Yeah, that looks nice. Btw, why didn't you put the new tests in the existing error-parsing-tests file?

cider-eval.el
```diff
@@ -302,8 +302,47 @@ It delegates the actual error content to the eval or op handler."
...
+                          (zero-or-more anything)
+                          ", compiling:("
+                          (group-n 2 (minimal-match (zero-or-more anything)))
+                          ":"
+                          (group-n 3 (one-or-more digit))
+                          ":"
+                          (group-n 4 (one-or-more digit))
+                          ")"))
+
+(setq clojure-1.9-warning `(sequence
```
bbatsov: If that's the same on all versions I guess the 1.9 is not needed in it's name.

cider-eval.el
```diff
@@ -302,8 +302,47 @@ It delegates the actual error content to the eval or op handler."
       (cider-default-err-op-handler)
     (cider-default-err-eval-handler)))
 
+(setq clojure-1.10-error `(sequence
```
bbatsov: I'd use `defconst` instead of `setq`, otherwise we'd get compilation warnings.

bbatsov: Also - all vars should be prefixed with `cider-`.

cider-eval.el
```diff
@@ -302,8 +302,47 @@ It delegates the actual error content to the eval or op handler."
...
+                            (group-n 1 "warning")
+                            ", "
+                            (group-n 2 (minimal-match (zero-or-more anything)))
+                            ":"
+                            (group-n 3 (one-or-more digit))
+                            ":"
+                            (group-n 4 (one-or-more digit))
+                            " - "))
+
+(setq clojure-compilation-regexp (rx bol (or (eval clojure-1.9-error)
```
bbatsov: I'd probably just fold this into the defvar that follows.

slipset: I chose not to, as it is easier to write tests for this now (at least with my very limited understanding of Elisp)

test/cider-error-parsing-tests.el
```diff
@@ -112,3 +112,30 @@
       (expect (line-num info) :to-equal 24)
       (expect (col-num info) :to-equal 43)
       (expect (face info) :to-equal 'cider-warning-highlight-face))))
+
+
+(setq clojure-compiler-warning "Reflection warning, /tmp/foo/src/foo/core.clj:14:1 - call to java.lang.Integer ctor can't be resolved.")
```
bbatsov: Here you're setting some non-existing var. Use `defvar` instead. Again - it's a bad idea not prefix this, as it might clash with something in `clojure-mode` down the road. Don't forget that Elisp doesn't have namespaces.

bbatsov:
🚀

PR #2727, opened 2019-10-09
Avoid hitting keyservers.net if key already in cache

dickmao:
Check we already have the desired elpa key before spamming keyservers.net.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,8 @@ elpa-key:
 	    if gpg -q --homedir $(HOME)/.emacs.d/elpa/gnupg -k | grep 81E42C40 ; then \
 	      exit 0 ; \
 	    fi ; \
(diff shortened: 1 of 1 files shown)


bbatsov:
Where exactly is the check? I just see that the sleep was made conditional now.

dickmao:
Before PR: 1. hit keyserver 2. check if key retrieved 3. if not, wait before retrying
After PR: 1. check if key retrieved 2. if not, wait if this is reattempt 3. hit keyserver

bbatsov:
Ah, got it. The way the loop's written I got confused. I'm too old for bash scripts. :D Guess similar update should be made to the docker images. //cc @Silex

Silex:
@bbatsov: fixed by https://github.com/Silex/docker-emacs/commit/36989a8f06483d0f3455d36028200ae5dc3b78cd

PR #2729, opened 2019-10-12
Add cider-inspector-def-current-val to cider-inspector

jpmonettas:
Add `cider-inspector-def-current-val` command to cider inspector so you can define a var with the inspector current value.
Also binds `d` in the inspector mode map to this command and adds "Define var" to the inspector menu.

Depends on https://github.com/clojure-emacs/cider-nrepl/pull/653/ 

-----------------


bbatsov:
Nicely done! Thanks! 🙇

PR #2730, opened 2019-10-12
Include ns when requiring repl utils

dpsutton:
Requiring repl utils did not send the current ns, so they were always required into the user namespace, not into the current ns.

```lisp
(-->
  id                       "55"
  op                       "eval"
  session                  "4ecbc22b-1107-4070-8c15-8659413e1479"
  time-stamp               "2019-10-12 09:26:43.125825653"
  code                     "(clojure.core/apply clojure.core/require clojure.main/repl-r..."
  inhibit-cider-middleware "true"
)
(<--
  id         "55"
  session    "4ecbc22b-1107-4070-8c15-8659413e1479"
  time-stamp "2019-10-12 09:26:43.130607785"
  ns         "user"    ;; note that these were evaled in user since no ns was supplied
  value      "nil"
)
(<--
  id         "55"
  session    "4ecbc22b-1107-4070-8c15-8659413e1479"
  time-stamp "2019-10-12 09:26:43.131230533"
  status     ("done")
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+* [#2730](https://github.com/clojure-emacs/cider/pull/2730) Require repl utilities into current namespace not just user ns
+
 ## 0.23.0 (2019-10-08)
 
 ### New features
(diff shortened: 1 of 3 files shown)


bbatsov:
Great catch! Thanks!

PR #2734, opened 2019-10-20
Remove all readthedocs pages and setup redirects to docs.cider.mx

### LINKED ISSUE #2695: Apply some SEO to cider.mx
The big problem is that right now Google search thinks that the domain should target just Mexico users and I couldn't find a way to tell it to target international users as well. As a result the official site can't discovered at all, but you'll get a million references to the legacy cider.readthedocs.io. 

I'd really appreciate some help with this, as I know next to nothing about SEO. Anyways, I did learn a valuable lesson - fun domain names might cause problems for you down the road. :) If we can't figure out what to do with the current domain I'll probably move cider under some more traditional domain like `.org` or `.net`.

danielcompton:
Once this is merged, we will need to:

- [ ] rebuild `latest` and `stable`. This _should_ delete all existing
HTML sources and cause every request to hit the RTD servers. Then those
requests will get a 302 redirect to their new counterpart on docs.cider.mx.
- [ ] request RTD staff to change our 302 temporary redirects to 301
redirects. This seems like a [manual process](https://github.com/readthedocs/readthedocs.org/pull/5727#pullrequestreview-241998379)
  > Users can request a 301 redirect via our support queue and we just need to make a change on those redirects after they set them up and are happy. 

Fixes #2695.

-----------------

- N/A You've added tests (if possible) to cover your change(s)
- N/A All tests are passing (`make test`)
- N/A All code passes the linter (`make lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes
- N/A You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)
- N/A You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)


danielcompton:
I've rebuilt the latest and stable docs. I had to quickly create a tag 0.23.1 to rebuild stable, which I've now removed.

PR #2735, opened 2019-10-21
Add P for inspect-prompt debugger command

yuhan0:
See https://github.com/clojure-emacs/cider-nrepl/pull/655

Also fixes a few references to :continue-all that were missed in the last PR.

Tests appear to fail locally in unrelated parts of Cider, it looks like a recent change introduced a dependency on `s.el` without explicitly requiring it in the Caskfile:

```
cider-extract-error-info extracts correct information from the error message

  (buttercup-expect (lambda nil (quote (file-name info)) (file-name info)) :...
  (buttercup-fail "%s" "Expected `(file-name info)' to be `equal' to `\"/som...
  (signal buttercup-failed "Expected `(file-name info)' to be `equal' to `\"...
FAILED: Expected `(file-name info)' to be `equal' to `"/some/test/file/core.clj"', but instead it was `nil' which does not match because: (different-types nil "/some/test/file/core.clj").

The cider compilation regex Recognizes a clojure warning message

  (buttercup-expect (lambda nil (quote (s-matches-p cider-clojure-compilatio...
  (buttercup--apply-matcher :to-equal ((lambda nil (quote (s-matches-p cider...
  (apply #[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\3...
  (#[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\305GD\...
  (mapcar buttercup--expr-and-value ((lambda nil (quote (s-matches-p cider-c...
  (buttercup--expr-and-value (lambda nil (quote (s-matches-p cider-clojure-c...
  ((lambda nil (quote (s-matches-p cider-clojure-compilation-regexp clojure-...
  (s-matches-p cider-clojure-compilation-regexp clojure-compiler-warning)
error: (void-function s-matches-p)


The cider compilation regex Recognizes a clojure-1.9 error message

  (buttercup-expect (lambda nil (quote (s-matches-p cider-clojure-compilatio...
  (buttercup--apply-matcher :to-equal ((lambda nil (quote (s-matches-p cider...
  (apply #[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\3...
  (#[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\305GD\...
  (mapcar buttercup--expr-and-value ((lambda nil (quote (s-matches-p cider-c...
  (buttercup--expr-and-value (lambda nil (quote (s-matches-p cider-clojure-c...
  ((lambda nil (quote (s-matches-p cider-clojure-compilation-regexp clojure-...
  (s-matches-p cider-clojure-compilation-regexp clojure-1\.9-compiler-error)
error: (void-function s-matches-p)


The cider compilation regex Recognizes a clojure-1.10 error message

  (buttercup-expect (lambda nil (quote (s-matches-p cider-clojure-compilatio...
  (buttercup--apply-matcher :to-equal ((lambda nil (quote (s-matches-p cider...
  (apply #[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\3...
  (#[514 "\300\301D\"\211G\302U\203\211A\262\242\202\303\304\305GD\...
  (mapcar buttercup--expr-and-value ((lambda nil (quote (s-matches-p cider-c...
  (buttercup--expr-and-value (lambda nil (quote (s-matches-p cider-clojure-c...
  ((lambda nil (quote (s-matches-p cider-clojure-compilation-regexp clojure-...
  (s-matches-p cider-clojure-compilation-regexp clojure-1\.10-compiler-error)
error: (void-function s-matches-p)


Ran 293 specs, 4 failed, in 2.4 seconds.
```

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2735](https://github.com/clojure-emacs/cider/pull/2735): New debugger command `P` to inspect an arbitrary expression, it was previously bound to `p` which now inspects the current value.
 * [#2729](https://github.com/clojure-emacs/cider/pull/2729): New cider inspector command `cider-inspector-def-current-val` lets you define a var with the current inspector value.
 
 ### Bugs fixed
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -205,6 +205,7 @@ Can be toggled at any time with `\\[cider-debug-toggle-locals]'."
     (?h "here" "here")
     (?e "eval" "eval")
     (?p "inspect" "inspect")
+    (?P "inspect-prompt" nil)
     (?l "locals" "locals")
     (?j "inject" "inject")
     (?s "stacktrace" "stacktrace")
@@ -405,13 +406,14 @@ In order to work properly, this mode must be activated by
   `("CIDER Debugger"
     ["Next step" (cider-debug-mode-send-reply ":next") :keys "n"]
     ["Continue" (cider-debug-mode-send-reply ":continue") :keys "c"]
-    ["Continue non-stop" (cider-debug-mode-send-reply ":Continue") :keys "C"]
+    ["Continue non-stop" (cider-debug-mode-send-reply ":continue-all") :keys "C"]
     ["Move out of sexp" (cider-debug-mode-send-reply ":out") :keys "o"]
     ["Quit" (cider-debug-mode-send-reply ":quit") :keys "q"]
     "--"
     ["Evaluate in current scope" (cider-debug-mode-send-reply ":eval") :keys "e"]
     ["Inject value" (cider-debug-mode-send-reply ":inject") :keys "i"]
-    ["Inspect value" (cider-debug-mode-send-reply ":inspect")]
+    ["Inspect current value" (cider-debug-mode-send-reply ":inspect") :keys "p"]
+    ["Inspect expression" (cider-debug-mode-send-reply ":inspect-prompt") :keys "P"]
     ["Inspect local variables" (cider-debug-mode-send-reply ":locals") :keys "l"]
     "--"
     ("Configure keys prompt"
--- a/doc/modules/ROOT/pages/debugging/debugger.adoc
+++ b/doc/modules/ROOT/pages/debugging/debugger.adoc
@@ -95,13 +95,19 @@ keys, although there are some differences.
 | Force-step to "`here`"
 
 | kbd:[c]
-| Continue without stopping
+| Continue without stopping in current breakpoint
+
+| kbd:[C]
+| Continue without stopping for all breakpoints
 
 | kbd:[e]
 | Eval code in current context
 
 | kbd:[p]
-| Inspect a value
+| Inspect the current value
+
+| kbd:[P]
+| Inspect an arbitrary expression
 
 | kbd:[l]
 | Inspect local variables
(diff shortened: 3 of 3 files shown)


bbatsov:
> Tests appear to fail locally in unrelated parts of Cider, it looks like a recent change introduced a dependency on s.el without explicitly requiring it in the Caskfile:

This was a mistake. Those usages of `s.el` should just be replaced with the built-in Emacs function. I guess I didn't notice this when I merged the PR in question.

bbatsov:
@yuhan0 Btw, I've noticed that the `here` commands seem to be missing from the debugger menu. I guess we should address this as well.

yuhan0:
It appears fine to me:

image

Although strangely, `here` doesn't appear to work if I precede the sexp with a #dbg, only when it's inserted by the middleware using `C-u C-M-x`. I'm never really used the `here` command before, so not sure if that is a regression.

bbatsov:
Sounds like a bug. I've also noticed there is `here` and `Here` in the docs.

But I was actually referring to the mode menu that's displayed in the menu bar, not the inline menu.

yuhan0:
Oh okay, I've never actually used the menu-bar before :) 

I don't quite understand the difference between `here` and `Here`, or the other uppercase variants like `o` vs `O`? On the few tests I've tried them on they behave exactly the same.

bbatsov:
@Malabarba do you remember the differences?

Malabarba:
The capital H version sends a "force" argument. The behaviour of that argument is somewhat described on the `skip-break?` documentation: https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/debug.clj#L79

>  For :deeper and :before, if we are not in the same code (i.e. we have stepped
>  into another instrumented function and code argument doesn't match old code in
>  *skip-breaks*), then return the value of `force?`.

In this, “:before” means the “here” command.

So my understanding is that `Here` avoids going into other functions while trying to reach the point of interest.

yuhan0:
Aha, so it has to be a different function context that it skips over - I was trying examples like the following, and found that both `h` and `H` would skip intermediate debugged forms.

```clojure
(+ #dbg (* 1 2)
   #dbg (* 3 4)
   #dbg (* 5 6)
   #dbg (* 7 8))
```

From what I understand, this illustrates the difference:
(instrument both functions, eval `(outer 5)`, navigate to the (* 2) form and press `h`/`H`) 
```clojure
(defn inner [x]
  (inc x))

(defn outer [x]
  (+ (inner x)
     (* 2 x)))

(outer 5)
```

Also, `O` now throws an error, I think it might be a regression due to the new commands API.
>=> Execution error at cider.nrepl.middleware.debug/read-debug-input (debug.clj:242).
Invalid token: :


My concern is if this uppercase `force?` behaviour is complected with the middleware implementation of `skip-breaks` that makes it difficult to document, or if it's somewhat of an obscure distinction. 

Wouldn't the less surprising default be _not_ jumping into another function context when the user specifies "jump here" or "jump out of this form"?

bbatsov:
I'm inclined to agree with you. Might be useful to track the history of the code changes that lead to the introduction of `H`. 

A good extra point is that we need to documented all of this in the user manual as well. :-)

Malabarba:
My guess is that, when we noticed this "stopping on other functions" behaviour we just fixed by adding a new key instead of incrementing the `h` functionality.
I'm fine with changing the default, as our `H` is actually the equivalent of `h` in edebug.

PR #2738, opened 2019-10-30
Allow lookup of function symbol when cursor is on opening paren

yuhan0:
A simple DWIM addition to allow for calling `cider-doc` and related commands on a symbol when the cursor is at its opening paren. (This is a common state to be in when using structure based navigation)

Eg. with the cursor at `|`, calling M-x cider-doc should bring up documentation for `g`
```clojure
(f |(g x))
```

-----------------

### DIFF
--- a/test/cider-util-tests.el
+++ b/test/cider-util-tests.el
@@ -98,6 +98,13 @@
       (spy-on 'thing-at-point :and-return-value nil)
       (expect (cider-symbol-at-point) :not :to-be-truthy)))
 
+  (describe "when on an opening paren"
+    (it "returns the following symbol"
+      (with-temp-buffer
+        (insert "(some function call)")
+        (goto-char (point-min))
+        (expect (cider-symbol-at-point 'look-back) :to-equal "some"))))
+
   (it "can identify symbols in a repl, ignoring the repl prompt"
     ;; ignores repl prompts
     (spy-on 'thing-at-point :and-return-value (propertize "user>" 'field 'cider-repl-prompt))
--- a/test/cider-util-tests.el
+++ b/test/cider-util-tests.el
@@ -30,6 +30,21 @@
 (require 'buttercup)
 (require 'cider-util)
 
+(defmacro with-clojure-buffer (contents &rest body)
+  "Execute BODY in a clojure-mode buffer with CONTENTS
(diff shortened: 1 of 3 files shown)


bbatsov:
Fine by me. Just add one unit test as well.

yuhan0:
Alright, added :)

test/cider-util-tests.el
```diff
@@ -97,6 +97,13 @@
       (spy-on 'thing-at-point :and-return-value nil)
       (expect (cider-symbol-at-point) :not :to-be-truthy)))
 
+  (describe "when on an opening paren"
+    (it "returns the following symbol"
+      (with-temp-buffer
```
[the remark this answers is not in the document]
yuhan0: I was just going by the other `cider-symbol-at-point` tests above which don't enable clojure-mode either. Maybe because it's a text-based util that doesn't need clojure-mode?

test/cider-util-tests.el
```diff
@@ -30,6 +30,22 @@
 (require 'buttercup)
 (require 'cider-util)
 
+(defmacro with-clojure-buffer (contents &rest body)
+  "Execute BODY in a clojure-mode with CONTENTS
+
+CONTENTS can be a single string to be inserted before the cursor, or a cons
+cell of strings (BEFORE . AFTER) where the cursor is placed in between
```
bbatsov: Might be better to directly support having  an (optional) `|` in the string and moving the point there instead of using a cons cell. The examples would read better IMO.

yuhan0: Okay, changed

yuhan0: Oh, I noticed something similar exists in clojure-mode's test suites, is it appropriate for the same macro name to be here as well? 
I suppose they don't conflict and there's no way to reuse a test namespace from another package?

https://github.com/clojure-emacs/clojure-mode/blob/147bf84189ef672161b2229ddc5c35713323ad59/test/utils/test-helper.el#L35

bbatsov: Yeah, that's what I had in mind. Unfortunately the test stuff can't be shared, so they have to be copied.

bbatsov:
Thanks!

PR #2739, opened 2019-10-30
Start `node-repl` and `browser-repl` shadow builds correctly

dpsutton:
Shadow can start your build with (shadow/watch
build) (shadow/nrepl-select build). But it provides two conveient
default builds on its own which are called in a different
manner: (shadow/node-repl) or (shadow/browser-repl). Need to invoke
correctly as right now we just get an error that there is no build
with id :browser-repl

tests are failing from unrelated changes :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
+* [#2739](https://github.com/clojure-emacs/cider/pull/2739): Start built-in shadow-cljs build profiles correctly (node-repl, browser-repl)
 * [#2730](https://github.com/clojure-emacs/cider/pull/2730): Require repl utilities into current namespace not just `user` ns.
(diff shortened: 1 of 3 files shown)


dpsutton:
1. `cider-shadow-select-cljs-init-form` does not need to be updated. 
The hard-coded builds can be `nrepl-select`ed but not `watched`. The shadow-select assumes something is running those builds and simply connects to them.

2. I'll add some unit tests now

3. Does the current build detection logic handle them properly
Yes with a caveat. These are hard-coded in shadow and always present. Our build detection appends them to the builds correctly. One divergence is that these builds are present even if there is no shadow-cljs.edn project file. This is a divergence that works for us because CIDER won't even know to use the shadow startup command without it so its kinda moot for our purposes.

dpsutton:
once https://github.com/clojure-emacs/cider/pull/2741 is merged i'll rebase and we should have a clean green CI machine

dpsutton:
rebased and green. i guess i can add a changelog real quick

bbatsov:
Thanks! 🙇

PR #2741, opened 2019-11-03
Tests/regex tests

dpsutton:
1. Fix tests for error message parsing: columns weren't treated as optional in the regex despite tests showing they were optional.
2. various doc fixups, punctuation, aligning to left of buffer, and marking bindings as unused.

One thing to note: in emacs 27 the linting is getting more strict when i run it locally. Its asking for doc strings on the generic functions from sesman and also a warning relating to seq functions:

> nrepl-client.el:1251:42:Error: ‘seq-contains’ is an obsolete function (as of 27.1); use ‘seq-contains-p’ instead.

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -212,16 +212,13 @@ Can be toggled at any time with `\\[cider-debug-toggle-locals]'."
     (?t "trace" "trace")
     (?q "quit" "quit"))
   "A list of debugger command specs.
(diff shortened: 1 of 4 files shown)


bbatsov:
I just realized that some reason the build status is no longer displayed on PRs, although it seems the builds are still running. Seems this has been causing constant CI regressions lately.

dpsutton:
It was running fine for me on my branch. I saw that a bit ago and I thought it was not running PRs from non-members or something.

PR #2744, opened 2019-11-08
Show startup commands in repl buffer

dpsutton:
Simple change to show the commands CIDER uses to get everything running to the user in the repl buffer.

image

There are a few good reasons to do this:
- we show the jack in command but its as a message and can quickly go away
- we never show the cljs command so people might forget how they are actually starting the cljs repl when CIDER doesn't display it
- it removes the "magic" of what CIDER does for beginners and trouble shooters.
- it gives someone access to the commands if they prefer to `cider-connect` instead of having clojure running as a subprocess to emacs.

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2744](https://github.com/clojure-emacs/cider/pull/2744): Add startup commands to repl banner.
 * [#2499](https://github.com/clojure-emacs/cider/issues/2499): Add `cider-jump-to-pop-to-buffer-actions`.
 * [#2738](https://github.com/clojure-emacs/cider/pull/2738): Add ability to lookup a function symbol when cursor is at the opening paren.
 * [#2735](https://github.com/clojure-emacs/cider/pull/2735): New debugger command `P` to inspect an arbitrary expression, it was previously bound to `p` which now inspects the current value.
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -699,6 +699,7 @@ PARAMS is a plist as received by `cider-repl-create'."
 (declare-function cider-repl-reset-markers "cider-repl")
 (defvar-local cider-session-name nil)
 (defvar-local cider-repl-init-function nil)
+(defvar-local cider-launch-params nil)
 (defun cider-repl-create (params)
   "Create new repl buffer.
 PARAMS is a plist which contains :repl-type, :host, :port, :project-dir,
@@ -727,7 +728,8 @@ function with the repl buffer set as current."
             ;; REPLs start with clj and then "upgrade" to a different type
             cider-repl-type (plist-get params :repl-type)
             ;; ran at the end of cider--connected-handler
-            cider-repl-init-function (plist-get params :repl-init-function))
+            cider-repl-init-function (plist-get params :repl-init-function)
+            cider-launch-params params)
       (cider-repl-reset-markers)
       (add-hook 'nrepl-response-handler-functions #'cider-repl--state-handler nil 'local)
       (add-hook 'nrepl-connected-hook #'cider--connected-handler nil 'local)
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -319,6 +319,7 @@ fully initialized."
     ((pred identity) (pop-to-buffer buffer)))
   (with-current-buffer buffer
     (cider-repl--insert-banner)
+    (cider-repl--insert-startup-commands)
     (when-let* ((window (get-buffer-window buffer t)))
       (with-selected-window window
         (recenter (- -1 scroll-margin))))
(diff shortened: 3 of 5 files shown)


bbatsov:
Nicely done! I think, however,  it's a bit misleading that the shell command that boots nREPL and the Clojure form that changes a REPL type are shown in the same way, because they clearly aren't. It also seems to me that this will work only for REPLs which came from cider-jack-in, so some command to describe a REPL might be useful as well. And perhaps we should show the ClojureScript version in use somewhere. Somehow I never thought about this.

dpsutton:
> It also seems to me that this will work only for REPLs which came from cider-jack-in

I was envisioning making sure it was clear how a repl is started. If someone is not using jack-in commands then it is already clear to them how it was started.

> And perhaps we should show the ClojureScript version in use somewhere

I don't follow this. I'm showing the clojurescript startup command. Or do you mean the cljs repl type? Like shadow or figwheel etc?

bbatsov:
I was thinking of:

```
ClojureScript: 1.10.x-fdfd (this can be different from the Clojure version)
ClojureScript REPL flavour: ...
ClojureScript REPL init form: ...
```

Or something along those lines. Similar to the info you get on top of the REPL. I guess it's easy to infer what's the flavour based on the form, but it might look nicer for the users. 

> I was envisioning making sure it was clear how a repl is started. If someone is not using jack-in commands then it is already clear to them how it was started.

Fair point. I guess the ClojureScript info, however, can always be displayed as we always upgrade a Clojure REPL and you can't really connect to a ClojureScript session.

dpsutton:
image

I've added the cljs repl type and done a bit of formatting so that its grouped. The only thing that is missing would be perhaps the clojurescript version and i guess the node version if running on node. I think these present a pretty annoying problem.

1. The other information appears at the top of the repl. The clojurescript version can't easily be gotten right now. nrepl provides some info in the describe op for clojure and nrepl, and cider-nrepl adds the cider-nrepl version. I see there's `cljs.util/*clojurescript-version*` that can be looked at but without nrepl support it will come back as edn rather than an nrepl dict.

bbatsov:
I guess we can have either the `cider-nrepl` version middleware return them on use the aux attribute of nREPL's own `describe`. There we can stuff whatever data we want to quite easily.

bbatsov:
@dpsutton One more thing - I notice that something's weird with the REPL prompt in your example. First there's the `shadow.user` prompt that's not on its own line, then the prompt changes to `cljs.user`. I guess you might have changed this explicitly, but I wanted to make sure the init is not messed up somehow.

cider-repl.el
```diff
@@ -333,6 +334,28 @@ fully initialized."
     (insert-before-markers
      (propertize (cider-repl--help-banner) 'font-lock-face 'font-lock-comment-face))))
 
+(defun cider-repl--insert-startup-commands ()
```
bbatsov: I thought a bit more about this and I'd prefer to split this in two - the jack-in command that gets displayed in the REPL banner if present and the ClojureScript details which can be placed/displayed as you've suggested. I just feel that the jack-in command is more fundamental than the ClojureScript form, that's why it should be listed more prominently. Potentially we can add there some info about the build system down the road - e.g. Leiningen 2.9, etc.

bbatsov:
@dpsutton ping :-)

dpsutton:
image

If I understand correctly you mean something like this? The startup command is in the banner and the cljs stuff is lower and more in-line?

bbatsov:
Yeah, something like this. We just need some fallback for the jack-in command in case it's not present. Probably we can just move it to the very top and prepend it to the rest of the banner if needed.

cider-repl.el
```diff
@@ -333,6 +335,26 @@ fully initialized."
     (insert-before-markers
      (propertize (cider-repl--help-banner) 'font-lock-face 'font-lock-comment-face))))
 
+(defun cider-repl--insert-param-values (param-tuples)
```
bbatsov: We don't call this often, so I'd go with explicit params here. I think it's going to read much nicer.

dpsutton:
@bbatsov this should be what you are looking for if i understand you correctly. The startup command is now in the banner along with the other info and the cljs stuff is outside and grouped together

dpsutton:
image

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### New features
 
+* [#2744](https://github.com/clojure-emacs/cider/pull/2744): Add startup commands to repl banner.
 * [#2499](https://github.com/clojure-emacs/cider/issues/2499): Add `cider-jump-to-pop-to-buffer-actions`.
+* Show clj and cljs startup commands in repl buffer
```
bbatsov: You've duplicated this in the changelog.

test/cider-repl-tests.el
```diff
@@ -38,7 +70,8 @@
     (spy-on 'cider--nrepl-version :and-return-value "0.5.3")
     (setq nrepl-endpoint (list :host "localhost" :port "54018"))
     (setq cider-version "0.12.0")
-    (setq cider-codename "Seattle"))
+    (setq cider-codename "Seattle")
+    (setq cider-saved-params (list :jack-in-cmd "startup command")))
```
bbatsov: I think you're not using this.

cider-connection.el
```diff
@@ -699,6 +699,7 @@ PARAMS is a plist as received by `cider-repl-create'."
 (declare-function cider-repl-reset-markers "cider-repl")
 (defvar-local cider-session-name nil)
 (defvar-local cider-repl-init-function nil)
+(defvar-local cider-saved-params nil)
```
bbatsov: I'd name this `cider-launch-params` or something like this.

cider-repl.el
```diff
@@ -333,6 +334,32 @@ fully initialized."
...
+    (let ((jack-in-command (plist-get cider-saved-params :jack-in-cmd))
+          (cljs-repl-type (plist-get cider-saved-params :cljs-repl-type))
+          (cljs-init-form (plist-get cider-saved-params :repl-init-form)))
+      (when jack-in-command
+        ;; spaces to align with the banner
+        (emit-comment (concat "  Startup: " jack-in-command)))
+      (when (or cljs-repl-type cljs-init-form)
+        (emit-comment "")
+        (when cljs-repl-type
+          (emit-comment (concat "cljs repl type: " (symbol-name cljs-repl-type))))
```
bbatsov: Let's use `ClojureScript REPL type:` here.

cider-repl.el
```diff
@@ -333,6 +334,32 @@ fully initialized."
...
+        (when cljs-init-form
+          (emit-comment (concat "cljs repl startup command: " cljs-init-form)))
```
bbatsov: `ClojureScript REPL init form:`

bbatsov:
@dpsutton Looks good. I've added a couple of tiny cosmetic remarks and we're good to go.

bbatsov:
Thanks for working on this! 🙇

dpsutton:
Thanks for the thorough checking of the code! much appreciated

Issue #2748, opened 2019-11-15
Evaluation indicators too subtle?


jr0cket:
The evaluation indicators in the fringe do not stand out very clearly and many people are unaware of this very useful feedback because they do not notice them.

I have found the evaluation indicators in the fringe particularly difficult to spot using either of the Spacemacs themes.  They are almost invisible on the spacemacs-light theme

Currently, `lightgreen` is used with light themes and `darkgreen` is used for dark themes.  Could these values be swapped?  That may make the indicators more obvious.

https://github.com/clojure-emacs/cider/blob/master/cider-overlays.el#L132

I would be happy to test out alternatives if there is a way to set the foreground colour, preferably setting that in the `init.el` or `.spacemacs` file than creating a fork of the CIDER package.

jr0cket:
Using `helm-faces`, selecting `cider-fringe-good-face` and clicking customize button seems a good way to experiment, as colours can be applied immediately.

With spacemacs-light and spacemacs-dark themes, the following stand out more.

`LimeGreen` #32cd32 works well on a dark background
`DarkGreen` #006400 works well on a light background

dpsutton:
can you throw up some screenshots? ideally with light and dark modes?

jr0cket:
Tested on the following themes with positive results, although light themes still seem a little faint
spacemacs-dark, spacemacs-light, adwaita, dichromancy, leuven, whiteboard.

Will add some screenshots of the changes soon.  In the mean time you can see how faint the existing colour is in this video https://youtu.be/o4Q-K3lzBRY?t=2132

Thank you.

jr0cket:
Spacemacs light and dark themes with suggested values for fringes

![spacemacs-dark-evaluation-indicators](https://user-images.githubusercontent.com/250870/68954898-d2cb3400-07bc-11ea-956c-38c4fe261e81.png)
![spacemacs-light-evaluation-indicators](https://user-images.githubusercontent.com/250870/68954899-d363ca80-07bc-11ea-9d73-9ab133b41ac9.png)

jr0cket:
An alternative would be to use a solid bar as a marker, or thicker lines, rather than very thin lines which blend into the background more readily.

bbatsov:
How would this look? 

Generally we wanted this to be something subtle and non-intrusive when the feature was introduced. I'm thinking that probably we never documented it anywhere, so I doubt that many people ever realized what those markers are about. :-)

jr0cket:
[...]

PR #2749, opened 2019-11-15
Fringe evaluation indicators colours more visible

jr0cket:
Using dark green for a light theme and lime green for a dark theme increases the visibility of the evaluation indicators in the fringe.

https://github.com/clojure-emacs/cider/issues/2748

Tested on the following themes with positive results
spacemacs-dark, spacemacs-light, adwaita, dichromancy, leuven, whiteboard.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -130,8 +130,8 @@ This function also removes itself from `post-command-hook'."
   (add-hook 'post-command-hook #'cider--remove-result-overlay nil 'local))
 
 (defface cider-fringe-good-face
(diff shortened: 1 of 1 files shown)


bbatsov:
Can you add some screenshots here of the state before and after? In general it seems to me that each theme should handle the optimal colours directly.

jr0cket:
Agreed that its more relevant to make a change in the specific themes.  I will create a PR for the spacemacs themes I use. 

Thank you

PR #2751, opened 2019-11-18
Clearer fringe evaluation indicator

### LINKED ISSUE #2748: Evaluation indicators too subtle?

jcaw:
Closes #2748

# Overview

Issue #2748 notes that the evaluation indicator in the fringe is very subtle. This PR makes three changes to address this, split into 3 commits:

1. Extract the bitmap displayed in the fringe into a user-configurable `custom`.
2. Inherit the indicator's color from the current theme, rather than hard-coding it.
3. Change the default bitmap used to `filled-rectangle`.

This is just an initial stab at changing the indicators, to prove the concept. Happy to rewrite as necessary. I haven't cleared the checklist yet - wanted to hear what you guys thought before formally cleaning things up.

# Current Look

The indicator is difficult to see in certain themes. Colors are hard-coded, which can result in an inconsistent palette. The bitmap itself creates a semi-transparent effect. Some examples:

before-spacemacs-dark
`spacemacs-dark`

before-spacemacs-light
`spacemacs-light`

before-nord
`doom-nord`

# Different Bitmap

Swapping to a different bitmap makes things a little bolder. I had a look through all the built-in bitmaps, in my opinion `filled-rectangle` looks best.

symbol-only-spacemacs-dark
`spacemacs-dark`

symbol-only-spacemacs-light
`spacemacs-light`

symbol-only-nord
`doom-nord`

# Dynamic Colors

Rather than expecting each theme to explicitly override the fringe colors, it would be better if we could dynamically adjust the color to match each theme. `diff-attach` is a built-in face which is almost always green, and should match the palette of the theme.

Inheriting the foreground from this face works across a wide range of themes:

both-spacemacs-dark
`spacemacs-dark`

both-spacemacs-light
`spacemacs-light`

both-nord
`doom-nord`

both-zenburn
`zenburn`

both-leuven
`leuven`

both-doom-one
`doom-one`

both-monokai
`monokai`

both-solarized-dark
`solarized-dark`

both-solarized-light
`solarized-light`

Many themes also set the background for `diff-attach`. Inheriting just the foreground reduces visual clutter, and seems to work consistently. We can make the operation safer with `:distant-foreground`, to defer to a hard-coded green if the diff foreground is not sufficiently visible.

## Scaling

An issue with `filled-rectangle` is it doesn't scale vertically. It seems the proportions are locked and the fringe is pegged to a specific pixel width. The original bitmap doesn't have this issue. For example:

large-text

I'm not sure if defining a new bitmap would allow us to adjust this property. Another solution might also be to adjust the background color of the fringe directly, rather than display a bitmap.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -135,9 +135,23 @@ This function also removes itself from `post-command-hook'."
   "Face used on the fringe indicator for successful evaluation."
   :group 'cider)
 
-(defconst cider--fringe-overlay-good
-  (propertize " " 'display '(left-fringe empty-line cider-fringe-good-face))
-  "The before-string property that adds a green indicator on the fringe.")
+(defcustom cider-fringe-bitmap 'empty-line
+  "Bitmap used on the fringe to indicate evaluation of an expression.
+
+The default choices are all quite aesthetically pleasing, and have no role
+in the fringe by default. See `fringe-bitmaps' in the Emacs manual for a
+full list of available bitmaps."
+  :type '(choice (const empty-line)
+                 (const filled-rectangle)
+                 (const filled-square)
+                 (const hollow-square)
+                 (const vertical-bar))
+  :group 'cider
+  :package-version '(cider . "0.23.1"))
+
+(defun cider--fringe-overlay-good ()
+  "Create the before-string property that adds a green indicator on the fringe."
+  (propertize " " 'display `(left-fringe ,cider-fringe-bitmap cider-fringe-good-face)))
 
 (defcustom cider-use-fringe-indicators t
   "Whether to display evaluation indicators on the left fringe."
@@ -160,7 +174,7 @@ END is the position where the sexp ends, and defaults to point."
         (clojure-forward-logical-sexp -1)
         ;; Create the green-circle overlay.
         (cider--make-overlay (point) end 'cider-fringe-indicator
-                             'before-string cider--fringe-overlay-good)))))
+                             'before-string (cider--fringe-overlay-good))))))
 
 (cl-defun cider--make-result-overlay (value &rest props &key where duration (type 'result)
                                             (format (concat " " cider-eval-result-prefix "%s "))
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -151,13 +151,21 @@ This function also removes itself from `post-command-hook'."
   "Face used on the fringe indicator for successful evaluation."
   :group 'cider)
 
-(defcustom cider-fringe-bitmap 'filled-rectangle
+;; Fringe bitmaps don't scale with the text, so we define a bitmap that's
+;; tall. This will appear to scale vertically when the text size is increased,
+;; but it's actually just showing more of the bitmap.
+(define-fringe-bitmap 'cider-filled-rectangle
+  (make-vector 100 #b01111110)
+  nil nil 'top)
+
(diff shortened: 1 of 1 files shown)


bbatsov:
Nicely done! A couple of small remarks:

* the CI is failing now
* I think we chose the current bitmap with the idea that it would look different from what some other modes use in the fringe from VC status. 
* It'd be nice if this was also documented in the manual - somewhere in basic usage would be best IMO.

cider-overlays.el
```diff
@@ -29,6 +29,7 @@
 (require 'subr-x)
 (require 'cider-compat)
 (require 'cl-lib)
+(require 'diff-mode)
```
bbatsov: Why did you add this?

cider-overlays.el
```diff
@@ -130,14 +131,32 @@ This function also removes itself from `post-command-hook'."
...
+The default choices are all quite aesthetically pleasing, and have no role
+in the fringe by default. See `fringe-bitmaps' in the Emacs manual for a
+full list of available bitmaps."
+  :type '(choice (const empty-line)
+                 (const filled-rectangle)
+                 (const filled-square)
+                 (const hollow-square)
+                 (const vertical-bar))
+  :group 'cider
+  :package-version '(cider . "0.23.1"))
```
bbatsov: Here it should say "0.24.0".

cider-overlays.el
jcaw: The `diff-added` face comes from this (built-in) package.

cider-overlays.el
jcaw: Perfect. Wasn't sure which version to use.

cider-overlays.el
bbatsov: Got it. Please add a comment next to the require, so it's clear.

jcaw:
Might not be able to get back to this for a bit, but I will eventually.

bbatsov:
No worries.

jr0cket:
@jcaw thanks for the work on the PR for evaluation markers in the fringe for CIDER.  Are you able to complete this PR or would you like some assistance?

Thank you.

jcaw:
Thanks for the reminder - I should be able to get back to it this week.

bbatsov:
@jcaw The build is failing, because by Emacs conventions you should put two spaces after `.` in docstrings and comments.

jr0cket:
@jcaw need any help with this PR?

bbatsov:
@jr0cket seem you can step in here. :-)

jcaw:
Sorry for the radio silence on this one - that's absolutely fine. Feel free to take over, and let me know if you have any questions.

PR #2757, opened 2019-11-25
Add `cider-jump-to-pop-to-buffer-actions`

simon-katz:
Provide a means to control what window to use when jumping to a definition with `cider-jump-to`.

By default, always use current window. (FWIW, note that this matches the behaviour of M-. in .el files.)

Provide an option to use an existing window in current frame when there is one that
is showing the target buffer. This is the old behaviour, but note that it does
not work well with `cider-pop-back` (M-,).

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2499](https://github.com/clojure-emacs/cider/issues/2499): Add `cider-jump-to-pop-to-buffer-actions`.
 * [#2738](https://github.com/clojure-emacs/cider/pull/2738): Add ability to lookup a function symbol when cursor is at the opening paren.
 * [#2735](https://github.com/clojure-emacs/cider/pull/2735): New debugger command `P` to inspect an arbitrary expression, it was previously bound to `p` which now inspects the current value.
 * [#2729](https://github.com/clojure-emacs/cider/pull/2729): New cider inspector command `cider-inspector-def-current-val` lets you define a var with the current inspector value.
--- a/cider-common.el
+++ b/cider-common.el
@@ -127,6 +127,18 @@ On failure, read a symbol name using PROMPT and call CALLBACK with that."
 
 (declare-function cider-mode "cider-mode")
 
+(defcustom cider-jump-to-pop-to-buffer-actions
+  '((display-buffer-same-window))
+  "Determines what window `cider-jump-to` uses.
+The value is passed as the `action` argument to `pop-to-buffer`.
+
+The default value means always use the current window.
+
+For further details, see https://docs.cider.mx/cider/repl/configuration.html#_control_what_window_to_use_when_jumping_to_a_definition"
+  :type 'sexp
+  :group 'cider
+  :package-version '(cider . "0.24.0"))
+
 (defun cider-jump-to (buffer &optional pos other-window)
   "Push current point onto marker ring, and jump to BUFFER and POS.
 POS can be either a number, a cons, or a symbol.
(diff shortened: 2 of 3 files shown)


cider-common.el
```diff
@@ -127,6 +127,14 @@ On failure, read a symbol name using PROMPT and call CALLBACK with that."
 
 (declare-function cider-mode "cider-mode")
 
+(defcustom cider-jump-to-always-use-same-window t
```
bbatsov: I think this should not be a boolean config, but rather a config with a symbol value like `'same-window`, etc.

bbatsov: Alternative we can just leave the user to specify the jump attributes list directly here.

cider-common.el
```diff
@@ -127,6 +127,14 @@ On failure, read a symbol name using PROMPT and call CALLBACK with that."
...
+  "Determines what window `cider-jump-to` uses.
+If non-nil, always use current window.
+If nil: if target buffer is being displayed in the current frame use its
+window, otherwise use the current window."
+  :type 'boolean
+  :group 'cider)
```
bbatsov: There should also be a `:package-version` property here.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2757](https://github.com/clojure-emacs/cider/pull/2757): Add `cider-jump-to-always-use-same-window`
```
bbatsov: You should refer to the original issue here, not the PR. Don't forget about the final `.`. :-)

cider-common.el
```diff
@@ -139,8 +147,10 @@ If OTHER-WINDOW is non-nil don't reuse current window."
     (ring-insert find-tag-marker-ring (point-marker)))
   (if other-window
       (pop-to-buffer buffer 'display-buffer-pop-up-window)
-    ;; like switch-to-buffer, but reuse existing window if BUFFER is visible
-    (pop-to-buffer buffer '((display-buffer-reuse-window display-buffer-same-window))))
+    (pop-to-buffer buffer (if cider-jump-to-always-use-same-window
```
bbatsov: Looking at this code it really seems to me it'd be easier to just pas the config list directly as a defcustom. This will give the users the maximum flexibility and eliminate the branching in the code.

doc/modules/ROOT/pages/repl/configuration.adoc
```diff
@@ -261,3 +261,15 @@ can set:
 ----
 (setq cider-repl-require-ns-on-set t)
 ----
+
+== Control what window to use when jumping to a definition
+
+By default M-. and other commands that jump to a definition use the current
+window to show the definition. An alternative is to use an existing window in
+the current frame when there is one that is showing the target buffer. If you
+want this alternative behaviour, set:
```
bbatsov: You might want to make this a bit more concrete with some example - e.g. you jump to `map` in `core.clj` and `core.clj` is already visible. Or something like this.

bbatsov:
Great work! I've left a few small comments, but I finally understand what the problem is (and why I never encountered it).

cider-common.el
simon-katz: Changed to allow user to specify the actions for `pop-to-buffer`.

cider-common.el
simon-katz: Done.

CHANGELOG.md
simon-katz: Done.

cider-common.el
simon-katz: Done.

doc/modules/ROOT/pages/repl/configuration.adoc
simon-katz: Done.

simon-katz:
Linting is complaining about this in a doc string:

```
A value of `((display-buffer-reuse-window display-buffer-same-window))`
means if target buffer is being displayed in the current frame use its
window, otherwise use the current window.
```

The complaint is: Lisp symbol `display-buffer-same-window' should appear in quotes.

Is this a bug in the linting?

bbatsov:
You can't have inlined code in the docstrings, just references to params and symbols. I guess you can either reword this or add a link to the manual.

simon-katz:
I've re-worded the doc string. Ready for re-review.

simon-katz:
Whoops — I've reworded things incorrectly. Let me fix that…

bbatsov:
Thanks!

simon-katz:
@bbatsov Thanks for your help.

PR #2759, opened 2019-11-29
Make it possible to `cider-find-var` in another window

joncol:
This fix makes it possible to use `-` prefix (`M-- t`) to jump to test definition in another window. Before this fix, pressing enter (or `t`) in the CIDER test report buffer always opened the definition of the failing test function in the same window as the test report.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2759](https://github.com/clojure-emacs/cider/pull/2759): Fix `cider-find-var` prefix behaviour when `var` is passed.
 * [#2744](https://github.com/clojure-emacs/cider/pull/2744): Add startup commands to repl banner.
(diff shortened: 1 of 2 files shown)


bbatsov:
Did you start experiencing this problem recently? It seems to me that changing the default here https://github.com/clojure-emacs/cider/pull/2757 might have been a bad idea for such use cases.

joncol:
> Did you start experiencing this problem recently? It seems to me that changing the default here #2757 might have been a bad idea for such use cases.

Yes, I think it started showing up around that time. Not 100 % sure though.

joncol:
OK, after reading #2757 and using
```
(setq cider-jump-to-pop-to-buffer-actions
      '((display-buffer-reuse-window display-buffer-same-window)))
```
it works as I want.

I guess this PR can be closed then.

bbatsov:
Let's keep it open for now until I restore the previous default value. When it was changed we didn't account the impact of the new default to cases like this one.

Issue #2756, opened 2019-11-24
Evaluate forms in multiple REPLS


simon-katz:
When both CLJ and CLJS REPLS exist, if I use an evaluation command such as C-M-x
in a .cljc file the form is sent only to the most-recently-visited REPL.

`cider-load-file` works in a similar way.

There is some high-level code that handles multiple REPLs, but the low-level
code (around Sesman sessions) doesn't handle multiple REPLs.

I suspect that at one time C-M-x may have sent forms to multiple REPLs,
and that there have been changes that break this.

I haven't looked at `cider-load-file`, but for C-M-x and other commands that go through `cider-interactive-eval` the problem is in the following call chain:

       cider-interactive-eval
    -> cider-map-repls
    -> cider-repls
    -> sesman-ensure-session or sesman-current-session

I have hacked a proof-of-concept fix for this — see https://github.com/simon-katz/nomis-emacs-configuration/blob/b64dcbdc18df4d96327519a5d7b2d3ce54adde40/emacs-installation/emacs-init-files/nomis-cider-eval-clj-and-cljs.el#L86-L133 (EDITED to a later, clearer version).

I've tested the hack on the current master of CIDER (commit dd74ddd4 — "Mention
the "State of CIDER" survey in the REPL's help banner").

My code arranges for `cider-repls` to return multiple REPLs when called using
the above call chain, by placing advice on `cider-interactive-eval` and
`cider-map-repls` and using alternative versions of `sesman-ensure-session` and
`sesman-current-session`.

simon-katz:
I've been looking in to this, and may have a fix — see https://github.com/simon-katz/cider/commit/ffc8107bd16e80e968dcf00c8a32529634f8c8a9

I'd appreciate any comments on what I've done so far. (Is this heading in the right direction? If so I'll get things finished and raise a PR.)

The change makes `cider-repls` return REPLS from all Sesman sessions unless the new optional argument `only-current-session` is non-nil. `cider-current-repl` calls `cider-repls` and passes `t` for the `only-current-session` argument.

All tests continue to pass.

Still to do:

- Check all calls of `cider-repls` — do any need to be changed to pass `t` for `only-current-session`?
- Make any necessary changes to documentation. I know some places explicitly mention the most recent session. Are changes needed?
- Add to the change log.

PR #2766, opened 2019-12-03
WORK PROBABLY STILL NEEDED — Evaluate forms in multiple REPLs when appropriate

simon-katz:
This addresses the issue of evaluating forms from .cljc files in both CLJ and CLJS REPLs — see https://github.com/clojure-emacs/cider/issues/2756

There was already high-level code to handle multiple REPLs, but the low-level code (around Sesman sessions) didn't handle multiple REPLs.

I've made changes to `cider-repls` and `cider-map-repls`, adding some more optional parameters. I don't much like the way things have turned out, but I've tried to make sure that things are backward-compatible. Perhaps things should be done differently.

WORK IS PROBABLY STILL NEEDED TO COMPLETE THIS, but I'd appreciate comments before going further.

These are the outstanding issues that I'm aware of:

- There are multiple evaluation results, from multiple REPLs. The result that happens second overwrites the result that happens first. I'm not sure what to do about this.
- The manual needs updating. The new variable `cider-eval-forms-in-all-sessions` needs to be documented, and I think there may be some documentation that talks about evaluation happening in the most recent REPL, which will need to be fixed.
- Perhaps new tests are needed. (I'm not sure I have the knowledge to do that — but I am happy to take a look if I get positive feedback on what I've done so far.)

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * [#2744](https://github.com/clojure-emacs/cider/pull/2744): Add startup commands to repl banner.
+* [#2756](https://github.com/clojure-emacs/cider/issues/2756): Evaluate forms in multiple REPLS.
 * [#2499](https://github.com/clojure-emacs/cider/issues/2499): Add `cider-jump-to-pop-to-buffer-actions`.
 * [#2738](https://github.com/clojure-emacs/cider/pull/2738): Add ability to lookup a function symbol when cursor is at the opening paren.
 * [#2735](https://github.com/clojure-emacs/cider/pull/2735): New debugger command `P` to inspect an arbitrary expression, it was previously bound to `p` which now inspects the current value.
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -66,6 +66,15 @@ available) and the matching REPL buffer."
   :safe #'booleanp
   :package-version '(cider . "0.9.0"))
 
+(defcustom cider-eval-forms-in-all-sessions t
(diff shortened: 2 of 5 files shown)


bbatsov:
What are you referring to as "all-sessions" in the PR? Evaluating code across multiple CIDER sessions or nREPL sessions doesn't make sense, so I'm assuming you're actually referring to all REPLs within one CIDER (sesman) session.

simon-katz:
I meant across multiple CIDER sessions, but I realise now that this probably doesn't make sense because different CIDER sessions could be for different projects.

This sprung from a question I asked on the the #cider Clojurians channel on Slack on 22nd November:

> I have a CLJ REPL and a CLJS REPL. My sesman-browser shows something like the following:
>      CIDER Sessions:
>       1: my-apps/my-app:localhost:63615
>           linked-to: proj(~/dev/my-apps/my-app/)
>             objects: *cider-repl %s(clj)*
>       2: my-apps/my-app:localhost:8778
>           linked-to: proj(~/dev/my-apps/my-app/)
>             objects: *cider-repl %s(cljs:shadow-select)*
> Is it possible to pass the same form to both REPLs with C-M-x etc?
> (Perhaps I’ve set up my sessions wrongly.)

So I have two CIDER sessions, one for CLJ and one for CLJS, in the same project.

The CLJ session was set up with a `cider-jack-in` and the CLJS session was set up with `cider-connect-cljs` to a server that was set up with shadow-cljs.

I was hoping for a way of evaluating a form in a.cljc file in both REPLs. Is there a way to do that?

PR #2767, opened 2019-12-08
Resolve relative paths in classpath

lins05:
Sometimes the classpath returned by `(cider-fallback-eval:classpath)` contains relative entries like `src/main` and we need to resolve those to absolute paths. Otherwise `sesman-friendly-session-p` would throw error because there would be `nil` in `classpath-roots`

https://github.com/clojure-emacs/cider/blob/fa7ffd45/cider-connection.el#L462-L465

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -503,18 +503,11 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "classpath")))
 
-(defun cider-sync-tooling-eval-get-value (expression)
-  "Send the EXPRESSION to the repl and get the returned value."
-  (thread-first expression
-    cider-sync-tooling-eval
-    (nrepl-dict-get "value")
-    read))
-
 (defun cider--get-abs-path (path project)
   "Resolve PATH to an absolute path relative to PROJECT.
 Do nothing if PATH is already absolute."
-  (if (not (string= (substring path 0 1) "/"))
-      (format "%s/%s" project path)
+  (if (not (file-name-absolute-p path))
+      (expand-file-name path project)
     path))
 
 (defun cider-fallback-eval:classpath ()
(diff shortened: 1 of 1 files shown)


bbatsov:
I think that a much better approach would be to simple check if some entry is a relative path and expand it against the project root. All of this can be done Emacs-side pretty easily.

lins05:
@bbatsov Updated, the project root is now read from emacs side.

lins05:
ping @bbatsov , any chance to take a second look?

cider-client.el
```diff
@@ -503,9 +503,27 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "classpath")))
 
+(defun cider-sync-tooling-eval-get-value (expression)
```
bbatsov: I think this helper doesn't make much sense, so let's remove it.

cider-client.el
```diff
@@ -503,9 +503,27 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
...
+  "Send the EXPRESSION to the repl and get the returned value."
+  (thread-first expression
+    cider-sync-tooling-eval
+    (nrepl-dict-get "value")
+    read))
+
+(defun cider--get-abs-path (path project)
```
bbatsov: Emacs already has the built-in functions `file-name-absolute-p` and `expand-file-name`, which you should use instead.

cider-client.el
```diff
@@ -503,9 +503,27 @@ Optional arguments include SEARCH-NS, DOCS-P, PRIVATES-P, CASE-SENSITIVE-P."
...
+
 (defun cider-fallback-eval:classpath ()
-  "Return a list of classpath entries using eval."
-  (read (nrepl-dict-get (cider-sync-tooling-eval "(seq (.split (System/getProperty \"java.class.path\") \":\"))") "value")))
+  "Return a list of classpath entries using eval.
+
+Sometimes the classpath contains entries like src/main and we need to
+resolve those to absolute paths."
+  (let ((classpath (cider-sync-tooling-eval-get-value "(seq (.split (System/getProperty \"java.class.path\") \":\"))")))
+    (mapcar (lambda (path) (cider--get-abs-path path default-directory)) classpath)))
```
bbatsov: I don't think you should expand in the `default-directory`. It makes more sense to expand in `(clojure-project-dir)` (or whatever this function was named).

bbatsov:
@lins05 ping :-)

lins05:
pong @bbatsov Thx for the review, updated.

bbatsov:
Thanks!

PR #2768, opened 2019-12-12
Fix the nrepl endpoint details in the mode line

thatismatt:
This fixes the display of the nrepl connection in the mode line.

It looks like `nrepl-endpoint` was historically a list, but is now a plist. This results in the mode line displaying as: `clj:my-project@:proc:nrepl-connection` with this fix it will now display as: `clj:my-project@:34969`.


bbatsov:
Thanks!

PR #2781, opened 2020-01-17
Extend the default doc xref regexp

yuhan0:
This allows cider-doc buffers to recognize `[[var]]` syntax and fully qualified symbols as xref links.

Both are fairly common in docstrings in core and 3rd-party libraries that I've personally encountered, and it's otherwise somewhat annoying to navigate to a referenced var or function.  (E.g. executing `M-x cider-doc`in a `*cider-doc*` buffer doesn't pick up the symbol at point).

If this is acceptable I'll go ahead with adding tests and changelogs :)

-----------------

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -511,9 +511,11 @@ restore it properly when going back."
     (setq help-xref-stack-item item)))
 
 (defcustom cider-doc-xref-regexp
-  (rx (or (: "`" (group-n 1 (+ (not space))) "`")   ; `var`
+  (eval-and-compile
+    (rx-to-string
+     `(or (: "`" (group-n 1 (+ (not space))) "`")  ; `var`
           (: "[[" (group-n 1 (+ (not space))) "]]") ; [[var]]
-          (group-n 1 (+ (not space)) "/" (+ (not space))))) ; Fully qualified
+          (group-n 1 (regexp ,clojure--sym-regexp) "/" (regexp ,clojure--sym-regexp))))) ;; Fully qualified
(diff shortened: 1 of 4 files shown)


bbatsov:
That's a nice improvement! I believe we did something similar a while ago in `clojure-mode` so that more xref formats are font-locked there properly.

cider-util.el
```diff
@@ -510,15 +510,17 @@ restore it properly when going back."
         (if tail (setcdr tail nil))))
     (setq help-xref-stack-item item)))
 
-(defcustom cider-doc-xref-regexp
-  (rx (or (: "`" (group-n 1 (+ (not space))) "`")   ; `var`
+(eval-and-compile
+  (defcustom cider-doc-xref-regexp
+    (rx-to-string
```
bbatsov: Shouldn't the `eval-and-compile` be here?

yuhan0: Ah of course, changing it now-

yuhan0:
I looked into writing tests for this, but realised there were none written for the entire `cider-doc` namespace-  also it's not clear how to test functions which rely on a REPL connection.

bbatsov:
The best approach is to simply stub the functions that require a REPL connection.

yuhan0:
I added a test for the regexp matching in cider-util, which should cover the changes for this PR.

bbatsov:
Great work! Thanks! 🙇

bbatsov:
@yuhan0 Btw, it'd be a good idea to update the font-locking of fully-qualified vars appearing in docstrings in `clojure-mode` as well for consistency.

PR #2782, opened 2020-01-20
Doc fixes

### LINKED ISSUE #2763: Unclear indentation specification example
## Expected behavior

I was expecting to make sense of the last example at:
https://docs.cider.mx/cider/indent_spec.html#_examples

## Actual behavior

I didn't make any sense of the following:

q

```
The first arg has an indent spec of , which means all forms inside the first arg have an indent spec of [:defn].
```
Is this a typo or is it intentional? It says that the first arg has an empty indent spec? Can this be made any clearer, perhaps?

yuhan0:
Just realised I made a typo in Changelog.md in my last PR, oops.
Also closes #2763 by using literal syntax for the indentation spec example

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #2785, opened 2020-01-24
Fix example in pretty_printing.adoc

jhacksworth:
The example setting for `cider-print-options` does not work for me. I think the issue is with `dict`. 

Using the example in the `defcustom` documentation works: https://github.com/clojure-emacs/cider/blob/4cc4280677e6eeb16cd55d9865c0ea9f9d141af3/cider-client.el#L248-L252


bbatsov:
Thanks!

Issue #2787, opened 2020-01-25
Cannot connect a sibling session if `nrepl-hide-special-buffers` is enabled


jmckitrick:
## Expected behavior
Both cider-jack-in-clj and cider-jack-in-cljs should work if special buffers are hidden.

## Actual behavior
If `nrepl-hide-special-buffers` is enabled, connecting to a sibling session calls `nrepl-start-server-process` which does not ’see’ the existing nrepl buffer and fails to generate a new buffer for the new session/connection.

## Steps to reproduce the problem
1. Set `nrepl-hide-special-buffers nil`
2. In a project using shadow-cljs, run `cider-jack-in-clj` first
3. After 2 succeeds, run `cider-jack-in-cljs`
4. Verify that 3 never completes successfully

This should be reproducible in either order: CLJ then CLJS, or vice-versa. The key is using the sibling session to connect, and the option to hide special nrepl buffers.

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; CIDER 0.24.0-snapshot, nREPL 0.6.0
;; Clojure 1.10.1, Java 1.8.0_74
```

### Lein/Boot version

Leiningen 2.9.1

### Emacs version

26.3

### Operating system

OS X 10.15.2 Catalina

dpsutton:
Want to add a minimal repro

in a project with just a deps.edn file consisting only of `{}`, with `nrepl-hide-special-buffers` set to `nil` if you jack in twice you will end up with two repls. When you set `nrepl-hide-special-buffers` to `t`, jacking in twice will create only a single repl. It _will_ start two jvms but only the first will connect. CIDER mistakenly reuses the first repl's nrepl buffer so the connection callback mechanism is never triggered.

So rather than two files with `nREPL server started on port 42097 on host localhost - nrepl://localhost:42097` you end up with one buffer with both connection statements in it:

```
nREPL server started on port 42097 on host localhost - nrepl://localhost:42097
nREPL server started on port 36653 on host localhost - nrepl://localhost:36653
```

PR #2788, opened 2020-01-25
Fix for 2787 when using nrepl-hide-special-buffers

dpsutton:
nrepl-hide-special-buffers would prepend a leading space to the nrepl
process buffer to hide it from the buffer list. If you created a new
sibling process there was a bug however. In order to uniquely name
buffers, CIDER uses `generate-new-buffer-name` to ensure a buffer is
uniquely named. However, the space was put on _after_ this
call. The sequence is as follows:

- jack in
- use template "*cider-nrepl blah" to see if existing process buffer
is around. Its not. Since we want these hidden, add a leading space so
now " *cider-nrepl blah*" process buffer exists
- watch for "nREPL server started on port \\([0-9]+\\)" and not
already a port saved
- start repl
- jack in again creating sibling process
- check if buffer "*cider-nrepl blah" exists. It doesn't since the
first one has a leading space
- add a leading space to the buffer name to get " *cider-nrepl blah"
which actually now is a collision
- lein/deps/shadow startup output goes in the same buffer as the other
process and therefore no new repl is started up when watching the the
process output since the following code is waiting:

```lisp
(when (and (null nrepl-endpoint)
           (string-match "nREPL server started on port \\([0-9]+\\)" output))
```

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#2787](https://github.com/clojure-emacs/cider/issues/2787): Fix nrepl process naming collision when using `nrepl-hide-special-buffers`
 * [#2744](https://github.com/clojure-emacs/cider/pull/2744): Add startup commands to repl banner.
 * [#2499](https://github.com/clojure-emacs/cider/issues/2499): Add `cider-jump-to-pop-to-buffer-actions`.
 * [#2738](https://github.com/clojure-emacs/cider/pull/2738): Add ability to lookup a function symbol when cursor is at the opening paren.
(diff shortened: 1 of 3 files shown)


dpsutton:
fixes https://github.com/clojure-emacs/cider/issues/2787

dpsutton:
as an aside, i recently switched to using fish shell and this was trivial to make a file watcher to run tests on every file save

```shell
while inotifywait -e modify -r . 
      make test
end

```

and in the makefile for make test you can add `--pattern "nrepl buffer"` or some other regex to rerun only tests matching that regex

bbatsov:
Nice catch! 

As for the tip on running tests - it might be good to add it to manual (ideally it's not hard to do something similar for bash/zsh, as I don't think fish is that popular).

Issue #2047, opened 2017-07-16
Result truncated. Type ‘M-x cider-inspect-last-result’ to inspect it. -- does not function as expected


rcullito:
## Expected behavior
inspecting the truncated object would display the object in full that was being inspected while the debugger is paused
## Actual behavior
information such as arg list and docstring is displayed for the currently executing function in inspector mode (rather than the immediate object of concern)

Emacs 25.1.1
cider - 0.15.0-SNAPSHOT

rcullito:
the closest discussion I can find regarding this behavior is here: https://github.com/clojure-emacs/cider/issues/1692

but it doesn't seem like the inspector implements the vision of folding that was discussed on the thread?

bbatsov:
I think @alexander-yakushev had some vision of folding for the inspector. Don't recall the details anymore.

bbatsov:
Btw, so repro steps for this problem would be useful.

bitti:
I experiencing this when setting `clojure.core/*print-length*` to a numeric value (not `nil`) in the repl. For `(set! *print-length* 20)` I see only the first three elements of the inspected object. In general it seems only about `*print-length*/6` elements are displayed. I just called `cider-inspect-last-result` and didn't even use the debugger. Only setting `*print-length*` to nil seems to help (even though after a `cider-inspector-pop` the results are truncated again), but there is a reason that I would like to limit the output length in the repl.

bbatsov:
@rcullito were you setting the *print-length* as well?

rcullito:
@bbatsov I'm currently running cider version 20170715.36 and the issue seems to be resolved.

When entering into the debugger in this version, a truncated result no longer prompts the user to enter the command for `M-x cider-inspect-last-result` but rather displays `...` to indicate more contents. 

Hitting `p` for inspect and entering the appropriate value does allow nested inspection as one might expect. From my end I believe that resolves the issue I originally raised. I had always kept _print-length_ at nill so it sounds like that may be a separate issue.

bbatsov:
Got it!

I'll close this ticket then - @bitti, please file a new one if you're on 0.16 (or the current snapshot).

bitti:
Ok, just opened https://github.com/clojure-emacs/cider/issues/2152. I had cider 0.16.0snapshot installed but I just upgraded to 0.17.0snapshot and the issue is still there.

rcullito:
thanks for following up, @bbatsov !

[...]

PR #2793, opened 2020-02-06
[debugger] Fix overlay message when the result is too large

alexander-yakushev:
Fixes https://github.com/clojure-emacs/cider/issues/2047#issuecomment-579715972.


bbatsov:
Thanks!

PR #2797, opened 2020-02-15
Docs - page not found on bottom link

kemalelmizan:
`basics/installation.adoc#setting-up-a-standalone-repl` does not exists, therefore it caused 404 page on https://docs.cider.mx/cider/basics/installation.adoc#setting-up-a-standalone-repl. I'd assume the correct link should be `basics/middleware_setup.adoc#setting-up-a-standalone-repl` so it goes to https://docs.cider.mx/cider/basics/middleware_setup.html#_setting_up_a_standalone_repl.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Thanks!

PR #2800, opened 2020-02-16
Add force-out support.

FiV0:
Adding the option to force out of a debugging operation. Also skips
instrumented functions lower in the call stack. Already implemented
at cider-nrepl level.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -202,7 +202,7 @@ Can be toggled at any time with `\\[cider-debug-toggle-locals]'."
     (?n "next" "next")
     (?i "in" "in")
     (?o "out" "out")
-    (?O "out" nil)
+    (?O "force-out" nil)
     (?h "here" "here")
     (?e "eval" "eval")
     (?p "inspect" "inspect")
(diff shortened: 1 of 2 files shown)


FiV0:
@bbatsov  Are the tests failure due to me or due to me or https://github.com/jorgenschaefer/emacs-buttercup/issues/122?

cider-debug.el
```diff
@@ -202,6 +202,7 @@ Can be toggled at any time with `\\[cider-debug-toggle-locals]'."
     (?n "next" "next")
     (?i "in" "in")
     (?o "out" "out")
+    (?O "out" nil)
```
bbatsov: As this is what people see in the menu, I guess we should use different wording here - otherwise it's not clear how it differs from the standard "out".

bbatsov:
I think we also list the debugger commands in its manual with some extra explanations.

FiV0:
@bbatsov What do you mean? The big O command is already in the docs.
https://docs.cider.mx/cider/debugging/debugger.html#_keys

bbatsov:
My bad. I didn't notice this. Thanks!

Issue #2799, opened 2020-02-16
`make test` issue related to buttercup


FiV0:
`make test` fails with 
```
cask exec buttercup -L . -L ./test/utils/
Spies can only be created in ‘before-each’
Makefile:68: recipe for target 'test' failed
make: *** [test] Error 255
```

I think this might be related to https://github.com/jorgenschaefer/emacs-buttercup/issues/122
as this was closed recently.  @bbatsov you also commented in that thread. 

### CIDER version information

```
;; CIDER 0.24.0 (package: 20191103.1528) (India), nREPL 0.6.0
;; Clojure 1.10.0, Java 11.0.6
```

### Lein/Boot version

not relevant

### Emacs version

26.3

### Operating system

Ubuntu 18.04

bbatsov:
Yeah, this definitely seems related. We'll probably have to update the tests to account for the upstream changes in buttercup.

bbatsov:
Fixed in https://github.com/clojure-emacs/cider/pull/2804

PR #2803, opened 2020-02-21
Fix buttercup issues

FiV0:
Should be fixing #2799.

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/test/cider-connection-tests.el
+++ b/test/cider-connection-tests.el
@@ -288,26 +288,27 @@
            (sesman-unregister 'CIDER session)))))))
 
 (describe "cider--connection-info"
-  (spy-on 'cider--java-version :and-return-value "1.7")
-  (spy-on 'cider--clojure-version :and-return-value "1.7.0")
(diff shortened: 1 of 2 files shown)


test/cider-connection-tests.el
```diff
@@ -288,26 +288,27 @@
            (sesman-unregister 'CIDER session)))))))
 
 (describe "cider--connection-info"
-  (spy-on 'cider--java-version :and-return-value "1.7")
```
FiV0: Sorry. I don't know if this is the correct way to do it, but at least these spy-on's need to get wrapped in a block.

bbatsov: It seems to me you're supposed to move the spies to `before-each` as illustrated here https://github.com/jorgenschaefer/emacs-buttercup/blob/master/docs/writing-tests.md#spies

bbatsov:
@dpsutton actually fixed the tests in https://github.com/clojure-emacs/cider/pull/2804, so I'll close this PR.

PR #2804, opened 2020-02-22
Fixup tests

dpsutton:
Fixup tests

buttercup pushes the spy information onto a context which is only available during `before-each` or `it` macros. I'm not sure what `before-all` does now but it doesn't allow for the spies.

I was looking for where the manual the following too helpful bits could go:

1. A filewatcher to rerun tests

```
while inotifywait -e modify -r .
  make test
end
```

2. supplying a pattern to the buttercup runner:

update the makefile with `--pattern` or `-p`:

```
test: version build
	$(CASK) exec buttercup -L . -L ./test/utils/ --pattern cider-eldoc
```

These two bits helped me track down these problem in just a few minutes.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/test/cider-connection-tests.el
+++ b/test/cider-connection-tests.el
@@ -288,9 +288,10 @@
            (sesman-unregister 'CIDER session)))))))
 
 (describe "cider--connection-info"
-  (spy-on 'cider--java-version :and-return-value "1.7")
-  (spy-on 'cider--clojure-version :and-return-value "1.7.0")
-  (spy-on 'cider--nrepl-version :and-return-value "0.2.1")
+  (before-each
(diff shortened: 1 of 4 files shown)


dpsutton:
tests-passing

not sure why the tests aren't running here but they're running on my circle version.

bbatsov:
> I'm not sure what before-all does now but it doesn't allow for the spies.

It's just a shared setup that's invoked once for the test suite. Seems to me that `spy-on` should work there as well, but maybe that's some buttercup bug. 

Thanks for fixing this! I'll think where we can incorporate your suggestions in the manual.

PR #2811, opened 2020-02-27
Refcard url parsing

tvirolai:
It seems that the refcard document was moved from the path `./doc/` to `./refcard/` in Cider 0.23. However, the `cider-view-refcard` function still uses a URL that points to the old path. For example, in Cider 0.25.0-snapshot, the refcard is opened to https://github.com/clojure-emacs/cider/raw/master/doc/cider-refcard.pdf, which doesn't exist.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -439,7 +439,7 @@ to."
                                    section-id))))
     (buffer-string)))
 
-(defconst cider-refcard-url "https://github.com/clojure-emacs/cider/raw/%s/doc/cider-refcard.pdf"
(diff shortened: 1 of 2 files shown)


cider-util.el
```diff
@@ -439,7 +439,7 @@ to."
                                    section-id))))
     (buffer-string)))
 
-(defconst cider-refcard-url "https://github.com/clojure-emacs/cider/raw/%s/doc/cider-refcard.pdf"
+(defconst cider-refcard-url "https://github.com/clojure-emacs/cider/raw/%s/%s/cider-refcard.pdf"
```
bbatsov: Why don't we simply hardcode the new URL? I don't see much point in making this depending on the CIDER version.

PR #2820, opened 2020-03-13
Avoid requiring org on load

yuhan0:
Currently requiring Cider causes the entire org-mode library to be loaded transitively via `org-table`, which is only required by one function in cider-doc. This can add several seconds to startup the first time one opens a Clojure buffer.

This change simply defers the require statement until the function `cider-docview-format-tables`
needs it.

-----------------

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -37,6 +37,11 @@
 (require 'easymenu)
 (require 'cider-browse-spec)
 
+(declare-function org-table-map-tables 'org-table)
+(declare-function org-table-align 'org-table)
+(declare-function org-table-begin 'org-table)
+(declare-function org-table-end 'org-table)
+
 
 ;;; Variables
 
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -33,7 +33,6 @@
 (require 'cider-client)
 (require 'cider-clojuredocs)
 (require 'nrepl-dict)
-(require 'org-table)
 (require 'button)
 (require 'easymenu)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Changes
 
 * [#2617](https://github.com/clojure-emacs/cider/pull/2617): Add menu bar entry for `Insert last sexp in REPL and eval` under `CIDER Eval`.
+* [#2820](https://github.com/clojure-emacs/cider/pull/2820): Defer loading of org-table library on startup.'
```
bbatsov: That's not a user-observable change, so it doesn't warrant a changelog entry.

yuhan0: Ok, I removed the commit

bbatsov:
@yuhan0 I saw the lint has failed and I remembered that this was probably the reason why the package is required. Probably we'll need to add a couple of declarations to appease the byte-compiler.

bbatsov:
Thanks!

yuhan0:
Hmm, the CI tests were still failing, and running `make lint` locally resulted in a "FAIL" for every single file

bbatsov:
The other two failures were unrelated. I fixed one and I'm not sure what to do with the other. @gonewest818 recently did some updates to `emacs-lint` and I guess this resulted in the extra errors. Not sure why this isn't working locally for you. Perhaps he can help with some ideas.

PR #2826, opened 2020-03-27
Changes to cider-symbol-at-point

yuhan0:
----

Trim leading quotes from symbol, now that clojure-mode treats it as a symbol constituent: 
https://github.com/clojure-emacs/clojure-mode/pull/559

This makes it possible to use cider-doc, eldoc, and various other features on symbols with quotes in them like `foo'`.

Also uses `macroexpand` to resolve ns-aliased keywords, allowing cider-browse-spec to work on aliased keywords at point. 
(Not sure how to write tests for this one)

Would this go under the "changes" or "new features" heading of the changelog? 

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
> Would this go under the "changes" or "new features" heading of the changelog?

Let's go with changes.

bbatsov:
Thanks!

PR #2833, opened 2020-04-18
Save command history for jack-in with universal arg

yuhan0:
I recently had to `C-u cider-jack-in` to a project with a custom set of command line params that took a few tries to get right, and found that I had to re-edit the command from scratch each time which was quite frustrating. This should solve the issue, I don't think tests are needed.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1221,6 +1221,12 @@ non-nil, don't start if ClojureScript requirements are not met."
   :safe #'booleanp
   :version '(cider . "0.22.0"))
 
+(defvar cider--jack-in-nrepl-params-history nil
+  "History list for user-specified jack-in nrepl command params.")
+
+(defvar cider--jack-in-cmd-history nil
+  "History list for user-specified jack-in commands.")
+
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -1258,7 +1265,7 @@ non-nil, don't start if ClojureScript requirements are not met."
                                                                 cmd-params))))
                     (plist-put params :jack-in-cmd (if (or cider-edit-jack-in-command
                                                            (plist-get params :edit-jack-in-command))
-                                                       (read-string "jack-in command: " cmd t)
+                                                       (read-string "jack-in command: " cmd 'cider--jack-in-cmd-history)
```
bbatsov, line 1268: I see you've quoted the symbol here, but not in the other `read-string`. Don't remember which usage is the correct one.

yuhan0: Whoops, good catch! It should be a quoted symbol, I just corrected it

bbatsov:
Thanks!

PR #2851, opened 2020-06-04
[Documentation] Configure Test report even on successful tests

jr0cket:
A test report is never created when all tests run are successful.  Add information on how to configure CIDER to create a test report when tests are successful.

Small documentation only PR, no tests required?

### DIFF
--- a/doc/modules/ROOT/pages/testing/test_reports.adoc
+++ b/doc/modules/ROOT/pages/testing/test_reports.adoc
@@ -8,8 +8,7 @@ and jump directly to the definition of failing tests.
 
 Calling test commands again will update the test report.
 
-Customise the `cider-test-show-report-on-success` and set to true 
-and a test report is also shown when tests pass.
+link:running_tests.html#display-test-report-on-success[Configure test reports on success]
 
 |===
 | Keyboard shortcut | Description
--- a/doc/modules/ROOT/pages/testing/test_reports.adoc
+++ b/doc/modules/ROOT/pages/testing/test_reports.adoc
@@ -1,10 +1,15 @@
 = Interacting with Test Result Reports
 :experimental:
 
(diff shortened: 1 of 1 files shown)


jr0cket:
The test-lint is failing on a line not changed by this pull request.
Line 30 (previously line 25) is longer than the 80 character limit that the test-lint checks for.

bbatsov:
We already have this info here https://docs.cider.mx/cider/testing/running_tests.html#display-test-report-on-success, but perhaps it's OK to duplicate it.

jr0cket:
Oops, I read the running tests page too, I just missed that section of the page.  Documentation is hard :)

jr0cket:
I've updated the PR with a link to the specific configuration section from the Test Reports page, to avoid duplication.

I have left the change to the first paragraph as only if you have read all the way down to almost the bottom of running tests page do you discover that the test report buffer only displays upon failure of tests.

Alternatively, you could make the whole configuration section from the Running Tests page a separate page and make the specific configuration items discoverable in the left hand menu navigation.

doc/modules/ROOT/pages/testing/test_reports.adoc
```diff
@@ -1,11 +1,15 @@
...
 
-After running your tests, CIDER displays a test result report in the
+Shoud tests fail then CIDER displays a test result report in the
 `+*cider-test-report*+` buffer. This buffer uses `cider-test-report-mode`,
 which makes it easy to review any failures that might have occurred
 and jump directly to the definition of failing tests.
 
+Calling test commands again will update the test report.
+
+link:running_tests.html#display-test-report-on-success[Configure test reports on success]
```
bbatsov, line 11: `link` is for external references, internally you should be using `xref`. I'd probably add `TIP: See also xref`, so it's consistent with the general style of the docs.

PR #2853, opened 2020-06-05
[Documentaiton] Mention the cider-nrepl requirement

philjackson:
Extend the shadow cljs section on cider-connect to mention the fact
cider-nrepl needs to be in the classpath.


bbatsov:
Well, that's a hard requirement to use CIDER with ClojureScript in general (it's possible to turn off the auto REPL type detection, though, and then you won't need cider-nrepl, only piggieback/shadow-cljs). It's not specific to shadow-cljs. :-) See https://docs.cider.mx/cider/cljs/overview.html#nrepl-and-clojurescript

bbatsov:
See also https://docs.cider.mx/cider/repl/configuration.html#repl-type-detection

PR #2855, opened 2020-06-07
Fix debug column numbers

### LINKED ISSUE #2852: Indented code cannot be debugged in-source
## Expected behavior
Executing cider-debug-defun-at-point (C-u C-M-x) should work in source for forms in an indented comment block when `clojure-toplevel-inside-comment-form` is true.
```
(comment
  (inc (inc 0))
  )
```

More generally, this occurs when evaluating any  indented form tagged with #dbg

```
(do
  #dbg (str (inc 0)) ;; cider-eval-last-sexp here
  ...)
```

## Actual behavior
The source "cannot be found" and a temporary debug buffer is created.

place cursor at | and eval last sexp
```
(do
  #dbg (str (inc 0)) | 
  ...)
```

## Solution

After some detective work I found the underlying issue to be the difference in column indexing between Emacs (0-indexed) and Clojure (1-indexed).

The outgoing conversion to 1-indexing happens in `cider-interactive-eval`, using the function `cider-column-number-at-pos`, so nrepl messages are sent as 1-indexed columns.
https://github.com/clojure-emacs/cider/blob/1911c474baefa1ef0f50274c7035a46949249d58/cider-util.el#L757

However for incoming responses, this line in cider-nrepl causes **only** toplevel forms (column = 1) to be converted to 0-indexing , and no further processing is done client-side.
https://github.com/clojure-emacs/cider-nrepl/commit/7660eb7698b0a708d2e4f6badf5e9afb3a300c4e#diff-0d65c570bef3dbf4640690a252e06f28R340

When column != 1 this causes an off-by-one error and the cider-debug code thinks the source has been edited.

I think the proper solution to this would be to remove the above line in cider-nrepl and offset the 1-indexed response in Emacs. This may break other clients, in particular Calva.

### CIDER version information
```
;; CIDER 0.25.0snapshot, nREPL 0.7.0
;; Clojure 1.10.1, Java 14.0.1
```
### Emacs version

26.3

### Operating system

macOS 10.14

yuhan0:
Fixes #2852. 

The information about column numbers seemed too implementation-specific to place in debugger docs, so I put it in Troubleshooting instead.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#2852](https://github.com/clojure-emacs/cider/issues/2852): Convert 1-based column numbers in response map to Emacs' 0-based system.
+
 ## 0.25.0 (2020-06-04)
 
 ### New features
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -624,7 +624,10 @@ is a coordinate measure in sexps."
               (save-excursion
                 ;; Get to the proper line & column in the file
                 (forward-line (- line (line-number-at-pos)))
-                (move-to-column column)
+                ;; Column numbers in the response start from 1.
+                ;; Convert to Emacs system which starts from 0
+                ;; Inverse of `cider-column-number-at-pos'.
+                (move-to-column (max 0 (1- column)))
                 ;; Check if it worked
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #2857, opened 2020-06-14
[Fix #2527] Enable configuring REPL auto-clear

### LINKED ISSUE #2527: Have a configuration option, so that cider auto-erases the repl buffer when it becomes too large
The well known problem of emacs/cider's  behavior on printing long or complex data structures in the repl 
is very frustrating.

There are several efforts on going to solve this on the level of truncating the output automatically, in the nrepl for example.
The latest attempt is this:
https://github.com/nrepl/nrepl/issues/75

From my proper experiments, I think it is not enough to limit this it the nrepl allone.
(as a nrepl solution does not solve the problem, that the repl buffer grows and grows over time, and sometimes it fills too quickly for the user to react)

See the reasoning and a potiental, very simple, solution, based on automatically erasing the 
repl buffer here:

https://github.com/nrepl/nrepl/issues/75#issuecomment-440999775

The proposed solution helps as well for #1115

cursande:
Enables setting a limit to the buffer size of the REPL buffer (off by default), which when exceeded after printing out a result or other output, will clear old repl output from the beginning to ensure it stays within the limit. When clearing it also ensures partial input or output is erased.

Tested it with logging smaller/larger outputs every millisecond via [tools.logging](https://github.com/clojure/tools.logging), haven't come across any performance issues so far. 

Currently it's calling clear after new output is printed. May make more sense for that call to come before printing new output?

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -670,7 +670,7 @@ Each functions takes a string and must return a modified string.  Also see
   "The max size of the REPL buffer.
 Setting this to nil removes the limit."
   :group 'cider
-  :type 'boolean
+  :type 'integer
   :package-version '(cider . "0.26.0"))
 
 (defun cider-start-of-next-prompt (point)
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -680,20 +680,22 @@ Setting this to nil removes the limit."
         next-prompt-or-input
       (next-single-char-property-change next-prompt-or-input 'field))))
 
-(defun cider-repl-trim-buffer-to-below-limit (buffer)
-  "Trims REPL output from beginning of BUFFER.
+(defun cider-repl-trim-top-of-buffer (buffer)
+  "Trims REPL output from beginning of BUFFER by one fifth of `cider-repl-buffer-size-limit'.
 Also clears remaining partial input or results."
-  (with-current-buffer buffer
+  (if cider-repl-buffer-size-limit
+      (with-current-buffer buffer
     (let* ((to-trim (ceiling (* cider-repl-buffer-size-limit 0.2)))
            (start-of-next-prompt (cider-start-of-next-prompt to-trim))
            (inhibit-read-only t))
-      (cider-repl--clear-region (point-min) start-of-next-prompt))))
+      (cider-repl--clear-region (point-min) start-of-next-prompt)))
+    (message "cider-repl-buffer-size-limit is not set.")))
 
 (defun cider-repl-trim-buffer ()
   "Trim the currently visited REPL buffer partially from the top.
 See also `cider-repl-clear-buffer'."
   (interactive)
-  (cider-repl-trim-buffer-to-below-limit (current-buffer)))
+  (cider-repl-trim-top-of-buffer (current-buffer)))
 
 (defun cider-repl-trim-buffer-if-limit-exceeded (buffer)
   "Clears portion of printed output in BUFFER when `cider-repl-buffer-size-limit' is exceeded."
@@ -1528,6 +1530,7 @@ constructs."
 (cider-repl-add-shortcut "clear" #'cider-repl-clear-buffer)
 (cider-repl-add-shortcut "clear-banners" #'cider-repl-clear-banners)
 (cider-repl-add-shortcut "clear-help-banner" #'cider-repl-clear-help-banner)
+(cider-repl-add-shortcut "trim" #'cider-repl-trim-buffer)
 (cider-repl-add-shortcut "ns" #'cider-repl-set-ns)
 (cider-repl-add-shortcut "toggle-pprint" #'cider-repl-toggle-pretty-printing)
 (cider-repl-add-shortcut "toggle-font-lock" #'cider-repl-toggle-clojure-font-lock)
@@ -1716,6 +1719,7 @@ constructs."
         ["Previous prompt" cider-repl-previous-prompt]
         ["Clear output" cider-repl-clear-output]
         ["Clear buffer" cider-repl-clear-buffer]
+        ["Trim buffer" cider-repl-trim-buffer]
         ["Clear banners" cider-repl-clear-banners]
         ["Clear help banner" cider-repl-clear-help-banner]
         ["Kill input" cider-repl-kill-input]
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -671,12 +671,11 @@ Each functions takes a string and must return a modified string.  Also see
 Setting this to nil removes the limit."
   :group 'cider
   :type 'boolean
-  :package-version '(cider . "0.25.0"))
+  :package-version '(cider . "0.26.0"))
 
 (defun cider-clear-old-repl-output-if-limit-exceeded ()
   "Clears printed output to meet `cider-repl-buffer-size-limit'.
-If limit exceeded, clears any partial input or results until under the limit.
-Clears partial input or results that remain."
+Also clears remaining partial input or results."
   (let ((size (buffer-size)))
     (when (> size cider-repl-buffer-size-limit)
       (let* ((over-limit (- size cider-repl-buffer-size-limit))
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -673,19 +673,32 @@ Setting this to nil removes the limit."
   :type 'boolean
   :package-version '(cider . "0.26.0"))
 
-(defun cider-clear-old-repl-output-if-limit-exceeded ()
-  "Clears printed output to meet `cider-repl-buffer-size-limit'.
+(defun cider-start-of-next-prompt (point)
+  "Return the position of the first char of the next prompt from POINT."
+  (let ((next-prompt-or-input (next-single-char-property-change point 'field)))
+    (if (eq (get-char-property next-prompt-or-input 'field) 'cider-repl-prompt)
+        next-prompt-or-input
+      (next-single-char-property-change next-prompt-or-input 'field))))
+
+(defun cider-repl-trim-buffer-to-below-limit (buffer)
+  "Trims REPL output from beginning of BUFFER.
 Also clears remaining partial input or results."
-  (let ((size (buffer-size)))
-    (when (> size cider-repl-buffer-size-limit)
-      (let* ((over-limit (- size cider-repl-buffer-size-limit))
-             (next-prompt-or-input (next-single-char-property-change over-limit 'field))
-             (next-prompt-after-limit (if (eq (get-char-property next-prompt-or-input 'field) 'cider-repl-prompt)
-                                          next-prompt-or-input
-                                        (next-single-char-property-change next-prompt-or-input 'field)))
-             (inhibit-read-only t))
-        (cider-repl--clear-region (point-min) next-prompt-after-limit)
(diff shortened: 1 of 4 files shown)


cider-repl.el
```diff
@@ -682,7 +704,9 @@ Before inserting, run `cider-repl-preoutput-hook' on STRING."
       (when (and (= (point) cider-repl-prompt-start-mark)
                  (not (bolp)))
         (insert-before-markers "\n")
-        (set-marker cider-repl-output-end (1- (point))))))
+        (set-marker cider-repl-output-end (1- (point))))
+      (when cider-repl-buffer-size-limit
```
bbatsov, line 708: I think this should go to the `done` handler, otherwise it would be called repeatedly for each chunks of output, which seems excessive to me.

cider-repl.el
```diff
@@ -734,7 +758,9 @@ of the line.  If BOL is non-nil insert at the beginning of the line."
             (insert-before-markers (cider-font-lock-as-clojure string))
           (cider-propertize-region
               '(font-lock-face cider-repl-result-face rear-nonsticky (font-lock-face))
-            (insert-before-markers string)))))))
+            (insert-before-markers string)))
+        (when cider-repl-buffer-size-limit
```
bbatsov, line 762: Move it to the done handler will also solve this as well (because done is triggered once the evaluation is done).

cider-repl.el
```diff
@@ -666,6 +666,28 @@ namespaces.  STRING is REPL's output."
 Each functions takes a string and must return a modified string.  Also see
 `cider-run-chained-hook'.")
 
+(defcustom cider-repl-buffer-size-limit nil
+  "The max size of the REPL buffer.
+Setting this to nil removes the limit."
+  :group 'cider
+  :type 'boolean
+  :package-version '(cider . "0.25.0"))
```
bbatsov, line 674: That should say 0.26.0 (0.25 was released already)

cider-repl.el
```diff
@@ -666,6 +666,28 @@ namespaces.  STRING is REPL's output."
...
+
+(defun cider-clear-old-repl-output-if-limit-exceeded ()
+  "Clears printed output to meet `cider-repl-buffer-size-limit'.
+If limit exceeded, clears any partial input or results until under the limit.
+Clears partial input or results that remain."
+  (let ((size (buffer-size)))
+    (when (> size cider-repl-buffer-size-limit)
+      (let* ((over-limit (- size cider-repl-buffer-size-limit))
+             (next-prompt-or-input (next-single-char-property-change over-limit 'field))
```
bbatsov, line 683: Not sure I get the idea here. Won't it be easier to search for some marker - e.g. `repl-input-start` or `repl-output-end`?

bbatsov:
> Currently it's calling clear after new output is printed. May make more sense for that call to come before printing new output?

Ah, I commented before seeing this. I'm assuming your logger output doesn't produce "done", which is why you put the trimming in both the output and the result printing functions, right?

cider-repl.el
cursande: That does sound easier, I was just ignorant. I was basing it on the implementation of `cider-repl--find-prompt` to walk forward to find the next prompt

cursande:
Currently I'm testing using a project just spun up, with java.util.logging driving it. The `("done")` message is not received until the evaluation is completed. I assume it's still the same behaviour across logging implementations like logback and SLF4J as far as CIDER is concerned? 
```
  (-->
  id                                 "13"
  op                                 "eval"
  session                            "5be5e907-d6cb-4e96-815d-6e9c2b5c7988"
  time-stamp                         "2020-06-14 22:06:20.239720587"
  code                               "(doseq [_ (range 10)]
    (Thread/sleep 10)
    (log/info (a..."
  column                             18
  file                               "*cider-repl projects/rando:localhost:36301(clj)*"
  line                               73
  nrepl.middleware.print/buffer-size 4096
  nrepl.middleware.print/options     (dict ...)
  nrepl.middleware.print/print       "cider.nrepl.pprint/pprint"
  nrepl.middleware.print/quota       1048576
  nrepl.middleware.print/stream?     "1"
  ns                                 "rando.core"
  )

  ...bunch of logging messages...

(<--
  id         "13"
  session    "5be5e907-d6cb-4e96-815d-6e9c2b5c7988"
  time-stamp "2020-06-14 22:06:20.423875248"
  value      "nil"
)
(<--
  id         "13"
  session    "5be5e907-d6cb-4e96-815d-6e9c2b5c7988"
  time-stamp "2020-06-14 22:06:20.424585793"
  ns         "rando.core"
)
(<--
  id         "13"
  session    "5be5e907-d6cb-4e96-815d-6e9c2b5c7988"
  time-stamp "2020-06-14 22:06:20.437525462"
  status     ("done")
)
(<--
  id                 "13"
  session            "5be5e907-d6cb-4e96-815d-6e9c2b5c7988"
  time-stamp         "2020-06-14 22:06:20.438275667"
  changed-namespaces (dict)
  repl-type          "clj"
  status             ("state")
)

```
I had added the fn for `cider-repl--emit-output` first to handle logging, only added after to accomodate printing eval results. I'm assuming having the done handler apply the limit would still be a better idea than doing it in `cider-repl-emit-result`? I'm still making sense of the broader code so definitely keen for input on what makes sense where

bbatsov:
> The ("done") message is not received until the evaluation is completed. I assume it's still the same behaviour across logging implementations like logback and SLF4J as far as CIDER is concerned?

Yeah, exactly. A single evaluation might generate a lot of output/result messages, but there's always going to be a single `done` message at the end. There are some edge cases where output can be delivered after "done" - e.g. if the main evaluation spins a background thread that keeps printing, but I guess those situations are not very common. 

> I had added the fn for cider-repl--emit-output first to handle logging, only added after to accomodate printing eval results. I'm assuming having the done handler apply the limit would still be a better idea than doing it in cider-repl-emit-result? I'm still making sense of the broader code so definitely keen for input on what makes sense where

Yep, definitely. For results it's certainly a bad idea to be doing some trimming in the result handler, as it'd be invoked many times for a big result that's printed in chunks to the REPL.

cider-repl.el
cursande: Was having a look at this earlier, think I'm still misunderstanding this. Is there a way to identify the earlier values of `repl-output-end` in the mark-ring?

cursande:
If it's also the frequency of calls to clear in `emit-output` that is problematic, would it be better to use a different strategy to clear e.g. clearing by some % of the limit instead of just trimming to bring it down to the current? I'm thinking specifically of small frequent log statements

cider-repl.el
bbatsov: If your current implementation works that's fine. Perhaps you can extract some helper function like "find-next-prompt" or just make `cider-repl--find-prompt` useful in your case. I'll need some time to figure out a simpler approach anyways, as I haven't touched this code in ages.

bbatsov:
> If it's also the frequency of calls to clear in emit-output that is problematic

Well, it really depends if people expect to gain something if output is trimmed in the middle of an evaluation or not. To me this sounds like an overkill, that's why I suggested to avoid it, but if there are some reasons to do so - I'd consider this approach as well.

> different strategy to clear e.g. clearing by some % of the limit instead of just trimming to bring it down to the current?

I'm not sure I follow, so some more details/examples would be useful.

bbatsov:
Btw, it might be nice to make the trim-to-limit thingy an interactive command, as some people might want to use this manually instead of clearing their REPL completely.

cursande:
> A single evaluation might generate a lot of output/result messages, but there's always going to be a single done message at the end

Case I'm thinking of is output generated from a mounted aleph server.

> Well, it really depends if people expect to gain something if output is trimmed in the middle of an evaluation or not. To me this sounds like an overkill, that's why I suggested to avoid it, but if there are some reasons to do so - I'd consider this approach as well.

For me I usually just clear the whole buffer when I feel it starts to lag, that's just how I've been working with it. Usually I use repl-history to re-eval if I need something from earlier in the session, so I don't have a strong feeling towards either approach. Whichever is less painful from a maintenance perspective I guess. 

> I'm not sure I follow, so some more details/examples would be useful.

At the minute it just finds the difference between the current buffer size and the limit, subtracts it and finds the next prompt after that. It crossed my mind that if you had a limit of say 10MB you could instead take say 20% of it, and so the buffer would not be cleared so often. But I don't know how expensive the calls to `buffer-size` would be effectively for every output print. I was generally thinking the sort of sliding window situation here could be replaced by something more aggressive.

> Btw, it might be nice to make the trim-to-limit thingy an interactive command, as some people might want to use this manually instead of clearing their REPL completely.
For sure, should be straightforward.

Also, thanks! Appreciate your patience with this, I know you've got a shitload to work on and oversee :smile:

cider-repl.el
cursande: Will do :+1:

bbatsov:
> Also, thanks! Appreciate your patience with this, I know you've got a shitload to work on and oversee 😄

No worries! I only wish I had enough time to oversee them more adequately. 😆 

> For me I usually just clear the whole buffer when I feel it starts to lag, that's just how I've been working with it. Usually I use repl-history to re-eval if I need something from earlier in the session, so I don't have a strong feeling towards either approach. Whichever is less painful from a maintenance perspective I guess.

Same here. From my perspective it's just simpler to have one place where the trimming happens instead of several place. I'm a big proponent of starting with the simplest solution first. 

> At the minute it just finds the difference between the current buffer size and the limit, subtracts it and finds the next prompt after that. It crossed my mind that if you had a limit of say 10MB you could instead take say 20% of it, and so the buffer would not be cleared so often. But I don't know how expensive the calls to buffer-size would be effectively for every output print. I was generally thinking the sort of sliding window situation here could be replaced by something more aggressive.

Yeah, that's a good point. I think that's going to be a more efficient approach that a simple sliding window and it doesn't really add complexity to the implementation.

cider-repl.el
```diff
@@ -666,6 +666,43 @@ namespaces.  STRING is REPL's output."
 Each functions takes a string and must return a modified string.  Also see
 `cider-run-chained-hook'.")
 
+(defcustom cider-repl-buffer-size-limit nil
```
bbatsov, line 669: You should also mention this in the REPL configuration section of the docs.

cider-repl.el
```diff
@@ -666,6 +666,43 @@ namespaces.  STRING is REPL's output."
...
+Also clears remaining partial input or results."
+  (if cider-repl-buffer-size-limit
+      (with-current-buffer buffer
+        (let* ((to-trim (ceiling (* cider-repl-buffer-size-limit 0.2)))
+               (start-of-next-prompt (cider-start-of-next-prompt to-trim))
+               (inhibit-read-only t))
+          (cider-repl--clear-region (point-min) start-of-next-prompt)))
+    (message "cider-repl-buffer-size-limit is not set.")))
+
+(defun cider-repl-trim-buffer ()
```
bbatsov, line 695: I'd mention this somewhere in the using the REPL docs.

bbatsov:
Great work! Thanks!

cursande:
Awesome :) I'll keep it in the back of my mind if I can see a nice way to clear stdout + stderr without calling too frequently as well

Issue #2845, opened 2020-05-18
Cider throws error when server is on Windows


Frozenlock:
## Expected behavior
Cider works as it would when running the server a linux machine.

## Actual behavior

Commands are sent but don't print anything back.
An error is thrown.

## Steps to reproduce the problem

Launch a server on a windows machine.

```clj
lein update-in :dependencies conj "[nrepl \"0.6.0\"]" -- update-in :plugins concat "[[cider/cider-nrepl \"0.24.0\"]]" -- update-in :pedantic? empty -- with-profile +dev,+test repl :headless :host localhost :port 6666
```
Then, from Emacs on your linux machine:
<kbd>M-x cider-connect</kbd> ...

Error:

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  compare-strings(nil 0 0 "/home/frozenlock/Docsolver/master/apps/basil/agent/" 0 0 nil)
  string-prefix-p(nil "/home/frozenlock/Docsolver/master/apps/basil/agent/")
  #f(compiled-function (path) #<bytecode 0x1136b4d>)(nil)
  #f(compiled-function (elt) #<bytecode 0x1136b65>)(nil)
  mapc(#f(compiled-function (elt) #<bytecode 0x1136b65>) (nil nil nil nil nil))
  seq-do(#f(compiled-function (elt) #<bytecode 0x1136b65>) (nil nil nil nil nil))
  seq-find(#f(compiled-function (path) #<bytecode 0x1136b4d>) (nil nil nil nil nil))
  #f(compiled-function (system session) #<bytecode 0x1bdf7d5>)(CIDER ("libs/bundle:vagrant-machine:6666" #<buffer *cider-repl libs/bundle:vagrant-machine:6666(clj)*>))
  apply(#f(compiled-function (system session) #<bytecode 0x1bdf7d5>) CIDER ("libs/bundle:vagrant-machine:6666" #<buffer *cider-repl libs/bundle:vagrant-machine:6666(clj)*>))
  sesman-friendly-session-p(CIDER ("libs/bundle:vagrant-machine:6666" #<buffer *cider-repl libs/bundle:vagrant-machine:6666(clj)*>))
```
### CIDER version information

```
;; CIDER 0.24.0 (India), nREPL 0.6.0
;; Clojure 1.10.0, Java 11.0.6
```
### Lein/Boot version
Leiningen 2.9.1 on Java 11.0.6 OpenJDK 64-bit Server VM

[...]

PR #2858, opened 2020-06-16
Workaround fix for #2845

kriyative:
Minor fix to filter out any errant `nil` entries in the classpath-roots introspected from the server.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -313,42 +313,43 @@ It delegates the actual error content to the eval or op handler."
 ;; Clojure 1.10.  That's why we're trying to match error messages to both the
 ;; old and the new format, by utilizing a combination of two different regular
 ;; expressions.
-(eval-when-compile
-  (defconst cider-clojure-1.10-error `(sequence
-                                       "Syntax error "
-                                       (minimal-match (zero-or-more anything))
-                                       "compiling "
-                                       (minimal-match (zero-or-more anything))
-                                       "at ("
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -313,36 +313,38 @@ It delegates the actual error content to the eval or op handler."
...
-                                     (group-n 2 (minimal-match (zero-or-more anything)))
-                                     ":"
-                                     (group-n 3 (one-or-more digit))
-                                     (optional ":" (group-n 4 (one-or-more digit)))
-                                     ")."))
-
-(defconst cider-clojure-1.9-error `(sequence
-                                    (zero-or-more anything)
-                                    ", compiling:("
+(eval-when-compile
```
bbatsov, line 316: Any idea why this is needed? I know the error it's supposed to solve, but I don't get how come this is not getting evaled normally anyways.

kriyative: Fair question - I added that somewhat reflexively but now that I think about it, this shouldn't be necessary. I'll try and look at this a bit more closely this weekend.

kriyative: I think the issue is that `cider-clojure-1.9-error` et al were being referenced at macro expansion time of the `rx` and the `defconst`s haven't yet been eval'ed. Moving the `eval` up around the whole `rx` macro fixes the lint issue and preserves the intended behavior.

bbatsov:
Thanks!

PR #2861, opened 2020-06-17
Added the Krell REPL support

dotemacs:
This change adds the support for Krell REPL.

Instead of having to pick a custom REPL and specifying a
cider-custom-cljs-repl-init-form, this is all now included. All the
user has to do is pick the REPL type, which in this case is krell.

I suspect that there might need to be issues to be considered:

- Is the way the "payload" is created, as in the require form, all on separate lines, is that OK? Or do you want it all on one line?
- If the above is OK, I'll move onto adding the documentation in the docs.

Please advise, thanks

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -689,6 +689,12 @@ Generally you should not disable this unless you run into some faulty check."
   (unless (cider-library-present-p "adzerk.boot-cljs-repl")
     (user-error "The Boot ClojureScript REPL is not available.  Please check https://github.com/adzerk-oss/boot-cljs-repl/blob/master/README.md for details")))
 
+(defun cider-check-krell-requirements ()
+  "Check whether we can start a Krell ClojureScript REPL."
+  (cider-verify-piggieback-is-present)
+  (unless (cider-library-present-p "krell.repl")
+    (user-error "The Krell ClojureScript REPL is not available.  Please check https://github.com/vouch-opensource/krell for details")))
+
 (defun cider-check-shadow-cljs-requirements ()
   "Check whether we can start a shadow-cljs REPL."
   (unless (cider-library-present-p "shadow.cljs.devtools.api")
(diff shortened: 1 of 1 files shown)


bbatsov:
From what I get `krell` can't define project types, so I'm guessing it should be defined as ClojureScript REPL type (e.g. the way figwheel is defined), not like something at the `cider-jack-in` level.

dotemacs:
The test failures are lint failures, related to the code that my change does not touch:

cider-eval.el:347 - https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider-eval.el#L347
cider.el:364 - https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider.el#L364

dotemacs:
> From what I get `krell` can't define project types, so I'm guessing it should be defined as ClojureScript REPL type (e.g. the way figwheel is defined), not like something at the `cider-jack-in` level.

Hi @bbatsov, do you mean these:

- https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider.el#L349
- https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider.el#L360
- https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider.el#L381

bbatsov:
Yep. I think that only the cljs part the setup is actually needed (https://github.com/clojure-emacs/cider/blob/fdc8d69b4434ff26c8ac8fc4d88670be77911c67/cider.el#L825 + the cljs requirements check). That way you'll be able to pick `krell` regardless of your build tool.

dotemacs:
Great, thanks for confirming!

dotemacs:
Cleaned up the PR, please have a look.

A things to consider:

The way the require clause is provided, it isn't all on one line like it is for the other REPL types. Would you like that all flattened out to a single line? I'm talking about this https://github.com/clojure-emacs/cider/pull/2861/files#diff-00bd1ee3b6816bf7f502bc824ec4a6e9R834

bbatsov:
> The way the require clause is provided, it isn't all on one line like it is for the other REPL types. Would you like that all flattened out to a single line? I'm talking about this https://github.com/clojure-emacs/cider/pull/2861/files#diff-00bd1ee3b6816bf7f502bc824ec4a6e9R834

It's fine as it is. 

Just update the changelog and add some docs about Krell (so that people know CIDER supports it now) and we're good to go.

bbatsov:
@dotemacs ping :-)

dotemacs:
Pong.

I’ll do it either today or this weekend.

Thanks

bbatsov:
I'm planning to cut a new release soon, so I'll merge this as is. You can add the docs later.

Issue #2864, opened 2020-06-19
Consider replacing Cask with Eldev


bbatsov:
Mostly a note to myself, as Cask seems to be close to abandonware these days. See https://github.com/doublep/eldev

@doublep has provided some example projects here https://github.com/doublep/eldev#example-projects Seems it's not hard to set this up.

doublep:
If you want, I can create a PR for this.

bbatsov:
@doublep That'd be great!

PR #2865, opened 2020-06-19
Use Eldev.

doublep:
As mentioned in #2864.

For now:
* configure Eldev, but don't remove anything else;
* MELPA Unstable is explicitly used (otherwise Eldev would prefer Stable, but at least one test fails with that; though this might also indicate too lax requirements in package declaration);
* added a loading root for tests (see e.g. `-L ./test/utils/` in `Makefile`: so it's nothing new);
* file `test/cider-tests--no-auto.el` has to be explicitly excluded, because otherwise Eldev would use everything in `test/` subdirectory;
* form `(setf eldev-build-load-before-byte-compiling t)` *is* needed, without it CIDER cannot be byte-compiled.

You can immediatly run e.g. `eldev test` - works out-of-the-box.  Use e.g. `eldev test cider-find-tests.el` to run tests from a single file. Use e.g. `eldev test nrepl` to run tests with string "nrepl" in their description.

One test is excluded if source code is not byte-compiled (stack overflow otherwise). With `eldev -c test` you can run tests on byte-compiled source code. It is also possible to make this the default if you want.

If this looks fine, I can continue with continuous integration. (Probably no point to merge yet.)


bbatsov:
> MELPA Unstable is explicitly used (otherwise Eldev would prefer Stable, but at least one test fails with that; though this might also indicate too lax requirements in package declaration);

That's fine. Any idea which package is causing the breakage?

> file test/cider-tests--no-auto.el has to be explicitly excluded, because otherwise Eldev would use everything in test/ subdirectory;

Probably you can add a comment about this, so it's clearer in the config file.

> You can immediatly run e.g. eldev test - works out-of-the-box. Use e.g. eldev test cider-find-tests.el to run tests from a single file. Use e.g. eldev test nrepl to run tests with string "nrepl" in their description.

Don't we need to specify somewhere dev deps like `buttercup`? 

> If this looks fine, I can continue with continuous integration. (Probably no point to merge yet.)

Let's make this happen! :-)

test/nrepl-bencode-tests.el
```diff
@@ -278,6 +278,9 @@ If object is incomplete, return a decoded path."
     :var (nrepl--toString-dicts nrepl--toString-strings)
 
     (it "decodes the structures"
+      (unless (byte-code-function-p (symbol-function 'nrepl--bdecode-message))
+        (buttercup-skip "[this test fails if source code is not byte-compiled]"))
```
bbatsov, line 282: I wonder why this works on the CI, though. I don't think we byte-compile the tests there.

doublep:
> Don't we need to specify somewhere dev deps like buttercup?

No, Eldev can detect itself. But it can be set explicitly if you want.

> I wonder why this works on the CI, though. I don't think we byte-compile the tests there.

In current setup all `.el` files are explicitly byte-compiled before testing. With Eldev I'd propose to leave it up to the tool. On CI server it can always be executed as `eldev -c` (byte-compiled) or `eldev -p` (packaged and "installed"), but I'd suggest to leave non-CI usage as not enforcing compilation. As I said, we can also change the *default* to be byte-compiled though.

doublep:
> That's fine. Any idea which package is causing the breakage?

No, sorry. It was something about `symbol-at-point`.

test/nrepl-bencode-tests.el
doublep: Ah, it's not about compiling the test files, it's about the source code.

doublep:
It now fails because of many compilation warnings (referencing undefined functions, variables). I'm not exactly sure how it worked before, might be because of different compilation order, for example. Or maybe because of `cider-autoloads.el` which was created by side-effect (?) at some point.

bbatsov:
> It now fails because of many compilation warnings (referencing undefined functions, variables). I'm not exactly sure how it worked before, might be because of different compilation order, for example.

No idea myself. :D That's one of the more annoying aspects of working with Elisp.

> Or maybe because of cider-autoloads.el which was created by side-effect (?) at some point.

I see that's created in the makefile explicitly, but probably mostly for the benefit of some end users who needed one. 

Apart from the warnings the output seems pretty clean and neat, which is a massive improvement over Cask in my book. :-) Might be a good idea to update the commands in the makefile to refer to eldev as well and we can purge cask completely.

doublep:
> > It now fails because of many compilation warnings (referencing undefined functions, variables). I'm not exactly sure how it worked before, might be because of different compilation order, for example.
>
> No idea myself. :D That's one of the more annoying aspects of working with Elisp.

Well, the question is how do we fix it? These warnings/errors also show up e.g. if you use Flycheck on the source code, so maybe it's better to make the code clean in this respect first (separate PR)? Another option is to try emulating the current approach — this would mean less changes in the source code, but more complicated infrastructure.

> Apart from the warnings the output seems pretty clean and neat, which is a massive improvement over Cask in my book. :-)

The main purpose of Eldev (why I even started it) was to run regression tests locally, i.e. not on a CI server, but also outside your normal Emacs, which quickly accumulates obsolete definitions, deleted functions (not present in source code, but evaluated once and thus staying in memory) and so on. So, I'm never confident of running tests in Emacs I use normally — I use that for debugging sometimes, but I don't trust the results to be "real".

By the way, for CI commands I usually pass `-t` (`--trace`) to Eldev which makes its output quite verbose: this way if something goes wrong during CI, I have more information to understand it.  But you can also remove it if you don't want it, to make the output even cleaner.

> Might be a good idea to update the commands in the makefile to refer to eldev as well and we can purge cask completely.

The idea of Eldev is to replace *both* Cask and `Makefile`, so that you can run the same or at least similar to CI commands locally before even commiting/pushing. But of course we can keep `Makefile` for compatibility. However, if you invoke Eldev via `Makefile`, you lose access to all the various options etc. Eldev provides.

Also, Eldev configuration is a program, so you can add project-specific commands, build targets etc. to it. Which in some way replaces functionality of `Makefile`, just in Elisp instead of shell script.

bbatsov:
Sorry for the slow response - had a few super busy days.

> Well, the question is how do we fix it? These warnings/errors also show up e.g. if you use Flycheck on the source code, so maybe it's better to make the code clean in this respect first (separate PR)? Another option is to try emulating the current approach — this would mean less changes in the source code, but more complicated infrastructure.

I'd go with a cleanup first approach. I assume tackling those errors won't be complex/a lot of work. 

> The main purpose of Eldev (why I even started it) was to run regression tests locally, i.e. not on a CI server, but also outside your normal Emacs, which quickly accumulates obsolete definitions, deleted functions (not present in source code, but evaluated once and thus staying in memory) and so on. So, I'm never confident of running tests in Emacs I use normally — I use that for debugging sometimes, but I don't trust the results to be "real".

That's a great main purpose. That's also the main reason I've adopted Cask + ert-runner in the past, and eventually `buttercup`. 

> By the way, for CI commands I usually pass -t (--trace) to Eldev which makes its output quite verbose: this way if something goes wrong during CI, I have more information to understand it. But you can also remove it if you don't want it, to make the output even cleaner.

Using `-t` is fine. The output seems pretty clean to me even now.

> The idea of Eldev is to replace both Cask and Makefile, so that you can run the same or at least similar to CI commands locally before even commiting/pushing. But of course we can keep Makefile for compatibility. However, if you invoke Eldev via Makefile, you lose access to all the various options etc. Eldev provides.

Pretty cool, I didn't realize this at first. Well, I guess it makes sense to eventually remove the makefile completely and update the dev docs to refer to equivalent `eldev` commands.

doublep:
I now better understand the purpose of `autoloads` in the `Makefile`. I think it is best if we postpone this a couple of weeks until I integrate this stuff to Eldev (and also the linter that you use).

bbatsov:
Sure. No rush.

doublep:
New iteration requires Eldev 0.6. I also fully removed `Makefile` and `Cask` as replaced. I tweaked `hacking.adoc` a bit, to replace references to the old commands.

Some changes from #2868 were actually not _really_ needed, because with autoloads they are not really warnings. If you use [flycheck-eldev](https://github.com/flycheck/flycheck-eldev), even Flycheck will now use autoload file (and even automatically updated) and thus not warn about undeclared, but autoloaded functions.

bbatsov:
> Some changes from #2868 were actually not really needed, because with autoloads they are not really warnings. If you use flycheck-eldev, even Flycheck will now use autoload file (and even automatically updated) and thus not warn about undeclared, but autoloaded functions.

Nice! I'll definitely check it out! 

Great work! :bowing_man:

PR #2868, opened 2020-06-25
Eliminate various byte-compilation warnings

doublep:
As decided in #2865. This cleans up byte-compilation in preparation for switch to Eldev. As a side-effect, Flycheck is also happier and doesn't report the warnings anymore.


bbatsov:
Thanks!

PR #2870, opened 2020-06-29
[Fix #2718] Avoid printing comment if empty result

### LINKED ISSUE #2718: Evaling an erroring form into a comment
Evaling the form `(throw (ex-info "foo" {}))` with `M-x cider-pprint-eval-last-sexp-to-comment` yields the empty comment string:

```
;; =>
```

Arguably it should either:

1. Not yield a `;; =>` at all (just the empty string) (and appropriate cider error buffer)

or it should:

2. Print the error in the comment.

The later might be useful when describing code interactions and REPL sessions; though I think the former might be a more sensible default for normal usage.

cursande:
Small fix for https://github.com/clojure-emacs/cider/issues/2718

If an error is returned when `pprint-eval-last-sexp-to-comment` is called, the result string will be empty. While it may occasionally make sense to print out the exception, typically people will only want the successful evaluation, without having to clean up the empty comment string.


bbatsov:
Thanks!

PR #2873, opened 2020-07-06
[Fix #2796] Disable debug if CIDER disconnected

### LINKED ISSUE #2796: Closing CIDER connection should end debugging session, quitting should clear menu overlays
# Expected behavior
Closing CIDER connection should end debugging session
## Actual behavior
**Part 1** As seen in this screenshot, I have an active rectangle showing that I am debugging `response`, but in the mode line cider is not connected.

response

**Part 2** And here is another one where I still have the overlay of commands, but have been able to edit the buffer.

overlay-debugger

**1.** Cider jack in, debug function, cider-quit. Notice that if you jack in again, the rectangle goes away, so it's not persisted across sessions - is it useful for anything?

**2.** Here I pressed "q" to quit the debugging session but it didn't quit properly for some reason (sorry, I'm not sure of reproduction steps in this case).

### CIDER version information

```
;; CIDER 0.24.0snapshot (package: 20200208.809), nREPL 0.6.0
;; Clojure 1.9.0, Java 1.8.0_242
```

### Lein/Boot version

Leiningen 2.9.1 on Java 1.8.0_242 OpenJDK 64-Bit Server VM

### Emacs version

GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30, cairo version 1.15.10) of 2020-02-04

### Operating system

Ubuntu 18.04

cursande:
On closing CIDER connection, we currently disable cider-mode across all clojure-mode buffers, however we don't disable debug-mode if active.

Even without this change I noticed that when closing the CIDER connection during a debugging sessoion, a buffer appears for a split second before being deleted, tracing it back I can see it's an unhandled `java.lang.InterruptedException` that gets thrown when the input promise is deref'd in cider-repl. This seems fine as is?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ macros, special forms and methods.
 * Show eldoc for `.` and `..`.
 * [#2860](https://github.com/clojure-emacs/cider/issues/2860): Don't send blank strings in `eldoc` requests.
 * [#2718](https://github.com/clojure-emacs/cider/issues/2718): When calling `cider-pprint-eval-last-sexp-to-comment`, avoid printing empty comment if eval throws error.
+* [#2796](https://github.com/clojure-emacs/cider/issues/2796): Closing CIDER connection will disable the debug minor mode on clojure buffers.
 
 ## 0.25.0 (2020-06-04)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

> Even without this change I noticed that when closing the CIDER connection during a debugging sessoion, a buffer appears for a split second before being deleted, tracing it back I can see it's an unhandled java.lang.InterruptedException that gets thrown when the input promise is deref'd in cider-repl. This seems fine as is?

Nobody has complained about it, so I guess we're good. :-)

PR #2878, opened 2020-07-23
Improve handling of carriage return ("\r")

### LINKED ISSUE #1677: REPL doesn't interpret carriage return

plexus:
A CARRIAGE RETURN character moves the head of the teletype printer (the
carriage) to the beginning of the line, without scrolling the paper (this is
done by the LINE FEED "\n")

In a contemporary terminal emulator this translates to moving the cursor to the
beginning of the current line. Any subsequent output then overwrites what was
already on that line. This is how terminal progress bars are implemented.

The existing implementation incorrectly treated "\r" as "\n", thus inserting a
newline, instead of moving to the beginning of the current line.

Instead iterate over the output and deal with "\r" and "\n" the way terminals
do, "\r" (ASCII 13) moves to the beginning of the line, "\n" (ASCII 10) moves to
the beginning of the next line. Any output overwrites output on the same line
after the cursor position.

Sample code:

```
(println "aaaa\rbb\ncccc\rdd")
;; Outputs:
;; bbaa
;; ddcc

(dotimes [i 20]
  (print "\r[" (inc i) "]")
  (flush)
  (Thread/sleep 500))
;; prints a counter in place
```

Feedback welcome as I'm not sure of the performance impact of dealing with each
character individually. Also not sure what the effect is of propertizing and
running cider-repl-preoutput-hook per character.

Fixes #1677 (which was closed but not actually implemented, only worked around)

-----------------

```
➜ make test
make: Nothing to be done for 'test'.

➜ make lint
make: *** No rule to make target 'lint'.  Stop.
```

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -745,11 +745,24 @@ Before inserting, run `cider-repl-preoutput-hook' on STRING."
     (save-excursion
       (cider-save-marker cider-repl-output-start
         (goto-char cider-repl-output-end)
-        (setq string (propertize string
-                                 'font-lock-face face
-                                 'rear-nonsticky '(font-lock-face)))
-        (setq string (cider-run-chained-hook 'cider-repl-preoutput-hook string))
-        (insert-before-markers string)
+        (seq-map
+         (lambda (char)
+           (cond
+            ((= 13 char)
+             (beginning-of-line))
+            ((= 10 char)
+             (end-of-line)
+             (insert-before-markers "\n"))
+            (t
+             (when (< (point) (line-end-position))
+               (delete-char 1))
+             (setq char (char-to-string char))
+             (setq char (propertize char
+                                    'font-lock-face face
+                                    'rear-nonsticky '(font-lock-face)))
+             (setq char (cider-run-chained-hook 'cider-repl-preoutput-hook char))
+             (insert-before-markers char))))
+         string)
         (cider-repl--flush-ansi-color-context))
       (when (and (= (point) cider-repl-prompt-start-mark)
                  (not (bolp)))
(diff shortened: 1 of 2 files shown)


bbatsov:
> Feedback welcome as I'm not sure of the performance impact of dealing with each
character individually.

I have a feeling the impact will be quite significant on big chunks of output, but without runnings some benchmarks that's just a guess. 

> Also not sure what the effect is of propertizing and
running cider-repl-preoutput-hook per character.

I guess it's going to break it, as currently the hook functions (like `ansi-color-apply` expect the complete output string). On the other hand that might be somewhat broken even now for streamed output, as we can't really know where the chunk boundaries are going to fall - in some cases the filter functions might produce weird results for the individual chunks. 

One idea that comes to my mind is to look for `\r` in the output before trying to do something special. This will probably cover most scenarios. I'm also wondering how is this handled in a typical shell, as this might give us some inspiration.

bbatsov:
> Uh are these instructions up to date? I'm getting:

Ops, sorry about that! We just switch to `eldev` and removed the Makefile.

PR #2881, opened 2020-08-01
Add feature to evaluate list around point

gveres:
This adds a new function cider-eval-list-at-point,  that evaluates a list around point [as in: code delimited by the nearest pair of parens surrounding point]. This comes handy to evaluate a function call while the point is anywhere within the function call.
It's modelled after the eval-sexp-at-point solution, and as such, deals with the "deref @" and other quirks intelligently.

In terms of the implementation, it's built on a thingatpoint built-in feature, so no magic text manipulation is happening. Although the thingatpoint code targets Emacs Lisp, it should work fine with Clojure code too.

Sorry, no tests (yet) - I'd appreciate some help or guidance on what to cover for this.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Isn't `cider-eval-sexp-at-point` pretty much the same?

gveres:
> Isn't `cider-eval-sexp-at-point` pretty much the same?

No: as I understand, a symbol is also a sexp in the implementation, so if point is on a symbol, that symbol get evaluated, not the surrounding list. list-at-point will expand to include the innermost surrounding paren pair, and sends that whole thing ("compound form"?) to eval.

bbatsov:
Yeah, I think you're right. I use only eval-last-sexp and eval-defun, so my memory on everything else is fuzzy. I'll merge this as is and I'll add a couple of tests myself.

gveres:
Amazing - thanks a lot, @bbatsov!

PR #2886, opened 2020-08-05
The "browser" cljs-repl-type does not require node

plexus:
Remove the check for node requirements


bbatsov:
Probably some copy/paste leftover. Thanks!

PR #2889, opened 2020-08-13
Fix a typo in cider-info-form

j-cr:
This also fixes babashka/babashka.nrepl#26

The commit where the bug was introduced: https://github.com/clojure-emacs/cider/commit/4fd24b1e767c776ca9fe913121ecc55abe103c67

`cider-info-form` now correctly reads as:
```
(do
  (require 'clojure.java.io)
  (require 'clojure.walk)

  (if-let [var (resolve '%s)]
    (let [info (meta var)]
      (-> info
          (update :ns str)
          (update :name str)
          (update :file (comp str clojure.java.io/resource))
          (cond-> (:macro info) (update :macro str))
          (cond-> (:special-form info) (update :special-form str))
          (assoc :arglists-str (str (:arglists info)))
          (clojure.walk/stringify-keys)))))
```


bbatsov:
Thanks!

PR #2897, opened 2020-08-27
Translate paths from CIDER to nREPL and vice-versa

iarenaza:
When `cider-path-translations` is in use (e.g., when CIDER is running
on the host, but the nREPL is running in a virtual machine or a
container) we need to translate from the host "CIDER-based" paths to
the container "nREPL-based" paths and vice-versa, depending on the
operation.

Translations from "nREPL-based" paths to "CIDER-based" paths were
already implemented, but not the other way around. Some operations in
clj-refactor need the "CIDER-based"to "nREPL-based" path translations
to work correctly (e.g., `cljr-rename-file-or-dir`,
`cljr-rename-symbol`, etc.).

This patch implements the missing path translations, and is a
pre-requisite for the complementary clj-refactor patch that is
proposed separately.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -276,32 +276,46 @@ otherwise, nil."
 
 (defcustom cider-path-translations nil
   "Alist of path prefixes to path prefixes.
-Useful to intercept the location of a path in a docker image and translate
-to the oringal location.  If your project is located at \"~/projects/foo\"
-and the src directory of foo is mounted at \"/src\" in the docker
-container, the alist would be `((\"/src\" \"~/projects/foo/src\"))"
+Useful to intercept the location of a path in a container (or virtual
+machine) and translate to the oringal location.  If your project is located
+at \"~/projects/foo\" and the src directory of foo is mounted at \"/src\"
+in the container, the alist would be `((\"/src\" \"~/projects/foo/src\"))."
   :type '(alist :key-type string :value-type string)
   :group 'cider
   :package-version '(cider . "0.23.0"))
 
-(defun cider--translate-path (path)
-  "Attempt to translate the PATH.
-Looks at `cider-path-translations' for (docker . host) alist of path
-prefixes."
-  (seq-some (lambda (translation)
-              (let ((prefix (file-name-as-directory (expand-file-name (car translation)))))
-                (when (string-prefix-p prefix path)
-                  (replace-regexp-in-string (format "^%s" (regexp-quote prefix))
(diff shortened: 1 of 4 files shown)


bbatsov:
The changes look good to me overall, but if this is going to be public API used by other packages the functions shouldn't have `--` in their names. I also think we can expand the relevant documentation.

iarenaza:
Hi @bbatsov 

I didn't intend any of the new functions to be public. I saw them as just internal implementation details for `cider-from-nrepl-filename-function` and `cider-to-nrepl-filename-function`, which I thought was the public API for this functionality. Do you want them to be public?

Regarding the documentation expansion, do you mean the functions documentation or other kind of documentation?

cider-common.el
```diff
@@ -276,32 +276,46 @@ otherwise, nil."
...
-                                            (file-name-as-directory
-                                             (expand-file-name (cdr translation)))
-                                            path))))
-            cider-path-translations))
+(defun cider--translate-path (path direction)
+  "Attempt to translate the PATH in the given DIRECTION.
+Looks at `cider-path-translations' for (container . host) alist of path
+prefixes and translates PATH from container to host or viceversa depending on
+whether DIRECTION is :from-nrepl or :to-nrepl"
+  (seq-let [from-fn to-fn path-fn] (cond ((eq direction :from-nrepl) '(car cdr identity))
```
bbatsov, line 292: In Emacs Lisp we normally use symbols instead of keywords - e.g. `'from-nrepl`.

cider-common.el
```diff
@@ -276,32 +276,46 @@ otherwise, nil."
...
-                  (replace-regexp-in-string (format "^%s" (regexp-quote prefix))
-                                            (file-name-as-directory
-                                             (expand-file-name (cdr translation)))
-                                            path))))
-            cider-path-translations))
+(defun cider--translate-path (path direction)
+  "Attempt to translate the PATH in the given DIRECTION.
+Looks at `cider-path-translations' for (container . host) alist of path
+prefixes and translates PATH from container to host or viceversa depending on
+whether DIRECTION is :from-nrepl or :to-nrepl"
```
bbatsov, line 291: Don't forget the final `.`.

bbatsov:
My bad. I misread the implementation.

bbatsov:
Thanks for tackling this!

PR #2910, opened 2020-10-16
[Fix #2909] Add option to control auto selection of the inspector buffer

### LINKED ISSUE #2909: Avoid focusing the cider-inspect window
I always have a `*cider-inspect*` buffer open and I would like to avoid having to switch back to my `clojure-mode` window whenever I inspect a new value. `*cider-inspect*` window should not be focused by default or an custom variable should control that behavior.

mallt:
This PR adds a a new customization variable `cider-auto-select-inspector-buffer` to control the auto selection of the inspector buffer.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#2909](https://github.com/clojure-emacs/cider/issues/2909): Add new customization variable `cider-inspector-auto-select-buffer` to control the auto selection of the inspector buffer.
+
 ### Bugs fixed
 
 * Fix broken links to the docs in REPL warnings (the REPL links included the full CIDER version, but the docs URLs are without the patch version).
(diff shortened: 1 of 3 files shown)


cider-inspector.el
```diff
@@ -67,6 +67,12 @@ by clicking or navigating to them by other means."
   :group 'cider-inspector
   :package-version '(cider . "0.25.0"))
 
+(defcustom cider-auto-select-inspector-buffer t
```
bbatsov, line 70: I'd name this `cider-inspector-auto-select-buffer` or something like this, so the prefix is consistent within this module.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### New features
+[#2909](https://github.com/clojure-emacs/cider/issues/2909): Add new customization variable `cider-auto-select-inspector-buffer` to control the auto selection of the inspector buffer.
```
bbatsov, line 6: You forgot the `*` before the entry. Also, please add a blank line after the heading.

bbatsov:
Don't forget to mention the new option somewhere in the docs.

mallt:
@bbatsov Thank you for the review! I addressed your comments in the new commit.

bbatsov:
Thanks!

PR #2919, opened 2020-10-23
Fix regression in cider-info-form arglists-str

llacom:
Fix a regression in commit 1be3b8 where arglists-str is not a string anymore.
I didn't update Changelog as the issue was introduced after the latest release version.

The current code produces an error on emacs when trying to use `cider-doc`:

```
Debugger entered--Lisp error: (wrong-type-argument stringp ([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls]))
  string-match("\n" ([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls]) 0)
  split-string(([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls]) "\n")
  cider-docview-render-info(#<buffer *cider-doc*> (dict "ns" "clojure.core" "name" "map" "built-in" t "arglists" "([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 &..." "doc" "Returns a lazy sequence consisting of the result o..." "file" "" "arglists-str" ([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls])))
  cider-docview-render(#<buffer *cider-doc*> "map" (dict "ns" "clojure.core" "name" "map" "built-in" t "arglists" "([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 &..." "doc" "Returns a lazy sequence consisting of the result o..." "file" "" "arglists-str" ([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls])))
  cider-create-doc-buffer("map")
  cider-doc-lookup("map")
  cider-try-symbol-at-point("Doc for" cider-doc-lookup)
  cider-doc(nil)
  funcall-interactively(cider-doc nil)
  call-interactively(cider-doc nil nil)
  command-execute(cider-doc)
```

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -434,7 +434,7 @@ contain a `candidates' key, it is returned as is."
           (cond-> (:special-form info) (update :special-form str))
           (cond-> (:protocol info) (update :protocol str))
           (cond-> (:arglists info) (update :arglists str))
-          (assoc :arglists-str (:arglists info))
+          (assoc :arglists-str (str (:arglists info)))
           (clojure.walk/stringify-keys)))))
 ")
 


cider-client.el
```diff
@@ -434,7 +434,7 @@ contain a `candidates' key, it is returned as is."
           (cond-> (:special-form info) (update :special-form str))
           (cond-> (:protocol info) (update :protocol str))
           (cond-> (:arglists info) (update :arglists str))
-          (assoc :arglists-str (:arglists info))
+          (assoc :arglists-str (str (:arglists info)))
```
bbatsov, line 437: The actual mistake here is that `:arglists` has to be fetched from the current object, not `info`. (as it's already converted to string there)

llacom: I saw that, but I'm now sure how to fetch it from the current object in a readable way.
How about this?
```
(as-> $ (assoc $ :arglists-str (:arglists $)))
```

bbatsov: Fair enough. That's quite ugly. :-)

PR #2920, opened 2020-10-23
added cache for previous eval

l0st3d:
Added feature to cache the previously eval'd form and a function to eval that cached form (`cider-re-eval-previous-sexp'). Also changed the default keybinding for `C-c C-e` to point to this new function rather than `cider-eval-last-sexp`. The new function will do the same as the old one without a prefix arg. With a prefix arg, it'll re-run the last sexp that was eval'd (instead of changing the `OUTPUT-TO-CURRENT-BUFFER` behaviour). The old behaviour is still available with the keyboard combination `C-x C-e`.

Also added `cider-re-eval-previous-sexp` (with a keybinding of `"a"/"C-a"` in `cider-eval-commands-map`) which re-evals previous sexp to allow access to the `OUTPUT-TO-CURRENT-BUFFER` functionality of `cider-eval-last-sexp`.

Also added a short description of `cider-eval-last-sexp-in-context`, because I referred to it in the documentation I wrote for `cider-eval-last-sexp-or-re-eval-previous-sexp`.

---

Apologies if I've overstepped the mark by changing an existing keyboard binding. I'm quite prepared to put this back if it upsets anyone, I just thought that seeing as there was another keyboard binding for the old function and the new function only differs by a prefix arg, nobody would mind too much. The shortcut `C-u C-c C-e` is pretty convenient for me to type (I'm using a dvorak layout - some of the hany qwerty shortcuts are incovenient to type) so I thought it'd be ok to have access to a cached sexp and a non-cached sexp version of evaluation and that might be useful. Given different keybindings for the same functionality I figured it'd be a small change ... ;)

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/modules/ROOT/pages/usage/code_evaluation.adoc
+++ b/doc/modules/ROOT/pages/usage/code_evaluation.adoc
@@ -173,9 +173,17 @@ Below is a listing of most keybindings for evaluation commands:
 | Command | Keyboard shortcut | Description
 
 | `cider-eval-last-sexp`
-| kbd:[C-x C-e] +
-kbd:[C-c C-e]
+| kbd:[C-x C-e]
 | Evaluate the form preceding point and display the result in the echo area and/or in an buffer overlay (according to `cider-use-overlays`).  If invoked with a prefix argument, insert the result into the current buffer.
+
+| `cider-eval-last-sexp-or-re-eval-previous-sexp`
+| kbd:[C-c C-e]
+| Evaluate the form preceding point like `cider-eval-last-sexp` unless invoked with a prefix argument, in which case it will evaluate the previously evaluated expression. This includes any context stored by commands like `cider-eval-last-sexp-in-context`.
+
+| `cider-eval-last-sexp-in-context`
+| kbd:[C-c C-v C-c] +
+kbd:[C-c C-v c]
+| Eval the form preceding point with a containing lexical scope. The context is defined interactively in the minibuffer and evaluated as if provided in as `let` bindings surrounding the form to be evaluated. Useful for debugging deeply nested forms.
 
 | `cider-eval-last-sexp-and-replace`
 | kbd:[C-c C-v w]
--- a/doc/modules/ROOT/pages/usage/code_evaluation.adoc
+++ b/doc/modules/ROOT/pages/usage/code_evaluation.adoc
@@ -67,20 +67,20 @@ example:
 ----
 (defn long-and-complicated-function [x y]
   (let [r (range x)
-        l (map #(+ % y) r)] ;; <- form we want to eval independently of it's context
+        l (map #(+ % y) r)] ;; <- form we want to eval independently of its context
     (reduce + l)))
 ----
 
 If you position the point after the `map` form and evaluate it with
 kbd:[C-c C-e], you will get an error because you have undefined
 symbols `y` and `r`. You can provide a context in which to evaluate
-this expression by using kbd:[C-c C-v C-c], which is bound to
-`cider-eval-last-sexp-in-context`. This will prompt you at the
-minibuffer for a lexical context to evaluate the form in. This is like
-wrapping the form in a `let` binding before sending it for evaluation.
-In this case you could bind `y` to `1` and `r` to `[0 1 2 3 4]` by
-entering these binding when prompted to produce the result `=> (1 2 3
-4 5)`.
+this expression by using kbd:[C-c C-v C-c] or kbd:[C-c C-v c], which
+is bound to `cider-eval-last-sexp-in-context`. This will prompt you at
+the minibuffer for a lexical context to evaluate the form in. This is
+like wrapping the form in a `let` binding before sending it for
+evaluation. In this case you could bind `y` to `1` and `r` to `[0 1 2
+3 4]` by entering these binding when prompted to produce the result
+`=> (1 2 3 4 5)`.
 
 [source]
 ----
@@ -88,7 +88,7 @@ Evaluation context (let-style) for `(map #(+ % y) r)': y 1 r [0 1 2 3 4]
 ----
 
 After evaluating an expression, you can re-evaluate this expression in
-the same buffer with kbd:[C-c C-v C-a] or kbd:[C-c C-v ], both of
+the same buffer with kbd:[C-c C-v C-a] or kbd:[C-c C-v a], both of
 which are bound to `cider-re-eval-previous-sexp`. This is useful when
 working on a function with some test code nearby in the same buffer.
 For example:
--- a/doc/modules/ROOT/pages/usage/code_evaluation.adoc
+++ b/doc/modules/ROOT/pages/usage/code_evaluation.adoc
@@ -226,7 +226,7 @@ Below is a listing of most keybindings for evaluation commands:
 
 | `cider-eval-last-sexp`
 | kbd:[C-x C-e] +
-[C-c C-e]
+kbd:[C-c C-e]
 | Evaluate the form preceding point and display the result in the echo area and/or in an buffer overlay (according to `cider-use-overlays`).  If invoked with a prefix argument, insert the result into the current buffer.
 
 | `cider-eval-last-sexp-and-replace`
--- a/doc/modules/ROOT/pages/usage/code_evaluation.adoc
+++ b/doc/modules/ROOT/pages/usage/code_evaluation.adoc
@@ -173,17 +173,9 @@ Below is a listing of most keybindings for evaluation commands:
 | Command | Keyboard shortcut | Description
 
 | `cider-eval-last-sexp`
-| kbd:[C-x C-e]
+| kbd:[C-x C-e] +
+[C-c C-e]
 | Evaluate the form preceding point and display the result in the echo area and/or in an buffer overlay (according to `cider-use-overlays`).  If invoked with a prefix argument, insert the result into the current buffer.
-
-| `cider-eval-last-sexp-or-re-eval-previous-sexp`
-| kbd:[C-c C-e]
-| Evaluate the form preceding point like `cider-eval-last-sexp` unless invoked with a prefix argument, in which case it will evaluate the previously evaluated expression. This includes any context stored by commands like `cider-eval-last-sexp-in-context`.
-
-| `cider-eval-last-sexp-in-context`
-| kbd:[C-c C-v C-c] +
-kbd:[C-c C-v c]
-| Eval the form preceding point with a containing lexical scope. The context is defined interactively in the minibuffer and evaluated as if provided in as `let` bindings surrounding the form to be evaluated. Useful for debugging deeply nested forms.
 
 | `cider-eval-last-sexp-and-replace`
 | kbd:[C-c C-v w]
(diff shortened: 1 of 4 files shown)


l0st3d:
The linter failures seem to be problems with code I haven't changed, like missing docstrings in `cider-connection,el` and warnings about org-table ... should I try and fix these?

bbatsov:
What's the problem you're trying to solve by adding this cache?

l0st3d:
Part of my workflow at the moment includes writing a rich comment with some test code in near the function I'm working on. I write a bit of the function, and then run the code to test it, then fix my code and repeat until I think it works. So I often eval same bit of code dozens of times while trying to get something to work. I wanted a quick way to just re-run the last thing I did without moving the cursor or my focus away from the code I'm writing, even by the few lines that it currently is. For example:
```
(defn sum-of-squares [n]
  (->> (range n)
       (map #(* % %))
       (reduce +)))

(comment
  (sum-of-squares 10)
  )
```
While editing the `sum-of-squares` fn, I want to run the `(sum-of-squares 10)` snippet of code repeatedly without having to move my cursor, it's a distraction that I thought could be easily solved. Does that make sense?

I brought it up on slack : https://clojurians.slack.com/archives/C0617A8PQ/p1603478744286200 and got one thumbs up for the suggestion of a PR, hence this ;)

yuhan0:
This seems like quite a good addition - my usual workflow for the above is to temporarily wrap both the function being defined and the test command in a `do` block, and eval the entire top-level form (`C-c C-c`) from anywhere in the function. But this solution is definitely less of a hack .

```clj
(do
  (defn sum-of-squares [n]
    (->> (range n)
      (map #(* % %)) ;; <- cursor in the guts of a complicated function
      (reduce +)))

  (sum-of-squares 10) ;; test code that I would like to run
  )
```

The keybinding change supposes that users primarily use `cider-eval-last-sexp` for evaluation, which might be a bit arbitrary - for example I use eval-sexp-at-point or eval-defun-at-point 99% of the time. It's easy enough for users to define their own wrapper functions or advices, perhaps an example snippet in the docs could show how:

```elisp

(defun my/cider-eval-sexp-at-point (rerun)
  (interactive "P")
  (call-interactively (if rerun #'cider-re-eval-previous-sexp #'cider-eval-sexp-at-point)))

(define-key 'cider-mode-map
  [remap cider-eval-sexp-at-point] #'my/cider-eval-sexp-at-point)

```

l0st3d:
I chose `C-c C-e` because my test is wrapped in a `comment` block, so it's easiest to execute with that. I would generally use `C--c C-c` to eval the `defn`, but thinking about that, I generally rely on my save-hook to run `cider-load-buffer` instead. I'm quite happy to back out the key binding change. I did also add a key binding into `cider-eval-commands-map` of `C-c C-v a/C-a` ("a" for again because "r" was already taken). It just seemed really useful to me in the excitement of the moment and I got carried away ;) ... You are correct that I can just set the keyboard combination in my own config, so I'll do that.

bbatsov:
Now that I understand a bit better the problem at hand I'll suggest to use some different terminology instead of "caching", as what the command actually does is to store some sexp for quick re-evaluation. I think proper naming will make it much easier to understand the purpose of those commands for most users. 

I also feel strongly about re-binding one of our essential keybindings, so let's not do this and use something else. As noted above it's easy for everyone to change this locally if they want to.

doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -59,6 +59,58 @@ the source buffer.
...
 
+Sometimes, when debugging a function, it is useful to provide a
+surrounding lexical context outside of the code you're evaluating. For
+example:
+
+[source,clojure]
+----
+(defn long-and-complicated-function [x y]
+  (let [r (range x)
+        l (map #(+ % y) r)] ;; <- form we want to eval independently of it's context
```
bbatsov, line 70: its

bbatsov:
I also wondering it's not better to have a separate recall command instead of having a single command that either evals the previous sexp or recalls the stored sexp. Keep in mind that you're storing the sexp regardless of which eval command you used, so I think the current "dual" behaviour is misleading when in comes to what you can re-eval.

l0st3d:
> Now that I understand a bit better the problem at hand I'll suggest to use some different
> terminology instead of "caching", as what the command actually does is to store some sexp for
> quick re-evaluation. I think proper naming will make it much easier to understand the purpose of
> those commands for most users.

I'm happy to change the wording used. What do you think makes sense? Should we just stick to
"re-evaluation" or "re-run"? In the docs I've written for this feature, I've used these words. I've
removed all references to the word "cache" from this PR (including in the CHANGELOG).

> I also feel strongly about re-binding one of our essential keybindings, so let's not do this and
> use something else. As noted above it's easy for everyone to change this locally if they want to.

Yes, this is totally reasonable and I suspected that would be the case. I've reverted that change.

> I also wondering it's not better to have a separate recall command instead of having a single
> command that either evals the previous sexp or recalls the stored sexp. Keep in mind that you're
> storing the sexp regardless of which eval command you used, so I think the current "dual"
> behaviour is misleading when in comes to what you can re-eval.

Agreed. I have added a keybinding that will do that (`C-c C-v a` and `C-c C-v C-a`) which is bound
to `cider-re-eval-previous-sexp` in `cider-eval.el`. In reverting the keyboard shortcut change I've
deleted the new function (`cider-eval-last-sexp-or-re-eval-previous-sexp`) that wrapped
`cider-re-eval-previous-sexp` and `cider-eval-last-sexp`. If no sexp has been evaled in the current
buffer, `cider-re-eval-previous-sexp` will respond with:

```
(user-error "No sexp has been run in this buffer")
```

Does this match what you would expect?

dpsutton:
I'm a bit confused about this. If i'm understanding correctly, you want to edit a function and then rerun some test or invocation of that function. The plan is to place this in a saved form of the last evaled form. What I'm conufsed about is wouldn't the modified function always be in the last saved form and not the exercising function?

I'm imagining you are working on some function `(defn foo [x] ...)` and make changes to it and want to re-evaluate some `(foo 3)` type form. But the only way for `(foo 3)` to produce new output is if you re-evaluate the `defn` form, at which point the cached value is now useless to you as its the `defn` form and not the exercising `(foo 3)` form.

My mind goes to an emacs feature of registers. It seems like being able to store registers, evaluate registers, and send registers to the repl would be quite nice and would solve this issue. It allows you to recall forms not based on the last one evaluated (which seems almost always the one you definitely don't want) but based on ones you've saved and named.

dpsutton:
Also, i'm going to pull these changes down locally in a bit but I am wondering what the UI looks like. The interactive handler stuff does a brilliant job of displaying the output in a natural manner after the current form being evaluated. This makes sense of "this form -> this value". When you get into evaluating a stored form, this paradigm no longer makes a lot of sense. Should this output into the minibuffer?

dpsutton:
Some other thoughts: each form sent over the nrepl connection really needs a namespace. `cider-interactive-eval` will grab the current ns from the buffer with point which seems to break the way this is expected to work. For instance, you evaluate the form and then hit your key command to rerun the last test form defined in another namespace and you get a var not defined error since there is no `foo-test` in the namespace where `foo` is defined.

dpsutton:
One possible way that i just threw together and like:

```emacs-lisp
(defun cider-insert-register-contents (register)
  (interactive (list (register-read-with-preview "From register")))
  (let ((form (get-register register)))
    ;; could put form into a buffer and check if its parens are
    ;; balanced
    (if form
        (cider-insert-in-repl form (not cider-invert-insert-eval-p)) ;; ensure its always evaluated
      (user-error "No saved form in register"))))
```

Add `(define-key map (kbd "x") #'cider-insert-register-contents)` to the definition of `cider-insert-commands-map`.

To save regions into registers, `C-x r s R` and then to insert into the repl `C-c C-j x` (i chose x because i couldn't think of anything better in 4 seconds). This solves a few issues.
1. namespace management is explicitly in my control. Its just going to throw a form in the repl. I can see the ns and make sure it is the right ns. I do this constantly because i almost always send forms to the repl. `C-c C-j` map is heavily exercised by me
2. presentation is solved. its just simple making it easy to throw things into the repl. inline-evaluation results aren't a thing, no minibuffer, etc.
3. i can easily choose which form is saved. no more juggling which form was last and in which buffer and which ns. The PR also makes this a local variable so its which was the last in this buffer which isn't adaptable to re-running test commands or forms in other buffers.

l0st3d:
> I'm a bit confused about this. If i'm understanding correctly, you want to edit a function and
> then rerun some test or invocation of that function. The plan is to place this in a saved form of
> the last evaled form. What I'm conufsed about is wouldn't the modified function always be in the
> last saved form and not the exercising function?

That isn't really a problem for me as I load the whole file with `cider-load-buffer` rather that
evaling individual forms. But I agree, given a different workflow, this feature wouldn't be that
useful. I did the first thing that solved my problem, without thinking about other people's
workflow. Apologies.

> My mind goes to an emacs feature of registers.

Registers didn't even occur to me. But, yes I agree, that might well be more generally useful.

> When you get into evaluating a stored form, this paradigm no longer makes a lot of sense. Should
> this output into the minibuffer?

I was happy with the minibuffer, which is where this currently displays to.

> Some other thoughts: each form sent over the nrepl connection really needs a namespace. 

I was happy assuming same buffer, same ns. But, I that's just an assumption I made that fitted how I
wanted to work. I feel like that restriction isn't really a negative, because the test code is
probably not very far from the code it's executing and will change with the actual code. However,
I'd also be perfectly happy with a more general feature.

> Some other thoughts: each form sent over the nrepl connection really needs a namespace.
> cider-interactive-eval will grab the current ns from the buffer with point which seems to break
> the way this is expected to work. For instance, you evaluate the form and then hit your key
> command to rerun the last test form defined in another namespace and you get a var not defined
> error since there is no foo-test in the namespace where foo is defined.

The previous command is totally buffer local and tied to that ns. I'm only expecting to re-use the
last command until I've finished my function. If I move on to another function in another ns I'm not
going to be trying to run the previous bit of code because I now want to exercise a new function in
the new buffer. So I'll write a new bit of code to do that and run it. Making the previous code
buffer local fits pretty well with how I'm working.

> To save regions into registers, C-x r s R and then to insert into the repl C-c C-j x (i chose x
> because i couldn't think of anything better in 4 seconds). This solves a few issues.

Cool. I've had a play with this too.

> - namespace management is explicitly in my control. Its just going to throw a form in the repl. I
>   can see the ns and make sure it is the right ns. I do this constantly because i almost always
>   send forms to the repl. C-c C-j map is heavily exercised by me

This creates some problems for me as I now need to remember to open the REPL buffer (which I rarely
look at) and change the ns of the REPL to the same as the buffer I'm currently working in. I don't
really use the `C-c C-j` keymap and generally just evaluate code in the buffer I'm working in.

> - presentation is solved. its just simple making it easy to throw things into the repl.
>   inline-evaluation results aren't a thing, no minibuffer, etc.

Sure, but if you're not looking at the REPL window, then it pops up suddenly, and if your code
throws an exception the (maybe because the repl ns is wrong and it can't find the symbol) then you
get a stack trace as well. On my laptop, this then hides the code that I was looking at and shows me
a REPL I don't want to see next to the stack trace that's caused by the REPL being in the wrong ns.
It means that every time I switch between buffers I need to change the REPL ns, which again is
something I rarely do.

> - i can easily choose which form is saved. no more juggling which form was last and in which
>   buffer and which ns. The PR also makes this a local variable so its which was the last in this
>   buffer which isn't adaptable to re-running test commands or forms in other buffers.

I like selecting between registers UI, that's useful, and saves you remembering what you last
evaluated. I did find myself evaluating something, just to put it in the buffer local, and registers
would improve that.

However, on my laptop, there's a 1 to 2 seconds delay between invoking the command and the register
list appearing. Not sure if this is something in my setup that's making that slow, but it happens
with `eldev` too. Maybe not too big an issue, but first time I ran it, I wondered if it worked ;)

Also, I guess you now need to remember which ns each form was from given a short prefix. That's
probably less of an issue, but something that's introduced by separating the form from the buffer it
came from. Could track which buffer each register came from and store the ns to evaluate it in
perhaps?

PR #2923, opened 2020-10-26
Handle multiple REPL types for single project

otwieracz:
## Handle multiple REPL types for single project

Goal of this PR is to handle the issue:

1. Create and open project with `lein` and `shadow-cljs` working in one directory tree:
```
src
├── clj  (lein source-paths)
├── cljc (lein & shadow-cljs source-paths)
└── cljs (shadow-cljs source-paths)
```
2. Open `.clj` file, jack-in `cider-jack-in-clj`
3. Switch to CLJ REPL buffer with `(cider-switch-to-repl-buffer)` - works fine
4. Open `.cljs` file, jack-in `cider-jack-in-cljs`, accept creation of sibling REPL.
5. Switch to CLJS REPL buffer with `(cider-switch-to-repl-buffer)` - fails with message `user-error: No clj REPLs in current session "project/project:localhost:63409"`

### Root cause
* There are two `sesman` sessions created linked to the same project: 
```
 CIDER Sessions:

  1: project/project:localhost:63409
      linked-to: proj(~/project/)  
        objects: *cider-repl %s(clj)*  

  2: project/project:localhost:63420
      linked-to: proj(~/project/)  
        objects: *cider-repl %s(cljs:shadow)*  
```
* `(cider-repls)` uses `(sesman-current-session)` to get only one matching session for project (even though there might be more than one) without taking type into consideration. Then it matches type against the single result.

### Proposed solution
* Look for specified type within all matching sesman sessions
* When deciding on buffer type, on `cljc` file detection, consult `(cider-repl-type)` function. This enables easy change of REPL type for CLJC buffer (which might be simultaneously used by both REPLs) with snippet like (in spacemacs):
```
  ;; Shortcut for easily changing REPL type in ClojureC mode
  (add-hook 'cider-mode-hook (lambda ()
                               (spacemacs/set-leader-keys-for-major-mode 'clojurec-mode "st" 'cider-set-repl-type)))
```

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -699,7 +699,8 @@ For the REPL type use the function `cider-repl-type'."
   (with-current-buffer (or buffer (current-buffer))
     (cond
      ((derived-mode-p 'clojurescript-mode) 'cljs)
-     ((derived-mode-p 'clojurec-mode) 'multi)
+     ((derived-mode-p 'clojurec-mode) (or (cider-repl-type (or buffer (current-buffer)))
+                                          'multi))
      ((derived-mode-p 'clojure-mode) 'clj)
      (cider-repl-type))))
 
@@ -858,9 +859,9 @@ throw an error if no linked session exists."
                ((listp type)
                 (mapcar #'cider-maybe-intern type))
                ((cider-maybe-intern type))))
-        (repls (cdr (if ensure
-                        (sesman-ensure-session 'CIDER)
-                      (sesman-current-session 'CIDER)))))
+        (repls (mapcar #'cadr (sesman-current-sessions 'CIDER))))
+    (when (and ensure (null repls))
+      (user-error "No linked CIDER sessions"))
     (or (seq-filter (lambda (b)
                       (cider--match-repl-type type b))
                     repls)


bbatsov:
Hmm, that's weird. If I recall correctly sibling REPLs were supposed to be in the same session, otherwise they are basically unrelated.

bbatsov:
@vspinu might be able to provide more insight here.

vspinu:
As it happens I have started working with clojure/Cider 2 days ago after a more than a year break and immediately stumbled over this (or maybe related)  issue but didn't have time to investigate.

One of the ways to reproduce my issue is to jack in twice, C-z works. Then C-c C-q on one of the processes, C-z will no longer work from the original file even though the linking is there.  The reason is that cider-mode is no longer installed in those buffers. One need to revert the buffer to get C-z working again.


@otwieracz Could you check if you really have cider-mode t in the buffer that you C-z from?

> Open .cljs file, jack-in cider-jack-in-cljs, accept creation of sibling REPL.

I think you are not starting the "sibling" repl, but a new repl if you press "y". That prompt is quite confusing. "Proceed Y/N?" means create a new server session altogether, not a sibling. 

I cannot judge the proposed solution as I forgot the details.  But the change seems unrelated to the "cider-mode disappearence" which is the culprit in my case.

otwieracz:
> @otwieracz Could you check if you really have cider-mode t in the buffer that you C-z from?
> 
> >

Yes, I've got `cider-mode` on `minor-modes-list`. Major mode is `clojure-mode`.


> I think you are not starting the "sibling" repl, but a new repl if you press "y". That prompt is quite confusing. "Proceed Y/N?" means create a new server session altogether, not a sibling.

Unfortunately, there's no other way than hitting `y`. `n`  is aborting the process and resulting in `Quit` message:
```
A session with the same parameters exists (project/project:localhost:51230).  You can connect a sibling instead.  Proceed? (y or n) n
Quit
```

bbatsov:
See the conversation in #2946 and the updated message. Basically after pressing "n" you're expected to run `cider-connect-sibling-clj(s)`, as nothing happens automatically. Perhaps when can reconsider this down the road and prompt the user to select what kind of sibling REPL they want.

otwieracz:
@bbatsov unfortunately I am still getting a lot of issues with this approach:

After commenting-out my modifeid `cider-repls` and `cider-repl-type-for-buffer` and then restarting spacemacs with `SPC q r`:

* I've stared `lein` clj REPL with `cider-jack-in`
* Trying to start another REPL while looking at `cljs` file kindly greets me with this confusing message. Hitting `no` ends up with no action taken.
* Doing `cider-connect-sibling-cljs` while looking at `cljs` file, having already `clj` REPL connected from `clj` file from the same project results in message: `: No cljs REPLs in current session "foo/bar:localhost:56670"`
* Trying to start `cljs` sibling from `clj` file with above command resulted in error: `error in process filter: user-error: The shadow-cljs ClojureScript REPL is not available.  Please check https://docs.cider.mx/cider/basics/clojurescript for details
`
* After killing `clj` REPL, `cider-jack-in-cljs` from `cljs` starts fine but it's type is set to `clj` - I have to manually switch REPL type `cider-set-repl-type` to make `sesman` match this session to `cljs` file (before that I'm getting: `user-error: No cljs REPLs in current session "foo/bar:localhost:56876"`)

otwieracz:
My setup (approximate):
```
myproj/
   project.clj ;; for clojure
   shadow-cljs.edn ;; for clojurescript
   src/
      myproj/
         foo.clj
         bar.cljs
```
I expect `lein` and `shadow-cljs` to *separately* provide clj and cljs REPLs.

FYI, with most recent CIDER in Emacs 28 the message indeed changed but other than that problems still persist... (and my hack stopped working):
* Start CLJ repl with `cider-jack-in-clj`
* CLJ repl starts fine:
```
 CIDER Sessions:

  1: cyberrr/myproj:localhost:53306
      linked-to: proj(~/Work/myproj/)  
        objects: *cider-repl %s(clj)*  
```
* Trying `cider-jack-in-cljs` now makes it clear what the behaviour will be:
`A CIDER session with the same connection parameters already exists (cyberrr/myproj:localhost:53306).  Are you sure you want to create a new session instead of using `cider-connect-sibling-clj(s)'?  (y or n) n`
* Trying `cider-connect-sibling-cljs` viewing `cljs` file in the same project:
```
condition-case: No cljs REPLs in current session "cyberrr/myproj:localhost:53306"
```
* Starting `sibling` with `cider-connect-sibling-cljs`, selecting `shadow-cljs` as REPL type and `:app` as env:
```
[nREPL] Establishing direct connection to localhost:53306 ...
[nREPL] Direct connection to localhost:53306 established
error in process filter: if: The shadow-cljs ClojureScript REPL is not available.  Please check https://docs.cider.mx/cider/basics/clojurescript for details
error in process filter: The shadow-cljs ClojureScript REPL is not available.  Please check https://docs.cider.mx/cider/basics/clojurescript for details
```
It should start new `shadow-cljs` instance? Here it's trying to use existing `lein` connection and pretends to be surprised for not finding `shadow-cljs` REPL here :)

Issue #2913, opened 2020-10-21
clojuredocs lookup not working with CLI based project under spacemacs


theophilusx:
## Expected behavior
After starting a repl, moving the cursor over a symbol in the clj file and hitting ', h d' (or SPC m h d) should open a cider docview buffer with the documentation for that symbol (assuming core symbols documented on clojuredocs). 

## Actual behavior
When entering ', h d' after moving the cursor over a symbol, you are prompted for the symbol to lookup (i.e. cider-try-symbol-at-point fails). You enter the symbol e.g. println and hit enter. You then get the following error

````
Wrong type argument: char-or-string-p, nil
````
The full backtrace is

````
Debugger entered--Lisp error: (wrong-type-argument char-or-string-p nil)
  cider-clojuredocs--content(nil)
  cider-clojuredocs-lookup("println")
  cider-try-symbol-at-point("ClojureDocs doc for" cider-clojuredocs-lookup)
  cider-clojuredocs(nil)
  funcall-interactively(cider-clojuredocs nil)
  call-interactively(cider-clojuredocs nil nil)
  command-execute(cider-clojuredocs)
````

## Steps to reproduce the problem

Setup a simple CLI based project. I use Sean Corfield's deps.edn from his dot-clojure repository, which includes his clj-new library that allows you to setup a project using lein templates. I used the basic app template to create a test CLI project. The deps.edn is

````
{:paths ["src" "resources"]
 :deps  {org.clojure/clojure {:mvn/version "1.10.1"}}
 :aliases
 {:test    {:extra-paths ["test"]
            :extra-deps  {org.clojure/test.check {:mvn/version "1.0.0"}}}
  :runner
  {:extra-deps {com.cognitect/test-runner
                {:git/url "https://github.com/cognitect-labs/test-runner"
                 :sha     "f7ef16dc3b8332b0d77bc0274578ad5270fbfedd"}}
   :main-opts  ["-m" "cognitect.test-runner"
                "-d" "test"]}
  :uberjar {:extra-deps {seancorfield/depstar {:mvn/version "1.1.104"}}
            :main-opts  ["-m" "hf.depstar.uberjar" "core.jar"
                         "-C" "-m" "doctester.core"]}}}
````

I also have a .dir-locals.el file which has the following

````
((clojure-mode . ((cider-clojure-cli-global-options . "-A:test")
                  (cider-preferred-build-tool . clojure-cli))))
````

I open the core.clj file and move the cursor to the 'println' symbol

[...]

PR #2926, opened 2020-11-05
[Documentation] Tip to evaluate namespace before lookup

jr0cket:
Add Tip to "Working with Documentation" page that a namespace should be
evaluated before lookup of documentation.

Resolves: #2913

-----------------

- [x] You've added tests - documentation only pr
- [x] All code passes the linter (`eldev lint`) - no lint errors on specific change

### DIFF
--- a/doc/modules/ROOT/pages/usage/working_with_documentation.adoc
+++ b/doc/modules/ROOT/pages/usage/working_with_documentation.adoc
@@ -11,6 +11,8 @@ press kbd:[C-c C-d C-d] then. This open a documentation buffer
 containing all the relevant information about the thing referenced
 by the symbol (special form, var, Java method, etc).
 
+TIP: Ensure the namespace of a var is evaluated before looking up the documentation. Cider relies on the runtime state to lookup documentation.
+
 == JavaDoc
 
 CIDER provides a quick access to the online Javadoc documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Bump the injected Piggieback version to 0.5.2. See [this issue](https://github.com/nrepl/piggieback/issues/118) for details.
 * [#2897](https://github.com/clojure-emacs/cider/pull/2897): Translate paths from CIDER to nREPL and vice-versa.
 * Set `cider-prompt-for-symbol` to `nil` by default.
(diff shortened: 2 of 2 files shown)


doc/modules/ROOT/pages/usage/working_with_documentation.adoc
```diff
@@ -11,6 +11,8 @@ press kbd:[C-c C-d C-d] then. This open a documentation buffer
 containing all the relevant information about the thing referenced
 by the symbol (special form, var, Java method, etc).
 
+TIP: Ensure the namespace of a var is evaluated before looking up the documentation. Cider relies on the runtime state to lookup documentation.
```
bbatsov, line 14: This applies to every feature in CIDER, so I'm not sure we need to repeat it everywhere.

jr0cket: Okay, so you just want to keep this on the one page in the docs.  I'll close the pr.

bbatsov:
For the record - I think it'd be best if we devised some way to warn people when they are trying to do something in an unevaluated ns. That's going to be way more effective way to get the message across. Evals handle this to some extent by auto-evaling the `ns` form so they can do something, but for other commands it's tricker.

jr0cket:
From the tests I've done this evening, it seems that help is only available for Clojure special forms until any namespace is evaluated.  Getting help for `clojure.core` functions therefore fails without evaluating a project namespace (unless I am missing something)

The biggest pain point I've experienced is receiving the `Doc for:` prompt in the mini-buffer and it then failing to work.  It would be less confusing just to get an error or warning rather than the `Doc for:` prompt (as it makes you think the name was incorrectly typed or is doesn't exist).

I've updated my own documentation to cover this in detail, as I didnt find this information in the official docs.
https://practicalli.github.io/spacemacs/documentation/cider-doc.html

bbatsov:
> From the tests I've done this evening, it seems that help is only available for Clojure special forms until any namespace is evaluated.

Special forms are not resolved, they are pretty much hardcoded, that's why they work everywhere. 

> The biggest pain point I've experienced is receiving the Doc for: prompt in the mini-buffer and it then failing to work. It would be less confusing just to get an error or warning rather than the Doc for: prompt (as it makes you think the name was incorrectly typed or is doesn't exist).

You mean you get it after the resolution fails, right?

PR #2932, opened 2020-11-21
Fix format of clojure-cli options because TRAMP

carrete:
This is /a/ solution for https://github.com/clojure-emacs/clj-refactor.el/pull/445#issuecomment-628206640. I do not know what the cause of this problem is. Perhaps `IFS` is different between cider-jack-in versus directly on the command-line? Perhaps it's because TRAMP uses `/bin/sh` to run remote commands, not `/bin/bash`? I don't know. Given that `,` is whitespace in clojure, this seems like a harmless patch.

These changes are consistent with the contribution guidelines. The results of the tests and linter are unchanged by this patch. I do not believe this requires a mention in the CHANGELOG nor user manual.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -364,6 +364,10 @@ Throws an error if PROJECT-TYPE is unknown."
 
 (defun cider-jack-in-params (project-type)
   "Determine the commands params for `cider-jack-in' for the PROJECT-TYPE."
+  ;; The format of these command-line strings must consider different shells,
+  ;; different values of IFS, and the possibility that they'll be run remotely
+  ;; (e.g. with TRAMP). Using `", "` causes problems with TRAMP, for example.
+  ;; Please be careful when changing them.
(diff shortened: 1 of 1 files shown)


cider.el
```diff
@@ -373,7 +373,7 @@ Throws an error if PROJECT-TYPE is unknown."
                            (mapconcat
                             (apply-partially #'format "\"%s\"")
                             (cider-jack-in-normalized-nrepl-middlewares)
-                            ", ")
+                            ",")
```
bbatsov, line 376: Might be a good idea to add a note here advising against changing this, as I'm certain someone will add back the space eventually otherwise. :-)

carrete: Good idea @bbatsov, done.

bbatsov:
Thanks!

Issue #2930, opened 2020-11-15
Non-interactive test selector filters


jellelicht:
**Is your feature request related to a problem? Please describe.**
I have to repeatedly add e.g. `:integration` to the list of exclude-selectors in order for cider-test-related functions to not run my `:integration`-tagged tests.

**Describe the solution you'd like**
I'd love to see support for non-interactive test selector filters, either by doing some `(setq ..)` magic or (even better) by reading the test selector configuration from Leiningen's `project.clj`. This would mean that e.g. `cider-auto-test-mode` would _ignore_ the tests that I want it to ignore by default, instead of bull-headedly trying to run tests that I know will fail.

**Describe alternatives you've considered**
Writing my own `cider-auto-test-mode`-esque mode, that always hard-codes `:integration` as being part of the `exclude-selectors` everywhere in cider.

PR #2955, opened 2020-12-24
Add variables for default test selectors (include/exclude)

zkry:
This PR addresses issue  #2930 by adding two variables (`cider-test-default-include-selectors` and `cider-test-default-exclude-selectors`) which provide default selectors in contexts where selectors are applicable (ex. ns, project ) and the selector prompt was not indicated (ie by prefix key).

-----------------

PR Checklist:

- [x] You've added tests (if possible) to cover your change
  - Note: This doesn't include unit tests but let me know if you think I should add some.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 * [#2909](https://github.com/clojure-emacs/cider/issues/2909): Add new customization variable `cider-inspector-auto-select-buffer` to control the auto selection of the inspector buffer.
 * [#2940](https://github.com/clojure-emacs/cider/pull/2940): Add a new customization variable cider-shadow-watched-builds to allow watching several shadow-cljs builds at the same time.
+* [#2930](https://github.com/clojure-emacs/cider/issues/2930): Add new customization variable `cider-test-default-include-selectors` and `cider-test-default-exclude-selectors` for specifying default test selectors when running commands such as `cider-test-run-ns-tests`.
 
 ### Bugs fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New features
 
+* [#2930](https://github.com/clojure-emacs/cider/issues/2930): Add new customization variable `cider-test-default-include-selectors` and `cider-test-default-exclude-selectors` for specifying default test selectors when running commands such as `cider-test-run-ns-tests`.
+
 ### Bugs fixed
 
 * [#2953](https://github.com/clojure-emacs/cider/issues/2953): Don't font-lock function/macro vars as vars.
@@ -16,7 +18,6 @@
 
 * [#2909](https://github.com/clojure-emacs/cider/issues/2909): Add new customization variable `cider-inspector-auto-select-buffer` to control the auto selection of the inspector buffer.
 * [#2940](https://github.com/clojure-emacs/cider/pull/2940): Add a new customization variable cider-shadow-watched-builds to allow watching several shadow-cljs builds at the same time.
-* [#2930](https://github.com/clojure-emacs/cider/issues/2930): Add new customization variable `cider-test-default-include-selectors` and `cider-test-default-exclude-selectors` for specifying default test selectors when running commands such as `cider-test-run-ns-tests`.
 
 ### Bugs fixed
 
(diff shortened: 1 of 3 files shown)


doc/modules/ROOT/pages/testing/running_tests.adoc
```diff
@@ -37,7 +37,11 @@ kbd:[C-c C-t C-l].
 
 If you invoke either of these commands with a prefix, CIDER will
 prompt for test selector filters and only run those tests that match
-the selector inclusions/exclusions.
+the selector inclusions/exclusions. If you have selectors you want
```
bbatsov, line 40: I guess you can provide some examples of the of the variables as well.

cider-test.el
```diff
@@ -617,6 +617,18 @@ This uses the Leiningen convention of appending '-test' to the namespace name."
 
 ;;; Test execution
 
+(defcustom cider-test-default-include-selectors '()
+  "List of include selector strings to use when executing tests if none provided."
+  :type '(repeat string)
+  :group 'cider-test
+  :package-version '(cider . "0.26.1"))
```
bbatsov, line 624: This should be "1.1.0".

cider-test.el
```diff
@@ -636,16 +648,28 @@ namespace is specified.  Upon test completion, results are echoed and a test
 report is optionally displayed.  When test failures/errors occur, their sources
 are highlighted.
 If SILENT is non-nil, suppress all messages other then test results.
-If PROMPT-FOR-FILTERS is non-nil, prompt the user for a test selector filters.
+If PROMPT-FOR-FILTERS is the keyword `:selector-vars' then use the variables
```
bbatsov, line 651: This part seems weird to me. I'd leave the behavior as it was and just apply the variables automatically when present. This will result in simpler code and more predictable behavior IMO.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [#2909](https://github.com/clojure-emacs/cider/issues/2909): Add new customization variable `cider-inspector-auto-select-buffer` to control the auto selection of the inspector buffer.
 * [#2940](https://github.com/clojure-emacs/cider/pull/2940): Add a new customization variable cider-shadow-watched-builds to allow watching several shadow-cljs builds at the same time.
+* [#2930](https://github.com/clojure-emacs/cider/issues/2930): Add new customization variable `cider-test-default-include-selectors` and `cider-test-default-exclude-selectors` for specifying default test selectors when running commands such as `cider-test-run-ns-tests`.
```
bbatsov, line 9: As 1.0 was just released, you'll have to move this changelog entry up.

cider-test.el
zkry: I could do that. I ended up doing it this way in order to distinguish it from the case where the user chooses to execute an individual test that happens to be excluded. So having something like `(if prompt-for-filters (cider-test--prompt-for-selectors "Test selectors...") cider-test-default-include-selectors)` would make it so that running an individual test (like with `cider-test-run-test`) would use the default selectors which might block the individual test trying to run. If you think this behavior is ok I can simplify the logic as you said.

zkry: I updated the PR to leave the behavior as it was for now though.

cider-test.el
```diff
@@ -617,6 +617,18 @@ This uses the Leiningen convention of appending '-test' to the namespace name."
...
+  "List of include selector strings to use when executing tests if none provided."
+  :type '(repeat string)
+  :group 'cider-test
+  :package-version '(cider . "1.1.0"))
+
+(defcustom cider-test-default-exclude-selectors '()
+  "List of exclude selector strings to use when executing tests if none provided."
+  :type '(repeat string)
+  :group 'cider-test
+  :package-version '(cider . "0.26.1"))
```
bbatsov, line 630: This shoud also say 1.1.0

bbatsov:
Your remark about what happens when you run a single test that's excluded by the default selectors is a valid one. I was thinking we might add some additional information in the test report buffer mentioning which selectors were applied (if any), which should probably help with that. Anyways, that's something we can do down the road.

zkry:
I agree that information in the test report buffer about applied selectors would be a good idea. After this PR I'll see if I can put in a PR for this.

bbatsov:
That'd be great! Thanks!

PR #2959, opened 2020-12-30
fix #2953

### LINKED ISSUE #2953: [syntax highlighting] functions highlighted as vars
I've disabled function highlighting, to bring color intensity a bit down by changing `cider-font-lock-dynamically`.
By default it is set to this:
```emacs-lisp
'(macro core function var)
```

And I get the usual highlighting:

image

However if I remove `core` and `function` from this setting:
```emacs-lisp
(setq cider-font-lock-dynamically '(macro var))
```

I get the following result after I load file:

image

## Expected behavior

image

1. Download [bug.zip](https://github.com/clojure-emacs/cider/files/5736985/bug.zip), extract and run Emacs with provided minimal `init.el`:

  ```
  $ unzip bug.zip
  $ cd bug
  $ emacs -q -l repro-init.el bug.clj
  ```
2. Jack in: <kbd>Ctrl-c</kbd><kbd>Alt-j</kbd><kbd>y</kbd>
3. Load `bug.clj` file with <kbd>Ctrl-c</kbd><kbd>Ctrl-k</kbd>

This results into the following highlighting:

image

### CIDER version information

```
;; CIDER 1.0.0snapshot (package: 20201216.1330), nREPL 0.8.3
;; Clojure 1.10.1, Java 11.0.9.1
```

### Emacs version

GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.13) of 2020-12-23

### Operating system

Fedora 33

andreyorst:

Thanks!

manual][1] extremely useful.*


cider-mode.el
```diff
@@ -784,7 +784,10 @@ with the given LIMIT."
                              ((and do-function (or (nrepl-dict-get meta "fn")
                                                    (nrepl-dict-get meta "arglists")))
                               (push sym functions))
-                             (do-var (push sym vars))))))))
+                             ((and do-var
+                                   (not (nrepl-dict-get meta "fn"))
```
bbatsov, line 788: Probably you should check for the `macro` meta as well.

bbatsov:
The proprosed fix seems reasonable. Might be good to reduce the duplication in the checks, but it's not a big deal.

andreyorst:
I'll look into it a bit later tooday

cider-mode.el
andreyorst: Should be fine now

bbatsov:
Thanks!

PR #2961, opened 2021-01-02
Fix cider-completion-flush-caches

### LINKED ISSUE #2914: Completion cache flush ends up with error
I'm not sure if that problem is not caused by some mismatching dependency versions because there is a deeper background on my machine (which I though could be fixed by the flush) - ~completion on namespaces and fully qualified symbols doesn't work. `(clojure.string/_)` or `(java.util._)` stopped working (afair it worked before). The only suggested autocompletion candidates come from `clojure.core` and the namespace I'm currently in~. Edit: that was a problem of my customization of `completion-styles` set to `'(flex)`.

## Expected behavior
`cider-completion-flush-caches` flushes completion cache.

## Actual behavior
`cider-completion-flush-caches` ends up with `nrepl-send-sync-request: Wrong type argument: stringp, abort-on-input`.

call `cider-completion-flush-caches` or invoke the flush from "CIDER interaction menu.

### CIDER version information

```
;; Connected to nREPL server - nrepl://localhost:48888
;; CIDER 1.0.0snapshot (package: 20201020.1058), nREPL 0.8.2
;; Clojure 1.10.1, Java 13.0.2
;;     Docs: (doc function-name)
;;           (find-doc part-of-name)
;;   Source: (source function-name)
;;     Exit: <C-c C-q>
;;  Results: Stored in vars *1, *2, *3, an exception in *e;
;;  Startup: /usr/local/bin/lein update-in :dependencies conj \[nrepl\ \"0.8.2\"\] -- update-in :plugins conj \[refactor-nrepl\ \"2.5.0\"\] -- update-in :plugins conj \[cider/cider-nrepl\ \"0.25.4\"\] -- repl :headless :host localhost
```
### Lein/Boot version

Leiningen 2.9.3 on Java 13.0.2 Java HotSpot(TM) 64-Bit Server VM

### Emacs version
GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.6.0, Carbon Version 162 AppKit 1894.6) of 2020-08-17

### Operating system
macOS Catalina 10.15.7

brool:
Fixes #2914 

In `cider-sync-request:complete-flush-caches` the call to `cider-nrepl-send-sync-request` was passing `'abort-on-input` in place of the connection parameter, which was causing problems down the line. 

Adding a nil should fix it, as `cider-nrepl-send-sync-request` will provide a connection if nil is passed in, and was consistent with the usage in `cider-sync-request:eldoc-datomic-query`.

Checked the other calls to `cider-nrepl-send-sync-request` but it looked like the others were okay.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -570,6 +570,7 @@ CONTEXT represents a completion context for compliment."
   "Send \"complete-flush-caches\" op to flush Compliment's caches."
   (cider-nrepl-send-sync-request (list "op" "complete-flush-caches"
                                        "session" (cider-nrepl-eval-session))
+                                 nil
                                  'abort-on-input))
 
 (defun cider-sync-request:info (symbol &optional class member)


bbatsov:
Thanks!

PR #2970, opened 2021-01-23
[Fix #2941] Use main args in :cider/nrepl alias for clojure cli

### LINKED ISSUE #2941: use `:main-opts` in `-Sdeps` map during jack in with clojure cli
if you jack in with other profiles that include main opts, the cider commandline main opts are ignored. One way around this is to use the merging resolution for multiple profiles: last main-opts "wins".

ie instead of 

```
/usr/local/bin/clojure -Sdeps '{:deps {nrepl {:mvn/version "0.8.3"} cider/cider-nrepl {:mvn/version "0.25.5"}}}' -A:dev:test -m nrepl.cmdline --middleware '["cider.nrepl/cider-middleware"]'
```

we can use 

```
clojure -Sdeps '{:deps ... :aliases {:cider/jack-in {:main-opts ["-m" ...]}}}' -M:dev:test:cider/jack-in --middleware ...
```

this ensures that if any main opts are in `dev` or `test`, the `cider/jack-in` alias will win.

The motivating example is the following alias:

```clojure
:test {:extra-paths ["test"]
         :extra-deps {lambdaisland/kaocha           {:mvn/version "1.0.690"}
                      lambdaisland/kaocha-cloverage {:mvn/version "1.0.63"}
                      lambdaisland/kaocha-junit-xml {:mvn/version "0.0.76"}
                      ring/ring-mock                {:mvn/version "0.3.2"}
                      mockery                       {:mvn/version "0.1.4"}
                      http-kit.fake/http-kit.fake    {:mvn/version "0.2.1"}}
         :main-opts  ["-m" "kaocha.runner"]}
```

where a repl is desired with the test library but without running the test lib's main or requiring splitting the main from the necessary dependencies.

dpsutton:
Addresses #2941.

Hopefully the string building is a bit more clear with the format string 

```emacs-lisp
    (format "%s-Sdeps '{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s]}}}' -M:cider/nrepl"
            (if global-opts (format "%s " global-opts) "")
            deps-string
            main-opts)
```

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -157,13 +157,15 @@ default to \"powershell\"."
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
 
-(defcustom cider-clojure-cli-global-options
+(defcustom cider-clojure-cli-aliases
   nil
   "Command line options used to execute clojure with tools.deps."
   :type 'string
   :group 'cider
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
+
+(make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-aliases "1.1")
 
 (defcustom cider-shadow-cljs-command
   "npx shadow-cljs"
@@ -346,7 +348,7 @@ Throws an error if PROJECT-TYPE is unknown."
   (pcase project-type
     ('lein        cider-lein-global-options)
     ('boot        cider-boot-global-options)
-    ('clojure-cli cider-clojure-cli-global-options)
+    ('clojure-cli cider-clojure-cli-aliases)
     ('shadow-cljs cider-shadow-cljs-global-options)
     ('gradle      cider-gradle-global-options)
     (_            (user-error "Unsupported project type `%S'" project-type))))
--- a/cider.el
+++ b/cider.el
@@ -166,7 +166,7 @@ default to \"powershell\"."
   :package-version '(cider . "0.17.0"))
 
 (defcustom cider-clojure-cli-parameters
-  "-m nrepl.cmdline --middleware '%s'"
+  ""
   "Params passed to clojure to start an nREPL server via `cider-jack-in'.
 This is evaluated using `format', with the first argument being the Clojure
 vector of middleware variables as a string."
@@ -370,14 +370,7 @@ Throws an error if PROJECT-TYPE is unknown."
   (pcase project-type
     ('lein        cider-lein-parameters)
     ('boot        cider-boot-parameters)
-    ('clojure-cli (format cider-clojure-cli-parameters
-                          (concat
-                           "["
-                           (mapconcat
-                            (apply-partially #'format "\"%s\"")
-                            (cider-jack-in-normalized-nrepl-middlewares)
-                            ",")
-                           "]")))
+    ('clojure-cli cider-clojure-cli-parameters)
     ('shadow-cljs cider-shadow-cljs-parameters)
     ('gradle      cider-gradle-parameters)
     (_            (user-error "Unsupported project type `%S'" project-type))))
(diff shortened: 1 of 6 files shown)


dpsutton:
weird. i'm not seeing any CI. tests pass for me and "Linter has no complaints"

cider.el
```diff
@@ -370,14 +370,7 @@ Throws an error if PROJECT-TYPE is unknown."
...
     ('boot        cider-boot-parameters)
-    ('clojure-cli (format cider-clojure-cli-parameters
-                          (concat
-                           "["
-                           (mapconcat
-                            (apply-partially #'format "\"%s\"")
-                            (cider-jack-in-normalized-nrepl-middlewares)
-                            ",")
-                           "]")))
+    ('clojure-cli cider-clojure-cli-parameters)
```
dpsutton, line 373: should this go away? i don't think it serves any purpose now. not sure the best cleanup of it. all the interesting stuff (refactor, etc) go in other places and the `cider-clojure-cli-global-options` is where you pass `-A:dev:test` and the like.

bbatsov:
> weird. i'm not seeing any CI. tests pass for me and "Linter has no complaints"

Perhaps we need to change something in the config? No idea why the CI results are not shown in the PR. I haven't changed the project config in ages.

cider.el
bbatsov: Yeah, I think it should go away. In general I still like the idea to simplify the config for each tool instead of just copying the original config for Lein everywhere.

bbatsov: Btw, if we're using the global options just for aliases, maybe we should rename it to something like `cider-clojure-cli-aliases` or whatever. I guess this might make things easier for the end users.

cider.el
```diff
@@ -157,23 +157,15 @@ default to \"powershell\"."
...
-(defcustom cider-clojure-cli-parameters
-  "-m nrepl.cmdline --middleware '%s'"
-  "Params passed to clojure to start an nREPL server via `cider-jack-in'.
-This is evaluated using `format', with the first argument being the Clojure
-vector of middleware variables as a string."
-  :type 'string
-  :group 'cider
-  :safe #'stringp
-  :package-version '(cider . "0.17.0"))
+(make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-aliases "1.1")
```
dpsutton, line 168: @bbatsov wasn't sure what version to put here.

dpsutton:
I think all i need from you is what version to put in the `make-obsolete-variable` and then to squash and we're good to go.

bbatsov:
The next release is going to be 1.1.

cider.el
```diff
@@ -157,23 +157,15 @@ default to \"powershell\"."
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
 
-(defcustom cider-clojure-cli-global-options
+(defcustom cider-clojure-cli-aliases
```
bbatsov, line 160: I think you should update the docstring as well. :-)

dpsutton: good catch. thanks

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -126,7 +126,7 @@ You can further customize the command line CIDER uses for `cider-jack-in` by
 modifying the following string options:
 
 * `cider-lein-global-options`, `cider-boot-global-options`,
-`cider-clojure-cli-global-options`, `cider-gradle-global-options`:
+`cider-clojure-cli-aliases`, `cider-gradle-global-options`:
```
bbatsov, line 129: And you should remove the params a couple of lines later. Now that the params are a bit different it might be a good idea to just list the params for each build tools separate (e.g. in subsections named "Leiningen", "Boot", etc).

dpsutton: which params should be removed? Not sure i'm following.

bbatsov: `cider-clojure-cli-parameters` is mentioned in the next paragraph.

dpsutton: oh totally. I was thinking this was in the code file `cider.el` and was super confused. Gotcha

bbatsov:
Thanks!

Bost:
I had in my .dir-locals.el
```elisp
((clojure-mode . ((cider-preferred-build-tool . clojure-cli)
                  (cider-clojure-cli-global-options . "-J-Djdk.attach.allowAttachSelf"))))
```
Changing it to `cider-clojure-cli-aliases` doesn't work. `cider-clojure-cli-parameters` doesn't work either.

dpsutton:
@Bost did you update? if you do `m-x describe-variable [enter] cider-clojure-cli-parameters` is the variable present? If so, does it say its current value due to the dir-locals? If not, try `m-: (hack-local-variables)` to try to make it pick up the new values.

image

then when i jack in its using `-M:test:cider/nrepl`:
image

> ;;  Startup: /usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.8.3"} cider/cider-nrepl {:mvn/version "0.25.8"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[\"cider.nrepl/cider-middleware\"]"]}}}' -M:test:cider/nrepl

dpsutton:
ah dang. i just read through your usecase there. I don't think my changes work for that. I forgot about jvm parameters like that. I'll need to work something out for your use case. I'm really sorry about that.

dpsutton:
the easiest workaround for you right now is to jack in with a prefix `C-u C-c M-j` and throw that string into the jack in command at the beginning. I had forgotten about those types of parameters.

dpsutton:
@Bost so sorry about that. These changes should fix it for you.

Bost:
Not a big deal. Thanks for fixing it quickly! :)

PR #2971, opened 2021-01-26
Put user supplied aliases in `-M[users]:cider/nrepl`

dpsutton:
`clojure -Sdeps {...} -M:[optional user aliases]:cider/nrepl` is the correct template string.

Do some string manipulation to strip off the leading `-A:` or `-M:` from user supplied deps.

The change relied on the main args of the last alias would win, but apparently aliases from `-A` always go last, defeating the purpose. So just throw them into the `-M` form and ensure `cider/nrepl` is last achieves this aim.

Long term we can go back to the `clojure [optional -A: user profiles] -Sdeps {...} -M:cider/nrepl` as indicated by the output of `clj --help`:

```
exec-opts:
 -Aaliases      Use concatenated aliases to modify classpath
 -X[aliases]    Use concatenated aliases to modify classpath or supply exec fn/args
 -M[aliases]    Use concatenated aliases to modify classpath or supply main opts
 -P             Prepare deps - download libs, cache classpath, but don't exec
```

Since it seems the real difference between -A and -M (will ultimately) be the "supply main opts". But that change isn't in effect yet so the `-M` stacking will work for now.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -160,7 +160,9 @@ default to \"powershell\"."
 (defcustom cider-clojure-cli-aliases
   nil
   "A list of aliases to include when using the clojure cli.
-Should be of the form `-A:foo:bar`."
+Should be of the form `foo:bar`.  Any leading \"-A\" or \"-M\" will be
+stripped as these are concatenated into the \"-M[your-deps]:cider/nrepl\"
+form."
   :type 'string
   :group 'cider
   :safe #'stringp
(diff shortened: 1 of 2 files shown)


bbatsov:
Seems to me you should also explain this in the docs, so it's clearer for the users. That's also why I suggested yesterday to put a separate section about the `clojure-cli` jack-in configuration, as it's a bit different from the config for everything else.

dpsutton:
@bbatsov sounds good. want to merge this now and i'll get the docs updated this evening after work?

bbatsov:
Sure!

Issue #2972, opened 2021-01-26
update docs for how clojure cli jack in works


dpsutton:
following the recent changes, update the docs so its clear what cider-jack-in is hoping to accomplish and how the user can customize this.

dpsutton:
note to self: 
- mention jvm options (eg, "-J-Djdk.attach.allowAttachSelf" see comments on https://github.com/clojure-emacs/cider/pull/2970)
- mention aliases in `cider-clojure-cli-global-options` (current behavior) and `cider-clojure-cli-aliases`
- this is a bit of a workaround until clojure cli is updated so that `-A` doesn't pass main args and is prevalent enough in the wild that this can all just go back to a single `cider-clojure-cli-global-options`

practicalli-john:
When defining user aliases a `:` is inserted after the -M and before the user defined alias string.
However, the alias names all start with a `:`, so the command line assembled by Cider contains `-M::user/alias-name

This feature does not seem to affect operation of cider-jack-in and aliases seem to be incorporated in the REPL.

Example:
I defined a `.dir-locals.el` file as follows
```
((nil . ((cider-clojure-cli-aliases . ":env/test:lib/ring-mock"))))
```
Running `cider-jack-in-clj` and CIDER command line is output in the `*messages*` buffer
```
[nREPL] Starting server via /usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.8.3"} cider/cider-nrepl {:mvn/version "0.26.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[cider.nrepl/cider-middleware]"]}}}' -M::env/test:lib/ring-mock:cider/nrepl
```

vemv:
The docs are probably better by now, however would either of you enjoy having a look again?

(I also will soon)

Cheers - V

PR #2974, opened 2021-01-27
Reintroduce main opts for jvm options

dpsutton:
these jvm options need to go at the beginning. Once the cli tools are
updated for `-A` to not use main opts all of this goes away and
everyone can just use the `cider-clojure-cli-global-options`. But
until that is the case people will need to use
`cider-clojure-cli-aliases` if they have aliases with mains.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -157,6 +157,14 @@ default to \"powershell\"."
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
 
+(defcustom cider-clojure-cli-global-options
+  nil
+  "Command line options used to execute clojure with tools.deps."
+  :type 'string
+  :group 'cider
+  :safe #'stringp
+  :package-version '(cider . "0.17.0"))
+
 (defcustom cider-clojure-cli-aliases
   nil
   "A list of aliases to include when using the clojure cli.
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -157,6 +157,14 @@ default to \"powershell\"."
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
 
+(defcustom cider-clojure-cli-global-options
+  nil
+  "Command line options used to execute clojure with tools.deps."
```
bbatsov, line 162: Probably we should add a couple of examples of those, as I had forgotten about the potential use-cases myself.

bbatsov:
I'll merge this as it is, and we can improve its docs as part of #2972.

cider.el
Bost: As for the examples, several profiling and measurement tools like those under [clojure-goes-fast](https://github.com/clojure-goes-fast) require using additional JVM options. And I personally pass them via my [.dir-locals.el](https://github.com/Bost/corona_cases/blob/master/.dir-locals.el). This might not be the best approach however. The .dir-locals.el is strictly speaking not a part of the project. You don't really want to keep it in the repo. On the other hand you don't want to see it popping up in every `git status` or `M-x magit-status`. So there might be a better solution than this one of mine.

Issue #2937, opened 2020-12-04
green fringe produced for extra line in rich comment block


dareno:
## Expected behavior
Evaluating just a single line in a Rich comment block will apply a green fringe to only the line being evaluated.
## Actual behavior
Evaluating just a single line in a Rich comment block will apply a green fringe to the line being evaluated as well as the following line.
## Steps to reproduce the problem
1. add the following comment block to a file:
```
(comment
  (+ 1 1)
  (+ 1 2)
  (+ 1 3))
```
2. evaluate the entire comment block to produce a green fringe just on the first line ("comment" line)
3. evaluate the second commented expression and see that both the second and third expressions will be marked as evaluated (this happens even when following line is blank).

### CIDER version information

```
;; CIDER 1.0.0snapshot (package: 20201125.1146), nREPL 0.8.3
;; Clojure 1.10.1, Java 1.8.0_271
```

### Lein/Boot version
```
Leiningen 2.9.4 on Java 1.8.0_271 Java HotSpot(TM) 64-Bit Server VM
```
### Emacs version
```
GNU Emacs 27.1 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95)) of 2020-08-12
```
### Operating system
macOS Catalina Version 10.15.7

fvides:
I just submitted a PR that hopefully fixes this issue.

fvides:
Should someone check this and if everything is okay close it?

dareno:
I just did some manual testing and can no longer reproduce my test case. Looks good to me. Closed by #2937.

PR #2976, opened 2021-01-30
Fix cider-last-sexp bounds on whitespaces after sexp

fvides:
Evaluating cider-last-sexp includes whitespaces and the first newline
after such sexp, which causes some weird behaviours evaluating
code (see https://github.com/clojure-emacs/cider/issues/2937).

Modified cider-last-sexp by removing two forms that included whitespaces and the first newline in the returned value (string
or bounds), so the result is exactly the last sexp as advertised.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,13 @@
 
 * [#2941](https://github.com/clojure-emacs/cider/issues/2941): Use main args in alias for clojure cli
 * [#2953](https://github.com/clojure-emacs/cider/issues/2953): Don't font-lock function/macro vars as vars.
+* [#2937](https://github.com/clojure-emacs/cider/issues/2937): Green fringe produced for extra line in rich comment block
 
 ### Changes
 
 * Removed `cider-clojure-cli-parameters` due to clojure-cli jack-in changes
 * Bump the injected `cider-nrepl` to 0.25.6. This should fix a compatibility issue with Java 15 and fetching fresh ClojureDocs data.
+* Changed the behaviour of cider-last-sexp so it returns only the sexp, excluding all whitespace and/or the first newline after.
 
 ## 1.0.0 (2020-28-12)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Did you check in the history why this was added? Seems oddly specific, so I'm assuming something else will break without it.

dpsutton:
Yeah this stuff gets complicated and there's lot of edge cases. Things like navigating to the next form, evaluating forms in comment forms when `clojure-toplevel-inside-comment-form` is `t`, etc. Also, need to read the docstring of the emacs stuff because there are some weird assumptions about newlines and such that it expects. All that says this might be right, but there is a lot of scariness in this code path.

fvides:
I did check history. The commit where these lines where added is f57abc7 and the comment was: "Make cider-last-sexp behave more like cider--region-for-defun..." I didn't find this function so seemed that mantaining a coherent behaviour with it wasn't very important anymore. On the other hand it doesn't feel right that the result of cider-last-sexp includes whitespaces and/or a newline. 

FWIW I've manually checked all functions that invoke cider-last-sexp:
* cider-eval-last-sexp-and-replace
* cider-eval-last-sexp-in-context
* cider-eval-last-sexp-to-comment
* cider-eval-last-sexp-to-repl
* cider-pprint-eval-last-sexp-to-repl
* cider-eval-print-last-sexp
* cider-pprint-eval-last-sexp
* cider-insert-last-sexp-in-repl
* cider-macroexpand-expr-inplace
* cider-macroexpand-1

And everything seems fine. Of course I'm willing to perform additional tests that you suggest.

bbatsov:
I just don't remember the rationale for the change in question and I doubt @Malabarba remembers it either after so many year. I agree this seems weird and I'll willing to accept to your patch. Just mention it in the changelog and we're good.

fvides:
Done, edited changelog.

bbatsov:
Thanks!

fvides:
Thank you for your great work and inspiration. I intend to keep helping, hopefully this is the first of many contributions.

PR #2977, opened 2021-01-31
Fixes #2964 -- exception when cider-complete is called without a repl…

### LINKED ISSUE #2964: Enabling cider-company-enable-fuzzy-completion breaks helm completions
### Expected behavior

When the REPL has *not* been stared, `M-x` and filtering *command* candidates should not throw an error in a `cider-mode` buffer with `cider-company-enable-fuzzy-completion`.

## Actual behavior

It throws the following:

2021-01-08-182059_1920x1080_scrot

It is not really clear from the screen but the error is basically "No linked CIDER session".

Install `helm` (you can run it in isolation with `emacs-helm.sh`) and enable `(cider-company-enable-fuzzy-completion)`. 

Execute `helm-M-x` and try `asd` or anything command prefix that does not exist. 

### CIDER version information
latest master

### Emacs version

*GNU Emacs 28.0.50*

### Operating system

Linux

---

I did not have time to investigate too much but I am wondering if we the `cider`, `completion-styles` function should not throw an error when the REPL is not connected. Throwing this there but I am not sure this is the actual solution without looking into it a bit more :smile:

brool:
Helm was causing errors when `cider-company-enable-fuzzy-completion` was enabled and a REPL was not active.

The issue here was that `cider-complete` called `cider-nrepl-op-supported-p`, which requires an active connection, and will signal otherwise -- but in the event of a completion failure it is better to just continue. 

The fix is just to return nil from `cider-complete` in the event there is not an active REPL. 

Example stack trace: 

```
  sesman-ensure-session(CIDER)
  cider-repls(nil ensure)
  cider-current-repl(nil ensure)
  cider-nrepl-op-supported-p("complete")
  cider-complete("as")
  cider-company-unfiltered-candidates("as" ("debug-on-entry") #f(compiled-function (str) #<bytecode 0x1ffd982b1305>) 0)
  #f(compiled-function (style) #<bytecode 0x1ffd9830d4d9>)(cider)
```

# Reproducing

Use `emacs -Q -l startup.el` with the following file, and then M-x to get the error.

```
(setq user-emacs-directory "/tmp/.emacs.d/")
(setq package-archives '(("melpa-stable" . "http://stable.melpa.org/packages/")
                         ("gnu" . "http://elpa.gnu.org/packages/")))

(package-initialize)
(package-refresh-contents)
(package-install 'helm)
(package-install 'cider)
(package-install 'company)

(require 'helm)
(require 'helm-config)
(define-key global-map [remap execute-extended-command] 'helm-M-x)
(helm-mode 1)
(require 'cider)
(require 'company)

(global-company-mode)
(cider-company-enable-fuzzy-completion)
```

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -162,7 +162,9 @@ we check if cider-nrepl's complete op is available
 and afterward we fallback on nREPL's built-in
 completion functionality."
   (cond
-   ;; First we try if cider-nrepl's completion is available
+   ;; if we don't have a connection, end early
+   ((not (cider-current-repl nil)) nil)
+   ;; next we try if cider-nrepl's completion is available
    ((cider-nrepl-op-supported-p "complete")
     (let* ((context (cider-completion-get-context))
            (candidates (cider-sync-request:complete prefix context)))
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -163,7 +163,7 @@ and afterward we fallback on nREPL's built-in
 completion functionality."
   (cond
    ;; if we don't have a connection, end early
-   ((not (cider-current-repl nil)) nil)
+   ((not (cider-connected-p)) nil)
    ;; next we try if cider-nrepl's completion is available
    ((cider-nrepl-op-supported-p "complete")
     (let* ((context (cider-completion-get-context))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bugs fixed
 
 * [#2953](https://github.com/clojure-emacs/cider/issues/2953): Don't font-lock function/macro vars as vars.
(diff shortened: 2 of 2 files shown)


cider-completion.el
```diff
@@ -162,7 +162,9 @@ we check if cider-nrepl's complete op is available
 and afterward we fallback on nREPL's built-in
 completion functionality."
   (cond
-   ;; First we try if cider-nrepl's completion is available
+   ;; if we don't have a connection, end early
+   ((not (cider-current-repl nil)) nil)
```
bbatsov, line 166: There's also `cider-connected-p`, which is probably a better fit here.

bbatsov:
The changes look good. Don't forget to reflect them in the changelog.

PR #2981, opened 2021-02-09
Quoting is not necessary for middleware list.

### LINKED ISSUE #2980: cider-jack-in with clojure-cli on windows doesn't work
`clojure-cli` installed using https://github.com/clojure/tools.deps.alpha/wiki/clj-on-Windows .

Quoting seems wrong for the following line. 
https://github.com/clojure-emacs/cider/blob/b82fd1cbee7ecac05939ab0d5009d700d1cc21eb/cider.el#L574

I changed it to 
````
(apply-partially #'format "%s")
````
It seems to work.

jpathy:
This also fixes cider-jack-in for `'clojure-cli` not working on windows. 

Fixes #2980.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -571,7 +571,7 @@ one used."
                                 (append dependencies cider-jack-in-lein-plugins))
                        " "))
          (middleware (mapconcat
-                      (apply-partially #'format "\\\"%s\\\"")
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

Issue #2907, opened 2020-10-10
Provide ability to customise cljfmt code formatting


BrunoBonacci:
**Describe the solution you'd like**

It should be possible to provide a custom formatting configuration for cider-nrepl/cljfmt.
The cljfmt configuration format is described here:

https://github.com/weavejester/cljfmt#configuration

The desired solution should have a `defcustom` variable where the configuration can be provided.

**Describe alternatives you've considered**

A workaround to this solution is to use an `after-save-hook` as described here:

https://github.com/dzer6/cljfmt-graalvm

However, in this case, the two formatters (cider, external cljfmt) will keep fighting each other causing an awful user experience.

bbatsov:
> However, in this case, the two formatters (cider, external cljfmt) will keep fighting each other causing an awful user experience.

How come? Unless you trigger manually `cljfmt` in CIDER it won't be invoked. Keep in mind that the default indentation engine is actually written in Emacs Lisp. 

As mentioned in the chat solving this issue is not particularly complex. Seems we already added the ability to pass configuration to the middleware op responsible for invoking `cljfmt`, but we never implemented this in the client.

BrunoBonacci:
So I guess it is fighting against the (elisp) Indentation engine.
Is there any way I can control/customise the behaviour of that one?

The issue I have is that the indentation works differently depending on whether you are connected to the repl or not.
If you are connected to the REPL and you have evaluated the namespace then things like `:style/indent` are respected;
while if the repl isn't running or the namespace isn't evaluated the indentation doesn't know about the `:style/indent`
value, so for obvious reasons, it will indent the forms differently.
If you add things like `aggressive-indent-mode` then the behaviour is even more bizarre. Forms start jumping left or right based on whether the engine has evaluated that form or not.

I'd like to replace the current behaviour with something that is a bit more deterministic and it works the same whether the repl  has evaluated a form or not.

bbatsov:
You can just disable dynamic (REPL-powered) indentation if you don't rely on it https://docs.cider.mx/cider/0.26/config/indentation.html Alternative you can completely override the indentation logic of `clojure-mode` and provide whatever implementation you want. CIDER simply hooks into it and adds new indentation rules there dynamically when connected.

BrunoBonacci:
Thanks for the link. I'll see if I can achieve what want with the configuration described there.

However, I still think that this ticket stands. Having the possibility to configure the `cljfmt` indentation rules is a nice-to-have feature.

[...]

PR #2982, opened 2021-02-16
Provide ability to customise cljfmt code formatting

iarenaza:

It should be possible to provide a custom formatting configuration for
cider-nrepl/cljfmt. The cljfmt configuration format is described here:

  https://github.com/weavejester/cljfmt#configuration

The desired solution should have a defcustom variable where the
configuration can be provided.

A workaround to this solution is to use an after-save-hook as
described here:

https://github.com/dzer6/cljfmt-graalvm

However, in this case, the two formatters (cider, external cljfmt)
will keep fighting each other causing an awful user experience.

**Additional context**

In https://github.com/clojure-emacs/cider/issues/2907 Bozhidar Batsov
commented:

 > You can just disable dynamic (REPL-powered) indentation if you
 > don't rely on it
 > https://docs.cider.mx/cider/0.26/config/indentation.html
 > Alternative you can completely override the indentation logic of
 > clojure-mode and provide whatever implementation you want. CIDER
 > simply hooks into it and adds new indentation rules there
 > dynamically when connected.

To which Brunno Bonacci responded:

  > However, I still think that this ticket stands. Having the
  > possibility to configure the cljfmt indentation rules is a
  > nice-to-have feature.

-----------------

Thanks!

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -210,6 +210,47 @@ buffer, defaults to (cider-current-repl)."
 If NS is non-nil, include it in the eval request."
   (nrepl-sync-request:eval input (or connection (cider-current-repl nil 'ensure)) ns))
 
+(defcustom cider-format-code-options nil
+  "A map of options that will be passed to `cljfmt' to format code.
+Assuming this is the Clojure map you want to use as `cljfmt options':
+
+  {:indents {org.me/foo [[:inner 0]]}
+   :alias-map {\"me\" \"org.me\"}}
+
+you need to encode it as the following plist:
+
+  '((\"indents\" ((\"org.me/foo\" ((\"inner\" 0))))) (\"alias-map\" ((\"me\" \"org.me\"))))"
+  :type 'list
+  :group 'cider
+  :package-version '(cider . "1.0.0"))
+
+(defun cider--nrepl-format-code-request-map (&optional format-options)
+  "Map to merge into requests that require code formatting.
+If non-nil, FORMAT-OPTIONS specifies the options cljfmt will use to format
+the code.  See `cider-format-code-options` for details."
+  (when format-options
+    (let* ((indents-dict (when (assoc "indents" format-options)
+                           (thread-last
+                               (cadr (assoc "indents" format-options))
+                             (map-pairs)
+                             (seq-mapcat #'identity)
+                             (apply #'nrepl-dict))))
+           (alias-map-dict (when (assoc "alias-map" format-options)
+                             (thread-last
+                                 (cadr (assoc "alias-map" format-options))
+                               (map-pairs)
+                               (seq-mapcat #'identity)
+                               (apply #'nrepl-dict)))))
+      (thread-last
+          (map-merge 'list
+                     (when indents-dict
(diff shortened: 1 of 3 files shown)


cider-client.el
```diff
@@ -210,6 +210,47 @@ buffer, defaults to (cider-current-repl)."
...
+
+  {:indents {org.me/foo [[:inner 0]]}
+   :alias-map {\"me\" \"org.me\"}}
+
+you need to encode it as the following plist:
+
+  '((\"indents\" ((\"org.me/foo\" ((\"inner\" 0))))) (\"alias-map\" ((\"me\" \"org.me\"))))"
+  :type 'list
+  :group 'cider
+  :package-version '(cider . "1.0.0"))
```
bbatsov, line 225: This should say `1.1.0`.

cider-client.el
```diff
@@ -210,6 +210,47 @@ buffer, defaults to (cider-current-repl)."
 If NS is non-nil, include it in the eval request."
   (nrepl-sync-request:eval input (or connection (cider-current-repl nil 'ensure)) ns))
 
+(defcustom cider-format-code-options nil
+  "A map of options that will be passed to `cljfmt' to format code.
+Assuming this is the Clojure map you want to use as `cljfmt options':
```
bbatsov, line 215: I think you probably wanted to put quotes only around `cljfmt` here.

cider-client.el
```diff
@@ -712,12 +753,19 @@ The result entries are relative to the classpath."
     (cider-nrepl-send-sync-request)
     (nrepl-dict-get "fn-deps")))
 
-(defun cider-sync-request:format-code (code)
-  "Perform nREPL \"format-code\" op with CODE."
-  (thread-first `("op" "format-code"
-                  "code" ,code)
-    (cider-nrepl-send-sync-request)
-    (nrepl-dict-get "formatted-code")))
+(defun cider-sync-request:format-code (code &optional format-options)
```
bbatsov, line 756: I think you can just stick the options directly in the request, as this is how we did it in other places. Technically passing this is a param is better, but it's unlikely to ever be used without the options in CIDER.

bbatsov:
Apart from my small remarks the changes look good. Don't forget to mention them in the changlog and the configuration docs.

bbatsov:
@iarenaza ping :-)

bbatsov:
@iarenaza Ping 2 :-)

iarenaza:
@bbatsov Sorry for the delay. Life's got in the way :disappointed: 

I already did the changes you suggested and had added the mention in the changelog. But I was unsure about where to put this in the documentation. There isn't any mention in the documentation at the moment of code formatting via `cljfmt`. So I was doubting between adding it in the "Using CIDER > Misc Features" and "Configuration > Indentation". How do you see it?

bbatsov:
> So I was doubting between adding it in the "Using CIDER > Misc Features" and "Configuration > Indentation". How do you see it?

I'm fine with both, but I'm leaning towards "Misc Features", as it's not directly related to indentation.

Issue #2983, opened 2021-02-17
annoying error when quiting cider in emacs-27.1 on windows


ikappaki:
There is an annoying issue when quiting cider in emacs 27.1 on
  windows. The following puzzling error message is displayed to the
  user with a loud bang:

  _error in process sentinel: Could not start nREPL server:_

  To reproduce with the minimum steps taken in emacs-27.1 under windows 10:

  1. Create a deps.edn file with an empty {} dictionary as contents and open it in emacs.
  2. Cider-jack-in-clj (C-c M-j); a connection to the REPL is opened.
  3. cider-quit (C-c C-q) the repl.
  4. above error message is displayed, there shouldn't be any errors since nothing has gone wrong.

The issue is caused by emacs bug #46284
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=46284 and only affects
27.1 (fixed in subsequent versions). 

PR to follow.

;; CIDER 1.0.0 (Sofia), nREPL 0.8.3
;; Clojure 1.10.1, Java 11.0.7

GNU Emacs 27.1 (build 1, x86_64-w64-mingw32) of 2020-11-19

Windows 10

ikappaki:
This issue has been fixed with PR #2984

PR #2984, opened 2021-02-17
correct signal descriptions in nrepl server sentinel as a workaround for emacs bug #46284

ikappaki:
Please consider a potential fix for issue #2983.

Emacs has switched to a new and improved dumper in 27.1 which caused a slight regression issue.  The unix signal descriptions faked under windows, are not carried over in the dump with new pdumper and as thus any attempt to convert a signal to a signal description (such as SIGKILL to "Killed") under windows results to an "uknown signal" being generated. The signal description is passed as the second argument to the process sentinel. Thus whoever uses that description to take decisions in their programs is potentially afffected by it. See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=46284 for more details.

When cider-quit's, it sends the interrupt signals to the process which triggers a call to the nrepl-server-sentinel. The latter, due to this emacs bug, receives an "uknown signal" EVENT instead of "Interrupt", an EVENT it knows nothing about and bails out with an empty problem description resulting to the following error being raised:

_error in process sentinel: Could not start nREPL server:_

The issue has been since fixed in 27.1 and the forthcoming 28 branch.

This fix consists of having nrepl-server-sentinel to correct the EVENT description at reception when necessary (if it is a signal and when it is running under 27.1/windows). The fix can be removed once 27.1 is past the end of its support life.

I have a test in the making  to validate the nrepl client lifecycle process, but it is taking me more time than anticipated due at least one more bug discovered in emacs under windows and some inconsistencies with running eldev tests under windows.

Anyway, I thought it would be good idea to take this patch out for review while putting the final touches to the test (which is much more involved).

thanks

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1063,6 +1063,13 @@ been determined."
             (when nrepl-on-port-callback
               (funcall nrepl-on-port-callback (process-buffer process)))))))))
 
+(defmacro emacs-bug-46284/when-27.1-windows-nt (&rest body)
+  "Only evaluate BODY when Emacs bug #46284 has been detected."
+  (when (and (eq system-type 'windows-nt)
+             (string= emacs-version "27.1"))
+    (cons 'progn body)))
+
+
 (declare-function cider--close-connection "cider-connection")
 (defun nrepl-server-sentinel (process event)
   "Handle nREPL server PROCESS EVENT."
@@ -1075,6 +1082,19 @@ been determined."
                       (with-current-buffer server-buffer
                         (buffer-substring (point-min) (point-max)))
                     "")))
+    (emacs-bug-46284/when-27.1-windows-nt
+     ;; There is a bug in emacs 27.1 (since fixed) that sets all EVENT
+     ;; descriptions for signals to "unknown signal". We correct this by
+     ;; reseting it back to its canonical value.
+     (when (eq (process-status process) 'signal)
+       (cl-case (process-exit-status process)
+         ;; SIGHUP==1 emacs nt/inc/ms-w32.h
+         (1 (setq event "Hangup"))
+         ;; SIGINT==2 x86_64-w64-mingw32/include/signal.h
+         (2 (setq event "Interrupt"))
+         ;; SIGKILL==9 emacs nt/inc/ms-w32.h
+         (9 (setq event "Killed")))))
+
     (when server-buffer
       (kill-buffer server-buffer))
     (cond
--- a/test/nrepl-server-mock.el
+++ b/test/nrepl-server-mock.el
@@ -1,0 +1,105 @@
+;; nrepl-server-mock.el
+
+;; Copyright © 2021 Ioannis Kappas
+
+;; This file is NOT part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation, either version 3 of the
+;; License, or (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see `http://www.gnu.org/licenses/'.
+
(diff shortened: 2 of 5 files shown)


nrepl-client.el
```diff
@@ -1063,6 +1063,13 @@ been determined."
             (when nrepl-on-port-callback
               (funcall nrepl-on-port-callback (process-buffer process)))))))))
 
+(defmacro emacs-bug-46284/when-27.1-windows-nt (&rest body)
```
bbatsov, line 1066: Isn't a macro an overkill for this? After all we only need this check at one place.

bbatsov:
I'm fine with the suggest change. Don't forget to add a changelog entry, please.

nrepl-client.el
ikappaki: I wanted to elide execution of the when condition on non 27.1/windows platforms, whence the macro, i.e. it's a non-op for everything else. Thus it is an optimisation rather than overkill?

bbatsov: Fair enough.

ikappaki:
Thanks will do. Let me finish with the accompanying test before merging, I should have something ready for review soon.

ikappaki:
Hi again,

Could you please review follow up buttercup test to test the lifecycle
(start and stop) of an nREPL client process? It currently fails to
bootstrap on emacs 25.3 and I am trying to get an older build up to
check what is wrong with it. Missing commentary to follow in the two 
newly created files

The test passes on both linux and windows, though running the full
eldev test suite under windows has a few tests failing, which seems to
suggest that the test suite is not supported under windows perhaps? If
so I will have a look at fixing these in a separate commit.

The nrepl-client-lifecycle test starts a mock nREPL server with
`nrepl-start-server-process', retrieves the address:port it is
listening at, and initiates a client nREPL connection to it with
`nrepl-start-client-process'. It then deletes the client process which
also brings down the mock server process.

The mock server is defined in nrepl-server-mock-filter.el and is just
a network process with a naive filter intelligent enough to respond
gracefully to the handshake request of the client connection
(basically implementing the clone and describe session ops) but can be
easily extended to support more.

The mock server is started as a separate process by the lifecycle test
via an emacs --batch command defined in
`nrepl-server-mock-invocation-string' at nrepl-test-utils.el file. The
command ensures that the mock server is invoked with the same emacs
version and user options used by the parent eldev process which
invoked the test. The server MESSAGEs out the address:port it is
listening to for the client to pick up. There is an unresolved emacs
bug #46388 (https://debbugs.gnu.org/cgi/bugreport.cgi?bug=46388) that
buffers stderr and as a consequence prevents messages to be received
by the parent process. The server works around this by introducing
enough padding characters after the message so that the messages
reaches the invoking process.

The mock server can also log messages to a file using a logging
facility introduced with the new `cider-tests-log/init!'. It is
currently disabled and all log statements are elided, but when
'enabled it logs to a file including an identifying NAME of the
process, which is useful for investigating message ordering when
multiple processes are writing to a file (e.g. an nrepl client process
and nrepl server process writing to the same log file).

Finally, the test uses the `nrepl-tests-sleep-until' macro to do the
necessary periodic status checking of the processes, so that it
carries on when the condition is met (such as when checking the
process status) rather than idle sleeping for a few seconds and
prolonging the test's duration.

Thanks

ikappaki:
25.3 test failure fixed.

ikappaki:
Hi I've updated the conventional headers and the change log, I don't expect to make any more  changes other than those coming from the review. Thanks

bbatsov:
You're changes seem fine to me. I appreciate the effort to implement the nREPL client lifecycle test. Just squash the related commits together and we're good to go.

test/nrepl-server-mock.el
```diff
@@ -0,0 +1,99 @@
...
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see `http://www.gnu.org/licenses/'.
+
+;;; Commentary:
+
+;; This file is part of CIDER
```
bbatsov, line 22: You might add some of your final PR comment here for easy future reference.

ikappaki:
Thanks a lot for the review! squashed, commited, tests passing and ready to merge.

bbatsov:
Thanks for tackling this!

PR #2985, opened 2021-02-17
Add command `cider-toggle-ignore-next-form`.

witek:
New command `cider-toggle-ignore-next-form` for toggling Clojure "ignore next form" reader macro `#_`.

When the cursor is on whitespace toggle on the next form. Otherwise toggle on the current form.

I use this a lot for debugging purposes. Especially to disable and enable the last item in a threading macro form like 
```clojure
(-> [:a :b nil :d]
    (remove nil?)
    #_(sort))
```
where a `;;` comment on the last line would break the S-Expression.

-----------------

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -836,6 +836,27 @@ file path otherwise."
                 (t (error "Unexpected archive: %s" d))))
         cider-jdk-src-paths)))))
 
+(defun cider-toggle-ignore-next-form ()
+  "Toggle \"Ignore next form\" reader macro.
+When on whitespace toggle on the next form. Otherwise toggle on the current form."
+  (interactive)
+  (cl-flet ((cursor-is-on-ignore-form ()
+                                      (equal "#_"
+                                             (buffer-substring-no-properties
+                                              (point) (+ (point) 2)))))
+    (save-excursion
+      (if (cursor-is-on-ignore-form)
+          (delete-char 2)
+        (progn
+          (if (string-match "[\s\n]" (char-to-string (char-after)))
+              (skip-chars-forward "[\s\n]")
+            (progn
+              (forward-char 1)
+              (thing-at-point--beginning-of-sexp)))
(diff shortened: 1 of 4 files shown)


bbatsov:
I don't see anything here that requires a REPL, so it's probably bettter to submit this to `clojure-mode` instead.

cider-util.el
```diff
@@ -836,6 +836,27 @@ file path otherwise."
                 (t (error "Unexpected archive: %s" d))))
         cider-jdk-src-paths)))))
 
+(defun cider-toggle-ignore-next-form ()
```
bbatsov, line 839: What's the value of doing this for the next form? Typing those characters is probably faster the typing the keybinding. As I said in the chat, it seems to me that would make more sense of the preceding or containing form, as it would save you from some movement.

yuhan0:
There are a few proposals for this over at the clojure-mode repo: 
https://github.com/clojure-emacs/clojure-mode/issues/567
https://github.com/clojure-emacs/clojure-mode/issues/411
(PS. for evil or other modal editing users, it does take more keystrokes to enter insert mode and type `#_`, especially if a numeric argument toggles several `#_#_#_...` in a row)

cider-util.el
witek: It works on the next form, even if the cursor is on the whitespace before the form. This is my usual usecase in the threading macro example.

It works on the form/symbol under cursor.

I am using evil key bindings and have it bound to `#` in normal mode. So it is just one keystroke instead of four `i#_<ESC>`.

witek:
> I don't see anything here that requires a REPL, so it's probably bettter to submit this to `clojure-mode` instead.

I have just discoverd the difference. Sorry for that.

bbatsov:
I'd suggest to continue the conversation in https://github.com/clojure-emacs/clojure-mode/issues/567 Sorry that I dropped the ball on this one.

PR #2989, opened 2021-02-25
Explain how to use eval for dynamic path translation configuration

iperdomo:
Using `eval` *pseudo-variable*, we can use `(clojure-project-dir)` and `(getenv)` functions to configure `cider-path-translations` custom variable


bbatsov:
Thanks!

PR #2990, opened 2021-02-26
Add svg rendering support to cider-repl

sritchie:
This PR gives Cider the ability to render SVG images at the REPL.

This came up while I was implementing https://github.com/applied-science/emacs-vega-view/pull/11 ;  I realized that with this small change, we could get both Vega(-lite) charts AND generated LaTeX rendering at the REPL.

Here are the results of this change, on a Vega-generated svg file:

image

-----------------


bbatsov:
Nice little change!

Issue #1981, opened 2017-04-28
cider-stack-trace-render-frame fails to handle nil frames


cddr:
Sometimes cider-stacktrace doesn't render the stack as nicely as it could and you get the message "error in process filter: Format specifier doesn’t match argument type". I toggled debug-on-error and it took me to the `cider-stacktrace-render-frame` function. If I just add `(when frame...)`, it seems to work but I'm not sure if that's just masking the real problem somewhere else in the stack.

This seems to happen when you have a chained exception with a NullPointerException somewhere in the chain. For example after adding the guard locally, I get this stacktrace

<img width="780" alt="screen shot 2017-04-28 at 10 39 39 am" src="https://cloud.githubusercontent.com/assets/48030/25523575/19bb761e-2bff-11e7-9b86-b4ac0207c6a1.png">

whereas before I'd get this

<img width="775" alt="screen shot 2017-04-28 at 10 40 10 am" src="https://cloud.githubusercontent.com/assets/48030/25523591/28815754-2bff-11e7-86b2-3342a88dbcc9.png">

If this change looks ok I can make a PR. Just wanted to check whether this was an ok way to fix it.

```clojure
(defun cider-stacktrace-render-frame (buffer frame)
  "Emit into BUFFER function call site info for the stack FRAME.
This associates text properties to enable filtering and source navigation."
  (when frame
    (with-current-buffer buffer
      (nrepl-dbind-response frame (file line flags class method name var ns fn)
        (let ((flags (mapcar 'intern flags))) ; strings -> symbols
          (insert-text-button (format "%26s:%5d  %s/%s"
                                      (if (member 'repl flags) "REPL" file) line
                                      (if (member 'clj flags) ns class)
                                      (if (member 'clj flags) fn method))
                              'var var 'class class 'method method
                              'name name 'file file 'line line
                              'flags flags 'follow-link t
                              'action 'cider-stacktrace-navigate
                              'help-echo "View source at this location"
                              'font-lock-face 'cider-stacktrace-face
                              'type 'cider-plain-button)
          (save-excursion
            (let ((p4 (point))
                  (p1 (search-backward " "))
                  (p2 (search-forward "/"))
                  (p3 (search-forward-regexp "[^/$]+")))
              (put-text-property p1 p4 'font-lock-face 'cider-stacktrace-ns-face)
              (put-text-property p2 p3 'font-lock-face 'cider-stacktrace-fn-face)))
          (insert "\n"))))))
```

[...]

PR #2991, opened 2021-03-03
Filter nil stack frames

rvlo:
When the stacktrace have nils, emacs freezes for a little while before
showing the error buffer.

Example code:

(defn slow-cider [x]
  (swap! x assoc :ok :not))

(slow-cider nil)

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -794,8 +794,9 @@ the NAME.  The whole group is prefixed by string INDENT."
 (defun cider-stacktrace-render-cause (buffer cause num note)
   "Emit into BUFFER the CAUSE NUM, exception class, message, data, and NOTE."
   (with-current-buffer buffer
-    (nrepl-dbind-response cause (class message data spec stacktrace)
-      (let ((indent "   ")
+    (nrepl-dbind-response cause (class message data spec stacktrace-with-nils)
+      (let ((stacktrace (seq-filter #'when stacktrace-with-nils))
+            (indent "   ")
             (class-face 'cider-stacktrace-error-class-face)
             (message-face 'cider-stacktrace-error-message-face))
         (cider-propertize-region `(cause ,num)


rvlo:
No good. Back to the drawing board.

bbatsov:
Are you sure the slowdown is related? What's your CIDER version?

rvlo:
> Are you sure the slowdown is related? What's your CIDER version?

I'm using the master branch.

Well, I get error messages in the messages buffer:
error in process filter: insert-text-button: Format specifier doesn’t match argument type
error in process filter: Format specifier doesn’t match argument type

Are you able to reproduce it with the example code?

I use doom emacs. Not sure if it's related, but with the following changes, it works fine now:
https://github.com/rvlo/cider/commit/ac68bf5b0f3369affebaf6a6e1e59655a52bc090

bbatsov:
> Are you able to reproduce it with the example code?

I haven't had the time to check this yet.

yuhan0:
This looks like a duplicate of #1981 ? 

The stacktrace parsing code should be made more robust against such cases, but in the meantime try adding `-XX:-OmitStackTraceInFastThrow` in your jvm-opts to avoid having nil stackframes in the first place.

bbatsov:
@yuhan0 Thanks for the reminder! Guess we never updated the docs to reflect this. Btw, which JDKs are affected by this?

yuhan0:
I always assumed it was a recent change, but a quick search says it's been in place since JDK5 in 2004 (!)
https://yoshihisaonoue.wordpress.com/2021/02/07/jvm-option-xx-omitstacktraceinfastthrow/
https://www.oracle.com/java/technologies/javase/release-notes-introduction.html

bbatsov:
Hmm, that's unexpected. I really wonder why so few people came across this problem if this has been the case forever.

rvlo:
Thanks for the tip, @yuhan0. I'm running with the flag and watching...


@bbatsov 
> Hmm, that's unexpected. I really wonder why so few people came across this problem if this has been the case forever.

It happens intermittently. It might be related to ns unloading, reloading... Not sure yet. But when it was happening, and it did a couple of different REPL sessions recently, the changes I made in the last link I posted prevented the delay and the error messages. I confirmed by switching branches back and forth and re-evaluating the cider buffer without restarting the repl, and then the behavior was consistent.

Issue #2993, opened 2021-03-17
cider-repl-set-ns can fail to work in cljs repl when cider-repl-require-ns-on-set is t


tomdl89:
## Expected behavior
`cider-repl-set-ns` sets the ns in cljs repl when `cider-repl-require-ns-on-set` is `t`

## Actual behavior
`cider-repl-set-ns` sets the ns in cljs repl only when `cider-repl-require-ns-on-set` is nil

## Steps to reproduce the problem
set `cider-repl-require-ns-on-set` to `t` and try to do `cider-repl-set-ns` in cljs buffer or repl

## Environment & Version information
clj & cljs project in Emacs 27

### CIDER version information
Master

thheller noted that `require` and `in-ns` are speical forms in cljs, so can't be wrapped in `do` (as they are in the current code)

PR #2994, opened 2021-03-17
require and in-ns ns in separate forms for cider-repl-set-ns

tomdl89:
I _think_ I had to create a new nrepl response handler in order to ensure it wouldn't cause _two_ new prompts. Tested in two cljs environments, where it works fine. Feedback very welcome!

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1183,7 +1183,9 @@ With a prefix argument CLEAR-REPL it will clear the entire REPL buffer instead."
         (cider-repl--clear-region start (1+ end))))))
 
 (defun cider-repl-require-ns-handler (buffer)
-  "Make an nREPL evaluation handler for the REPL BUFFER's ns require."
+  "Make an nREPL evaluation handler for the REPL BUFFER's ns require.
+Ignores a `done' response, so as to avoid a double prompt when called
+alongside `cider-repl-switch-ns-handler'."
   (nrepl-make-response-handler buffer
                                (lambda (_buffer _value))
                                (lambda (buffer out)
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1182,6 +1182,16 @@ With a prefix argument CLEAR-REPL it will clear the entire REPL buffer instead."
       (let ((inhibit-read-only t))
         (cider-repl--clear-region start (1+ end))))))
 
+(defun cider-repl-require-ns-handler (buffer)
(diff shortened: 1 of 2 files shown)


dpsutton:
That handler should have a docstring indicating its subtle behavior: to swallow the nil that comes back from requiring. I always go back and forth if the name should be what it does, ie `cider-repl-value-ignoring-repl-handler` instead of `cider-repl-require-ns-handler` and then the call site gives it its meaning. Oh, this code requires the namespace but ignores the return value (which is always nil?)

tomdl89:
fix for #2993

bbatsov:
Do we really need an extra handler for this? Seems to me that we can just do a `sync` eval for the `require` and then trigger the async eval for the `in-ns`. This should result in code that's more straightforward IMO.

cider-repl.el
```diff
@@ -1207,9 +1219,10 @@ command will prompt for the name of the namespace to switch to."
     (user-error "No namespace selected"))
   (cider-map-repls :auto
     (lambda (connection)
-      (cider-nrepl-request:eval (if cider-repl-require-ns-on-set
-                                    (format "(do (require '%s) (in-ns '%s))" ns ns)
-                                  (format "(in-ns '%s)" ns))
+      (when cider-repl-require-ns-on-set
+        (cider-nrepl-request:eval (format "(require '%s)" ns)
```
bbatsov, line 1223: Here I'd just do a sync eval of the require and we should be good.

tomdl89: awesome! I had a feeling there would be a more concise way of doing this. thanks

cider-repl.el
```diff
@@ -1207,9 +1207,9 @@ command will prompt for the name of the namespace to switch to."
     (user-error "No namespace selected"))
   (cider-map-repls :auto
     (lambda (connection)
-      (cider-nrepl-request:eval (if cider-repl-require-ns-on-set
-                                    (format "(do (require '%s) (in-ns '%s))" ns ns)
-                                  (format "(in-ns '%s)" ns))
+      (when cider-repl-require-ns-on-set
+        (cider-nrepl-sync-request:eval (format "(require '%s)" ns)))
```
bbatsov, line 1211: Don't forget the connection param here.

tomdl89: and the ns param?

bbatsov: It's not needed in this case.

bbatsov:
Thanks!

PR #2997, opened 2021-03-29
Fix debugger incorrectly locating #_ ignored forms

### LINKED ISSUE #2996: Debugger doesn't ignore commented out expressions
I am getting a really confusing behaviour when using the Cider debugger on a code with commented out blocks.
The debugger seems to stop on expressions inside the commented out block, yet it does things which are outside the block.

## Expected behavior

I am trying to debug into the following piece of code (a minimal example):

```
#dbg
(defn test-function []
  #_
  (let [a (java.lang.System/currentTimeMillis)]
    (println "This should have never happened!" a))
  (let [b (java.lang.System/currentTimeMillis)]
    (println "This is what I want to see!" b))
  (let [c (java.lang.System/currentTimeMillis)]
    (println "Let's finish this, alright!" c)))
```

I would expect the debugger to ignore the first `let` block commented out by the `#_` reader macro,
and stop first only at the second `let` block, so I could step through the rest of the code from there,
to execute the remaining `println` expressions. 

## Actual behavior

The debugger seems to stop first at the `println` statement inside the commented out `let` block.
Seemingly I can step through all the expressions in the commented out block twice, yet the lines which get printed out by the `println` statement are:
```
This is what I want to see! 1616675364833
Let's finish this, alright! 1616675369010
```

Even though the results seem correct, the experience was rather confusing. I definitely thought the debugger is showing something the wrong way.

As written above

### CIDER version information

```
;; CIDER 1.0.0snapshot, nREPL 0.8.3
;; Clojure 1.10.2, Java 1.8.0_275
```

### Lein/Boot version

`Leiningen 2.9.5 on Java 1.8.0_275 OpenJDK 64-Bit Server VM`

### Emacs version

`25.1.1`

### Operating system

Debian 9.13 (Stretch)

yuhan0:
This was a quick fix I made over a year ago which closes #2996, putting it up here without any tests for now.

Also includes a menubar commit I forgot to include in a previous PR #2735

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -414,6 +414,7 @@ In order to work properly, this mode must be activated by
     ["Continue non-stop" (cider-debug-mode-send-reply ":continue-all") :keys "C"]
     ["Move out of sexp" (cider-debug-mode-send-reply ":out") :keys "o"]
     ["Forced move out of sexp" (cider-debug-mode-send-reply ":out" nil true) :keys "O"]
+    ["Move to current position" (cider-debug-mode-send-reply ":here") :keys "h"]
     ["Quit" (cider-debug-mode-send-reply ":quit") :keys "q"]
     "--"
     ["Evaluate in current scope" (cider-debug-mode-send-reply ":eval") :keys "e"]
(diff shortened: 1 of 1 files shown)


bbatsov:
The suggested fix seems reasonable. Thanks!

PR #3002, opened 2021-04-09
[Inspector] Configure truncation limits

pdbrown:
TODO: adjust cider-required-middleware-version when published

The limits after which the inspector truncates collection members are now
configurable. Previously they were hardcoded to 5 and 150 for collection and
atom (non-collection) members.

Add keybindings to cider-inspector-mode:
- a  adjust atom truncation limit (default 150)
- c  adjust nested collection truncation limit (default 5)

and add defcustoms to adjust the defaults
- cider-inspector-max-atom-length
- cider-inspector-max-coll-size

---------------------------------

This continues the work in https://github.com/clojure-emacs/cider-nrepl/pull/694 and needs to be updated with the right cider-nrepl version number before it can merge.

-----------------

### DIFF
--- a/doc/modules/ROOT/pages/debugging/inspector.adoc
+++ b/doc/modules/ROOT/pages/debugging/inspector.adoc
@@ -49,6 +49,12 @@ You'll have access to additional keybindings in the inspector buffer
 | kbd:[s]
 | Set a new page size in paginated view
 
+| kbd:[c]
+| Set a new maximum size above which nested collections are truncated
+
+| kbd:[a]
+| Set a new maximum length above which nested atoms (non-collections) are truncated
+
 | kbd:[d]
 | Defines a var in the REPL namespace with current inspector value
 |===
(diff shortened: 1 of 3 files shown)


doc/modules/ROOT/pages/debugging/inspector.adoc
```diff
@@ -49,6 +49,12 @@ You'll have access to additional keybindings in the inspector buffer
 | kbd:[s]
 | Set a new page size in paginated view
 
+| kbd:[c]
```
bbatsov, line 52: I'd also mention the new defcustoms in the "Config" section. Seems we forgot to document any of the existing config options, but we should change this.

cider.el
```diff
@@ -406,7 +406,7 @@ Elements of the list are artifact name and list of exclusions to apply for the a
 (defconst cider-latest-clojure-version "1.10.1"
   "Latest supported version of Clojure.")
 
-(defconst cider-required-middleware-version "0.25.10"
+(defconst cider-required-middleware-version "0.26-SNAPSHOT-TODO"
```
bbatsov, line 409: Here it should be 0.26.0-SNAPSHOT

doc/modules/ROOT/pages/debugging/inspector.adoc
pdbrown: Sounds good! I added those and a brief mention of the fill-frame setting, so all `cider-inspector.el` defcustoms are covered.

bbatsov:
Looks good. Thanks!

PR #3004, opened 2021-04-16
Use appropriate coding system when unzipping jars

pdbrown:
in `cider-find-file`. This prevents CRLF encoded files from appearing with ^M
characters at the end of each line, which, in turn, avoids confusing
`clojure-find-ns` and breaking `cider-find-var`.
`clojure-find-ns` uses Emacs' `(thing-at-point 'symbol)` as part of identifying
a file's namespace, and when a file isn't decoded properly, namespaces can be
reported as `my.lib^M` which `cider-find-var` won't know what to do with.

-----------------

Hey again! I came across https://github.com/clojure-emacs/cider/discussions/2957 today, where @expez mentions `cider-find-var` is broken in jars. I do this all the time and it works fine, so I looked into the repro a little and found that `cider-find-var` was getting confused in `cljfx.api`, as extracted from its jar, because it has windows line endings.

The nrepl message trace:
```
(-->
  id         "14"
  op         "info"
  session    "5cf7e082-fe85-448e-bc8f-9bb99e038211"
  time-stamp "2021-04-15 14:33:35.951811315"

  ns         #("cljfx.api^M" 0 10 (fontified nil))
;;                        ^--- broken!

  sym        "event-handler/dispatch-effect"
)
(<--
  id         "14"
  session    "5cf7e082-fe85-448e-bc8f-9bb99e038211"
  time-stamp "2021-04-15 14:33:35.954095391"
  status     ("done" "no-info")
)
```

The ns value is obtained from `clojure-find-ns` via `(thing-at-point 'symbol)`, which will include the `^M` if it's there.
While we could wrap a `(string-trim-right ...)` around the ns name somewhere, that shouldn't really be necessary, and `clojure-mode` uses `thing-at-point` in several other ways so this might not be the only broken thing. The `^M`s are also a bit of an eyesore, so I think attempting to get rid of them is the way to go. People might still be broken by files with mixed encoding though.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@
 * [#2953](https://github.com/clojure-emacs/cider/issues/2953): Don't font-lock function/macro vars as generic vars.
 * [#2964](https://github.com/clojure-emacs/cider/issues/2964): Fix issue with `cider-company-enable-fuzzy-completion` and Helm.
 * [#2937](https://github.com/clojure-emacs/cider/issues/2937): Green fringe produced for extra line in rich comment block.
-* [#2996]((https://github.com/clojure-emacs/cider/issues/2937)): Fix debugger incorrectly locating `#_` ignored forms.
+* [#2996](https://github.com/clojure-emacs/cider/issues/2937): Fix debugger incorrectly locating `#_` ignored forms.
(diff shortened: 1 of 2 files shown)


bbatsov:
The fix looks good. Just mention it in the changelog and we're good to go.

bbatsov:
Thanks!

Issue #2934, opened 2020-11-23
eldoc-mode disabled, global-eldoc-mode ignored.


convexstrictly:
I upgraded to emacs 27.1 and cider 20201122.1448.  For some reason, eldoc-mode is always nil before and after jack-in.  global-eldoc-mode is t.  However until I enable eldoc-mode, I don't see any function signatures in the minibuffer.  

System Information
```
;; CIDER 1.0.0snapshot (package: 20201122.1448), nREPL 0.8.3
;; Clojure 1.10.1, Java 13

Linux desktop 4.4.0-194-generic #226-Ubuntu SMP Wed Oct 21 10:19:36 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

```

I even reduced my .emacs to just the following lines to ensure it wasn't some side-effect of some other configuration that was causing these problems.

```
(package-initialize)
(add-to-list 'package-archives 
	     '("melpa" . "http://melpa.org/packages/")
	     '("marmalade" . "http://marmalade-repo.org/packages/"))

(eval-when-compile
  (require 'use-package))
```

bbatsov:
I'm guessing this might be something to do with the eldoc setup function:

```
(defun cider-eldoc-setup ()
  "Setup eldoc in the current buffer.
eldoc mode has to be enabled for this to have any effect."
  ;; Emacs 28.1 changes the way eldoc is setup.
  ;; There you can have multiple eldoc functions.
  (if (boundp 'eldoc-documentation-functions)
      (add-hook 'eldoc-documentation-functions #'cider-eldoc nil t)
    (setq-local eldoc-documentation-function #'cider-eldoc))
  (apply #'eldoc-add-command cider-extra-eldoc-commands))
```

I'll investigate this further when I can.

bbatsov:
Btw, I can't reproduce this issue on Emacs 27. Everything works fine for me with only `global-eldoc-mode` enabled.

pdbrown:
I noticed this too, on Emacs 28: Before cider starts, clojure-mode buffers don't have eldoc as expected. Once cider is running, new buffers enable eldoc, also as expected. But after starting cider, we run `cider-enable-on-existing-clojure-buffers`, which does `cider-eldoc-setup` but doesn't retrigger `turn-on-eldoc-mode`.

PR #3008, opened 2021-04-21
Enable eldoc-mode in existing clojure buffers

pdbrown:
Fixes https://github.com/clojure-emacs/cider/issues/2934

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -261,7 +261,18 @@ See command `cider-mode'."
   (add-hook 'clojure-mode-hook #'cider-mode)
   (dolist (buffer (cider-util--clojure-buffers))
     (with-current-buffer buffer
-      (cider-mode +1))))
+      (cider-mode +1)
+      ;; In global-eldoc-mode, a new file-visiting buffer calls
+      ;; `turn-on-eldoc-mode' which enables eldoc-mode if it's supported in that
+      ;; buffer as determined by `eldoc--supported-p'.  Cider's eldoc support
+      ;; allows new buffers in cider-mode to enable eldoc-mode.  As of 2021-04,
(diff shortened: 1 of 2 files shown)


cider-connection.el
```diff
@@ -261,7 +261,12 @@ See command `cider-mode'."
   (add-hook 'clojure-mode-hook #'cider-mode)
   (dolist (buffer (cider-util--clojure-buffers))
     (with-current-buffer buffer
-      (cider-mode +1))))
+      (cider-mode +1)
+      (when global-eldoc-mode
```
bbatsov, line 265: I'm really puzzled by this part. In theory `global-eldoc-mode` should have trigged `eldoc-mode`, so it seems very weird to enable it manually if the global mode is enabled.

pdbrown: Yeah this does look very weird, so I just pushed an update to the inline comment. I looked into globalized minor modes a bit, and from what I can tell, they call their e.g. `turn-on-eldoc-mode` function when the global mode is toggled on or when new file-visiting buffers are created while it's on, and `turn-on-eldoc-mode` only actually enables eldoc-mode if `eldoc--supported-p` returns true when `turn-on-eldoc-mode` is called. So old clojure buffers did call `turn-on-eldoc-mode`, but it didn't do anything because eldoc wasn't supported before cider was enabled. Now that cider is enabled, eldoc is supported, but global minor modes won't re-run any `turn-on` functions in reponse to just enabling another minor mode (cider-mode), so we have to do it ourselves, but only if the user hasn't disabled global-eldoc-mode.

bbatsov:
Okay, that makes sense. Thanks for looking into this!

pdbrown:
No problem!

PR #3018, opened 2021-05-16
Port CIDER tests to MS-Windows

ikappaki:
Hi,

can you please consider patch to port the CIDER tests to MS-Windows, given the recent support of `Eldev` for windows.

There are basically two compatibility changes:

1. Root test paths (such as "/tmp/a-dir", "/home/host/project/src" or "/docker/src") are expanded with `expand-file-name`, so that they are also valid root paths on windows by the addition of the drive letter (e.g. "/tmp/a-dir" expands to "c:\temp\a-dir").
2. Command line arguments are quoted with `shell-quote-argument` rather than just `\`, given quoting rules are different on windows.

In addition, the CircleCI `.circleci/config.yml` configuration file has been updated to include two new jobs for running existing `test` and `lint` jobs on windows with `test-windows` and `test-windows-lint` respectively, using the latest version of Emacs available in the `Chocolatey` package manager. I suggest, to begin with, we only support the latest Emacs version (though we won't have control when the version changes, and CI won't kick in automatically when it happens, so there is some risk to be taken).

For the root test paths update, I've chosen to bind the expanded paths to symbols that resemble the original paths (e.g. `"/tmp/a-dir"` is bound to symbol `/tmp/a-dir`), so that the changes are easier to be validated by the reviewer. Though, it might be confusing to a user  who is just reading the tests and is unfamiliar with this change why `expand-file-name' has been used.

Thanks


test/cider-common-tests.el
```diff
@@ -88,71 +88,89 @@
     (cider--translate-path-to-nrepl file)))
 
 (describe "cider--translate-container-vm"
+  :var ((/docker/src (expand-file-name "/docker/src"))
```
bbatsov, line 91: I'd add some comments here to explain the need for this.

bbatsov:
The changes seem reasonable, although I'm concerned that there won't be anyone who'd like to look into Windows-build related problems in the long run. 

I wonder if it wouldn't be useful to have some section in the docs "Hacking CIDER on Windows" or something along those line. I'm using Windows myself, but I do all my development using WSL, and I've never had much of an incentive to run the native Emacs build for Windows.

test/cider-common-tests.el
ikappaki: Done.

ikappaki:
> The changes seem reasonable, although I'm concerned that there won't be anyone who'd like to look into Windows-build related problems in the long run.
> 
> I wonder if it wouldn't be useful to have some section in the docs "Hacking CIDER on Windows" or something along those line. I'm using Windows myself, but I do all my development using WSL, and I've never had much of an incentive to run the native Emacs build for Windows.

I've added an `MS-Windows compatibility` section under hacking mentioning the cases that required remediation for tests to work under windows. Luckily, there are only two simple mechanical rules to follow that hopefully also make sense to people not familiar with windows. If windows test runs ever fail, the devs can be pointed to that section on how to fix. If tests fail for other reasons, perhaps there has been a more serious flaw that would have prevented CIDER to run under windows, and thus the code should be reconsidered before merging, avoiding breaking people's environments picking up a broken release.

I believe CIDER should support windows natively (at least the latest Emacs version), since Emacs is both officially and well supported on that operating system. There are many people who either haven't tried or don't want to switch to WSL as their primary development environment.  IMHO we shouldn't treat them as second class citizens, that might perhaps make them miss the opportunity of using such a magnificent tool.

(and we can always later remove windows CI support, if it turns out to be a burden hampering evolution) 

Thanks!

bbatsov:
> I believe CIDER should support windows natively (at least the latest Emacs version), since Emacs is both officially and well supported on that operating system.

No argument from me on that point. Thanks for tackling this!

PR #3021, opened 2021-06-24
[Fix #3020] Session linking on Windows 

### LINKED ISSUE #3020: Session linking doesn't work right on Windows
When jumping to files on classpath they aren't properly linked with the current buffer in Windows.

Previously discussed [here](https://github.com/clojure-emacs/cider/discussions/2957) because I wasn't sure if this was a bug or not.

expez:
When jumping into a file on the classpath it isn't properly linked with
the current buffer at present.  This commit fixes that, by normalizing
the file paths returned by `cider-classpath-entries`.

Without this normalization step the comparison in
`sesman-friendly-session-p` returns the wrong results when comparing the
current buffer with entries on the class path.

The normalized path will look like this `c:/myFile` whereas the raw one,
on Windows, will look like this: `c:\\myFile`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#3020](https://github.com/clojure-emacs/cider/issues/3020): Fix session linking on Windows, e.g. when jumping into a library on the classpath.
+
 ## 1.1.1 (2021-05-24)
 
 ### Bugs fixed
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -542,7 +542,8 @@ REPL defaults to the current REPL."
                                   (let ((cp (thread-last classpath
                                               (seq-filter (lambda (path) (not (string-match-p "\\.jar$" path))))
                                               (mapcar #'file-name-directory)
-                                              (seq-remove  #'null))))
+                                              (seq-remove  #'null)
+                                              (seq-uniq))))
                                     (process-put proc :cached-classpath-roots cp)
                                     cp))))
         (or (seq-find (lambda (path) (string-prefix-p path file))
--- a/cider-client.el
+++ b/cider-client.el
@@ -582,9 +582,10 @@ resolve those to absolute paths."
 
 (defun cider-classpath-entries ()
   "Return a list of classpath entries."
-  (if (cider-nrepl-op-supported-p "classpath")
-      (cider-sync-request:classpath)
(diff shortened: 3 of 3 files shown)


bbatsov:
Are the returned paths relative to the project root or what?

expez:
> Are the returned paths relative to the project root or what?

As mentioned in the commit message it just changes: `c:\\myFile` to this  `c:/myFile` . All the paths that showed up in the debugger as I investigated this were absolute, it was just the path separators that were different.

bbatsov:
I'm asking mostly because I'm not sure if `expand-file-name` is the way to handle this, as it needs absolute paths to do what you want it to do. Seems in our scenario it should get the job done.

expez:
Yeah, the doc says it does two things: expand and normalize.  We really just need the normalization step, as all paths are already expanded, but I couldn't find a function that just does normalization.  If you know of a good alternative--maybe CIDER already has such a util?--swapping would improve the code.  I added a comment stating that we call this for the normalization that function affords, but it would've been better to call a function which did only that.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#3020](https://github.com/clojure-emacs/cider/issues/3020): Fix session linking on Windows, e.g. when jumping into a library on the classpath.
```
bbatsov, line 5: Just add a "Bugs fixed" heading as well.

cider-connection.el
```diff
@@ -542,7 +542,8 @@ REPL defaults to the current REPL."
                                   (let ((cp (thread-last classpath
                                               (seq-filter (lambda (path) (not (string-match-p "\\.jar$" path))))
                                               (mapcar #'file-name-directory)
-                                              (seq-remove  #'null))))
+                                              (seq-remove  #'null)
+                                              (seq-uniq))))
```
bbatsov, line 546: Is this somehow related to the normalization?

expez: No, I added this as its own commit https://github.com/clojure-emacs/cider/pull/3021/commits/13d4b7f7851694aebcd6157589ed9f8565770a56 

This list had mostly duplicates on my system.  We later loop over this list many times and were done 10x the amount of work needed (as we were comparing with all the duplicates over and over)

CHANGELOG.md
expez: Done

bbatsov:
Thanks!

Issue #3022, opened 2021-06-26
CIDER 1.1.1 Can't Handle Errors Without Stack Traces


timvisher:
## Expected behavior
Be able to eval `(< 1 nil)` and get a reasonable stack trace.

## Actual behavior
CIDER panics with `*Messages*`

```
error in process filter: cider-stacktrace-render-frame: Format specifier doesn’t match argument type
error in process filter: Format specifier doesn’t match argument type
```

and a Stacktrace buffer like

```
1. Unhandled java.lang.NullPointerException
   (No message)
```

## Steps to reproduce the problem

1. Open a clojure file against a JVM _without_ `-XX:-OmitStackTraceInFastThrow` passed.
2. Insert `(< 1 nil)`
3. Whack `M-x cider-eval-last-sexp RET`

This can be fixed by passing `-XX:-OmitStackTraceInFastThrow` to the underlying JVM via the appropriate channel.

For instance:

```
$ cat deps.edn
{:deps
 …
 :aliases {:dev {:jvm-opts ["-XX:-OmitStackTraceInFastThrow"]}}}
```

Followed by `M-0 M-x cider-jack-in RET` and appending`:dev` to the CL like

```
jack-in command: /usr/local/bin/clojure … -M:cider/nrepl:dev`
```

### CIDER version information

```
;; CIDER 1.1.1 (Plovdiv), nREPL 0.8.3
;; Clojure 1.10.3, Java 16.0.1
```

### `clojure` version

```
$ clojure --version
Clojure CLI version 1.10.3.855
```

### Emacs version

```
GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0) of 2021-03-25
```

### Operating system

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.14.6
BuildVersion:   18G9216
```

bbatsov:
I'm not sure what's the best behaviour in such scenarios, as some stack frames seem to be missing. I recall this problem has been around for ages, but we never decided on the best approach to tackle it. E.g.:

- adding the JVM opt automatically on jack-in
- displaying some more informative message that the stacktrace is looking offf
- try to somehow salvage the situation and display some stacktrace in CIDER

timvisher:
_Personally_ speaking I think adding that option by default seems like a great addition. I can't think of any interactive development situation where I _wouldn't_ want it on, but maybe I'm not thinking of something common?

timvisher:
[...]

PR #3023, opened 2021-06-28
cider-jack-in: Add `-XX:-OmitStackTraceInFastThrow` by default

timvisher:
Expands `cider-clojure-cli-jack-in-dependencies` to also add `:jvm-opts
["-XX:-OmitStackTraceInFastThrow"]` in the `:cider/nrepl` alias so that
there are less situations where CIDER encounters a stack trace that has no
traceback.

May be worth extending to the other `cider-*-jack-in-dependencies`
functions.

Fixes https://github.com/clojure-emacs/cider/issues/3022

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -575,7 +575,7 @@ one used."
                       (cider-jack-in-normalized-nrepl-middlewares)
                       ","))
          (main-opts (format "\"-m\" \"nrepl.cmdline\" \"--middleware\" \"[%s]\"" middleware)))
-    (format "%s-Sdeps '{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s]}}}' -M%s:cider/nrepl"
+    (format "%s-Sdeps '{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s] :jvm-opts [\"-XX:-OmitStackTraceInFastThrow\"]}}}' -M%s:cider/nrepl"
             (if global-options (format "%s " global-options) "")
             deps-string
             main-opts
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -371,7 +371,8 @@
       (let ((expected (string-join '("clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version \"0.8.3\"} "
                                      "cider/cider-nrepl {:mvn/version \"0.25.7\"}} "
                                      ":aliases {:cider/nrepl {:main-opts [\"-m\" \"nrepl.cmdline\" \"--middleware\""
-                                     " \"[cider.nrepl/cider-middleware]\"]}}}' -M:cider/nrepl")
+                                     " \"[cider.nrepl/cider-middleware]\"] :jvm-opts [\"-XX:-OmitStackTraceInFastThrow\"]}}}' "
+                                     "-M:cider/nrepl")
                                    "")))
         (setq-local cider-allow-jack-in-without-project t)
         (setq-local cider-clojure-cli-command "clojure")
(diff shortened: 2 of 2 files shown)


bbatsov:
I think it'd be weird if we only harcoded this for people using `clj`. Ideally the defaults should be consistent across all build tools.

dpsutton:
I think the only thing to do here is gracefully handle empty stack traces. We should not add jvm arguments. CIDER already injects quite a bit in order to run, requiring jvm args seems too far. Especially if people are already specifying this behavior. Not sure if both args is an error or one silently overrides the other.

Seems far preferable to just add some kind of nil check on the stacktrace and show what information is available, if any.

timvisher:
@dpsutton That seems like a good idea. @bbatsov Are you down with switching to that direction?

timvisher:
@dpsutton I will say that an argument for adding this as a default argument is that I can't think of many times when I would want my stack traces omitted _in dev._ It feels like a better default for a development environment.

Interestingly, `lein` appears to [set this by default already](https://github.com/technomancy/leiningen/blob/e270102447fb9ca1f33591d39b72b15d06cdc0ef/leiningen-core/src/leiningen/core/project.clj#L544-L546).

Maybe that points to the idea that `clj` is actually where this should be set by default rather than patched in via CIDER?

dpsutton:
> I will say that an argument for adding this as a default argument is that I can't think of many times when I would want my stack traces omitted in dev. 

Exceptions in hot paths will pay a penalty. Core.match uses this for backtracking I believe. I understand the gist of what you are saying but I disagree. The reason I think this is a bad default is that you've already shown where some tooling has a default. Having several layers to check for this (lein, CIDER) makes it _harder_ to get the behavior you want. I'd far prefer someone setting this flag (if desired) in the standard way and then no tooling mucking with it and confusing them why it has been overriden, or startup throws an error because the arg for ommitting stacktraces has been included twice, or they said don't create stacktraces and CIDER includes an option to create stacktraces.

I'm a big fan of CIDER helping out on top of existing tooling. If lein sets this, then that sets the expectation and its documentation will explain why this is set, and people have to live with their choices. I cannot imagine `clj` will add this by default. But adding another layer where options are set sounds more complicated than necessary.

geokon-gh:
As you say, at the very least I'd think CIDER should notice that your NullPointerException has no stack and suggest the potentially reason and how to modify things so that  the stacks appear again. I've had these empty NullPointerException errors pop up for months with no clue as to why there was no stack. I just figured my CIDER was somehow broken or misconfigured or it was some issue in my emacs init file. I finally spent some time searching online today and figured it out I need to add this magic flag. But I had no idea this was an "optimization" :roll_eyes: . Again, the way it's handled at the moment just makes CIDER look broken (though it's not really CIDER's fault.. :sweat_smile:  )

timvisher:
> @bbatsov Are you down with switching to that direction?

@bbatsov Sorry if I'm being too pushy here. Just a quick ping. :)

bbatsov:
Yeah, I'm totally find with adding some handling of those empty stack frames in CIDER. I assume that should be pretty empty to do.

timvisher:
> I assume that should be pretty _empty_ to do.

Was this a bit of freudian sarcasm? xD

I'll see if I have some time to dig into this. Is there any chance you could point me at some likely spots to handle this as well as a rough description of that way you'd like to see it implemented?

Initially my instinct is to have it as some kind of message directly in the stack trace buffer with a description of what could be done to turn the fast trace behavior off.

bbatsov:
> Was this a bit of freudian sarcasm? xD

Just an epic typo. 😆 

> Is there any chance you could point me at some likely spots to handle this as well as a rough description of that way you'd like to see it implemented?

I don't remember the code well, but I assume this should be handled somewhere in/near `cider-stacktrace-render-frame`. I guess we can just display some message in the minibuffer or something along those lines.

bbatsov:
Superseded by #3042.

PR #3024, opened 2021-07-03
[fix #3017] Annotate company completion kinds

### LINKED ISSUE #3017: company-mode now supports icons, but CIDER's backend needs adjusting
Company mode recently added new feature, which provides icons in the popup. CIDER's backend doesn't yet return icon information back to company, and company displays no icons, when connected to CIDER.

As @dgutov mentioned in the https://github.com/company-mode/company-mode/issues/1102#issuecomment-841711072, CIDER needs some adjusting to make icons work.

andreyorst:
Fix #3017

This adds icon mapping for various completion kinds, allowing company-mode users to enable icons for completion items.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#3017](https://github.com/clojure-emacs/cider/issues/3017): Annotate company completion kinds.
 * [#3020](https://github.com/clojure-emacs/cider/issues/3020): Fix session linking on Windows, e.g. when jumping into a library on the classpath.
 
 ## 1.1.1 (2021-05-24)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## master (unreleased)
 
-### Bugs fixed
+### New features
 
 * [#3017](https://github.com/clojure-emacs/cider/issues/3017): Annotate company completion kinds.
+
+### Bugs fixed
+
 * [#3020](https://github.com/clojure-emacs/cider/issues/3020): Fix session linking on Windows, e.g. when jumping into a library on the classpath.
 
 ## 1.1.1 (2021-05-24)
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -84,6 +84,26 @@ backend, and ABBREVIATION is a short form of that type."
   :group 'cider
   :package-version '(cider . "0.9.0"))
 
+(defconst cider-completion-kind-alist
+  '(("class" class)
+    ("field" field)
+    ("function" function)
+    ("import" class)
+    ("keyword" keyword)
+    ("local" variable)
+    ("macro" macro)
+    ("method" method)
+    ("namespace" module)
+    ("protocol" enum)
+    ("protocol-function" enum-member)
+    ("record" struct)
+    ("special-form" keyword)
+    ("static-field" field)
+    ("static-method" interface)
+    ("type" parameter)
+    ("var" variable))
+  "Icon mapping for company-mode.")
+
 (defcustom cider-completion-annotations-include-ns 'unqualified
   "Controls passing of namespaces to `cider-annotate-completion-function'.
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#3017](https://github.com/clojure-emacs/cider/issues/3017): Annotate company completion kinds.
```
bbatsov, line 7: I'd put this under "New features" or "Changes".

bbatsov:
Can you attach some screenshot of the end result?

andreyorst:
> Can you attach some screenshot of the end result?

sure!

methods variables and macros
image

namespace:

image

defrecord:

image

I'm not sure how to get to other icons, I don't have the code that will trigger those :)

cider-completion.el
```diff
@@ -84,6 +84,26 @@ backend, and ABBREVIATION is a short form of that type."
...
+    ("method" method)
+    ("namespace" module)
+    ("protocol" enum)
+    ("protocol-function" enum-member)
+    ("record" struct)
+    ("special-form" keyword)
+    ("static-field" field)
+    ("static-method" interface)
+    ("type" parameter)
+    ("var" variable))
```
expez, line 104: Maybe use `constant` for this (or vice versa) so you can easily tell the difference between local and global variables in the completion list.

bbatsov: I was going to change most of those afterwards anyways. E.g. I don't think that `macro` is a misc, but rather a built-in and so on. But I think there's little point in debating this too much in this particular PR.

andreyorst: Like this?
image

I think `constant` may be a bit misleading, but maybe it's ok. I'll let @bbatsov decide.

bbatsov: There are many nuances to consider and this has to be aligned with how clojure-mode works internally as well. That's why I think we can sort it out later. Just update the changelog and for now we're good to go.

andreyorst: > E.g. I don't think that macro is a misc

Btw I've changed this to use `macro` icon, but given that there's no `macro` icon in company mode, it will default to `misc` icon automatically. Originally I've used `misc` because I was not aware of this mechanism. Maybe we can submit an issue to company mode so they could include a macro icon? Or maybe this should be done to vscode, because the icons are from there, and this will benefit to other plugins that use NREPL, like Calva.

Also triangle square and circle kinda seem like a good icon to represent macros, as this is really puzzling :D

bbatsov: Emacs modes normally font-lock macros and built-in keywords the same. I assume that's why @dgutov opted not to add such an icon.

bbatsov:
Thanks!

cider-completion.el
dgutov: Right. We basically mirror VS Code's type. Not to say we can't extend the list, but getting nice icons can be a problem.

Anyway, check out https://github.com/emacs-mirror/emacs/blob/6ec3cf1ccb5380acc376e89140b8d3a7fa4e471a/lisp/progmodes/elisp-mode.el#L647-L655

Since there are no "real" keywords in Lisp, reusing this kind for macros and special forms seemed like a good idea.

dgutov: Regarding local variables, we use the `value` kind (following the example of some LSP servers). Which differentiates them nicely.

andreyorst: > Since there are no "real" keywords in Lisp, reusing this kind for macros and special forms seemed like a good idea.

Clojure actually does have real keywords, and CIDER annotates those with `<k>`:

image

bbatsov: In this context I actually meant keywords in the sense of "built-ins"/"reserved words". That's how elisp-mode and clojure-mode work. Anyways, not a big deal.

dgutov: @andreyorst These kind of keywords are present in Elisp as well.

But that's not what is usually called a keyword in most programming languages (in Ruby, we call them "symbols", for example). And as someone who routinely works with several languages, I like to see annotations (and icons) have more universal meanings.

PR #3026, opened 2021-07-18
[Fix #2831] add xref integration

### LINKED ISSUE #2831: Could CIDER provide an xref backend instead of hijacking `M-.` and `M-,`?

I'm frustrated with the way CIDER hijacks `M-.` in all clojure-mode buffers, even for projects where you don't connect to CIDER. I don't always want to connect to CIDER to navigate projects efficiently.

I would love for CIDER to provide an xref backend instead of hijacking `M-.`, similar to how `clojure-lsp` does. 

My hope is that this would mean xref would present results both from CIDER and `clojure-lsp`, and other backends.

A simpler (though less ideal) fix would perhaps be to fall back to `xref-find-definitions` if there's no CIDER connection for the project.

**Additional context**

jiacai2050:
-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-find.el
+++ b/cider-find.el
@@ -241,22 +241,26 @@ thing at point."
   "Return the relevant identifier at point."
   (cider--kw-to-symbol (cider-symbol-at-point 'look-back)))
 
+(defun cider--var-to-xref-location (var)
+  "Get location of definition of VAR."
+  (when-let* ((info (cider-var-info var))
+              (line (nrepl-dict-get info "line"))
+              (file (nrepl-dict-get info "file"))
+              (buf (cider--find-buffer-for-file file)))
+    (xref-make-buffer-location
+     buf
+     (with-current-buffer buf
+       (save-excursion
+         (goto-char 0)
+         (forward-line (1- line))
+         (back-to-indentation)
+         (point))))))
+
 (cl-defmethod xref-backend-definitions ((_backend (eql cider)) var)
   "Find definitions of VAR."
   (cider-ensure-connected)
   (cider-ensure-op-supported "ns-path")
-  (when-let* ((info (cider-var-info var))
-              (line (nrepl-dict-get info "line"))
-              (file (nrepl-dict-get info "file"))
-              (buf (cider--find-buffer-for-file file))
-              (loc (xref-make-buffer-location
-                    buf
-                    (with-current-buffer buf
-                      (save-excursion
-                        (goto-char 0)
-                        (forward-line (1- line))
-                        (back-to-indentation)
-                        (point))))))
+  (when-let* ((loc (cider--var-to-xref-location var)))
     (list (xref-make var loc))))
 
 (cl-defmethod xref-backend-identifier-completion-table ((_backend (eql cider)))
@@ -278,6 +282,21 @@ thing at point."
                      (line (nrepl-dict-get info "line"))
                      (loc (xref-make-file-location filename line column)))
                 (xref-make filename loc)))
+            results)))
+
+(cl-defmethod xref-backend-apropos ((_backend (eql cider)) pattern)
+  "Find all symbols that match regexp PATTERN."
+  (cider-ensure-connected)
+  (cider-ensure-op-supported "apropos")
+  (when-let* ((ns (cider-current-ns))
+              (results (cider-sync-request:apropos pattern ns t t completion-ignore-case)))
+    (mapcar (lambda (info)
+              (let* ((symbol (nrepl-dict-get info "name"))
+                     (loc (cider--var-to-xref-location symbol))
+                     (type (nrepl-dict-get info "type"))
+                     (doc (nrepl-dict-get info "doc")))
+                (xref-make (format "[%s] %s\n  %s" (propertize symbol 'face 'bold) (capitalize type) doc)
+                           loc)))
             results)))
 
 (provide 'cider-find)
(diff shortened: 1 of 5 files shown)


bbatsov:
You should also mention this in the docs.

cider-find.el
```diff
@@ -32,6 +32,13 @@
 
 (require 'thingatpt)
 
+(defcustom cider-enable-xref t
```
bbatsov, line 35: The way things are currently implemented this defcustom will do something only before `cider-mode` is started. Not a big deal, but it should be noted in the docs at least. 

I'm also thinking that a name like `cider-use-xref` might be more consistent with the current names we use. I'm also worried that some people might be confused with CIDER's own `xref` functions, so we have to be a bit more descriptive in the docs I guess.

bbatsov:
I'm assuming you'll also have to make the explicit keybindings for `M-.` and `M-,` conditional, if `xref` takes care of those.

cider-find.el
jiacai2050: Fixed.

doc/modules/ROOT/pages/usage/misc_features.adoc
```diff
@@ -343,3 +343,7 @@ to format EDN using `clojure.tools.reader.edn`. The following commands are provi
 * `cider-format-edn-defun`
 * `cider-format-edn-region`
 * `cider-format-edn-buffer`
+
+== https://www.gnu.org/software/emacs/manual/html_node/emacs/Xref.html[Xref] integration
+
+Beginning with version 1.2.0, cider support Emacs built-in xref, which means `M-.` (xref-find-definitions) will just work as expected.
```
bbatsov, line 349: You should probably mention the configuration variable here as well.

bbatsov:
One thing I don't quite get is how is this going to play with the `cider-mode` keybindings, as there are some explicit keybindings for `M-.` and `M-,` currently. @dgutov I'd appreciate any feedback from you as well, as I still haven't had the time to play with xref myself.

dgutov:
Looking at the current patch, the user will have change the bindings in `cider-mode-map` for it to work. So you should probably change the bindings if `cider-use-xref` is non-nil too.

Even better, though, would be for Cider to migrate to Xref conditionally. If there is no important functionality that will be missed this way, of course.

cider-find.el
```diff
@@ -226,5 +233,56 @@ thing at point."
...
+              (let* ((filename (nrepl-dict-get info "file"))
+                     (column (nrepl-dict-get info "column"))
+                     (line (nrepl-dict-get info "line"))
+                     (loc (xref-make-file-location filename line column)))
+                (xref-make filename loc)))
+            results)))
+
+(cl-defmethod xref-backend-apropos ((_backend (eql cider)) pattern)
+  "Find all symbols that match regexp PATTERN."
+  (user-error "xref-backend-apropos not supported"))
```
jiacai2050, line 285: For apropos, I try this:

```
(when-let* ((ns (cider-current-ns))
            (results (cider-sync-request:apropos pattern ns nil t completion-ignore-case)))
  results)
```
results have no file/line/column infomation, so cannot new xref location here. Any ideas?

jiacai2050:
I have implemented other xref functions, For apropos, I try this:

```
(cl-defmethod xref-backend-apropos ((_backend (eql cider)) pattern)
  "Find all symbols that match regexp PATTERN."
  (when-let* ((ns (cider-current-ns))
              (results (cider-sync-request:apropos pattern ns nil t completion-ignore-case)))
  results))
```
but results have no file/line/column infomation, so cannot new xref location here. Any ideas?

dgutov:
> but results have no file/line/column infomation, so cannot new xref location here. Any ideas?

If *some* information is there, perhaps in a different form so that you can't reuse xref-file-location or xref-buffer-location, you can create your own xref types and instantiate them, like elisp-mode does, see `elisp--xref-make-xref` and the constructor that is calls.

jiacai2050:
@dgutov Thanks for your guide.

I just realize I can reuse xref-buffer-location with another call to cider-var-info. 
PTAL

cider-find.el
```diff
@@ -226,5 +233,71 @@ thing at point."
...
+       (save-excursion
+         (goto-char 0)
+         (forward-line (1- line))
+         (back-to-indentation)
+         (point))))))
+
+(cl-defmethod xref-backend-definitions ((_backend (eql cider)) var)
+  "Find definitions of VAR."
+  (cider-ensure-connected)
+  (cider-ensure-op-supported "ns-path")
```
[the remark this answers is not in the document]
jiacai2050: I'm not quiet sure. I add this assert with reference to `cider-find-ns`

cider-find.el
```diff
@@ -226,5 +233,71 @@ thing at point."
     (cider--find-ns kw-ns arg)
     (search-forward-regexp kw-to-find nil 'noerror)))
 
+(defun cider--xref-backend ()
+  "Used for xref integration."
+  'cider)
+
+(cl-defmethod xref-backend-identifier-at-point ((_backend (eql cider)))
+  "Return the relevant identifier at point."
+  (cider--kw-to-symbol (cider-symbol-at-point 'look-back)))
```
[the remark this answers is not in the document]
jiacai2050: Fixed.

bbatsov:
I guess that's good enough for a first version. Let's see what kind of feedback we'll get from the end users. Thanks for tackling the ticket!

manuel-uberti:
Thanks to everyone involved in this.

PR #3039, opened 2021-08-31
Allow starting the sideloader for the tooling session

plexus:
CIDER's deferred loading of middleware means that requires can happen during all
kinds of nREPL ops, many of which are performed over the tooling session. To
make this work via the sideloader we need it to be active both on the regular
session and on the tooling session, so provide a flag for starting it in either
session.

-----------------

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -162,13 +162,14 @@ Signal an error if it is not supported."
   (unless (cider-nrepl-op-supported-p op)
     (user-error "`%s' requires the nREPL op \"%s\" (provided by cider-nrepl)" this-command op)))
 
-(defun cider-nrepl-send-request (request callback &optional connection)
+(defun cider-nrepl-send-request (request callback &optional connection tooling)
   "Send REQUEST and register response handler CALLBACK.
 REQUEST is a pair list of the form (\"op\" \"operation\" \"par1-name\"
                                     \"par1\" ... ).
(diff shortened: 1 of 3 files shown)


plexus:
I originally had the op automatically go to both sessions, the more granular version seems preferable though. Not sure if it really makes sense to add tests here or doc updates.

cider-client.el
```diff
@@ -162,13 +162,14 @@ Signal an error if it is not supported."
   (unless (cider-nrepl-op-supported-p op)
     (user-error "`%s' requires the nREPL op \"%s\" (provided by cider-nrepl)" this-command op)))
 
-(defun cider-nrepl-send-request (request callback &optional connection)
+(defun cider-nrepl-send-request (request callback &optional connection tooling)
```
bbatsov, line 165: I wonder if a more generic `session` parameter wouldn't be more appropriate here. Technically a connection can have other sessions than the two created by CIDER by default, and this would allow us to interact with them. I guess this will require a small change in `nrepl-send-request` as well, though.

bbatsov:
I don't think tests/doc updates for this are needed. I agree it makes sense for the tooling session to enable sideloading as well.

cider-client.el
bbatsov: Anyways, I can just do this myself down the road. Let's keep your change focused to the problem at hand.

PR #3040, opened 2021-08-31
Allow clojuredocs navigation on qualified symbols and within *cider-clojuredocs* buffer

ag91:
This is a feature suggested by user @dotemacs: when we are looking at
clojuredocs, it is often useful to jump to the next related function.

Before this change, cider-clojuredocs could not reach symbols like
true? or clojure.core/map.

Also in buffers with cider-mode enabled we can navigate with the
keybinding C-c C-d C-c. Here I try to support the same also in the
temporary buffer the cider creates for clojuredocs.

-----------------

Thanks!

manual][1] extremely useful.*


yuhan0:
I don't think this is the right approach -- `(thing-at-point 'symbol)` works as expected in clojure-mode buffers and doesn't need a separate regex based approach to replace it.
The problem is that the clojuredocs buffer uses the default fundamental-mode syntax table, which doesn't recognize `.` and `?` as symbol constituents. 
Executing `(set-syntax-table clojure-mode-syntax-table)` in the buffer solves the issue.

The question might be whether to apply this more generally to other popups, in `cider-make-popup-buffer`, or only to 
 `cider-create-clojuredocs-buffer`.

bbatsov:
I'd start small. :-) 

And I agree with the rest of your suggestions. In the clojuredocs buffers it might also be a good idea to convert related symbols to links that people can simply press.

yuhan0:
I was thinking the same thing :)
There's some relevant "See also" button code already implemented at the bottom of `cider-docview-render-info`.

ag91:
hey @yuhan0 thanks so much for your comment: I just learned that thing-at-point changes according to the mode AND how to set the syntax table! This PR was just worth the effort for your comment.
@bbatsov maybe I can try to add links in another PR? I am happy to try.

bbatsov:
Yeah, you can tackle this in separate PR. Thanks!

Issue #3037, opened 2021-08-31
Allow CIDER to load its own middleware via the sideloader


plexus:
nREPL now has [dynamic insertion of middleware](https://github.com/nrepl/nrepl/pull/185), as well a [sideloader](https://github.com/nrepl/nrepl/pull/162) for requesting resources from the client. CIDER has [basic support for sideloading](https://github.com/clojure-emacs/cider/commit/ac4fedf8b29519f4b4ccac4aa40770d0090810e6), so it would seem that all the pieces are there to be able to take a vanilla nREPL connection, and upgrade it to a fully CIDER-compliant connection.

After initial testing it seems this isn't quite there yet, there are small tweaks needed in CIDER, cider-nrepl, and nREPL, so this is a meta-issue to keep track of the overall goal and progress.

- [x] https://github.com/clojure-emacs/cider/pull/3038
- [x] https://github.com/clojure-emacs/cider/pull/3039
- [x] https://github.com/nrepl/nrepl/pull/243
- [x] https://github.com/clojure-emacs/cider-nrepl/pull/705
- [x] https://github.com/clojure-emacs/cider/pull/3041
- [x] https://github.com/nrepl/nrepl/pull/244
- [x] https://github.com/clojure-emacs/cider/pull/3044
- [ ] https://github.com/nrepl/nrepl/pull/246

Related

- https://github.com/clojure-emacs/cider/issues/3049

## How to try it out

Start a vanilla nREPL server:

```
clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.9.0-beta1"}}}' -M: -m nrepl.cmdline --port 1234
```

Connect to it from Emacs:

```emacs-lisp
(setq
 ;; Make sure we can inspect the nREPL traffic
 nrepl-log-messages t
 ;; Prevent the nrepl pprint middleware from looking for a print function that
 ;; isn't (yet) there. Not necessary but prevents some noise.
 cider-print-fn nil
 ;; Use the cider-nrepl jar that I `make install'ed locally
 ;; we need the classloader-related fixes in the deferred middleware
 cider-dynload-cider-nrepl-version "0.27.0-SNAPSHOT"
 ;; Make sure caching doesn't bite us
 cider-jar-content-cache (make-hash-table :test #'equal))

;; Start the connection
(cider-nrepl-connect
 '( :project-dir "/home/arne/clj-projects/timezoner"
    :host "localhost"
    :port 1234
    :repl-init-function nil
[...]

PR #3041, opened 2021-09-01
Sideloader: handle binary files, support multiple directories

plexus:
Make sure that files are served byte-for-byte as they are on disk, without emacs
doing any coding system conversion. This also prevents base64-encode from
complaining about multibyte characters.

Support multiple directories as sources for the sideloader, e.g.
cider-nrepl/src, cider-nrepl/resources, orchard/src

Part of https://github.com/clojure-emacs/cider/issues/3037

Discovered this because cider.nrepl.middleware.track-state contains a U+2019 RIGHT SINGLE QUOTATION MARK :)

-----------------

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -189,14 +189,29 @@ When invoked with a prefix ARG the command doesn't prompt for confirmation."
 
 ;;; Sideloader
 
-(defvar cider-sideloader-dir (file-name-directory load-file-name))
+(defvar cider-sideloader-dirs
+  (list (file-name-directory load-file-name))
+  "Directories where we look for resources requested by the sideloader")
+
+;; based on f-read-bytes
+(defun cider-read-bytes (path)
+  "Read binary data from PATH.
+Return the binary data as unibyte string."
+  (with-temp-buffer
+    (set-buffer-multibyte nil)
+    (setq buffer-file-coding-system 'binary)
+    (insert-file-contents-literally path nil)
+    (buffer-substring-no-properties (point-min) (point-max))))
 
 (defun cider-provide-file (file)
   "Provide FILE in a format suitable for sideloading."
-  (let ((file (expand-file-name file cider-sideloader-dir)))
-    (if (file-exists-p file)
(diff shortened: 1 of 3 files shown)


plexus:
Oops there's actually another important change in there, I should split that up. (serving from multiple directories)

cider-eval.el
```diff
@@ -189,14 +189,29 @@ When invoked with a prefix ARG the command doesn't prompt for confirmation."
 
 ;;; Sideloader
 
-(defvar cider-sideloader-dir (file-name-directory load-file-name))
+(defvar cider-sideloader-dirs
+  (list (file-name-directory load-file-name))
+  "Directories where we look for resources requested by the sideloader")
```
bbatsov, line 194: I think the missing `.` is breaking the CI.

bbatsov:
> Discovered this because cider.nrepl.middleware.track-state contains a U+2019 RIGHT SINGLE QUOTATION MARK :)

Wow! 🤣 

> Support multiple directories as sources for the sideloader, e.g., cider-nrepl/src, cider-nrepl/resources, orchard/src

Seems you've address my comment from the meta issue already. :D Probably we should make this eventually a defcustom to which we always append CIDER's source dir + lib (or something along those lines). 

Seems the win build has hit some CircleCI limit... No love for OSS. :-)

plexus:
That is some funny math going on at CircleCI

bbatsov:
That's pretty weird, as supposedly we should get a pretty high limit on an OSS plan https://circleci.com/open-source/ I guess I'll have to contact their support (if they provide any for the OSS projects).

plexus:
My contact at CircleCI says the limits for Windows builds are different since they're more expensive. Seems we should use them sparingly.

- Do we need to lint separately on Windows?
- Maybe we can convert the windows jobs to GH actions?

plexus:
OMG it worked... ok going to squash this. See https://github.com/clojure-emacs/cider/issues/3046 regarding the windows builds.

PR #3042, opened 2021-09-01
Fix errors on empty stacktraces

### LINKED ISSUE #3022: CIDER 1.1.1 Can't Handle Errors Without Stack Traces

yuhan0:
Fixes #3022, succeeds #3023.

I also edited the docs to reflect that Cider no longer throws an error, and added beginner-friendly instructions for setting the JVM opts. 
Note: only added examples for Clojure CLI, since Leiningen enables the flag by default and I'm not familiar with Boot. 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/modules/ROOT/pages/troubleshooting.adoc
+++ b/doc/modules/ROOT/pages/troubleshooting.adoc
@@ -281,4 +281,20 @@ to be cached and have no stacktrace.
 You can turn off the optimization by adding the JVM flag `-XX:-OmitStackTraceInFastThrow` to whatever
 command you're using to start nREPL.
 
+In particular, by adding it to a deps.edn file under an alias (eg. `:dev`)
+[source,lisp]
+---
+{:aliases
+ {:dev
+  {:jvm-opts ["-XX:-OmitStackTraceInFastThrow"]
+  ...}}}
+---
+
+Or by customising the jack-in options.
+[source,lisp]
+---
+(setq cider-clojure-cli-global-options "-J-XX:-OmitStackTraceInFastThrow")
+---
+
+
 TIP: You can read more about this problem https://github.com/clojure-emacs/cider/issues/3022[here].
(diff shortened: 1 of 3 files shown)


doc/modules/ROOT/pages/troubleshooting.adoc
```diff
@@ -269,16 +269,10 @@ counting from 0 while Clojure starts from 1. Accordingly, the `cider-nrepl`
 middleware uses 1-based indexing and all conversions to 0-based indexing should
 be handled client-side. See https://github.com/clojure-emacs/cider/issues/2852.
 
-=== Empty Java stacktraces resulting in `NullPointerException`
```
yuhan0, line 272: (I removed this clause since it's not clear that the stacktraces are only omitted for NPEs or cause them to occur)

bbatsov:
Thanks!

PR #3043, opened 2021-09-02
Fix typo in variable name

stig:
Fix typo in the variable name to disable CIDER's Xref integration

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Thanks!

PR #3044, opened 2021-09-07
Dynamically upgrade nrepl connection

plexus:
Piece together the final pieces to make dynamic loading + sideloading of
middleware work.

This introduces `M-x cider-upgrade-nrepl-connection`, which combines the different steps needed
- fetch the cider-nrepl jar
- start the sideloader
- insert the middleware
- reload the nREPL capabilities

This introduces a cider-jar mini library for looking up jars and files in jars fairly efficiently and in a cross-platform manner.

Part of https://github.com/clojure-emacs/cider/issues/3037

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,10 @@ commands:
       - run:
           name: Install Eldev
           command: curl -fsSL https://raw.github.com/doublep/eldev/master/webinstall/circle-eldev > x.sh && source ./x.sh
+      - run:
+          name: Install unzip
+          command: apt-get update && apt-get install unzip
+
   setup-windows:
     steps:
       - checkout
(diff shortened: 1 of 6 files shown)


plexus:
Apparently `string-replace` is an Emacs 28 thing.

plexus:
And it seems relying on internals of `arc-mode` wasn't smart either :sweat_smile:

plexus:
Good thing we test on Emacs 25/26/27, and I use 28. Seems in 28 the implementation of arc-mode was changed to use structs instead of vectors. Will look into the remaining failures tomorrow.

plexus:
God I hate linters.

plexus:
I give up...

``` emacs-lisp
     (if (functionp #'archive--file-desc-ext-file-name)
         #'archive--file-desc-ext-file-name
```

eldev lint is complaining about this, because it's not defined. Which is correct, it's defined on Emacs 28, which is why it's there. meanwhile I can only find out through CI because locally eldev lint gives me completely different results, and it seems there's no way to give it a hint to ignore this.

plexus:
Linter passed... IMO the code has gotten worse in the process but the machines are happy.

plexus:
That... actually worked :D CI is looking ok now, apart from the invisible windows builds.

bbatsov:
Great work! 🚀

PR #3047, opened 2021-09-08
Fix info lookup

plexus:
When doing a var-info lookup we potentially try three different strategies:
info, lookup, and eval. The second currently always fails, because the "info"
dict is not at the top level in the response message, but nested under an "info"
key.

Compare and contrast:

```emacs-lisp
 (cider-sync-request:info "clojure.core/str")
 (cider-sync-request:lookup "clojure.core/str")
 (cider-fallback-eval:info "clojure.core/str")
```

Running tests locally I get a failure, but it seems unrelated.

```
cider-repls when multiple sessions exist always returns the most recently used connection

  buttercup-fail("%s" "Expected `(cider-repls)' to be `equal' to `(#<buffer *b-session:b2:cljs*(8228)> #<buffer *b-session:b1:clj*(7432)>)', but instead it was `(#<buffer *a-session:bb2:cljs*(8909)> #<buffer *a-session:bb1:clj*(4337)>)' which does not match because: (list-elt 0 (different-atoms #<buffer *a-session:bb2:cljs*(8909)> #<buffer *b-session:b2:cljs*(8228)>)).")
  signal(buttercup-failed "Expected `(cider-repls)' to be `equal' to `(#<buffer *b-session:b2:cljs*(8228)> #<buffer *b-session:b1:clj*(7432)>)', but instead it was `(#<buffer *a-session:bb2:cljs*(8909)> #<buffer *a-session:bb1:clj*(4337)>)' which does not match because: (list-elt 0 (different-atoms #<buffer *a-session:bb2:cljs*(8909)> #<buffer *b-session:b2:cljs*(8228)>)).")
FAILED: Expected `(cider-repls)' to be `equal' to `(#<buffer *b-session:b2:cljs*(8228)> #<buffer *b-session:b1:clj*(7432)>)', but instead it was `(#<buffer *a-session:bb2:cljs*(8909)> #<buffer *a-session:bb1:clj*(4337)>)' which does not match because: (list-elt 0 (different-atoms #<buffer *a-session:bb2:cljs*(8909)> #<buffer *b-session:b2:cljs*(8228)>)).
```

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#3039](https://github.com/clojure-emacs/cider/pull/3039): Allow starting the sideloader for the tooling session.
 * [#3041](https://github.com/clojure-emacs/cider/pull/3041): Sideloader: handle binary files, support multiple directories
 * [#3044](https://github.com/clojure-emacs/cider/pull/3044): Dynamically upgrade nREPL connection
+* [#3047](https://github.com/clojure-emacs/cider/pull/3047): Fix info/lookup fallback: response has an extra level
 
 ## 1.1.1 (2021-05-24)
 
--- a/cider-client.el
+++ b/cider-client.el
@@ -637,7 +637,7 @@ CONTEXT represents a completion context for compliment."
                     (cider-nrepl-send-sync-request (cider-current-repl)))))
     (if (member "lookup-error" (nrepl-dict-get var-info "status"))
         nil
-      var-info)))
+      (nrepl-dict-get var-info "info"))))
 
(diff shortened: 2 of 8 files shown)


bbatsov:
Looks good, but you will need to rebase.

plexus:
Rebased!

plexus:
You'll notice I also fixed some more linter warnings that I was getting locally but not on CI. Maybe cause I'm on a more recent Emacs build? (I tend to build from master, last time was end of July).

bbatsov:
Thanks!

PR #3050, opened 2021-09-14
[Fix #3045] Add customizable variable cider-xref-fn-depth

### LINKED ISSUE #3045: Unpredictable behaviour with cider-use-xref, if LSP is also used.
The new cider-use-xref functionality is great, but the behaviour is unpredictable when lsp is also being used, depending on which order the CIDER and LSP backends appear in `xref-backend-functions`.

## Expected behavior
Probably undefined. I'm not sure anything can be done about this, other than add some documentation about it.

## Actual behavior
Depending on which backend function appears first in `xref-backend-functions`, either CIDER or LSP will be used for xrefs. Sometimes the behaviour between the two is a little different, e.g. with `xref-find-references`, IIRC LSP will pick up references to baz in `(ns foo (:require [bar :refer [baz]]))`, whereas CIDER will not find these, but only the places where baz is called. (No big deal, although the CIDER behaviour is preferable, IMO).

- Use both CIDER and LSP together, with the default configuration (`cider-use-xref` and `lsp-enable-xref` set to true).
- Add `lsp` to `clojure-mode-hook`, open a Clojure file, and run `cider-jack-in-clj`.
- Use some xref command, e.g. `xref-find-references`, and depending on the order of `xref-backend-functions`, either CIDER or LSP will be used. On two difference machines, I have seen LSP being used one one, and CIDER on the other.

### CIDER version information

```
;; CIDER 1.2.0snapshot (package: 20210902.1412), nREPL 0.9.0-beta1
;; Clojure 1.10.1, Java 1.8.0_302
```

### Lein/Boot version

Lein 2.9.6

### Emacs version

27.2

### Operating system

Latest macOS and latest Ubuntu

deejayem:
This allows the depth to be specified when adding to
xref-backend-functions, giving control over whether
cider--xref-backend is added before or after other functions. This will
determine which function in the hook is tried first (i.e. which will
provide the xref functionality, e.g. to ensure is used instead of
lsp-mode).

This fixes https://github.com/clojure-emacs/cider/issues/3045 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#3044](https://github.com/clojure-emacs/cider/pull/3044): Dynamically upgrade nREPL connection
 * [#3047](https://github.com/clojure-emacs/cider/pull/3047): Fix info/lookup fallback: response has an extra level
 * [#2746](https://github.com/clojure-emacs/cider/issues/2746): Handle gracefully Clojure versions with non-standard qualifiers (e.g. `1.11.0-master-SNAPSHOT`).
+* [#3045] (https://github.com/clojure-emacs/cider/issues/3045): Add customizable variable cider-xref-fn-depth for specifying depth when adding to xref-backend-functions.
 
 ## 1.1.1 (2021-05-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New features
 
-* [#2831](https://github.com/clojure-emacs/cider/issues/2831): Add xref integration.
+* [#2831](https://github.com/clojure-emacs/cider/issues/2831): Add xref integration, configured with customizable variables cider-use-xref and cider-xref-fn-depth.
 * [#3017](https://github.com/clojure-emacs/cider/issues/3017): Annotate company completion kinds.
 * [#3040](https://github.com/clojure-emacs/cider/pull/3040): Support invoking `cider-clojuredocs` within the `*clojuredocs*` buffer.
 
@@ -18,7 +18,6 @@
 * [#3044](https://github.com/clojure-emacs/cider/pull/3044): Dynamically upgrade nREPL connection
 * [#3047](https://github.com/clojure-emacs/cider/pull/3047): Fix info/lookup fallback: response has an extra level
 * [#2746](https://github.com/clojure-emacs/cider/issues/2746): Handle gracefully Clojure versions with non-standard qualifiers (e.g. `1.11.0-master-SNAPSHOT`).
-* [#3045] (https://github.com/clojure-emacs/cider/issues/3045): Add customizable variable cider-xref-fn-depth for specifying depth when adding to xref-backend-functions.
 
 ## 1.1.1 (2021-05-24)
 
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -483,6 +483,14 @@ As it stands Emacs fires these events on <mouse-8> and <mouse-9> on 'x' and
   :group 'cider
   :version '(cider . "1.2.0"))
 
+(defcustom cider-xref-fn-depth nil
+  "The depth to use when adding the CIDER xref function to the relevant hook.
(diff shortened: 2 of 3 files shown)


cider-mode.el
```diff
@@ -483,6 +483,14 @@ As it stands Emacs fires these events on <mouse-8> and <mouse-9> on 'x' and
   :group 'cider
   :version '(cider . "1.2.0"))
 
+(defcustom cider-xref-fn-depth nil
```
bbatsov, line 486: Shouldn't we set this to `-100` by default?

doc/modules/ROOT/pages/usage/misc_features.adoc
```diff
@@ -352,3 +352,12 @@ Beginning with version 1.2.0, cider support Emacs built-in xref, which means `M-
 ----
 (setq cider-use-xref nil)
 ----
+
+If you use other minor modes (e.g. lsp-mode) that also integrate with xref, you may with to customize the precedence of CIDER, e.g. to use CIDER rather than LSP (when available), you can set the following:
```
bbatsov, line 356: I'm assuming that major modes may integrate with xref as well.

cider-mode.el
deejayem: It could be -90 (-100 and 100 are discouraged - https://www.gnu.org/software/emacs/manual/html_node/elisp/Setting-Hooks.html - I thought about documenting that, but didn't want to duplicate what's covered elsewhere too much).

I went for nil rather than -90, to leave it up to the user to decide, rather than making the choice for them.

I'm happy to change it though, if you'd prefer that.

doc/modules/ROOT/pages/usage/misc_features.adoc
deejayem: Yes. Maybe "other packages" would be a better way to word it. Otherwise I could just remove "minor".

cider-mode.el
bbatsov: My point was that we should probably resolve the ambiguity immediately, but I'd be fine with keep this as `nil`.

doc/modules/ROOT/pages/usage/misc_features.adoc
bbatsov: I like "other packages".

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#3044](https://github.com/clojure-emacs/cider/pull/3044): Dynamically upgrade nREPL connection
 * [#3047](https://github.com/clojure-emacs/cider/pull/3047): Fix info/lookup fallback: response has an extra level
 * [#2746](https://github.com/clojure-emacs/cider/issues/2746): Handle gracefully Clojure versions with non-standard qualifiers (e.g. `1.11.0-master-SNAPSHOT`).
+* [#3045] (https://github.com/clojure-emacs/cider/issues/3045): Add customizable variable cider-xref-fn-depth for specifying depth when adding to xref-backend-functions.
```
bbatsov, line 21: Technically speaking - `xref` integration hasn't been released. :-) You can just add to the other changelog entry that it's controlled by `cider-use-xref` and `cider-xref-fn-depth`.

doc/modules/ROOT/pages/usage/misc_features.adoc
```diff
@@ -352,3 +352,12 @@ Beginning with version 1.2.0, cider support Emacs built-in xref, which means `M-
...
+
+[source,lisp]
+----
+(setq cider-xref-fn-depth -90)
+----
+
+See the documentation for `add-hook` for more information about depth.
```
bbatsov, line 363: I guess you can just add a link to the relevant documentation here.

cider-mode.el
deejayem: I decided -90 is a good idea, as if it's nil the order ends up being different for different buffers, which is probably not what anyone wants.

bbatsov:
Thanks!

PR #3053, opened 2021-09-17
Doc: suggest in advance using stable releases

vemv:
It is plausible that people can cut to the chase when reading this doc and simply perform `M-x package-install <RET> cider <RET>`.

By pointing out the stability options in advance, it is more likely that they are considered.

Honestly I'd consider pusing a little harder for stability (as it means a happier user base and less support burden) but this PR would seem a smaller, agnostic improvement for the time being.

### DIFF
--- a/doc/modules/ROOT/pages/basics/installation.adoc
+++ b/doc/modules/ROOT/pages/basics/installation.adoc
@@ -29,6 +29,8 @@ maintained repos -
 http://stable.melpa.org[MELPA Stable]
 and http://melpa.org[MELPA].
 
+WARNING: Please read #using-stable-releases[using stable releases] before proceeding to installation.
+
 You can install CIDER with the following command:
 
 kbd:[M-x package-install <RET> cider <RET>]
@@ -45,6 +47,8 @@ or by adding this bit of Emacs Lisp code to your Emacs initialization file
 If the installation doesn't work try refreshing the package list:
 
 kbd:[M-x package-refresh-contents <RET>]
+
+== Using stable releases
 
 It's important to note that MELPA packages are built automatically
 from the `master` branch, and that means you'll be right on the


bbatsov:
The problem is that relatively few people use MELPA Stable, and if someone does - they'll probably get the stable version anyways. Once CIDER's on NonGNU ELPA the situation will change, but that's contingent on some pending changes to `parseedn` and `parseclj`, so I can submit them there as well. See https://github.com/clojure-emacs/parseclj/pull/29 and https://github.com/melpa/melpa/issues/6656

I'm all for promoting the "stable" releases more, but we need to do a bit of extra legwork.

vemv:
Thanks! Yes taking a look, it appears Melpa Stable is not all rainbows and unicorns. I was aware of some of it pitfalls however I discovered even more reading the thread you've linked to.

In face of it, maybe we can forward their desire to not promote Melpa Stable at all?

Maybe we could mention straight.el instead. I can't personally vouch for it, however I can definitely vouch for using sha-pinned git clones as a means of consumption. It's what I've done for approximately my latest 9 years of emacsing.

bbatsov:
> In face of it, maybe we can forward their desire to not promote Melpa Stable at all?

That's my plan in general - once CIDER's here https://elpa.nongnu.org/nongnu/ I'll just remove the mentions of MELPA Stable and focus on the new official repo (as a bonus it's enabled by default in Emacs 28). 

Maybe we could mention straight.el instead. I can't personally vouch for it, however I can definitely vouch for using sha-pinned git clones as a means of consumption. It's what I've done for approximately my latest 9 years of emacsing.

I've never used `straight.el`, so I can't say much on the topic. As `package.el` is the most popular/built-in package manager I tried to keep the focus on it.

vemv:
Cheers. Let's see how it goes.

Hope nongnu doesn't represent yet another rabbit hole / sticky situation in the package mgmt saga 🌀

bbatsov:
Time will tell. I hope I'll manage to get CIDER on NonGNU ELPA this week.

PR #3057, opened 2021-09-27
fix deprecated positional arguments warning in emacs 28

zhscn:
in emacs 28, there are some warning messages:
```
.emacs.d/straight/build/cider/cider-test.el: Warning: Use keywords rather than deprecated positional arguments to `define-minor-mode'
.emacs.d/straight/build/cider/cider-mode.el: Warning: Use keywords rather than deprecated positional arguments to `define-minor-mode'
.emacs.d/straight/build/cider/cider-debug.el: Warning: Use keywords rather than deprecated positional arguments to `define-minor-mode'
```
use keywords could fix these warning.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -340,7 +340,7 @@ of `cider-interactive-eval' in debug sessions."
   "Mode active during debug sessions.
 In order to work properly, this mode must be activated by
 `cider--turn-on-debug-mode'."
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #3062, opened 2021-10-02
Upgrade cider-nrepl

vemv:
Upgrades `cider-nrepl`, `Orchard` and `clj-suitable` for pulling their latest bugfixes.


bbatsov:
Thanks!

vemv:
🍻! Could we let me know as it hits MELPA? Not familiar with the release process

bbatsov:
MELPA rebuilds the packages automatically every few hours.

vemv:
Ah nice. What about the CHANGELOG? Should it be updated to say sth other than `master`?

bbatsov:
Nope, this is just a snapshot release built automatically from `master`. The 1.2 release will likely happen around the end of the month, provided nREPL 0.9 is finalized by then.

PR #3063, opened 2021-10-03
Upgrade cider-nrepl

vemv:
Brings in https://github.com/clojure-emacs/cider-nrepl/pull/718


vemv:
Merging so that MELPA picks this up asap

PR #3064, opened 2021-10-05
Soften the warning around cider-nrepl version mismatches

vemv:
Especially given we try to inflict as few breaking changes as possible (https://github.com/clojure-emacs/cider/discussions/2960), a version mismatch might simply mean that one is using a more recent cider-nrepl version (without one having bothered to customize the related `defcustom` to silence this).

This commit proposes a softer wording for the warning, so that people feel less alarmed in face of these.

Else one gives the quite untruthful impression that things are more fragile/coupled than they actually are.


bbatsov:
I like the current message more, as it's both short, informative and neutral. Once we get to the point of a really stable API we might not show any warning for newer cider-nREPL at all, but I don't think that we're quite there yet. At the very least we have to get to a 1.0 release soon.

It might be a good idea to add a way to suppress those messages for users who really know what they are doing. Still, my observation is that almost no one outside of the people hacking on CIDER changes those versions, so such a configuration option might be an overkill.

Btw, we never use terms like cider.el in the messages - it's always CIDER everywhere.

PR #3069, opened 2021-10-11
Don't call `fundamental-mode` if we already are in `fundamental-mode`

### LINKED ISSUE #3068: CIDER doesn't play well with `evil-set-cursor-color`
## Expected behavior

When configuring different cursor colors for evil states, the different states should have, well, different colors.

## Actual behavior

Typing anything in a CIDER-connected buffer will reset the cursor color.

1.
```
(setq evil-insert-state-cursor '(box "#f19618"
                                            ;; orange
                                            ))

(setq evil-normal-state-cursor '(box "#c7ceff"))
```

2. Open a CIDER buffer, jack in, go into the insert mode, observe the cursor color changing:
image

3. Move around, type sth, etc. The color will reset:
image

4. Disabling `eldoc-mode` solves the problem, and the color reset seems to happen when eldoc is triggered.

PS: I have tested this bug on both the GUI emacs and the TUI, using the following to enable the cursor colors for the TUI version:
```
       (require 'evil-terminal-cursor-changer)
       (evil-terminal-cursor-changer-activate)
```

### CIDER version information

```
nREPL server started on port 56118 on host localhost - nrepl://localhost:56118
;; Connected to nREPL server - nrepl://localhost:56118
;; CIDER 1.1.0snapshot, nREPL 0.8.2
;; Clojure 1.10.3, Java 15
;;     Docs: (doc function-name)
;;           (find-doc part-of-name)
;;   Source: (source function-name)
;;     Exit: <C-c C-q>
;;  Results: Stored in vars *1, *2, *3, an exception in *e;
;;  Startup: /usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.8.3"} me.raynes/fs {:mvn/version "1.4.6"} compliment {:mvn/version "0.3.11"} com.cemerick/pomegranate {:mvn/version "0.4.0"} org.tcrawley/dynapath {:mvn/version "0.2.5"} nrepl {:mvn/version "0.8.2"} refactor-nrepl/refactor-nrepl {:mvn/version "2.5.1"} cider/cider-nrepl {:mvn/version "0.25.9"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[refactor-nrepl.middleware/wrap-refactor,cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
```

### Emacs version

28.0.50

### Operating system

macOS 11.2.1

plexus:
If the `" *nrepl-decoding*"` buffer already exists then it is likely already in
the right mode. Not re-initializing the mode means we don't re-trigger hooks,
which in this case in particular prevents an unfortunate interaction with
evil-mode, where the cursor color gets updated even though it shouldn't.

Fixes #3068 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#3044](https://github.com/clojure-emacs/cider/pull/3044): Dynamically upgrade nREPL connection
 * [#3047](https://github.com/clojure-emacs/cider/pull/3047): Fix info/lookup fallback: response has an extra level
 * [#2746](https://github.com/clojure-emacs/cider/issues/2746): Handle gracefully Clojure versions with non-standard qualifiers (e.g. `1.11.0-master-SNAPSHOT`).
+* [#3069](https://github.com/clojure-emacs/cider/pull/3069): Fix cursor color changing when it shouldn't in evil-mode
 
 ## 1.1.1 (2021-05-24)
 
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -362,6 +362,11 @@ STACK is as in `nrepl--bdecode-1'.  Return a cons (INFO . STACK)."
             stack (cdr istack)))
     istack))
 
+(defun nrepl--ensure-fundamental-mode ()
+  "Enable `fundamental-mode' if it is not enabled already."
+  (when (not (eq 'fundamental-mode major-mode))
+    (fundamental-mode)))
+
 (defun nrepl-bdecode (string-q &optional response-q)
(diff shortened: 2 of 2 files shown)


vemv:
There's another call at https://github.com/clojure-emacs/cider/blob/69606faf05fffe8187ee30ec8c42c36553c1ec23/nrepl-client.el#L1359 - maybe we can extract a helper and apply it in both places?

plexus:
>  - maybe we can extract a helper and apply it in both places?

Done!

PR #3077, opened 2021-10-21
[Fix #3074] Makes cider recognize 'pwsh' as powershell.

### LINKED ISSUE #3074: Recognize both 'powershell' and 'pwsh' as powershell commands
## Expected behavior
`M-x cider-jack-in` base64-encodes a powershell command when I've set `cider-clojure-cli-command` to "pwsh" for powershell core. 

## Actual behavior
`M-x cider-jack-in` tries to call pwsh as if it's the clojure cli, without base64-encoding a call to clj, and it fails.

I had to edit cider.el  `cider--update-jack-in-cmd` to check for both "powershell" and "pwsh", after which it appears to be working.   I don't know if hardcoding both names is the kind of solution you'd want to merge, but realistically, powershell will be running under one of these two names in almost every case.

Install the powershell module under pwsh (powershell core) on Windows as described in the docs (at [https://github.com/clojure/tools.deps.alpha/wiki/clj-on-Windows](https://github.com/clojure/tools.deps.alpha/wiki/clj-on-Windows)).

Now install cider in emacs, and try to use it.  `cider-jack-in` tries to call `powershell`, where the module isn't installed, and fails.  So, set your `cider-clojure-cli-command` to "pwsh" to get past that.  At this point, cider doesn't realize it's dealing with powershell anymore, and issues a non-base64 command that fails.

So, to work around the issue, edit your cider.el `cider--update-jack-in-cmd`  to check for both "powershell" and "pswh" in the if-check.  Now it works.

### CIDER version information

```
;; CIDER 1.2.0snapshot (package: 20211013.2024), nREPL 0.9.0-beta3
;; Clojure 1.10.3, Java 17
```

### Emacs version

*27.2*

### Operating system

Windows 10

rwtodd:
The recent powershell executables (in the .NET Core era) are named pwsh, and should be treated the same as powershell for the purposes of cider-jack-in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * [#2746](https://github.com/clojure-emacs/cider/issues/2746): Handle gracefully Clojure versions with non-standard qualifiers (e.g. `1.11.0-master-SNAPSHOT`).
 * [#3069](https://github.com/clojure-emacs/cider/pull/3069): Fix cursor color changing when it shouldn't in evil-mode
 * [#3071](https://github.com/clojure-emacs/cider/issues/3071): Use xref instead of etags to push point to marker stack
+* [#3074](https://github.com/clojure-emacs/cider/issues/3074): Recognize pwsh as a powershell executable
 
 ## 1.1.1 (2021-05-24)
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #3083, opened 2021-10-31
Remove CircleCI MS-Windows-linter job, fix doc formatting

ikappaki:
Hi,

could you please consider the following changes.

The CircleCI MS-Windows linter check is not necessary, since there already exists a job to lint the codebase. It was blindly added when introducing the corresponding MS-Windows test job back at the time.

Also fixed a formatting issue with the doc hacking section.

All CircleCI tests are passing.

Thanks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,20 +81,6 @@ jobs:
       - setup
       - lint
 
-  test-windows-lint:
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #3084, opened 2021-11-01
Add before and after eval hooks

ericdallo:
This PR adds 2 new hooks:
- `cider-before-eval-hook` which runs **before** the request is sent to nrepl
- `cider-after-eval-done-hook` which runs **after** the nrepl response handle with status "done".

This is useful so users can listen to eval hooks and do customizations on UI, for example doom-modeline which show the REPL icon:
image

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -70,6 +70,18 @@ resulting value are used to compute completions."
   :group 'cider
   :package-version '(cider . "0.23.0"))
 
+(defcustom cider-before-eval-hook nil
+  "List of functions to call before eval request is sent to nrepl."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "1.2.0"))
+
+(defcustom cider-after-eval-done-hook nil
+  "List of functions to call after eval was responded by nrepl with done status."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "1.2.0"))
+
 (defun cider-spinner-start (buffer)
   "Start the evaluation spinner in BUFFER.
 Do nothing if `cider-show-eval-spinner' is nil."
@@ -78,21 +90,19 @@ Do nothing if `cider-show-eval-spinner' is nil."
       (spinner-start cider-eval-spinner-type nil
                      cider-eval-spinner-delay))))
 
-(defun cider-eval-spinner-handler (eval-buffer original-callback)
-  "Return a response handler to stop the spinner and call ORIGINAL-CALLBACK.
+(defun cider-eval-spinner (eval-buffer response)
+  "Handle RESPONSE stopping the spinner.
 EVAL-BUFFER is the buffer where the spinner was started."
-  (lambda (response)
-    ;; buffer still exists and
-    ;; we've got status "done" from nrepl
-    ;; stop the spinner
-    (when (and (buffer-live-p eval-buffer)
-               (let ((status (nrepl-dict-get response "status")))
-                 (or (member "done" status)
-                     (member "eval-error" status)
-                     (member "error" status))))
-      (with-current-buffer eval-buffer
-        (when spinner-current (spinner-stop))))
-    (funcall original-callback response)))
+  ;; buffer still exists and
+  ;; we've got status "done" from nrepl
+  ;; stop the spinner
+  (when (and (buffer-live-p eval-buffer)
+             (let ((status (nrepl-dict-get response "status")))
+               (or (member "done" status)
+                   (member "eval-error" status)
+                   (member "error" status))))
+    (with-current-buffer eval-buffer
+      (when spinner-current (spinner-stop)))))
 
 
 ;;; Evaluation helpers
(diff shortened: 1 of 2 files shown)


bbatsov:
If think that if you rebase on top of master the Windows failures will disappear.

cider-client.el
```diff
@@ -70,6 +70,18 @@ resulting value are used to compute completions."
...
   :package-version '(cider . "0.23.0"))
 
+(defcustom cider-before-eval-hook nil
+  "List of functions to call before evaluate."
+  :type 'hook
+  :group 'cider
+  :package-version '(cider . "1.2.0"))
+
+(defcustom cider-after-eval-hook nil
+  "List of functions to call after evaluate."
```
bbatsov, line 80: Might be a good idea to specify what do we mean as after evaluate here - e.g. is it linked to the receival of `done`, as there are several ways to interpret "after eval".

cider-client.el
```diff
@@ -78,21 +90,19 @@ Do nothing if `cider-show-eval-spinner' is nil."
       (spinner-start cider-eval-spinner-type nil
                      cider-eval-spinner-delay))))
 
-(defun cider-eval-spinner-handler (eval-buffer original-callback)
-  "Return a response handler to stop the spinner and call ORIGINAL-CALLBACK.
+(defun cider-eval-spinner (eval-buffer response)
```
bbatsov, line 93: What prompted you to rename the function and it's param here? I'm asking mostly because I think @Malabarba came up with those names (and with `spinner.el` as well). :-)

cider-client.el
```diff
@@ -198,10 +208,13 @@ define the position of INPUT in its buffer.  ADDITIONAL-PARAMS is a plist
...
   (let ((connection (or connection (cider-current-repl nil 'ensure))))
+    (run-hooks 'cider-before-eval-hook)
     (nrepl-request:eval input
-                        (if cider-show-eval-spinner
-                            (cider-eval-spinner-handler connection callback)
-                          callback)
+                        (lambda (response)
+                          (when cider-show-eval-spinner
+                            (cider-eval-spinner connection response))
+                          (run-hooks 'cider-after-eval-hook)
```
bbatsov, line 216: Shouldn't you check here for what's the actual message on which this is being called? I don't remember very well if that branch was called only in case of "done"/"error" - if that's the case fine, if not we should be careful not to trigger the hook multiple times.

bbatsov:
The changes seem reasonable. Just make sure to mention the new hooks in the eval hooks documentation https://docs.cider.mx/cider/usage/code_evaluation.html#evaluation-hooks.

cider-client.el
bbatsov: My bad, I see you moved the callback lambda to a different place in the code. Now this names sense.

vemv: Maybe I'm having a dense morning but I do see `cider-eval-spinner-handler` as removed, which would be breaking change

bbatsov: It got renamed, as it's not a handler anymore, but rather a regular function. Technically speaking that's a breaking change, but in practice I'm fairly certain that no one used this handler outside of CIDER's code, so I guess the change is acceptable.

bbatsov: All the usages I see are just people who committed their installed packages https://github.com/search?q=cider-eval-spinner-handler&type=code

ericdallo: Yeah, I saw the same and thought nobody was using this, actually I think this function should be private 😅

cider-client.el
ericdallo: Yes, I just realized it will be called multiple times, because of that I will rename the after hook to `cider-after-eval-done-hook` and only fire it when response status is "done", since it seems to always be available even when before error response was received

ericdallo:
Done @bbatsov, I update PR description as well, LMK what you think

cider-client.el
vemv: Personally IDK, this is a discussion that comes up time and again https://github.com/clojure-emacs/cider/discussions/2960

It would be greatly simplifying to commit to zero breaking changes, period. Searches over github cannot be comprehensive (there are private repos, gitlab, etc).

It takes 1 instance of breakage for a developer to have their trust in us broken. I'd like to see CIDER as something (mainly) targeted for professional developers, and consequently we don't possibly break people's work tools.

Else there's a slippery slope of applying similar reasoning to other defuns in future PRs. So in the end one isn't keeping a real no-breakage policy.

bbatsov: I'm aiming for a more practical policy, simply because I know that there are few extensions outside of those that we maintain ourselves and only a handful of common extension points that people use in their configs. 

I definitely don't want to frustrate people, but for most of the project's life there were a ton of massive breaking changes in every release and still we got where we got. Right now we can keep this handler as some dead code, or add more unrelated logic to it. I don't see how being dogmatic about a relatively safe change helps us in any way. If this was a function that's likely to be used, than my perspective would have been different. 

Btw, snapshot releases are great for uncovering unexpected breakages.

vemv: Alright no biggie, we can keep an eye for more instances in a future and keep/adjust the policy accordingly

bbatsov:
@ericdallo Looks good to me. Thanks for tackling this!

cider-client.el
plexus: As someone who's dealt with a lot of frustration because of CIDER breakage and still feels a little sore about it I will second @vemv's concerns and general sentiment. I also agree with @bbatsov that as far as breaking changes go this one seems fairly safe.

Some actionable things we could consider:

- make more use of "private" function names, e.g. in this case since we're already making an API change to what we consider to be an internal function, it could have been renamed to `cider--eval-spinner`
- be more diligent about marking API changes (even minor ones) as BREAKING in the changelog, to raise awareness, but also so that if someone suddenly starts getting a `(void-function cider-eval-spinner-handler)`, they can search for that in the CHANGELOG and see what's up

bbatsov: Fair point. I try to be pragmatic simply because CIDER is full of horrible, but low-impact public APIs, that would be very hard to undo if we decide to really go by the book about every potentially breaking change. I was thinking that at some point it'd be nice to write some guide for extensions and highlight there explicitly the part of the API that's meant to be used by third-parties and is considered the official API. Probably that's a tiny fraction (e.g. 10%) of everything that's public right now. 

I also remind everyone there are no truly private definitions in Emacs, other than a naming convention. You can mark something with `cider--`, but people can still use it and get breakages. But at least they will know what they signed up for. :-)

PR #3087, opened 2021-11-01
Detect host and port number from babashka nREPL conn msg

### LINKED ISSUE #3086: Can't jack-in to babashka
## Expected behavior
Open a client repl buffer to the babashka nREPL server.

## Actual behavior
The babashka nREPL-server is started but CIDER fails to connect displaying a message

`[nREPL] Starting server via /usr/local/bin/bb nrepl-server`

Open any `bb.edn` file and `M-x cider-jack-in-clj`

### CIDER version information

```
;; CIDER 1.2.0snapshot (package: 20211028.1231), nREPL 0.9.0-beta3
;; Clojure 1.10.3, Java 1.8.0_131
```

Issue is that CIDER can't detect babashka's nREPL port number from its opening message.

patch to follow.

ikappaki:
Hi,

could you please consider fix for jacking-in into the babashka prompt. 

There appears to be an issue with the current code, that the nrepl client can't parse the port of host number from the babashka nREPL opening message ([as defined in babashka nrepl code](https://github.com/babashka/babashka.nrepl/blob/fa3a55ae58f7bcbcfcfaf10fab98b412220f733f/src/babashka/nrepl/server.clj#L34)) , example:

```
Started nREPL server at 127.0.0.1:1667
For more info visit: https://book.babashka.org/#_nrepl
```

which is a differnet format from the standard clojure nREPL opening message, example:

```
nREPL server started on port 57721 on host localhost - nrepl://localhost:57721
```

The patch combines the two into a single regexp using a shy group:
```emacs-lisp
"\\(?:nREPL server started on port \\(?1:[0-9]+\\)\\|Started nREPL server at \\(?2:.*?\\):\\(?1:[0-9]+\\)\\)"
```

I can further refactor it if required.

Fixes #3086.

Thanks,

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1043,6 +1043,20 @@ been determined."
                (propertize cmd 'face 'font-lock-keyword-face))
       serv-proc)))
 
+(defconst nrepl-listening-address-regexp
+  (rx (or
+       ;; standard
+       (and "nREPL server started on port " (group-n 1 (+ (any "0-9"))))
+       ;; babashka
+       (and "Started nREPL server at "
+            (group-n 2 (+? any)) ":" (group-n 1 (+ (any "0-9"))))))
+  "A regexp to search an nREPL's stdout for the address it is listening on.
+
+If it matches, the address components can be extracted using the following
+match groups:
+1  for the port, and
+2  for the host (babashka only).")
+
 (defun nrepl-server-filter (process output)
   "Process nREPL server output from PROCESS contained in OUTPUT."
   ;; In Windows this can be false:
@@ -1062,8 +1076,7 @@ been determined."
               (set-window-point win (point)))))
         ;; detect the port the server is listening on from its output
         (when (and (null nrepl-endpoint)
-                   (string-match "\\(?:nREPL server started on port \\(?1:[0-9]+\\)\\|Started nREPL server at \\(?2:.*?\\):\\(?1:[0-9]+\\)\\)"
-                                 output))
+                   (string-match nrepl-listening-address-regexp output))
           (let ((host (or (match-string 2 output)
                           (file-remote-p default-directory 'host)
                           "localhost"))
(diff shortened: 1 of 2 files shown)


nrepl-client.el
```diff
@@ -1062,10 +1062,13 @@ been determined."
               (set-window-point win (point)))))
         ;; detect the port the server is listening on from its output
         (when (and (null nrepl-endpoint)
-                   (string-match "nREPL server started on port \\([0-9]+\\)" output))
-          (let ((port (string-to-number (match-string 1 output))))
-            (setq nrepl-endpoint (list :host (or (file-remote-p default-directory 'host)
-                                                 "localhost")
+                   (string-match "\\(?:nREPL server started on port \\(?1:[0-9]+\\)\\|Started nREPL server at \\(?2:.*?\\):\\(?1:[0-9]+\\)\\)"
```
vemv, line 1065: It would sound sensible to extract this bit to a defun that would be unit-tested (for both the traditional and bb formats)

bbatsov:
I've noticed the problem, but I wasn't sure if @borkdude wasn't going to just align the format to the standard one. I guess we still have to support their old format, though, so the patch is most welcome.

nrepl-client.el
bbatsov: Yeah, that'd be nice indeed.

ikappaki: sure, will do. Currently there's only indirect test coverage for the existing case.

borkdude:
@bbatsov I think changing the current message would break tooling that people already wrote and also creates a divide between older bb versions and newer bb versions, so I'd rather just let it be this way. For newer SCI-based nREPLs I'll go with the "lein" format, nbb has already done this.

vemv (approved):
Very neat! TIL about `rx` :)

nrepl-client.el
ikappaki: Done. Moved regexp to a const, and added some tests for the three types I could think of
1. the standard output (lein, deps.edn).
2. babashka.
3. shadow-cljs.

ikappaki:
> 
> 
> Very neat! TIL about `rx` :) 
`rx` :fire:

bbatsov:
Thanks!

Issue #3032, opened 2021-08-21
Add support for UNIX socket connections


bbatsov:
Support for UNIX socket connections was recently added to nREPL 0.9 (currently in beta). We should make it possible to connect to UNIX sockets from CIDER.  This will require some changes to the connection functions (mostly in `nrepl-client.el`). We can either expose a `socket` param there directly or recycle `host` if it looks like a file name.

bbatsov:
@rlbdv, btw is this something you plan to tackle? I was just reminded you're an Emacs user. :D

bendlas:
I did a proof of concept, that let me successfully connect to a unix domain socket: https://github.com/bendlas/cider/commit/03675162be3d30624563718cd4030d035c1ac780

```sh
clj -R:nrepl-latest -m nrepl.cmdline --socket /tmp/clj.sock
```

```el
(cider-connect-unix (list :socket-file "/tmp/clj.sock"))
```

Not sure about how to best integrate this into the menu structure of cider though ...

bbatsov:
Cool!

I think the bigger question is how to integrate this in the existing connect command, so we don't have to add new commands and create more confusion with them. Perhaps the current command can just prompt for socket as well as host and port and do something differently if there's a socket? I'll have to think a bit about this, but if anyone has great ideas - please, share them. :-)

//cc @vspinu

bbatsov:
I left a bit more feedback on your commit.

bendlas:
I'd be more than happy with a version, that only implements `cider-connect` support.

My use case is about developing in a VPN, and were it not for the REPL port, I could run my dev server in a completely isolated (linux) networking ns. Thus, cider-jack-in and other convenience isn't applicable anyway.

I suspect that many other use cases for domain- instead network-sockets will be similarly advanced.

rlbdv:
Fair enough, and thanks for working on it.

bbatsov:
Okay, guess we're on the same page, though. @bendlas Feel free to submit a PR when you're happy with your draft.

bendlas:
Sure, I could use some help with the design for `cider-connect`, though.

nrepl doesn't seem to write `.nrepl-port` for socket servers, so auto-finding the socket seems out of scope for this for now.

How do you feel about this design:

- Instead of `Host:`, `cider-connect` prompts for `TCP Host or Protocol:`, and allow to enter `nrepl+unix:` (with auto-completion).
[...]

PR #3088, opened 2021-11-02
Unix domain socket connection

bendlas:
Preview for https://github.com/clojure-emacs/cider/issues/3032

- [x] nrepl-client adaptions to connect via socket file
- [x] rudimentary menu integration, by overloading on a special hostname
- [x] select socket file with auto completion
- [ ] update menu structure to not overload host, instead take explicit protocol with `C-u` prefix arg

Please note that as per https://github.com/clojure-emacs/cider/issues/3032#issuecomment-958133135, this is not yet considered ready.

-----------------

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -492,18 +492,20 @@ and kill the process buffer."
 ;;; Network
 
 (defun nrepl--unix-connect (socket-file &optional no-error)
+  "If SOCKET-FILE is given, try to `make-network-process'.
+If NO-ERROR is non-nil, show messages instead of throwing an error."
   (if (not socket-file)
       (unless no-error
         (error "[nREPL] Socket file not provided"))
-    (message "[nREPL] Establishing unix connection to %s ..." socket-file)
+    (message "[nREPL] Establishing unix socket connection to %s ..." socket-file)
     (condition-case nil
         (prog1 (list :proc (make-network-process :name "nrepl-connection" :buffer nil
-			                         :family 'local :service socket-file)
-                     :host "unix:"
+                                                 :family 'local :service socket-file)
+                     :host "local-unix-domain-socket"
                      :port socket-file
                      :socket-file socket-file)
-          (message "[nREPL] Direct connection to %s established" socket-file))
-      (error (let ((msg (format "[nREPL] Direct connection to %s failed" socket-file)))
+          (message "[nREPL] Unix socket connection to %s established" socket-file))
+      (error (let ((msg (format "[nREPL] Unix socket connection to %s failed" socket-file)))
                (if no-error
                    (message msg)
                  (error msg))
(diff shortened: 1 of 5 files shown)


nrepl-client.el
```diff
@@ -483,6 +483,24 @@ and kill the process buffer."
...
+      (unless no-error
+        (error "[nREPL] Socket file not provided"))
+    (message "[nREPL] Establishing unix connection to %s ..." socket-file)
+    (condition-case nil
+        (prog1 (list :proc (make-network-process :name "nrepl-connection" :buffer nil
+			                         :family 'local :service socket-file)
+                     :host "unix:"
+                     :port socket-file
+                     :socket-file socket-file)
+          (message "[nREPL] Direct connection to %s established" socket-file))
```
bbatsov, line 497: Perhaps we can say "Unix socket connection" here?

nrepl-client.el
```diff
@@ -621,14 +639,16 @@ Do not kill the server if there is a REPL connected to that server."
                            (buffer-list)))
         (nrepl-kill-server-buffer server-buf)))))
 
-(defun nrepl-start-client-process (&optional host port server-proc buffer-builder)
+(defun nrepl-start-client-process (&optional host port socket-file server-proc buffer-builder)
```
bbatsov, line 642: You'll have to mention the new param somewhere in the docstring, otherwise checkdoc won't be happy.

nrepl-client.el
```diff
@@ -621,14 +641,16 @@ Do not kill the server if there is a REPL connected to that server."
...
 
-(defun nrepl-start-client-process (&optional host port server-proc buffer-builder)
-  "Create new client process identified by HOST and PORT.
-In remote buffers, HOST and PORT are taken from the current tramp
-connection.  SERVER-PROC must be a running nREPL server process within
-Emacs.  BUFFER-BUILDER is a function of one argument (endpoint returned by
-`nrepl-connect') which returns a client buffer.  Return the newly created
-client process."
-  (let* ((endpoint (nrepl-connect host port))
+(defun nrepl-start-client-process (&optional host port socket-file server-proc buffer-builder)
```
bbatsov, line 644: One more thing - we'll likely have to move `socket-file` to the end of the function, even if this will make for a very weird API invocation. (I googled for uses of the function and I'm fairly certain this will break some clients, which we want to avoid) Or deprecate this function and replace it with a newer version that uses keyword params instead of named params.

bbatsov:
Btw, you also have to rebase on top of `master`. I believe this will address the CI failures.

bbatsov:
We'll also need to add a bit of documentation for starting a server that listens on a socket and connecting to it here https://docs.cider.mx/cider/1.1/basics/up_and_running.html#connect-to-a-running-nrepl-server Nothing fancy - a couple of paragraphs with code snippets will be enough.

bendlas:
Rebased and documented.

For now, I documented the current state for consistency. I don't think using this `local-unix-domain-socket` as-is would be a good idea. I'll change docu when I change impl.

bbatsov:
We can just add a note to the docs that this feature is experimental and subject to changes, that won't be considered breaking. That's how I typically approach the issue when I want to ship something sooner, but I know that its implementation is not ideal and will likely see some changes down the road.

bbatsov:
@bendlas Are things working properly with the latest nREPL beta?

bendlas:
Right, so I've been daily driving `-beta5` with jdk11

I just gave it a try with jdk17 and everything seems to be in order.

I tried to take another timeboxed stab at integrating a protocol chooser in order to get rid of the magic hostname, but couldn't complete it. I have a rough Idea, but no capacity right now, and would be happy to hand that part off.

bbatsov:
Okay, I'll merge this as it is and I'll add a note that the current behavior is subject to changes. Thanks for the getting it to this point!

PR #3092, opened 2021-11-23
Update *print-meta* bug comment in usage/pretty_printing

linkenneth:
https://clojure.atlassian.net/browse/CLJ-1445 is fixed as of Clojure 1.10.2, so this is no longer an active bug for the latest versions of Clojure. I added a note about versioning here, but we can also choose to remove this note entirely.

Verification:
```clj
user> (set! *print-meta* true)
;; => true
user> 
user> (with-meta [1 2 3] {:foo "bar"})
;; => ^{:foo "bar"} [1 2 3]
user> (def xs (with-meta [1 2 3] {:foo "bar"}))
;; => ^{:line 36, :column 7, :file "*cider-repl ml/tool-server:localhost:51070(clj)*", :name xs, :ns #namespace[user]} #'user/xs
user> xs
;; => ^{:foo "bar"} [1 2 3] ;; notice that meta is printed
```

### DIFF
--- a/doc/modules/ROOT/pages/usage/pretty_printing.adoc
+++ b/doc/modules/ROOT/pages/usage/pretty_printing.adoc
@@ -49,9 +49,10 @@ Here's one example:
   (apply pp/write value (mapcat identity (assoc options :stream writer))))
 ----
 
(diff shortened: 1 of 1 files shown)


linkenneth:
Test failure seems to be due to MELPA being unreachable; unrelated to doc changes.

bbatsov:
Thanks!

PR #3093, opened 2021-11-24
Convert "See Also" text to buttons in *cider-clojuredocs* buffer

tongjie-chen:
This "See Also" part in *cider-clojuredocs* buffer becomes clickable

### DIFF
--- a/cider-clojuredocs.el
+++ b/cider-clojuredocs.el
@@ -125,7 +125,7 @@ opposite of what that option dictates."
                               'sym see-also
                               'action (lambda (btn)
                                         (cider-clojuredocs-lookup (button-get btn 'sym)))
-                              'help-echo (format "Middle click to jump to %s" see-also)))
+                              'help-echo (format "Press Enter or middle click to jump to %s" see-also)))
       (insert "Not available\n"))
     (insert "\n== Examples\n\n")
     (if-let ((examples (nrepl-dict-get dict "examples")))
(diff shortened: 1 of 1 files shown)


cider-clojuredocs.el
```diff
@@ -121,7 +121,11 @@ opposite of what that option dictates."
     (insert "\n== See Also\n\n")
     (if-let ((see-alsos (nrepl-dict-get dict "see-alsos")))
         (dolist (see-also see-alsos)
-          (insert (format "* %s\n" see-also)))
+          (insert-text-button (format "* %s\n" see-also)
```
bbatsov, line 124: The indentation seems broken here.

bbatsov:
I guess the CI is failing because of the broken indentation. Don't forget to also mention this in the changelog.

cider-clojuredocs.el
```diff
@@ -140,7 +144,8 @@ opposite of what that option dictates."
 (defun cider-clojuredocs-lookup (sym)
   "Look up the ClojureDocs documentation for SYM."
   (let ((docs (cider-sync-request:clojuredocs-lookup (cider-current-ns) sym)))
-    (pop-to-buffer (cider-create-clojuredocs-buffer (cider-clojuredocs--content docs)))))
+    (pop-to-buffer (cider-create-clojuredocs-buffer (cider-clojuredocs--content docs)))
+    (highlight-regexp (car (last (split-string sym "/"))) 'bold)))
```
bbatsov, line 148: This seems unrelated to the button to me. Also if you know that `split-string` returns just two elements you can use directly `cadr` or `nth`.

Issue #3106, opened 2021-12-21
Use lein without bootclasspath optimization


vemv:
## Context

cider uses enrich-classpath, which uses fipp, which requires `clojure.instant`, which can't be loaded under the bootclasspath approach.

This is reflected here: https://github.com/brandonbloom/fipp/issues/60#issuecomment-470468952

## Proposal

Always run lein prefixed by `LEIN_USE_BOOTCLASSPATH=no`, which I have verified that improves things.

(A full fix for the issues expressed here https://clojurians.slack.com/archives/C0617A8PQ/p1640101913375900 would also have to bump rrb-vector, see https://github.com/brandonbloom/fipp/pull/79)

bbatsov:
What do you need `fipp` for? Perhaps we can just replace it with something else that doesn't have deps?

vemv:
fipp shines in that it's fast and scalable, whereas `pprint`ing huge values can easily cause issues.

Issue #2666, opened 2019-07-01
Allow setting environment variables for cider-jack-in


xificurC:
**Is your feature request related to a problem? Please describe.**

Boot honors some environment variables for configuring the java executable that is about to be started. There's a [longstanding issue](https://github.com/boot-clj/boot/issues/433) where these variables cannot be set from `boot.properties`. There's also no way to give environment variables to the command that `cider-jack-in` is about to start.

**Describe the solution you'd like**

Here's a quick hack that works. Normally the `defvar` would be a `defcustom`. The changes in `cider--update-jack-in-cmd` are minimal, just another local variable that isn't even really necessary and an `if` checking if our variable isn't `nil`. I'm not sure how much `env` as a command is portable or if the command could be decorated differently, but this is a working solution for Linux users.

If you're interested I wanted to pass OmitStackTraceInFastThrow to the java command when debugging [this issue](https://github.com/clojure-emacs/cider-nrepl/issues/621).

```clojure
(defvar cider-jack-in-env-vars nil)
(defun cider--env-vars-to-string (alist)
  (string-join (mapcar (lambda (x)
                         (format "%s=%s"
                                 (shell-quote-argument (car x))
                                 (shell-quote-argument (cdr x))))
                       alist)
               " "))

(defun cider--update-jack-in-cmd (params)
  "Update :jack-in-cmd key in PARAMS."
  (let* ((params (cider--update-do-prompt params))
         (project-dir (plist-get params :project-dir))
         (project-type (cider-project-type project-dir))
         (command (cider-jack-in-command project-type))
         (command-resolved (cider-jack-in-resolve-command project-type))
         (command-global-opts (cider-jack-in-global-options project-type))
         (command-params (cider-jack-in-params project-type))
         (env-vars cider-jack-in-env-vars))
    (if command-resolved
        (with-current-buffer (or (plist-get params :--context-buffer)
                                 (current-buffer))
          (let* ((command-params (if (plist-get params :do-prompt)
                                     (read-string (format "nREPL server command: %s " command-params)
                                                  command-params)
                                   command-params))
                 (cmd-params (if cider-inject-dependencies-at-jack-in
                                 (cider-inject-jack-in-dependencies command-global-opts command-params project-type)
                               command-params)))
            (if (or project-dir cider-allow-jack-in-without-project)
                (when (or project-dir
                          (eq cider-allow-jack-in-without-project t)
                          (and (null project-dir)
[...]

PR #3107, opened 2021-12-21
Disable Lein bootclasspath optimization by default

vemv:
* Disable Lein bootclasspath optimization by default
  * Fixes https://github.com/clojure-emacs/cider/issues/3106
* Bump enrich-classpath
  * Delivers https://github.com/clojure-emacs/enrich-classpath/commit/3f5d4fc7350911740ff42b195cefca905c43102f

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -521,7 +521,7 @@ where it doesn't make sense."
   (let* ((corpus (if (and cider-enrich-classpath
                           (eq project-type 'lein))
                      (append cider-jack-in-lein-plugins
-                             '(("mx.cider/enrich-classpath" "1.5.1")))
+                             '(("mx.cider/enrich-classpath" "1.5.2")))
                    cider-jack-in-lein-plugins)))
     (thread-last corpus
       (seq-filter
--- a/cider.el
+++ b/cider.el
@@ -101,7 +101,9 @@ version from the CIDER package or library.")
   "Codename used to denote stable releases.")
 
 (defcustom cider-lein-command
-  "lein"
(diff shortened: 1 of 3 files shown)


cider.el
```diff
@@ -101,7 +101,9 @@ version from the CIDER package or library.")
   "Codename used to denote stable releases.")
 
 (defcustom cider-lein-command
-  "lein"
+  ;; bootclasspath is an optimization that makes the `clojure.instant' ns unreachable,
+  ;; and can be problematic for Lein plugins (like enrich-classpath, or any user-specified plugins)
+  "LEIN_USE_BOOTCLASSPATH=no lein"
```
vemv, line 106: Not really sure if this works as intended, I mean will `VAR=42` syntax work?

As for the underlying fix, I did verify that it does work over https://github.com/dpassen/cider-doo-repro

bbatsov: This will break the command discovery, as somewhere in the code we check if `cider-lein-command` is on the PATH. We can always have generic way to prepend ENV variables to commands.

vemv:
I'll see what I can do, maybe the issue can be solved some other way.

bbatsov:
Btw, here's also an old patch showing how simple it'd be to prepend some ENV vars to the jack in command https://github.com/clojure-emacs/cider/issues/2666

Given this doesn't add much complexity we might consider doing it down the road. I know you have another solution in mind and it seems like a better solution, but I'm adding this comment here just to remind myself to look into better support for env vars.

PR #3111, opened 2021-12-22
Enable enrich-classpath middleware for tests that expect it.

jellelicht:
Some tests fail currently because of the addition of the `cider-enrich-classpath` setting. These tests assume that the enriching of classpaths takes place, which it currently does not by default.

If this fix is not acceptable, I can simply edit the expected test output instead.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Thanks! We had to disable this feature last minute (quite literally) before the 1.2 release, so I ignored the failed test with the intention to do something about it today, but you beat me to it. :-)

Issue #3113, opened 2021-12-28
wrong-type-argument when jacking into a REPL


dpassen:
## Expected behavior

cider-jack-in-clj spawns a new REPL

## Actual behavior

REPL starts up but CIDER fails to connect due to a wrong-type-argument

## Steps to reproduce the problem

M-x cider-jack-in-clj or C-c C-x j j 

*This is extremely important! Providing us with a reliable way to reproduce
a problem will expedite its solution.*

### CIDER version information

REPL fails to launch so:
```
Package cider is installed.

     Status: Installed in ‘cider-20211228.1132/’ (unsigned). Delete
    Version: 20211228.1132
     Commit: 42958551cf306bb3579e8669e8c2da922331bdba
```

### Emacs version

GNU Emacs 27.2 (build 1, x86_64-apple-darwin20.6.0, Carbon Version 164 AppKit 2022.6) of 2021-11-16

### Operating system

macOS Monterey 12.1

## Debug Log

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("-snapshot" nil nil)
  cider--version()
  cider-repl--clojure-banner()
  cider-repl--banner()
  cider-repl--insert-banner()
  cider-repl-init(#<buffer *cider-repl apps/bowling-scorer:localhost:53565(clj)*> #f(compiled-function () #<bytecode 0x1ffcb88388d9>))
  cider--connected-handler()
  run-hooks(nrepl-connected-hook)
  nrepl-start-client-process("localhost" 53565 #<process nrepl-server> #f(compiled-function (_) #<bytecode 0x1ffca9e020f5>) nil)
  cider-nrepl-connect((:project-dir "~/apps/bowling-scorer/" :jack-in-cmd "/usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl..." :host "localhost" :port 53565 :project-dir "~/apps/bowling-scorer/" :server #<process nrepl-server> :server-command "/usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl..." :repl-init-function nil :repl-type clj :session-name nil))
  cider-connect-sibling-clj((:project-dir "~/apps/bowling-scorer/" :jack-in-cmd "/usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl...") #<buffer *nrepl-server apps/bowling-scorer:localhost*>)
[...]

PR #3115, opened 2021-12-28
Use .el instead of .elc file for lm-version

Whoops:
Digging into #3113, when requiring cider on my machine `load-file-name` points to `cider.elc` rather than `cider.el`. It seems `lm-version` can't extract the version from the `.elc` file. I couldn't find a better way of getting at the `.el` file, so just replacing `.elc` with `.el` *seems* to work. I don't have a strong background in Emacs programming so I'm not sure if I've done evil or not ;-).

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -95,7 +95,10 @@
 (require 'package)
 
 (defconst cider-version
-  (lm-version (or load-file-name buffer-file-name))
+  (lm-version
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks for the PR, but it's even more complex, so I just hardcoded the version for now.

Issue #3118, opened 2021-12-30
`package-get-version` might not always return the cider version.


ikappaki:
## Expected behavior

`package-get-version` was recently used in `CIDER` to simplify the logic to retrieve the current CIDER version.

It should return the CIDER package version as set in the `cider.el` package header.

## Actual behavior

The fn uses various heuristics to locate the the package file (in our case `cider.el`) where the package version is located at, and it fails if the parent directory name is not named after the package file, i.e. `cider'.

This for example, causes the Eldev compilation test to issue a warning in `cider-util.el` causing the corresponding CircleCI job [to fail](https://circleci.com/api/v1.1/project/github/clojure-emacs/cider/4919/output/105/0?file=true&allocation-id=61cc63d35cb92944382f8d41-0-build%2F5A9B3B3D):

```
[00:00.929]  
             In toplevel form:
             cider-util.el:377:8:Error: value returned from (fboundp 'package-get-version) is unused
[00:00.930]  Saving target dependencies to file `.eldev/27.2/target-dependencies.build'...
```

This is because the the call to `(package-get-version)` is expanded at compile time to nil (`package-get-version` is marked as `pure` and as such its value is resolved at compile time. CircleCi checks out the code at `/root/project`. Because `project` != `cider` the fn returns nil since it looks for a `project.el` file to extract the version number from), thus the `if` statement returns nil in both true and false cases causing the above warning:
```emacs-lisp
(defun cider--version ()
  "Retrieve CIDER's version.
A codename is added to stable versions."
  (if (string-match-p "-snapshot" cider-version)
      (let ((pkg-version (if (fboundp 'package-get-version)
                             (package-get-version)
                           nil)))
        (if pkg-version
            ;; snapshot versions include the MELPA package version
            (format "%s (package: %s)" cider-version pkg-version)
          cider-version))
    ;; stable versions include the codename of the release
    (format "%s (%s)" cider-version cider-codename)))
```

See https://lists.gnu.org/archive/html/emacs-devel/2021-12/msg02963.html for a discussion.

## Steps to reproduce the problem

1. Checkout CIDER to a directory other than `cider`, e.g. `git clone https://github.com/clojure-emacs/cider.git cider-other`
2. checkout a commit which has the above code, e.g. `git checkout 8a8ceb63f8549a9e4be05e7c649b96847a3ac5dc`
3. byte compile `cider-util.el`: `eldev compile -f cider-util.el --force`

You should get the compilation warning mentioned earlier.

[...]

PR #3119, opened 2021-12-30
Set project directory in CircleCI to `cider`

ikappaki:
Hi,

a simple hack to work around the unused value compilation warning as described in #3118.

Basically, it sets the directory where the code is checked out to `cider`, so that the `package-get-version` can figure out the filename where to look for the cider package version (i.e. `cider.el`).

`package-get-version` does not look very mature at the moment, given the unexpected issue that is caused with the compilation warning (and not so easy to track), and it appears there still some work to be done as per this discussion https://lists.gnu.org/archive/html/emacs-devel/2021-12/msg02963.html. Would it be a better idea not to use it yet perhaps than trying to patch its shortcomings?

Linter tests now pass on Ubuntu.

----------------

- [ x] You've added tests (if possible) to cover your change(s)
- [x ] All tests are passing (`eldev test`)
- [x ] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes
- [ x] You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,6 +77,17 @@ jobs:
       - test
 
   test-lint:
+    # There's issue with `package-get-version' that could fail to
+    # return the cider package version, if the parent directory where
+    # it is called from, is not named `cider', i.e. it does not have
+    # the same name as the file wherein the package version is defined
+    # at. In our case, the package version is defined in `cider.el',
+    # thus the parent directory of the checkout should be named
+    # `cider'. Otherwise it could cause compilation warnings, thus we
+    # fix the directory name here. See
+    # https://lists.gnu.org/archive/html/emacs-devel/2021-12/msg02963.html
+    # for a discussion.
+    working_directory: ~/cider
     docker:
       - image: silex/emacs:27-dev
     steps:


.circleci/config.yml
```diff
@@ -77,6 +77,17 @@ jobs:
       - test
 
   test-lint:
+    # There's issue with `package-get-version' that could fail to
+    # return the cider package version, if the parent directory where
+    # it is called from, is not named `cider', i.e. it does not have
+    # the same name as the file wherein the package version is defined
+    # at. In our case, the package version is defined in `cider.el',
+    # thus the parent directory of the checkout should be named
+    # `cider'. Otherwise it could cause compilation warnings, thus we
```
bbatsov, line 86: Or `cider-some-version-number`. :-)

bbatsov:
Hmm, for some reason I see only the GitHub Actions check for the PR:

image

bbatsov:
> Would it be a better idea not to use it yet perhaps than trying to patch its shortcomings?

Yeah, I think so - probably it'd be to copy this to `cider-util.el` and adjust it there.

ikappaki:
> Hmm, for some reason I see only the GitHub Actions check for the PR:
> 
> image

Indeed, not sure what is happening here, a CircleCI permissioning issue? I get the same when I view it inside the PR, though it looks fine when I click on the branch @ https://github.com/ikappaki/cider/tree/issue/compile-unused

image

bbatsov:
No idea myself. Probably something's wrong with the CircleCI configuration, but I haven't changed it it a while.

bbatsov:
I opted to just wrap and stub the problematic function in the tests. Eventually we'll roll out our own implementation of the pkg version extraction code, but this will do for now.

ikappaki:
> I opted to just wrap and stub the problematic function in the tests. Eventually we'll roll out our own implementation of the pkg version extraction code, but this will do for now.

Great, thanks! Just to summarise
1. The compilation warning with `package-get-version` mentioned in #3118 has been solved with https://github.com/clojure-emacs/cider/commit/30d425b398cd0790934a74f04eb0c2c2e383911f by completely removing `package-get-version` from `cider.el` (thus this PR hack is not necessary).
2. There was also a separate `cider-util-tests.el/cider--version` test failing while is `spy-on 'package-get-version`, this was fixed with https://github.com/clojure-emacs/cider/commit/4ce7e80bda7537717c66b451022a1350805c2ac0, by moving the inlined code into its own function.

Btw, looking at the `package-get-version` doc string in 27.2, it does say that it assumes that is called from the installed ELPA package directory, thus it does not lent itself to unit testing. Perhaps a better name for it would have been `package-get-installed-version`. This would have immediately saved me some of the confusion what is supposed to do :)

```
Return the version number of the package in which this is used.
Assumes it is used from an Elisp file placed inside the top-level directory
of an installed ELPA package.
The return value is a string (or nil in case we can’t find it).
```

PR #3122, opened 2021-12-30
Get :socket-file from `server-endpoint'

### LINKED ISSUE #3121: Failing integration test
@ikappaki Can you take a look at this failure (I think you're the author of this test):

```
nrepl-client-lifecycle start and stop nrepl client process

  error((void-variable params))
  apply(format-message (void-variable params))
  format-message((void-variable params))
error: (wrong-type-argument stringp (void-variable params))
```

I see the test refers to an unset `params` variable and I guess the byte-compilation fixes broke it somehow, although I'm not sure how it passed in the first place.

ikappaki:
Fixes #3121, whereby a `params`variable was recently introduced in the test that was defined in another test using buttercup's `:var`  construct. Buttercup suggest the use of [`lexicial scope`](https://github.com/jorgenschaefer/emacs-buttercup/blob/master/docs/writing-tests.md) if this construct is used:

```
it's important to note that lexical-binding must be non-nil for :var and :var* to work properly. Within a test file this is usually set using a local file variable.
```

thus this variable disappeared with the change to switch the file to lexical scope, and whence the test failure.

-----------------

- [x ] You've added tests (if possible) to cover your change(s)
- [x ] All tests are passing (`eldev test`)
- [x ] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes
- [x ] You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/test/nrepl-client-tests.el
+++ b/test/nrepl-client-tests.el
@@ -169,7 +169,7 @@
                                     server-process
                                     (lambda (client-endpoint)
                                       client-buffer)
-                                    (plist-get params :socket-file))))
+                                    (plist-get server-endpoint :socket-file))))
 
               ;; client connection is open
               (expect (process-status process-client)


bbatsov:
I figured as much. Seems we fixed a real issue then. :-) Thanks!

PR #3132, opened 2022-01-13
Fix profiling summary command name in dox

camsaul:
Minor dox fix.

[`cider-profile-display-stats` isn't `interactive`](https://github.com/clojure-emacs/cider/blob/master/cider-profile.el#L163-L175) and [`C- C-= S` is currently bound to `cider-profile-summary`](https://github.com/clojure-emacs/cider/blob/master/cider-profile.el#L41). I'm guessing maybe this code was reworked at some point but the dox didn't get updated?


bbatsov:
Thanks!

PR #3138, opened 2022-01-24
Fix broken doc link

jmckitrick:
In 'customize group' for CIDER, there is a setting 'Cider Jump To Pop To Buffer Actions' with a link for further details. That link is broken, and is corrected with this fix.

-------------------

No code changes needed.

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -136,7 +136,7 @@ The default value means:
 - If the target file is already visible in a window, reuse it (switch to it).
 - Otherwise, open the target buffer in the current window.
 
-For further details, see https://docs.cider.mx/cider/config/basic_config.html#control_what_window_to_use_when_jumping_to_a_definition"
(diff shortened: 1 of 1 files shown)


cider-common.el
```diff
@@ -136,7 +136,7 @@ The default value means:
 - If the target file is already visible in a window, reuse it (switch to it).
 - Otherwise, open the target buffer in the current window.
 
-For further details, see https://docs.cider.mx/cider/repl/configuration.html#_control_what_window_to_use_when_jumping_to_a_definition"
+For further details, see https://docs.cider.mx/cider/config/basic_config.html#control_what_window_to_use_when_jumping_to_a_definition"
```
vemv, line 139:
```suggestion
For further details, see https://docs.cider.mx/cider/config/basic_config.html#control-what-window-to-use-when-jumping-to-a-definition"
```

Only this one worked for me

vemv (commented):
Thanks for the heads up!

jmckitrick:
Oops! Copy-pasta fixed.

vemv:
🍻!

Issue #2580, opened 2019-02-12
Changed handling of stdout for cider-pprint-eval-last-sexp


Macroz:
Is nobody troubled by the change that `cider-pprint-eval-last-sexp` does not direct stdout to the same buffer as the eval result? Or there doesn't seem to be an option to enable this old behavior.

I.e. try to eval this with `C-c C-p` and see that the print is not shown in the buffer.
`(list (prn 1) 2)`

I use `C-c C-p` constantly and never look at the REPL or some separate stdout buffer. I used to be able to add debug prints here and there and keep track of the results with the debug prints.

## Expected behavior

A new buffer opens with `1` and `(nil 2)` shown.

## Actual behavior

A new buffer opens with only `(nil 2)` shown. The stdout `1` goes to REPL buffer.

## Steps to reproduce the problem

`cider-pprint-eval-last-sexp` this code `(list (prn 1) 2)`

## Environment & Version information

CIDER 0.21.0snapshot 20190125.1339

aiba:
Yes, me too.  I much preferred having stdout go to *cider-result*.  I would love to have that functionality back.

cichli:
I agree. If you rely on `C-c C-p` a lot it's extremely convenient to have the output from `*out*` and `*err*` in the popup as well. Having the option to redirect it to a different buffer is useful sometimes (for example if it ends up interleaved with the printed result due to laziness) but in most cases that isn't a problem, so I'm inclined to just revert to the old behaviour for now.

There are [smarter ways](https://github.com/Unrepl/spiral#standard-output-strings-can-be-grouped-into-their-respective-repl-entries) to deal with the interleaving problem, which weren't possible with `cider.nrepl.middleware.pprint`, but are now with the new version of nREPL.

cichli:
The redirecting is also (possibly) useful if you kill the result buffer. Maybe some `future` body or thread is still printing to `*out*`, even long after the evaluation returned. What should we do with that output? Two possible strategies:

* Use result buffer if it's alive, else `cider-interactive-eval-output-destination`
* Use _both_ result buffer (if it's alive) and `cider-interactive-eval-output-destination`

This would only apply to `*out*`/`*err*`. Killing the buffer should always stop printing the value.

cichli:
Just for my own reference, relevant change is 34783f5551c656667c20217171179769a3bc6bc8.

bbatsov:
[...]

PR #3143, opened 2022-01-28
add an interactive function cider-eval-last-sexp-to-kill-ring

blak3mill3r:
This is my first attempt to contribute to cider, please consider this PR a draft. I am no elisp expert, I welcome feedback. I read the CONTRIBUTING guidelines but I must confess that I did not understand all of them.

This is a feature that I'd been wanting: eval, storing the eval result in the kill ring

takes an optional `pretty` argument

stores the eval result in the kill ring

there was some discussion about this feature here
https://github.com/clojure-emacs/cider/issues/2580

It's based on `master` as of this morning, I'm using emacs built from source fairly recently:

```
emacs-27.1-11299-g5b250ca79b
```

I'm using native-compilation

I did not add a default keybinding in this commit. The keybindings I use for cider are very different from the defaults and so I don't feel qualified to suggest a reasonable default keybinding for this.

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -771,6 +771,22 @@ Optional argument DONE-HANDLER lambda will be run once load is complete."
                                  (cider-emit-interactive-eval-err-output err))
                                '()))
 
+(defun cider-eval-kill-ring-handler ()
+  "Make a handler for evaluating and storing the eval result in the kill ring."
+  (nrepl-make-response-handler (current-buffer)
+                               (lambda (_buffer value)
+                                 (with-temp-buffer
+                                   (insert
+                                    (if (derived-mode-p 'cider-clojure-interaction-mode)
+                                        (format "\n%s\n" value)
+                                      value))
+                                   (copy-region-as-kill (point-min) (point-max))))
+                               (lambda (_buffer out)
+                                 (cider-emit-interactive-eval-output out))
+                               (lambda (_buffer err)
+                                 (cider-emit-interactive-eval-err-output err))
+                               '()))
+
 (defun cider-eval-print-with-comment-handler (buffer location comment-prefix)
   "Make a handler for evaluating and printing commented results in BUFFER.
 LOCATION is the location marker at which to insert.  COMMENT-PREFIX is the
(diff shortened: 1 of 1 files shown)


cider-eval.el
```diff
@@ -771,6 +771,22 @@ Optional argument DONE-HANDLER lambda will be run once load is complete."
...
+(defun cider-eval-kill-ring-handler ()
+  "Make a handler for evaluating and storing the eval result in the kill ring."
+  (nrepl-make-response-handler (current-buffer)
+                               (lambda (_buffer value)
+                                 (with-temp-buffer
+                                   (insert
+                                    (if (derived-mode-p 'cider-clojure-interaction-mode)
+                                        (format "\n%s\n" value)
+                                      value))
+                                   (copy-region-as-kill (point-min) (point-max))))
```
bbatsov, line 783: This doesn't look correct to me as you'll be adding to the clipboard each chunk of the response (provided it's big enough to the split into several response messages). Normally you'd want to collect all the chunks until you get "done" and only then copy something to the clipboard.

bbatsov: Also - it seems to me it would still be a goo idea to echo somewhere the result, so you'd know what ended up in the kill-ring without having to check it.

cider-eval.el
```diff
@@ -933,6 +949,17 @@ buffer."
...
 
+(defun cider-eval-last-sexp-to-kill-ring (&optional pretty)
+  "Evaluate the expression preceding point.
+Store the evaluation result in the kill ring.
+Pretty print the result if PRETTY is non-nil."
+  (interactive "P")
+  (cider-interactive-eval nil
+                          (cider-eval-kill-ring-handler)
+                          (cider-last-sexp 'bounds)
+                          ;; this says that it does NOT require pretty printing... add an option?
```
bbatsov, line 960: I'm not quite sure what this is supposed to mean.

yuhan0:
I think it would be better to reserve one of Emacs' built-in registers for this, and _always_ copy the result of any interactive eval into it - something like

```elisp
(defcustom cider-eval-register ?e
  "When non-nil, save the last eval result into this register"
  :type '(choice character (const nil)))
```

Then instead of having to consciously call a dedicated eval-to-kill-ring command/ family of commands, one could decide *after* any regular `cider-eval-defun` or `eval-buffer` or `eval-last-sexp-up-to-point` etc. that the result was worth saving, and simply yank it from the `e` register. (`C-x r i e` / `" e p` with evil)

yuhan0:
Or alternatively, have a command which prints the repl's `*1` value into the kill ring, similar to `cider-inspect-last-result`.

bbatsov:
Yeah, I think that's a better and simpler approach. PR welcome!

yuhan0:
Which approach would you prefer? Auto saving to register (might be less discoverable) or adding a `cider-kill-last-result` command

bbatsov:
@yuhan0 I was actually thinking of doing both, as they are not mutually exclusive. :D Probably with a defcustom to enable the auto-save to register behavior.

bbatsov:
@yuhan0 Any chance you'll find a time for this soon? I'm thinking of cutting a new CIDER release in March, and it'd be nice if this made the cut.

yuhan0:
Oh yes, I'll submit a PR in the next day or two.

bbatsov:
Closing this in favour of #3162.

PR #3148, opened 2022-02-10
Eval UX improvements

yuhan0:
This is the first of a bunch of tweaks on my local copy of Cider that I've been using for almost 2 years now and never got around to submitting. I hope it's alright if I group several related changes together in a single PR, much of the friction comes from having to rebase commits and write multiple changelogs. 

- Overlay text takes on the face of the following text - this is especially noticeable with coloured parens and `cider-result-overlay-position` set to `'at-point'`.
Before (using cider-eval-last-sexp, the text takes on the yellow of the closing paren next to it):
image
After:
image

- When an error occurs *and* `cider-show-error-buffer` is set to nil, an overlay is created with an error indicating face. Currently there is no indication at all of the type or presence of error. 
I've been using this setting for some time now and find it much more responsive and less distracting than a stacktrace buffer popping up and needing to be dismissed. Most of the time a simple error message is enough information, and one can use `cider-selector` to bring up the stacktrace buffer if needed.

image

- Similarly with `cider-eval-to-comment` commands, it helps (especially for teaching purposes) to have the error output to the comment:
image

I don't think tests or user manual updates are needed here? 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#3127](https://github.com/clojure-emacs/cider/pull/3040): Strip all exec-opts flags (`-A` `-M` `-T` `-X`) if they exist in `cider-clojure-cli-aliases`. Also addresses a duplicate `:` in the generated `clj` command.
 * Enable `cider-enrich-classpath` by default.
+* [#3148](https://github.com/clojure-emacs/cider/pull/3148): Display error messages in multiline comment eval results, and in result overlays when `cider-show-error-buffer` is set to nil.
 
 ### Bugs fixed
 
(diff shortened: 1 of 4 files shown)


yuhan0:
The only thing it does is print out the error message exactly as sent to stderr, so naturally it handles all kinds of errors:

image

image

image

A slight visual nit is the trailing newline being rendered as a hanging background-coloured box - if `string-trim-right` is applied it would look like this:
image
I personally prefer the former for consistency of the extended faces.



>  - The final cause is displayed (might be under a chain of prior causes)

This is only a few keystrokes away with `cider-selector` to bring up the full stacktrace of exceptions - I've found that the majority of errors can be identified by their type and message alone.

>  - One or two lines are displayed at most

I don't see the advantage in truncating multiline error messages, after all the overlay can be dismissed easily. 

>  - Width isn't excessive

Linebreaks and indentation in the error message are meaningful, so trying to fill-paragraph could run into undesired formatting. Within the window the width is naturally wrapped around or extend offscreen depending on `truncate-line-mode.`

>   - Ideally, any irrelevant info would be trimmed away (e.g. Long is in module java.base...)

Again I think it's best to have as little "magic" as needed here - simply echoing the stderr instead of trying to determine what is relevant / irrelevant to show. (There are already external libraries like expound which simplify / pretty print error messages, which is a separate concern)

bbatsov:
I love the proposed changes and I don't think that any of them are controversial, so let's just merge this and polish it as we go based on whatever user feedback we receive. Thanks for tackling this! 🙇‍♂️

PR #3152, opened 2022-02-16
Update docs and variables related to enrich-classpath

vemv:
* Soft-deprecate variables/defuns made redundant by enrich-classpath
* Remove/update docs around enrich-classpath

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -542,7 +542,6 @@ Achieved by destructively manipulating `cider-stacktrace-suppressed-errors'."
          (line-shift (- (or (button-get button 'line) 0)
                         (or (nrepl-dict-get info "line") 1)))
          (file (or
-                (and (null var) (cider-resolve-java-class class))
                 (nrepl-dict-get info "file")
                 (button-get button 'file)))
          ;; give priority to `info` files as `info` returns full paths.
--- a/cider-util.el
+++ b/cider-util.el
@@ -799,10 +799,6 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
 
 (define-obsolete-variable-alias 'cider-jdk-src-paths nil "CIDER 1.3")
 
-(defun cider-resolve-java-class (class)
-  "Deprecated"
-  nil)
-
 (provide 'cider-util)
 
 ;;; cider-util.el ends here
--- a/cider-util.el
+++ b/cider-util.el
@@ -797,8 +797,6 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
-(define-obsolete-variable-alias 'cider-jdk-src-paths nil "CIDER 1.3")
-
 (provide 'cider-util)
 
 ;;; cider-util.el ends here
--- a/cider-util.el
+++ b/cider-util.el
@@ -797,11 +797,7 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
-(defcustom cider-jdk-src-paths '()
-  "Deprecated"
-  :group 'cider
-  :package-version '(cider . "0.17.0")
-  :type '(list string))
+(define-obsolete-variable-alias 'cider-jdk-src-paths nil "CIDER 1.3")
 
 (defun cider-resolve-java-class (class)
   "Deprecated"
--- a/cider-util.el
+++ b/cider-util.el
@@ -797,33 +797,15 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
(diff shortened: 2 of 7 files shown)


cider-stacktrace.el
```diff
@@ -542,7 +542,6 @@ Achieved by destructively manipulating `cider-stacktrace-suppressed-errors'."
          (line-shift (- (or (button-get button 'line) 0)
                         (or (nrepl-dict-get info "line") 1)))
          (file (or
-                (and (null var) (cider-resolve-java-class class))
```
vemv, line 545: The idea is that a cider-nrepl with an enriched classpath would already have all the info it needs.

doc/modules/ROOT/pages/about/compatibility.adoc
```diff
@@ -17,13 +17,16 @@ and Boot releases.
...
-release version (e.g. Java 15).  Generally speaking, we aim
+CIDER officially targets Java 8, Java 11, Java 17, and the most recent rapid
+release version (e.g. Java 18).  Generally speaking, we aim
 to support all Java releases that are currently officially supported
 by Oracle.
 
 NOTE: The requirements for Java are pretty much ``cider-nrepl``'s requirements.
 
+On Linux you are also strongly advised to make sure that JDK sources and javadocs are installed.
+You can find example commands in https://docs.cider.mx/cider/troubleshooting.html#navigation-to-jdk-sources-doesnt-work[Troubleshooting].
```
vemv, line 28: No idea if there's a fancier way to interlink

cider-util.el
```diff
@@ -797,33 +797,15 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
-(defcustom cider-jdk-src-paths '("/usr/lib/jvm/openjdk-8/src.zip")
-  "Used by `cider-stacktrace-navigate'.
-Zip/jar files work, but it's better to extract them and put the directory
-paths here.  Clojure sources here:
-https://repo1.maven.org/maven2/org/clojure/clojure/1.8.0/."
+(defcustom cider-jdk-src-paths '()
+  "Deprecated"
```
bbatsov, line 801: Keep the values as they were, but use `define-obsolete-variable-alias` (or however it was named) to warn the users that something shouldn't be used.

cider-util.el
```diff
@@ -797,33 +797,15 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
...
   :group 'cider
   :package-version '(cider . "0.17.0")
   :type '(list string))
 
 (defun cider-resolve-java-class (class)
```
bbatsov, line 806: I'm 99.9% that's not used outside CIDER, so you can just delete it.

bbatsov:
What's about people who aren't using Leiningen? Isn't the deprecated code still useful for them?

cider-util.el
vemv: Alright, considered this but it didn't seem to explicitly allow deprecating from x to nil.

I guess it accepts nil, worst-case the CI suite would show otherwise

vemv:
> What's about people who aren't using Leiningen? Isn't the deprecated code still useful for them?

enrich-classpath has finally a full deps.edn story, related PR to follow (I preferred to favor 2 separate PRs here). 

enrich-classpath is in general a generic Clojure program that could be used for further use cases.

cider-util.el
```diff
@@ -797,33 +797,7 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
...
-       (mapcar
-        (lambda (d)
-          (cond ((file-directory-p d)
-                 (expand-file-name file-name d))
-                ((and (file-exists-p d)
-                      (member (file-name-extension d) '("jar" "zip")))
-                 (format "zip:file:%s!/%s" d file-name))
-                (t (error "Unexpected archive: %s" d))))
-        cider-jdk-src-paths)))))
+(define-obsolete-variable-alias 'cider-jdk-src-paths nil "CIDER 1.3")
```
bbatsov, line 800: I think `nil` can actually be any string like "This was rendered obsolete by enrich-classpath" or whatever.

doc/modules/ROOT/pages/about/compatibility.adoc
bbatsov: You can do something like `xref:troubleshooting.adoc#navigation-to-jdk-sources-doesnt-work[Troubleshooting]`.

bbatsov:
Got it. I added a couple of small remarks, but overall the changes look good.

vemv:
A string argument won't compile:

```
ielm> (define-obsolete-variable-alias 'an-example "A string" "CIDER 1.3")
*** Eval error ***  Wrong type argument: symbolp, "A string"
```

...`nil` works, however when it's set via `custom-set-variables`, it will print `Error setting nil: (setting-constant nil)` (without throwing an error).

Maybe we can simply remove the variable? That way, `custom-set-variables` will print nothing, which seems less confusing.

Issue #3153, opened 2022-02-20
CIDER nrepl client can fail the ansi colorisation process in the REPL output


ikappaki:
## Expected behavior

CIDER should be able to colorize REPL output consisting of valid ansi color control characters (aka [SGR sequences](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters)).

## Actual behavior

CIDER can produce SGR characters verbatim, instead of using the code to colorize rest of the output stream.

## Steps to reproduce the problem

Jack in to an empty deps.edn file, and repeatedly print out the 0 to 9 preceded by the bold intensity SGR seq (i.e `ESC [ 1 m`):
```clojure 
user> (doseq [i (range 500)] (print (format "\u001b[1m%d" (mod i 10))))
```
Somewhere midway in the below, the SGR control sequence is printed verbatim (look for `�[1m9`):

```
0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678[1m9012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
```

### CIDER version information

```
;; CIDER 1.2.0 (Nice), nREPL 0.9.0
;; Clojure 1.10.3, Java 17
```

### Emacs version

GNU Emacs 27.2 (build 1, x86_64-w64-mingw32) of 2021-03-26

### Operating system

*Windows 10*

----

The issue is a side effect of the #1813 workaround put in place sometime ago to fix #1794 whereby the cider nrepl client was stop printing output on stray ESC characters, caused by what is now known as [bug#53808](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=53808) and fixed in Emacs 29. The cider fix is flushing out verbatim any residual ansi control seq caught up between nrepl output fragments, and thus when the output is send in chunks by the REPL, it is likely that the control chars will leak out.

PR for Emacs versions < 29 to follow.

Thanks

vemv:
https://github.com/clojure-emacs/cider/pull/3154

PR #3154, opened 2022-02-20
Workaround for `ansi-color-apply' Emacs bug#53808

ikappaki:
Hi,

could you please consider proper fix for #3153 which addresses Emacs [bug#53808](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=53808), whereby stray ESC characters in the nREPL output can block the `ansi-color-apply` Emacs fn used by the CIDER nREPL client to colorise the incoming REPL output.

The proposed patch supersedes @dpsutton 's  #1813 excelled work, which worked around the chunked nREPL output split on partial ANSI control sequences by immediately flushing out any partial control sequence caught up between the chunks (and this is why we see valid ANSI color control chars printout in the #3153 example).

The patch improves on #1813 by only flushing out the partial ANSI control seq if it determines it is definitely not a partial or full [C1 SGR](https://en.wikipedia.org/w/index.php?title=ANSI_escape_code&oldid=1070369816#Description) seq (i.e. `ESC` followed by `[` followed by zero or more `[:digit:]+;` followed by one or more digits, followed by `m`), otherwise it waits for the next chunk until the SGR seq is complete or becomes invalid.

Added additional test assertions to catch the case where ANSI color output is split at various points in the SGR seq.

Enabled only in Emacs versions < 29.

Also fixed a couple of old linter docstring warnings, while at it.

Thanks

-----------------

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
         (set-marker cider-repl-prompt-start-mark prompt-start)
         prompt-start))))
 
-(defun cider-repl--flush-ansi-color-context ()
-  "Flush ansi color context after printing.
-When there is a possible unfinished ansi control sequence,
- `ansi-color-context` maintains this list."
-  (when (and ansi-color-context (stringp (cadr ansi-color-context)))
-    (insert-before-markers (cadr ansi-color-context))
-    (setq ansi-color-context nil)))
+(defun ansi-color-apply--emacs-bug-53808-workaround (string)
+  "Like `ansi-color-apply', but does not block on stray ESC in STRING.
+
+Workaround for Emacs bug#53808 to also include any non-SGR ANSI control
+sequence found at the end of STRING."
+  (let* ((result (ansi-color-apply string))
+
+         (context-flush?
+          (when-let (fragment (and ansi-color-context (cadr ansi-color-context)))
+            (save-match-data
+              ;; An SGR seq is defined as ESC [ (:digit:+;)* :digit:+ m
+              ;; e.g. "\e[1m" or "\e[3;30m"
+              (when (string-match
+                     (rx (sequence ?\e
+                                   (? (and (or ?\[ eol)
+                                           (or (+ (any (?0 . ?9))) eol)
+                                           (* (sequence ?\; (+ (any (?0 . ?9)))))
+                                           (or ?\; (? (group-n 1 ?m)))))))
+                     fragment)
+                (let* ((sgr-end-pos (match-end 0))
+                       (sgr-whole? (match-string 1))
+                       (fragment-matches-whole? (or (= sgr-end-pos 0)
+                                                    (= sgr-end-pos (length fragment)))))
+                  (when (and (not sgr-whole?)
+                             (not fragment-matches-whole?))
+                    ;; Definitely not an SGR seq.
+                    t)))))))
+
+    (if (not context-flush?)
+        result
+
+      (progn
+        ;; Temporarily replace ESC char to flush it out, and append to result.
+        (aset (cadr ansi-color-context) 0 ?\0)
+        (let ((result-context (ansi-color-apply "")))
+          (aset result-context 0 ?\e)
+          (concat result result-context))))))
 
 (defvar-local cider-repl--ns-forms-plist nil
   "Plist holding ns->ns-form mappings within each connection.")
@@ -672,7 +702,9 @@ namespaces.  STRING is REPL's output."
   (put-text-property 0 (length string) 'help-echo 'cider-locref-help-echo string)
   string)
 
-(defvar cider-repl-preoutput-hook '(ansi-color-apply
+(defvar cider-repl-preoutput-hook `(,(if (< emacs-major-version 29)
+                                       'ansi-color-apply--emacs-bug-53808-workaround
+                                      'ansi-color-apply)
                                     cider-repl-highlight-current-project
                                     cider-repl-highlight-spec-keywords
                                     cider-repl-add-locref-help-echo)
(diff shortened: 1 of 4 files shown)


cider-repl.el
```diff
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
...
-    (insert-before-markers (cadr ansi-color-context))
-    (setq ansi-color-context nil)))
+(defun ansi-color-apply--emacs-bug-53808-workaround (string)
+  "Like `ansi-color-apply', but does not block on stray ESC in STRING.
+
+Workaround for Emacs bug#53808 to also include any non-SGR ANSI control
+sequence found at the end of STRING."
+  (let* ((result (ansi-color-apply string))
+
+         (context-flush?
```
vemv, line 603: Suggested comment: what is context flushing? Why do we possibly perform it?

cider-repl.el
```diff
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
...
+          (when-let (fragment (and ansi-color-context (cadr ansi-color-context)))
+            (save-match-data
+              ;; An SGR seq is defined as ESC [ (:digit:+;)* :digit:+ m
+              ;; e.g. "\e[1m" or "\e[3;30m"
+              (when (string-match
+                     (rx (sequence ?\e
```
vemv, line 609: Suggested comment: what are we looking for here? Why?

cider-repl.el
```diff
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
...
+                                   (? (and (or ?\[ eol)
+                                           (or (+ (any (?0 . ?9))) eol)
+                                           (* (sequence ?\; (+ (any (?0 . ?9)))))
+                                           (or ?\; (? (group-n 1 ?m)))))))
+                     fragment)
+                (let* ((sgr-end-pos (match-end 0))
+                       (sgr-whole? (match-string 1))
+                       (fragment-matches-whole? (or (= sgr-end-pos 0)
+                                                    (= sgr-end-pos (length fragment)))))
```
vemv, line 618: This logic seems worth commenting; while the variable names are useful one might not necessarily follow the intent

cider-repl.el
```diff
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
...
         prompt-start))))
 
-(defun cider-repl--flush-ansi-color-context ()
-  "Flush ansi color context after printing.
-When there is a possible unfinished ansi control sequence,
- `ansi-color-context` maintains this list."
-  (when (and ansi-color-context (stringp (cadr ansi-color-context)))
-    (insert-before-markers (cadr ansi-color-context))
-    (setq ansi-color-context nil)))
+(defun ansi-color-apply--emacs-bug-53808-workaround (string)
```
vemv, line 596: Ideally this defun (or the core of it) would be _unit_ tested. I realise you have expanded the existing tests, but those look integration-ish (they are concerned with buffers etc; a more direct test would be simply show an input->output mapping for a pure function)

cider-repl.el
```diff
@@ -593,13 +593,43 @@ Return the position of the prompt beginning."
...
+  "Like `ansi-color-apply', but does not block on stray ESC in STRING.
+
+Workaround for Emacs bug#53808 to also include any non-SGR ANSI control
```
vemv, line 599: Use a link instead of `bug#53808` if possible

vemv (commented):
Seems reasonable to me, thanks for your work!

Some suggestions.

cider-repl.el
bbatsov: I'd name this `cider-repl--ansi-color-apply`. I don't like having in the package functions that don't have a `cider-` prefix, as this makes it harder for users to figure out where they come from.

cider-repl.el
ikappaki: Added comments around the ansi color context and the process of flushing.

cider-repl.el
ikappaki: Added comments on what we are looking for.

cider-repl.el
ikappaki: This format is the standard way Emacs backs are referred to. I personally find links in doc strings as adding two much text for the eye to see. Though I made sure the link is in the check in comment. Nevertheless I can add it in the doc string if you insist :)

ikappaki:
Hi, I've updated the comments are per feedback, I've also realised that there is no need to check for the presence of a fully realised SGR seq in the `ansi-color-context` fragment, since by definition the fragment only holds partial ANSI control seq of which the SGR seq is a subset of).

Thanks

cider-repl.el
vemv: No issue! TIL

cider-repl.el
ikappaki: @vemv Not sure if the mention of "pure function" refers to that function, it is not pure, since it calls on `ansi-color-apply` which keeps state in `ansi-color-context`. My opinion is that we are fine with the current testing since the tests are meant to be exercising this function, and is the only way it can get through to it. I think i have to duplicate exactly the same input seq, which I would recommend. Furthermore, this function will eventually become inactive in Emacs 29, in which case the tests will not run, though the "integration" tests will be still valid.

@bbatsov, name changed.

cider-repl.el
ikappaki: Logic has been simplified to only check for a partial SGR seq. I realised that the looking for a full SGR expression in the fragment is redundant, since the `ansi-color-context` fragment can only hold partial ansi control seq by definition.

Please let me know if the comments are still not self explanatory.

cider-repl.el
vemv: I realise `cider-repl--ansi-color-apply` is not pure, however often these defuns can be decomposed into pure defuns, which can be tested in a particularly illustrative way. Which would be useful here, since this particular domain seems rather dense.

Please don't consider my suggestion a blocker, you are right in that the CIDER test suite is what it is (and it's ok, just not _awesome_, arguably).

Feel free to give it a shot if it doesn't turn out to be a disproportionate effort.

ikappaki: I think unless there is a strong recommendation otherwise, I will pass on this one. It is as you said that this particular domain is dense, but I don't think it worth spending more time trying to structure a throwaway hack around a bug into something that spans multiple functions that are each well tested. I find with the update that the fn is now well documented, and whoever wants to look into the full details, they  need have a read through the ansi-color lib, the ansi control sequences, and in particular the SGR subset. I also think that the fun is well tested as part of exercising the functionality above it that uses it with all possible combination of ansi color sequences. Nevertheless I can still have a look at breaking it up if you or somebody else insists. Thanks :)

vemv (approved):
Thanks much for your contribution!

PR #3155, opened 2022-02-21
`cider-clojure-cli-jack-in-dependencies`: always remove duplicates

vemv:
Prevents all sorts of issues, particularly as this area is changing as of lately.

This way we guard against any misconfiguration, misunderstanding, subtle bug, etc.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -608,17 +608,22 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
    " -- "
    params))
 
+(defun cider--dedupe-deps (deps)
+  (cl-delete-duplicates deps :test 'equal))
+
 (defun cider-clojure-cli-jack-in-dependencies (global-options _params dependencies)
   "Create Clojure tools.deps jack-in dependencies.
 Does so by concatenating DEPENDENCIES and GLOBAL-OPTIONS into a suitable
 `clojure` invocation.  The main is placed in an inline alias :cider/nrepl
 so that if your aliases contain any mains, the cider/nrepl one will be the
 one used."
-  (let* ((deps-string (string-join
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -614,19 +614,21 @@ Does so by concatenating DEPENDENCIES and GLOBAL-OPTIONS into a suitable
...
-  (let* ((deps-string (string-join
-                       (seq-map (lambda (dep)
-                                  (format "%s {:mvn/version \"%s\"}" (car dep) (cadr dep)))
-                                (append (cider--jack-in-required-dependencies) dependencies))
-                       " "))
+  (let* ((all-deps (thread-last dependencies
+                     (append (cider--jack-in-required-dependencies))
+                     ;; Duplicates are never OK since they would result in
+                     ;; `java.lang.IllegalArgumentException: Duplicate key [...]`:
+                     (cl-delete-duplicates)
```
vemv, line 621: Apparently cl-delete-duplicates doesn't work with nested lists, can revisit this PR when the build is green

test/cider-tests.el
```diff
@@ -435,6 +435,10 @@
         (let ((cider-clojure-cli-aliases ":test"))
           (expect (cider-clojure-cli-jack-in-dependencies nil nil deps)
                   :to-equal expected))
+        (describe "should remove duplicates, yielding the same result"
```
bbatsov, line 438: Seems a bit weird to nest this under "handles aliases correctly". I think a dedicated `it` block would be better here.

bbatsov:
Good change. This dep dedup should be applied for all build tools I guess.

vemv:
> Good change. This dep dedup should be applied for all build tools I guess.

Guess so! Perhaps in a subsequent pr, that way we can think about Lein impact particularly.

PR #3165, opened 2022-03-05
Fix most of the build

vemv:
Fixes linting and tests but not `cider-util.el:377:8:Error: value returned from (fboundp 'package-get-version) is unused` (https://app.circleci.com/pipelines/github/clojure-emacs/cider/1157/workflows/80e0b8e4-6f6d-4497-81a6-5268ffac52a1/jobs/5337)

### DIFF
--- a/test/cider-tests.el
+++ b/test/cider-tests.el
@@ -425,15 +425,15 @@
         (spy-on 'cider-jack-in-resolve-command :and-return-value "clojure")
         (expect (plist-get (cider--update-jack-in-cmd nil) :jack-in-cmd)
                 :to-equal expected)))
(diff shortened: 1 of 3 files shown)


test/cider-tests.el
```diff
@@ -210,6 +210,7 @@
     (before-each
       ;; FIXME: Needed because its set in an earlier test
       (setq-local cider-jack-in-lein-plugins nil)
+      (setq-local cider-jack-in-dependencies nil)
```
vemv, line 213: Similar to the ones above per `Needed because its set in an earlier test`

vemv:
Iterated a few times until the build was not only green but it also made sense :)

PR #3171, opened 2022-03-24
[Fixes #3170] Skip ensure of repl available on xref functions

### LINKED ISSUE #3170: Sideeffect checking `cider--xref-backend`
When doom runs `(xref-find-backend)` to find current xref backend, cider gives a error: `No linked CIDER sessions`. xref backend functions should not have side-effects like that, this is making doom's code not be able to check for other backends like lsp making users need to jack in repl to use doom's lsp find-definition for example even if not using cider find definition.

## Expected behavior
`(cider--xref-backend)` not throw any side effects like checking if repl is available.

## Actual behavior
`(cider--xref-backend)` checks if repl is available, throwing a error message if not.

`M-x (cider--xref-backend)` with a repl not enabled in a project.

### CIDER version information

```
;; CIDER 1.3.0 (Ukraine), nREPL 0.9.0
;; Clojure 1.11.0, Java 11.0.12
```

### Emacs version

28.0.91

### Operating system

NixOS

ericdallo:
Fixes #3170 adding a extra arg to skip the ensure for xref functions.

-----------------

Thanks!

### DIFF
--- a/cider-find.el
+++ b/cider-find.el
@@ -234,7 +234,7 @@ thing at point."
   "Used for xref integration."
   ;; Check if `cider-nrepl` middleware is loaded. Allows fallback to other xref
   ;; backends, if cider-nrepl is not loaded.
-  (when (cider-nrepl-op-supported-p "ns-path")
+  (when (cider-nrepl-op-supported-p "ns-path" 'skip-ensure)
     'cider))
 
 (cl-defmethod xref-backend-identifier-at-point ((_backend (eql cider)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * Use clojure-mode [5.14.0](https://github.com/clojure-emacs/clojure-mode/blob/v5.14.0/CHANGELOG.md#5140-2022-03-07).
+* [#3170](https://github.com/clojure-emacs/cider/issues/3170) Skip ensure repl available on xref functions.
 
 ## 1.3.0 (2021-03-07)
 
--- a/cider-client.el
+++ b/cider-client.el
@@ -161,9 +161,12 @@ nREPL connection."
           (clojure-expected-ns path)))
     (clojure-expected-ns path)))
 
-(defun cider-nrepl-op-supported-p (op &optional connection)
-  "Check whether the CONNECTION supports the nREPL middleware OP."
-  (nrepl-op-supported-p op (or connection (cider-current-repl nil 'ensure))))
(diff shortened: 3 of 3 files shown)


vemv (commented):
Looks like it would work!

Did you get to try it? (Of course, without any other workaround in place)

ericdallo:
Yes @vemv, I called that function and it did return `nil` instead of throwing a error :)

vemv (approved):

ericdallo:
Thank you!

ericdallo:
that's what I just did! 😄

vemv:
Just in case, I meant upstream e.g. https://github.com/hlissner/doom-emacs/pull/6183

ericdallo:
yes, I was talking with Henrik on Doom's discord, he said will bump cider soon with this fix :)

vemv:
awesome!

cider-find.el
```diff
@@ -234,7 +234,7 @@ thing at point."
   "Used for xref integration."
   ;; Check if `cider-nrepl` middleware is loaded. Allows fallback to other xref
   ;; backends, if cider-nrepl is not loaded.
-  (when (cider-nrepl-op-supported-p "ns-path")
+  (when (cider-nrepl-op-supported-p "ns-path" 'skip-ensure)
```
vemv, line 237: This change had an oversight fixed in https://github.com/clojure-emacs/cider/commit/36277802a47b0dbc9c1fd60f60908963c59c58d6

bbatsov:
I'm not sure this was the right approach to solving the problem, but you merged it before I saw the PR. Not that big of deal in general, but I was thinking that actually the predicate function should always return `t/nil` and there would be a `cider-ensure-op-supported` matching it for the cases where the other behavior is needed.

bbatsov:
(likely the behavior you're trying to fix is somewhat accidental, given where it's coming from)

ericdallo:
Yeah, I like precicate functions without side-effects, your idea sounds better indeed, but probably will require a little bit more refactor I think

bbatsov:
We can always simply check if the test suite works fine if we make this the default behavior (suppressing the exception from `cider-current-repl`). I think this function didn't exists back when `cider-nrepl-op-supported-p` was written and its introduction altered the API. I assume that code that actually care about getting an exception is using `cider-ensure-connected` directly.

vemv:
> I'm not sure this was the right approach to solving the problem, but you merged it before I saw the PR.

Yeah that wasn't ideal, sorry, I guess I wanted to leave it bug-free for Doom users asap.

Given this 'api' has not been shared as a tagged release, I'd be happy for the solution to be changed if that's better.

Issue #3178, opened 2022-04-05
CI: Failing lint step


vemv:
The `test-lint` job keeps failing, I wonder what could we do about it?

https://app.circleci.com/pipelines/github/clojure-emacs/cider/1179/workflows/2744caf7-cdf6-4ed2-aca8-8eafe0f7324c/jobs/5440

The error in question (`Error: value returned from (fboundp 'package-get-version) is unused`) seems a false positive, where would we report that? Is there something we could upgrade first?

Worst-case scenario, we could remove the `--warnings-as-errors` flag and do our own grepping with some minimal blaklisting logic.

ikappaki:
Hi @vemv,

this is a bit tricky ... The warning is legit, but the actual text is misleading. What is happening here is that at compile time the compiler encountered a conditional whose both branches return `nil`.

``` elisp
(if (fboundp 'package-get-version)
      (package-get-version)
    nil))
``` 
Or to simplify what the compiler sees:

``` elisp
(if (<runtime-predicate>)
     nil
   nil)
```
and thus it complains that the return value of the `(runtime-predicate)` is unused since the if block will return `nil`, no need to look at it.

I suppose you might have tried to run the compilation locally, and most likely it succeed without a warning/error
``` batch
d:/src/cider $ rm cider-util.elc; eldev.bat -dtT compile --warnings-as-errors cider-util.el
...
[00:01.099]  Byte-compiling file `cider-util.el'...
[00:01.099]  Loading file `cider-util.el' before byte-compiling it...
[00:01.273]  Loading the byte-compiled file `cider-util.elc'...
[00:01.278]  Saving target dependencies to file `.eldev/28.1/target-dependencies.build'...
```

This is because `(package-get-version)` does return a non-nil value on your local checkout, i.e. the `(<runtime-predicate>)` has to be checked at runtime to check which branch to take, so it is going to be used:

``` elisp
(if (<runtime-predicate>)
     "<package-version>"
   nil)
```

Why does it fail on the Circle CI? because `package-get-version` returns nil when it runs over there! Why does it only return `nil` on Circle? Because `package-get-version` on Emacs 27.x will only consider getting package-versions if the parent directory is name the same as the package name, i.e. `cider` in our case  (This has been improved on 28.x and should not be a problem any more). On circle ci the parnet directory is named "project" (there is nothing wrong with that):

``` bash
#!/bin/bash -eo pipefail
eldev -dtT compile --warnings-as-errors

[...]

PR #3181, opened 2022-04-14
Fix unused returned value compilation warning

ikappaki:
> Fixes https://github.com/clojure-emacs/cider/issues/3178

Fixes `cider--pkg-version` unused return value compilation warning, see #3178 for details.

`Error: value returned from (fboundp 'package-get-version) is unused`

Converts `if` conditional to `cond`, and fixes unused param warnings by prefixing with  param names with `_`.

-----------------

Thanks!

### DIFF
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -223,7 +223,7 @@ be displayed."
         (cider-find-ns nil value)
       (cider-find-var current-prefix-arg value))))
 
-(defun cider-browse-ns-handle-mouse (event) ;FIXME: Unused arg!
+(defun cider-browse-ns-handle-mouse (_event) ;FIXME: Unused arg!
   "Handle mouse click EVENT."
   (interactive "e")
   (cider-browse-ns-operate-at-point))
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -223,7 +223,7 @@ be displayed."
         (cider-find-ns nil value)
       (cider-find-var current-prefix-arg value))))
 
-(defun cider-browse-ns-handle-mouse (_event) ;FIXME: Unused arg!
+(defun cider-browse-ns-handle-mouse (_event)
   "Handle mouse click EVENT."
   (interactive "e")
   (cider-browse-ns-operate-at-point))
--- a/cider-classpath.el
+++ b/cider-classpath.el
@@ -80,7 +80,7 @@
          (line (buffer-substring-no-properties bol eol)))
     (find-file-other-window line)))
 
-(defun cider-classpath-handle-mouse (_event) ;FIXME: Unused arg!
+(defun cider-classpath-handle-mouse (_event)
   "Handle mouse click EVENT."
   (interactive "e")
   (cider-classpath-operate-on-point))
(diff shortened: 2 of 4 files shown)


cider-browse-ns.el
```diff
@@ -223,7 +223,7 @@ be displayed."
         (cider-find-ns nil value)
       (cider-find-var current-prefix-arg value))))
 
-(defun cider-browse-ns-handle-mouse (event) ;FIXME: Unused arg!
```
bbatsov, line 226: You can remove the fixme here.

cider-classpath.el
```diff
@@ -80,7 +80,7 @@
          (line (buffer-substring-no-properties bol eol)))
     (find-file-other-window line)))
 
-(defun cider-classpath-handle-mouse (event) ;FIXME: Unused arg!
+(defun cider-classpath-handle-mouse (_event) ;FIXME: Unused arg!
```
bbatsov, line 83: And here.

cider-util.el
```diff
@@ -377,9 +377,8 @@ propertized (defaults to current buffer)."
 (defun cider--pkg-version ()
   "Extract CIDER's package version from its package metadata."
   ;; FIXME: Inline the logic from package-get-version and adapt it
-  (if (fboundp 'package-get-version)
-      (package-get-version)
-    nil))
+  (cond ((fboundp 'package-get-version)
```
bbatsov, line 380: I have to say I'm really puzzled why this work and the previous version doesn't. Seems like some bug in Emacs's own linter or something.

ikappaki: I believe this is due to the compiler optimisations:

The compiler appears to be optimising the `if` case, the `fboundp` value is discarded and a `constant nil` returned:

``` elisp
(defun if-ex ()
  (if (fboundp 'messsage)
      nil
    nil))

(disassemble 'if-ex)
;; byte code for if-ex:
;;   args: nil
;; 0       constant  fboundp
;; 1       constant  messsage
;; 2       call      1
;; 3       discard   
;; 4       constant  nil
;; 5       return    
```

while for the `cond` case, the compiler does not optimise the call, and is concerned with the return value (the `  goto-if-nil-else-pop 1` part):

``` elisp 
(defun cond-ex ()
  (cond ((fboundp 'message)
         nil)))

(disassemble 'cond-ex)

;; byte code for cond-ex:
;;   args: nil
;; 0       constant  fboundp
;; 1       constant  message
;; 2       call      1
;; 3       goto-if-nil-else-pop 1
;; 6       constant  nil
;; 7:1     return    
```
Thus i believe, without looking at the compiler details, the warning is the fact that the return value in the first case is discarded i.e. unused. The warning operates on the compiled code, not the sexp tree.

cider-classpath.el
ikappaki: done

cider-browse-ns.el
ikappaki: done

cider-util.el
vemv: Impressive work!

Maybe worth adding as a comment that only `cond` works here (and not `when` etc) as you noted in https://github.com/clojure-emacs/cider/issues/3178#issuecomment-1099526826.

bbatsov: Yeah, a comment would be useful, otherwise someone might decide to improve the readability of the code.

ikappaki: Comment added.

bbatsov:
Thanks!

PR #3185, opened 2022-04-29
Automatically extract let bindings in cider-eval-in-context

yuhan0:
A little convenience function I wrote a while ago for `cider-eval-in-context` commands, for the common use case of evaluating some nested sub-expression that depends on outer let bindings.

eg:
```clj
(defn complicated-fn [xs]
  (if-let [[x & r] (seq xs)]
    (let [m {:id x}]
      (cond (pred1 x) f1
            (pred2 x)
            (let [v (str x)]
              (fn []
                (assoc m :val v) ; <-- cursor here
                ))
            (pred3 x) :something-else))
    identity))
```

Here I can call `cider-eval-last-sexp-in-context` with `C-u`, letting it automatically traverse and pick up outer let bindings. The prompt now reads
```
Evaluation context (let-style): [x & r] (seq xs), m {:id x}, v (str x), 
                               ↑
                               point here
```

with the cursor at the beginning of the prompt by default, I can fill additional context eg. `xs [5]`.

https://user-images.githubusercontent.com/22216124/165936461-c637b59b-bf04-4550-a9a6-157318b4aa06.mp4

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## New features
 
+* [#3185](https://github.com/clojure-emacs/cider/pull/3185): Add feature to `cider-eval-in-context` for automatically extracting parent let bindings when called with `C-u` prefix argument.
 * Add new interactive command `cider-inspire-me`. It does what you'd expect.
 * [#3162](https://github.com/clojure-emacs/cider/pull/3162): Save eval results into kill ring and registers.
   * Add new customization variable `cider-eval-register` to automatically store the last interactive eval result into the specified register.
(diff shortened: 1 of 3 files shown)


yuhan0:
Note: The second commit changes the API of a private `--` function in order to display the eval overlays at the correct location.

bbatsov:
> cider-eval.el:1037:0 (checkdoc)  Argument `guess' should appear (as GUESS) in the doc string

I guess that's coming from your changes.

bbatsov:
I like the proposed change, but please mention it somewhere in the docs for `cider-eval-in-context`, otherwise few people will discover it.

yuhan0:
The remaining lint failure is a bug I accidentally introduced with my last PR, I'll group it with a couple of other non user-facing bugfixes in a separate PR

bbatsov:
Thanks!

PR #3188, opened 2022-04-30
Add support for undef-all op

yuhan0:
See https://github.com/clojure-emacs/cider-nrepl/pull/698.

Added prefix argument behaviors to `cider-load-buffer`, `-load-file`, `-load-all-files`, and `-eval-ns-form` commands, since these are likely to throw errors when some redefined namespace alias conflicts with the hidden REPL state. Conceptually something like a "hard reload".  

Also adds a standalone command `cider-undef-all`, which is somewhat less useful on its own and would usually be followed by a `load` anyway. Proposed keybinding: `C-c C-M-u` by extension of `C-c C-u`.

Also fixes a bug in my last PR #3185 (I should really avoid doing these last minute refactorings 😶) 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1025,7 +1025,7 @@ That's set by commands like `cider-eval-last-sexp-in-context'.")
   "Return context for `cider--eval-in-context' by extracting all parent let bindings."
   (save-excursion
     (let ((res ""))
-      (condition-case er
+      (condition-case nil
           (while t
             (backward-up-list)
             (when (looking-at (rx "(" (or "when-let" "if-let" "let") (opt "*")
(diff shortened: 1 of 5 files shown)


cider-eval.el
```diff
@@ -1329,12 +1329,15 @@ buffer, else display in a popup buffer."
       (cider-pprint-eval-defun-to-comment)
     (cider--pprint-eval-form (cider-defun-at-point 'bounds))))
 
-(defun cider-eval-ns-form ()
-  "Evaluate the current buffer's namespace form."
-  (interactive)
+(defun cider-eval-ns-form (&optional undef-all)
```
bbatsov, line 1332: I don't see this mentioned in the docs. Perhaps it wasn't there in the first place?

yuhan0: Added, thanks for the catch!

PR #3196, opened 2022-05-04
Issue template refinements

vemv:
* Replace Boot reference with more usual tools.deps reference
* Ask for more comprehensive JDK info (although `CIDER version information` already provides to us the most important part)
  * I think this can occasionally be super useful - and from time to time it can prevent an issue from being created altogether.


bbatsov:
Looks good!

Issue #3195, opened 2022-05-03
`(error "Cyclic keymap inheritance")` on `cider-test-run-test`


rrudakov:
## Expected behavior

No emacs errors after running `cider-test-run-test` or any other `cider-test-*` funciton.

## Actual behavior

```
Debugger entered--Lisp error: (error "Cyclic keymap inheritance")
  cider-test-report-mode()
  cider-test-render-report("*cider-test-report*" (dict "error" 0 "fail" 0 "ns" 1 "pass" 1 "test" 1 "var" 1) (dict "core-test" (dict "simple-test" ((dict "context" nil "index" 0 "message" "" "ns" "core-test" "type" "pass" "var" "simple-test")))))
  #f(compiled-function (response) #<bytecode 0xdc993a82ffbcc89>)((dict "gen-input" nil "id" "33" "results" (dict "core-test" (dict "simple-test" ((dict "context" nil "index" 0 "message" "" "ns" "core-test" "type" "pass" "var" "simple-test")))) "session" "5c50f48b-7e2b-4566-bbe4-3d90fbfb4aca" "summary" (dict "error" 0 "fail" 0 "ns" 1 "pass" 1 "test" 1 "var" 1) "testing-ns" "core-test"))
  nrepl--dispatch-response((dict "gen-input" nil "id" "33" "results" (dict "core-test" (dict "simple-test" ((dict "context" nil "index" 0 "message" "" "ns" "core-test" "type" "pass" "var" "simple-test")))) "session" "5c50f48b-7e2b-4566-bbe4-3d90fbfb4aca" "summary" (dict "error" 0 "fail" 0 "ns" 1 "pass" 1 "test" 1 "var" 1) "testing-ns" "core-test"))
  nrepl-client-filter(#<process nrepl-connection<1>> "d9:gen-inputle2:id2:337:resultsd9:core-testd11:sim...")
```

## Steps to reproduce the problem
1. Create a simple test project with failing test:
```clojure
;; deps.edn
{:paths ["test"]}
;; ./test/core-test.clj
(ns core-test
  (:require [clojure.test :refer [deftest is]]))

(deftest simple-test
  (is (= 5 (* 2 2))))
```
2. Run the test using `C-c C-t C-t` or `cider-test-run-test`
3. Test fails, dismiss test report window by pressing `q`
4. Fix the test
```clojure
(ns core-test
  (:require [clojure.test :refer [deftest is]]))

(deftest simple-test
[...]

PR #3197, opened 2022-05-04
Revert "Inherit test-report mode from cider-popup-mode"

yuhan0:
This change resulted in "Cyclic keymap inheritance" errors for some users, see #3195

I can't reproduce the error, but it seems many people are affected and it's probably not worth the breakage ultimately.
(so much for non-user-visible changes!)

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Hmm, but I was under the impression this was just a byte-compilation issue, so I wouldn't revert the change just because of this.

bbatsov:
On the other hand - it might be something to do with Emacs 28. I'm also on Emacs 29 myself. I'll think a bit about this.

Issue #3198, opened 2022-05-07
Fix emacs 29 CI breaks 07 May 2022


ikappaki:
## Expected behavior

all CI tests should pass for Emacs 29

## Actual behavior

There are a few errors around queue identification and ansi color mismatch

## Steps to reproduce the problem

Failures evident on every new commit

## Environment & Version information

Emacs-29 on debian, CIDER latest.

PR to follow.

PR #3199, opened 2022-05-07
Fix Emacs-29 latest breaks

ikappaki:
Fixes Emacs 29 breaks #3198

1. ANSI context inspection test have been removed, since this is internal to Emacs and shouldn't be relied upon. It so happened that the internal implementation has changed with Emacs-29. The recent tests that look at the external behavior of the REPL are sufficient and cover the cases just removed.
2. The  `(clojure-mode)` is required for the `cider-eldoc-beginning-of-sexp` to properly work, otherwise it gives "Use: No comment syntax is defined" warnings.
3. The current snapshot image used by https://github.com/Silex/docker-emacs has a bug that returns strange tag symbols for cl structs when `eval-and-compile`'d, which is the case for `queue.el`, and as a consequence throws back `(wrong-type-argument queue #s(queue nil nil)` errors for a few tests that use queues indirectly. I have provisionally put a hack in to just force the correct tag symbols to be generated. Please do not merge as yet, I've only included here to make the build green. A new Emacs snapshot has been cut, waiting for the Silex to pick it up overnight at which point will remove this, see  https://github.com/Silex/docker-emacs/issues/86.

-----------------

Thanks!

### DIFF
--- a/test/cider-eldoc-tests.el
+++ b/test/cider-eldoc-tests.el
@@ -122,6 +122,7 @@
 (describe "cider-eldoc-beginning-of-sexp"
   (it "moves to the beginning of the sexp"
     (with-temp-buffer
+      (clojure-mode)
       (save-excursion
         (insert "(a (b b) (c c) d)"))
       (search-forward "d")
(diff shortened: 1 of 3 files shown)


bbatsov:
Nicely done! Thanks! 🙇‍♂️

Issue #2946, opened 2020-12-12
CIDER gets confused with both a clj and a cljs repl open


dpsutton:
CIDER gets confused when you have both a clj and a cljs repl open. Ideally evaluating code in a cljs file would always use the cljs repl and code evaluated in a clj file would always use the clj file. However, the bug is that CIDER attempts to use only the last repl your point was in. Meaning if your point was in the cljs repl and you attempt to evaluate cljs code, it works. But if point was last in the clj buffer nothing happens.

## Repo
```
.
├── deps.edn
├── shadow-cljs.edn
└── src
    ├── clj.clj
    └── cljs.cljs
```
### deps.edn
```clojure
{}
```
### shadow-cljs.edn
```clojure
{:source-paths ["src"]}
```

### src/clj.clj
```clojure
(ns clj)

(+ 1 1)
```

### src/cljs.cljs
```clojure
(ns cljs)

(+ 1 1)
```

## Starting the repls

Just `cider-jack-in` as normal. Then clojurescript, selecting shadow and the browser repl. Then put point in one of the repls and evaluate code in the corresponding source file. 

### point last in clj repl and evaluating in the clojure file
<img width="1904" alt="image" src="https://user-images.githubusercontent.com/6377293/101991026-a14ac380-3c6f-11eb-897a-85af0eca1c0d.png">

When moving the point into the clojurescript file and doing the inline evaluation, nothing happens.

# Suspected cause

This functionality hits `cider-interactive-eval` which uses `(cider-map-repls :auto ...)` This calls `(cider-repls type ensure))` with type `cljs` which returns no repls.

<img width="961" alt="image" src="https://user-images.githubusercontent.com/6377293/101991145-6ac17880-3c70-11eb-87c0-d52c0d2c48ce.png">

`(sesman-current-session 'CIDER)` returns only the clj repl.

The docstring of that function is 
> "Get the most relevant current session for the SYSTEM. CXT-TYPES is a list of context types to consider."

And that's our problem. There's no way to tell sesman what CIDER thinks is relevant (aka, the cljs repl). We can only ask sesman for what _it_ thinks is relevant, which is the most recently focused buffer.

```emacs-lisp
(defun sesman-current-session (system &optional cxt-types)
  "Get the most relevant current session for the SYSTEM.
CXT-TYPES is a list of context types to consider."
  (or (car (sesman--linked-sessions system 'sort cxt-types))   ;; car returns only the first "most" relevant
[...]

PR #3201, opened 2022-05-16
Allow combining all sesman sessions of a project

lycheese:
- Makes cider-repls return the combination of all sesman sessions
  associated with a project --> allows using multiple repls with different dependency files (e.g. clojure-cli for clj files and shadow-cljs for cljs files) and have CIDER still select the correct repl without having to move the point into the repl first (see issue referenced below)
- Adds the custom var `cider-merge-sessions` for toggling the new functionality to avoid changing the defaults and allow for making it project-specific via `.dir-locals.el`

Closes clojure-emacs/cider#2946.
-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -61,6 +61,19 @@ available) and the matching REPL buffer."
   :safe #'booleanp
   :package-version '(cider . "0.9.0"))
 
+;;;###autoload
+(defcustom cider-merge-sessions nil
+  "Controls session combination behaviour.
+
+Symbol `host' combines all sessions of a project associated with the same host.
+Symbol `project' combines all sessions of a project.
+
+All other values do not combine any sessions."
+  :type 'symbol
+  :group 'cider
+  :safe #'symbolp
+  :package-version '(cider . "1.5"))
+
 (defconst cider-required-nrepl-version "0.6.0"
   "The minimum nREPL version that's known to work properly with CIDER.")
 
@@ -862,6 +875,33 @@ no linked session or there is no REPL of TYPE within the current session."
           ((listp type) (member buffer-repl-type type))
           (t (string= type buffer-repl-type)))))
 
+(defun cider--get-host-from-session (session)
+  "Returns the host associated with SESSION."
+  (plist-get (cider--gather-session-params session)
+             :host))
+
+(defun cider--make-sessions-list-with-hosts (sessions)
+  "Makes a list of SESSIONS and their hosts.
+Returns a list of the form ((session1 host1) (session2 host2) ...)."
+  (mapcar (lambda (session)
+            (list session (cider--get-host-from-session session)))
+          sessions))
+
+(defun cider--get-sessions-with-same-host (session sessions)
+  "Returns a list of SESSIONS with the same host as SESSION."
+  (mapcar #'car
+          (seq-filter (lambda (x)
+                        (string-equal (cadr x)
+                                      (cider--get-host-from-session session)))
+                      (cider--make-sessions-list-with-hosts sessions))))
+
+(defun cider--extract-connections (sessions)
+  "Returns a flattened list of all session buffers in SESSIONS."
+  (cl-reduce (lambda (x y)
+               (append x (cdr y)))
+             sessions
+             :initial-value '()))
+
 (defun cider-repls (&optional type ensure)
   "Return cider REPLs of TYPE from the current session.
 If TYPE is nil or multi, return all REPLs.  If TYPE is a list of types,
@@ -871,9 +911,24 @@ throw an error if no linked session exists."
                ((listp type)
                 (mapcar #'cider-maybe-intern type))
                ((cider-maybe-intern type))))
-        (repls (cdr (if ensure
-                        (sesman-ensure-session 'CIDER)
-                      (sesman-current-session 'CIDER)))))
+        (repls (pcase cider-merge-sessions
+                 ('host
+                  (if ensure
+                      (or (cider--extract-connections (cider--get-sessions-with-same-host
+                                                       (sesman-current-session 'CIDER)
+                                                       (sesman-current-sessions 'CIDER)))
+                          (user-error "No linked %s sessions" 'CIDER))
+                    (cider--extract-connections (cider--get-sessions-with-same-host
+                                                 (sesman-current-session 'CIDER)
+                                                 (sesman-current-sessions 'CIDER)))))
+                 ('project
+                  (if ensure
+                      (or (cider--extract-connections (sesman-current-sessions 'CIDER))
+                          (user-error "No linked %s sessions" 'CIDER))
+                    (cider--extract-connections (sesman-current-sessions 'CIDER))))
+                 (_ (cdr (if ensure
+                             (sesman-ensure-session 'CIDER)
+                           (sesman-current-session 'CIDER)))))))
     (or (seq-filter (lambda (b)
                       (cider--match-repl-type type b))
                     repls)
--- a/doc/modules/ROOT/pages/usage/managing_connections.adoc
+++ b/doc/modules/ROOT/pages/usage/managing_connections.adoc
@@ -71,10 +71,22 @@ You can add new REPLs to the current session with:
 
 A very common use-case would be to run `cider-jack-in-clj` for some project and then follow up with `cider-connect-sibling-cljs`.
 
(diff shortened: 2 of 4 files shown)


lycheese:
I have used this configuration for the past week and haven't noticed any problems so far. I tried replacing `(flatten-list (mapcar #'cdr (sesman-current-sessions 'CIDER)))` with `(mapcan #'cdr (sesman-current-sessions 'CIDER))`, but this somehow duplicated one repl into another session multiple times leading to all commands being evaluated multiple times. Not sure how it managed to do that.

lycheese:
Added tests.

bbatsov:
This will also have to be documented, likely in the connection management section.

lycheese:
> This will also have to be documented, likely in the connection management section.

I added a short explanation and warning to the note on separate sessions. Not sure if it should be more prominently featured, but since it is quite an ugly hack it probably shouldn't.

lycheese:
Added custom var `cider-combine-sesman-sessions-per-project-per-host` which should allow the most common uses for having multiple sessions for a single project. This should be quite easy to extend to allow for custom combination logic (e.g. only combining sessions with certain ports), but I'm not sure if that will be necessary.

lycheese:
@bbatsov Could I get a review on this?

bbatsov:
@lycheese Yep, totally. I waited to cut 1.4.1 before reviewing more carefully any new features. I'll take a closer look later today or tomorrow.

doc/modules/ROOT/pages/usage/managing_connections.adoc
```diff
@@ -71,11 +71,33 @@ You can add new REPLs to the current session with:
...
+Unless there are both Clojure and ClojureScript REPLs in the same
+session smart-dispatch commands (e.g. evaluate the code in the right
+Clojure/ClojureScript REPL, toggle between Clojure and ClojureScript REPL) won't
+work. A very common problem newcomers experience is to create a Clojure REPL and
+a ClojureScript REPL in separate sessions and wonder why those are not
 interacting properly with one another.
 
+In the case of using separate config files for the clj and cljs dependencies
+(e.g. clj dependencies in `deps.edn` and cljs dependencies in `shadow-cljs.edn`)
+it is currently impossible to group those two repls in the same session.
```
bbatsov, line 85: Is it really impossible? It's definitely not possible with jack-in, but it should be possible when using `cider-connect-*` commands.

cider-connection.el
```diff
@@ -61,6 +61,22 @@ available) and the matching REPL buffer."
   :safe #'booleanp
   :package-version '(cider . "0.9.0"))
 
+;;;###autoload
+(defcustom cider-combine-sesman-sessions-per-project nil
+  "Whether to treat all sesman sessions associated with a project as one session."
+  :type 'boolean
+  :group 'cider
+  :safe #'booleanp)
```
bbatsov, line 69: Don't forget to add `:package-version` properties to both defcustoms.

cider-connection.el
```diff
@@ -61,6 +61,22 @@ available) and the matching REPL buffer."
   :safe #'booleanp
   :package-version '(cider . "0.9.0"))
 
+;;;###autoload
+(defcustom cider-combine-sesman-sessions-per-project nil
```
bbatsov, line 65: I'm not sure I like this name. Perhaps `cider-merge-(all)-project-sessions` would be better, but there's some ambiguity to it as well.

cider-connection.el
```diff
@@ -61,6 +61,22 @@ available) and the matching REPL buffer."
...
+  "Whether to treat all sesman sessions associated with a project as one session."
+  :type 'boolean
+  :group 'cider
+  :safe #'booleanp)
+
+;;;###autoload
+(defcustom cider-combine-sesman-sessions-per-project-per-host nil
```
bbatsov, line 72: I definitely don't like this name as well. I mean it's descriptive, but it sounds like some Java variable from an enterprise project to me. 😆

cider-connection.el
```diff
@@ -867,18 +883,54 @@ no linked session or there is no REPL of TYPE within the current session."
...
-               ((cider-maybe-intern type))))
-        (repls (cdr (if ensure
-                        (sesman-ensure-session 'CIDER)
-                      (sesman-current-session 'CIDER)))))
-    (or (seq-filter (lambda (b)
-                      (cider--match-repl-type type b))
-                    repls)
-        (when ensure
-          (cider--no-repls-user-error type)))))
+  (cl-flet* ((get-host-from-session
```
bbatsov, line 886: I'm not a big fan of `cl-flet*` as it makes debugging a bit harder and it obscures what a function does with all the inner definitions in it. I'd just make all of those "private" functions, defined right before they are used.

bbatsov:
A few related thoughts:

- I'm pondering if won't be easier for such cases to just disregard sessions altogether (similar to the old behavior pre-sesman). I wonder how often people work on multiple projects at the same time in general, and my guess is that this doesn't happen very often. 
- I'm also wondering if we should have one defcustom with multiple symbol options instead of several boolean flags.

doc/modules/ROOT/pages/usage/managing_connections.adoc
lycheese: If I do:
1. `cider-jack-in-clj` (with `clojure-cli`)
2. `npx shadow-cljs server` in `vterm`
3. `cider-connect-cljs` to `localhost:<nrepl port>`

Then `sesman-info` returns two sessions, one for clj, one for cljs.

cider-connection.el
lycheese: `1.5`?

bbatsov: Yep.

bbatsov:
@vspinu Feel free to share your thoughts on this as well.

cider-connection.el
lycheese: Fixed in newest version. Not sure if the docstrings are expressive enough. Thoughts?

lycheese:
>     * I'm pondering if won't be easier for such cases to just disregard sessions altogether (similar to the old behavior pre-sesman). I wonder how often people work on multiple projects at the same time in general, and my guess is that this doesn't happen very often.

At least in my case I sometimes work on 2 projects. One sort of scratch project to test stuff that I can kill quickly and add dependencies. And another with the actual production code. So, I think that the project isolation is quite useful (as well as host isolation).

>     * I'm also wondering if we should have one defcustom with multiple symbol options instead of several boolean flags.

I agree that that's probably more elegant. I'll try doing that. Maybe I'll also be inspired and find a less stupid name for it.

lycheese:
I combined the two options into `cider-merge-sessions` with the options being `:host`, `:project` or anything else for the default behaviour.

cider-connection.el
```diff
@@ -61,6 +61,19 @@ available) and the matching REPL buffer."
   :safe #'booleanp
   :package-version '(cider . "0.9.0"))
 
+;;;###autoload
+(defcustom cider-merge-sessions nil
+  "Controls session combination behaviour.
+
+`:host' combines all sessions of a project associated with the same host.
```
bbatsov, line 68: Using symbols (e.g. `'host`) is more idiomatic in Emacs Lisp.

lycheese: Done. Is there a better way to quote symbols in elisp docstring than just `` `symbol' `` (to get `'symbol` instead of `symbol` in the describe-variable)?

lycheese: Oh, of course the manual has a section on this:
> If a symbol has a function definition and/or a variable definition, but those are irrelevant to the use of the symbol that you are documenting, you can write the words ‘symbol’ or ‘program’ before the symbol name to prevent making any hyperlink. 
[Source](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html)

doc/modules/ROOT/pages/usage/managing_connections.adoc
bbatsov: Yeah, my bad. Only the `connect-sibling` commands attach to the existing session, but I think this also requires the use of the same nREPL server for all connections.

cider-connection.el
bbatsov: They look good to me.

bbatsov:
I think the PR's in a good shape and we can merge at this point. Thanks!

lycheese:
@bbatsov I just realized that the changelog entry for this is listed under 1.4.1 and not master. Sorry about that.

bbatsov:
No worries. I'll fix this in a bit.

Issue #3194, opened 2022-05-03
Unexpected cider-undef-all outcome wrt. java.lang classes


codeasone:
I've hit upon an issue with `cider-undef-all`

Minimal context:

    {:paths ["src"]
     :deps  {org.clojure/clojure {:mvn/version "1.10.3"}}}

with `src/core.clj`

    (ns core)
    
    (print (System/getenv "HOME"))

Custom binding I established to evaluate the new `undef-all` functionality:

    :bind (:map cider-mode-map
                ("C-c C-k" . (lambda () (interactive)
                               (cider-undef-all)
                               (cider-load-buffer)
                               (message "Reset vars in namespace")))
                ...)

## Expected behavior

Invoking `C-c C-k` should `dwim` and I should see the message in the mini-buffer. 

## Actual behavior

Invoking `C-c C-k` gives rise to:

      Show: Project-Only All
      Hide: Clojure Java REPL Tooling Duplicates  (33 frames hidden)
    
    2. Unhandled clojure.lang.Compiler$CompilerException
       Error compiling src/core.clj at (3:8)
    
       No such namespace: System
    
                     Util.java:  221  clojure.lang.Util/runtimeException
                 Compiler.java: 7388  clojure.lang.Compiler/resolveIn
                 Compiler.java: 7362  clojure.lang.Compiler/resolve
                 Compiler.java: 7323  clojure.lang.Compiler/analyzeSymbol
                 Compiler.java: 6772  clojure.lang.Compiler/analyze
                 Compiler.java: 6749  clojure.lang.Compiler/analyze
                 Compiler.java: 3824  clojure.lang.Compiler$InvokeExpr/parse
                 Compiler.java: 7113  clojure.lang.Compiler/analyzeSeq
                 Compiler.java: 6793  clojure.lang.Compiler/analyze
                 Compiler.java: 6749  clojure.lang.Compiler/analyze
                 Compiler.java: 3892  clojure.lang.Compiler$InvokeExpr/parse
                 Compiler.java: 7113  clojure.lang.Compiler/analyzeSeq
                 Compiler.java: 6793  clojure.lang.Compiler/analyze
                 Compiler.java: 6749  clojure.lang.Compiler/analyze
                 Compiler.java: 6124  clojure.lang.Compiler$BodyExpr$Parser/parse
                 Compiler.java: 5471  clojure.lang.Compiler$FnMethod/parse
[...]

PR #3202, opened 2022-05-18
Fix broken cider-eval-ns-form

danskarda:
This pull request fixes bugs in `cider-eval-ns-form` introduced by recent addition of `undef-all` parameter:

1. There was a typo in `interactive` parameter. `"p"` is a number which evaluates to true even if it is 0. Parameter is used in `when` condition, so undef-all branch is always executed.

2. It uses magic value of `(match-string 0)`. For me it returns string `(ns` and undef fails. Probably regexp in `clojure-find-ns` was recently changed, which broke `cider-eval-ns-form` without anybody noticing. I do not think it is a good idea to rely on magic side-effects of previous functions, so I replaced match-string with return value of `cider-find-ns`.

There is a last issue with undef in general. It undefs symbols of classes which are imported by default (eg `Throwable`). But I guess this is more an issue of nrepl middleware.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 * [#3195](https://github.com/clojure-emacs/cider/issues/3195): Revert the change that resulted in `(error "Cyclic keymap inheritance")` on `cider-test-run-test`.
 * [#3182](https://github.com/clojure-emacs/cider/issues/3182): Don't try to invoke
 JVM-specific code outside of JVM Clojure.
+* [#3202](https://github.com/clojure-emacs/cider/pull/3202): Fix `cider-eval-ns-form`
+  * Do not always perform `undef-all`. Undef only with `C-u` prefix.
+  * Fix extraction of namespace name.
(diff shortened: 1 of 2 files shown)


danskarda:
Changelog updated.

yuhan0:
Sorry about introducing so many bugs! I've just checked the other functions affected by that PR and they seem to be fine, it's just `cider-eval-ns-form` which escaped my attention because I've never used it.

The last point about symbols like Throwable being undef'ed was brought up in #3194, unfortunately it requires a update to cider-nrepl version after the fix is merged.

bbatsov:
No worries! @danskarda Thanks for fixing this!

bbatsov:
(I'll likely cut CIDER 1.4.1 after we merge the `undef-all` cider-nrepl update)

vemv:
Done https://github.com/clojure-emacs/cider/blob/v1.4.1/CHANGELOG.md#141-2022-05-25

PR #3203, opened 2022-05-18
Use cider-nrepl 0.28.4

vemv:
https://github.com/clojure-emacs/cider-nrepl/blob/v0.28.4/CHANGELOG.md#0284-2022-05-18
https://clojars.org/cider/cider-nrepl/versions/0.28.4


bbatsov (approved):
Looks good.

PR #3208, opened 2022-05-28
Style thread-first/last with new Emacs 28 indent format

### LINKED ISSUE #3204: thread-first/last indentation change in Emacs 28 breaking the linter
The linter complains about `thread-first` argument indentation in a recent commit:

```
cider-client.el:600:0 (indent)   !                                    (cider-sync-tooling-eval)
cider-client.el:601:0 (indent)   !                                    (nrepl-dict-get "value")
cider-client.el:602:0 (indent)   !                                    read))
Found 3 warnings in file `cider-client.el'
```

This is because the indentation method has changed for `thread-first` and `thread-last` in Emacs 28 with [this commit](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=2a736738095c313ccef07d074aac4c5467b750e0) from `(indent 1)` to `(indent 0)`, and as a result their arguments are indented under the first argument, rather than the two spaces after the function:

auto indentation, before `(indent 1)`:
```emacs-lisp
(thread-first "(seq (.split (System/getProperty \"java.class.path\") \":\"))"
  (cider-sync-tooling-eval)
  (nrepl-dict-get "value")
  read)
```
after, Emacs 28, `(indent 0)`:
```emacs-lisp
(thread-first "(seq (.split (System/getProperty \"java.class.path\") \":\"))"
              (cider-sync-tooling-eval)
              (nrepl-dict-get "value")
              read)
```

There's a setting in `.dir-locals.el` for the linter to check the correct thread- indentation is at `(ident 1)`:
https://github.com/clojure-emacs/cider/blob/e86f2f74f65eaa3aa2dd088ad2e0b1426a04c3e1/.dir-locals.el#L30

any new `thread-*` calls written in Emacs 28 though, are most likely going to be indented with `(indent 0)` since this is the default.

As such, we have our first break with a recent check in identified by the linter (they are now indented under the first argument rather than two spaces after the function name):
```
cider-client.el:600:0 (indent)   !                                    (cider-sync-tooling-eval)
cider-client.el:601:0 (indent)   !                                    (nrepl-dict-get "value")
cider-client.el:602:0 (indent)   !                                    read))
Found 3 warnings in file `cider-client.el'
```

Shall I go ahead and change the `.dir-locals.el` `elisp-lint-indent-spaces` for `thread-first` and `thread-last` to be indent 0 (Emacs 28 setting), as well change every `thread-first` and `thread-last` call in CIDER to conform to this new indentation format? I much prefer the Emacs 27 format, but I don't think there is much we can do going forward ...

## Expected behavior

Linter should report any errors when checking the indentation for `thread-first`, `thread-last`

## Actual behavior

Linter complains about indentation of recent `thread-first` call.

run `eldev lint`

### CIDER version information

cider master

### Emacs version

Emacs 28

### Operating system

Ubuntu 20.04

ikappaki:
Hi,

could you please consider patch to change the indentation style for `thread-first` and `thread-last` to `(indent 0)` from `(indent 1)`  throughout the cider code? (fixes #3204). These are non-functional changes.

An [Emacs 28 change](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=2a736738095c313ccef07d074aac4c5467b750e0) has switched the indentation style for these two functions to `0`, and this means that any new code with `thread-*` is most likely to be formatted using the now default style `0` while the user is typing the code.

The linter though is instructed in `.dir-locals.el` to look for style `1`, and as such it will issue a warning, though the user will have no clue where the indentation warning is coming from (as far as they are concerned Emacs 28 has formatted the sexp fine).

With this patch, we instruct the linter to switch to style `0`, and also make all the necessary changes in the cider code to conform to that style. It turns out this is a lot of indentation changes to the existing code.

The alternative is to keep the old style `1` when indenting the code, this will mean Emacs will indent the functions with style `1`, and thus the linter will be happy. This can be done by just adding the following line to `.dir-locals.el` (just showing thread-first here):

```
(eval . (put 'thread-first 'lisp-indent-function 1))
```

Let me know what you think, I prefer more the current patch so that we are compatible with latest Emacs changes (although I don't like there was a sudden switch to `0` out of the blue).

-----------------

Thanks!

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -27,8 +27,8 @@
                               (insert-rect . defun)               ;; cl-flet
                               (cl-defun . 2)
                               (with-parsed-tramp-file-name . 2)
-                              (thread-first . 1)
-                              (thread-last . 1)))))
+                              (thread-first . 0)
+                              (thread-last . 0)))))
 
 ;; To use the bug-reference stuff, do:
 ;;     (add-hook 'text-mode-hook #'bug-reference-mode)
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -129,7 +129,8 @@ Display TITLE at the top and SPECS are indented underneath."
       (insert (cider-propertize title 'emph) "\n")
       (dolist (spec-name specs)
         (insert (propertize "  " 'spec-name spec-name))
-        (thread-first (cider-font-lock-as-clojure spec-name)
+        (thread-first
+          (cider-font-lock-as-clojure spec-name)
           (insert-text-button 'type 'cider-browse-spec--spec)
           (button-put 'spec-name spec-name))
         (insert (propertize "\n" 'spec-name spec-name)))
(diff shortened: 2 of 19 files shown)


bbatsov:
Thanks!

PR #3212, opened 2022-06-04
clojure-cli: provide a mechanism to configure default aliases

pyr:
The clojure CLI does not impose any convention on what the default
alias for a project should be. It is also common practice to
only have the necessary environment for production in the
root dependency expressions.

As a result, jacking in without modification often results in
a REPL environment that is too constrained.

This patch looks up the current `deps.edn` file to look for a
`:cider/default-aliases` key, when that key is present and
contains a collection, it is processed to produce the correct
alias stanza for the clojure CLI commandline.

- [X] Commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -615,6 +615,15 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   "Removes the duplicates in DEPS."
   (cl-delete-duplicates deps :test 'equal))
 
+(defun cider-clojure-cli-jack-in-user-supplied-aliases ()
+  "Load the current `deps.edn` to look for default aliases.
+In the file, if a `:cider/default-aliases` keyword is found, use the
+corresponding value to build an aliases list."
+  (thread-last
+    (with-temp-buffer (insert-file-contents "deps.edn") (buffer-string))
+    (parseedn-read-str)
+    (gethash :cider/default-aliases)))
+
 (defun cider-clojure-cli-jack-in-dependencies (global-options _params dependencies)
   "Create Clojure tools.deps jack-in dependencies.
 Does so by concatenating DEPENDENCIES and GLOBAL-OPTIONS into a suitable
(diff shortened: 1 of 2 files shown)


pyr:
I am in the process of figuring out how to run tests since my file-mocking macro does not seem to make eldev happy for now. Before going further I wanted a bit of feedback on the idea and whether it was worth following up on it.

plexus:
For what it's worth, we've been using `.dir-locals.el` for this for a long time, works like a charm. This has the benefit that you can set up other project-specific config, e.g. we sometimes have a non-standard cljs-repl setup.

E.g.

```lisp
((nil . ((cider-clojure-cli-global-options . "-A:dev:test"))))
```

or

```lisp
((nil . ((cider-clojure-cli-global-options     . "-A:dev:test")
         (cider-custom-cljs-repl-init-form     . "(user/shadow-cljs-repl)")
         (cider-default-cljs-repl              . custom)
         (cider-preferred-build-tool           . clojure-cli)
         (eval . (progn
                   (make-variable-buffer-local 'cider-jack-in-nrepl-middlewares)
                   (add-to-list 'cider-jack-in-nrepl-middlewares "shadow.cljs.devtools.server.nrepl/middleware"))))))
```

pyr:
Hi @plexus, thinking about it, your approach is probably the best, I'll go with that, thanks!
Something like my patch would only make sense if most tooling agreed to honor it, and then `:cider/` would be the wrong namespace, I'll drop it.

Cheers

vemv:
btw, if you think the doc could be improved somewher in this regard, a suggetion would be much welcome.

I believe people often trip up with clj CLI aliases.

bbatsov:
We can always add something here https://docs.cider.mx/cider/basics/up_and_running.html#customizing-the-jack-in-command-behavior or here https://docs.cider.mx/cider/basics/up_and_running.html#customizing-the-jack-in-command-behavior

Not sure if someone will find it, though. :-)

PR #3213, opened 2022-06-05
Allow ignoring top-level comments in eval-defun-up-to-point

lycheese:
Adds custom var cider-eval-defun-up-to-point-in-comment to make
cider-eval-defun-up-to-point ignore top-level comments and treat
the form before point as top-level.

To try it out:
- Set `cider-eval-defun-up-to-point-in-comment` (suggestions for a better name welcome) to `t`
- Go to a comment form and put point at the `^`
```
(comment
  (* 5 25)
  (let [b "str"]
    (-> b
        keyword^
        name))
  *e)
```
- Execute `cider-eval-defun-up-to-point`
- Return value of `:str` instead of `nil`

Not sure how to write tests for this, but it seems none of the eval commands in `cider-eval.el` have tests so maybe that is okay?
I did not add a section to the user manual since the command itself only shows up in a table. I can write something up though if it is deemed necessary.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+## New features
+
+* Add `cider-eval-defun-up-to-point-in-comment` which when set to true makes `cider-eval-defun-up-to-point` ignore top-level comment blocks and treats the next-to-top-level form before point as top-level.
+
 ## 1.4.1 (2022-05-25)
 
 ## Changes
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -176,6 +176,13 @@ automatically written into this register."
   :group 'cider
   :package-version '(cider . "1.4.0"))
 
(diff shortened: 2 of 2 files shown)


yuhan0:
I believe this functionality is already present as `clojure-toplevel-inside-comment-form` - when set to `t` and executing the above snippet I get `:str`.

Maybe the docs could be improved to mention this - users would understandably expect to find it as a Cider configuration variable instead of clojure-mode.

lycheese:
> I believe this functionality is already present as `clojure-toplevel-inside-comment-form` - when set to `t` and executing the above snippet I get `:str`.
> 
> Maybe the docs could be improved to mention this - users would understandably expect to find it as a Cider configuration variable instead of clojure-mode.

And that happens if you write PRs late at night. Should have searched more carefully. Sorry for the noise.

PR #3214, opened 2022-06-06
Add tip to docs for evaluating in comments

lycheese:
Document clojure-toplevel-inside-comment-form for changing the
behaviour of defun-style commands.

Result of #3213.

@yuhan0 is this expressive enough or should I add a codeblock to demonstrate?

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/modules/ROOT/pages/usage/code_evaluation.adoc
+++ b/doc/modules/ROOT/pages/usage/code_evaluation.adoc
@@ -302,6 +302,44 @@ You can also use the command `cider-kill-last-result`(kbd:[C-c C-v k]) after any
 eval command to store its result in the kill ring. This works even when the
 `cider-eval-register` feature is disabled.
 
+=== Evaluating inside comments
+
+By default, when using the defun-style eval commands inside a comment the return value is always `nil`. Calling `cider-eval-defun-up-to-point` with point at `<point>` also returns `nil`:
+
+[source,clojure]
+----
+(comment
+  (let [b "str"]
+    (-> b
+        keyword<point> ;;=> nil
+        name))
+  *e)
+----
+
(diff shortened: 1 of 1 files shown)


doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -418,3 +418,8 @@ kbd:[C-u C-c C-c]
 |===
 
 TIP: You'll find all evaluation commands and their keybindings in the `CIDER Eval` menu.
+
+TIP: Setting `clojure-toplevel-inside-comment-form` to `t` makes
```
bbatsov, line 422: I'd probably move this to some dedicated section earlier in the document and add a couple of code examples, so it's easier to understand how this works in practice.

bbatsov:
There are some examples you can borrow from here https://camdez.com/blog/2020/07/03/interactively-evaluating-clojure-code-in-comments-in-emacs/ Maybe there's something in clojure-mode's README as well.

yuhan0:
I do wonder if the var should be set to true by default - it was labeled as "Experimental" 4 years ago but I've been using it all this time without any issues. I believe Calva has a similar feature enabled by default.

lycheese:
> I believe Calva has a similar feature enabled by default.

Calva explicitly states that it treats all forms in comments as top-level in the `Getting Started REPL`:

```
;; == Rich Comments Support ==
;; Repeating an important concept: Forms inside
;; `(comment ...)` are also considered top level
;; by Calva. Alt+Enter at different places below
;; to get a feel for it.
```

bbatsov:
Btw, It was labeled as "experimental" because we weren't certain how well it works, not because of plans to make it the default at some point.

lycheese:
I have been unsuccessful so far to get my asciidoc to generate any highlighted code blocks, but referencing others it should be ok the way it is.

I think the duplication of code blocks is fine to make the message more clear, but the section could be made more concise, if desired.

bbatsov:
Thanks!

PR #3215, opened 2022-06-13
Consecutive result overlays can now seen

pestctrl:
This is a minor bug that has just bothered me for years, and I've finally decided to fix it. 

When I run cider-eval-last-sexp twice, the second overlay never shows up. This is because we're letting the hook that's supposed to kill overlay #1 also kill overlay #2. When creating a second overlay right after the first one, we already remove overlay #1 in #'cider--make-result-overlay, and thus we can safely remove #'cider--remove-result-overlay from the post-command-hook. 

-----------------

Previously, if multiple overlays were created one after another that
are meant to last for one command, every second overlay wouldn't show
up. This is because there's a hook in the post-command-hook that will
delete all overlays, and it runs right after the second overlay is
created.

Instead, if we're creating an overlay as we speak (ie. we are in the
function 'cider--make-result-overlay), we already know that the
previous overlay has been deleted, and thus we can remove
'cider--remove-result-overlay from the 'post-command-hook.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -129,7 +129,7 @@ PROPS is a plist of properties and values to add to the overlay."
     (push #'cider--delete-overlay (overlay-get o 'modification-hooks))
     o))
 
-(defun cider--remove-result-overlay (&rest _)
+(defun cider--remove-result-overlay (&optional beg end &rest _)
   "Remove result overlay from current buffer.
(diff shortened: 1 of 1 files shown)


pestctrl:
I will write some tests this weekend, and re-submit the pull request.

PR #3217, opened 2022-06-22
[Fix #3200] Add private var listing below public vars

### LINKED ISSUE #3200: Enable cider-ns-browser to show private vars via toggle
**Problem**: It seems good practice to mark your non-public vars as `^:private` (including fns from `defn-`). But then `cider-ns-browser` (a very useful view) doesn't show them. And it's still useful to document private vars.

**Solution**: Enable the private vars to be shown, with their docstrings. Could be a toggle command like `a` (for "show All") in the map (https://github.com/clojure-emacs/cider/blob/master/cider-browse-ns.el#L53).

Should probably still _not_ show vars marked as `^:deprecated`.

zkry:
This PR modifies the behavior of `cider-browse-ns` command to display the private vars below the public ones.  This is accomplished by sending an additional nrepl request to fetch only the private vars, and then having the display function (cider-browse-ns--list) be able to optionally add this section for the private vars.

The following are points I wasn't entirely sure about:
- I wasn't sure the best way to format the output data. I'm definitely open to suggestions for how to make this look better.
- I ended up only able to get this to work by making a second nrepl request. I couldn't find a way to shape the query to get all the items from the namespace with the additional private metadata in the response.

Screenshot of feature:
Screen Shot

-----------------

- [ ] N/A You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)


bbatsov:
I think that ideally we should have some configuration settings about this - e.g. alphabetical listing, visibility grouping listing (what you propose),  type grouping listing (var, function, macro, etc).

cider-repl.el
```diff
@@ -784,6 +784,29 @@ Before inserting, run `cider-repl-preoutput-hook' on STRING."
       (with-selected-window window
         (scroll-down 1)))))
 
+(defun cider-repl-syntax-propertize (beg end)
```
bbatsov, line 787: This seems a bit unrelated to me. Perhaps part of something else you were working on?

cider-browse-ns.el
```diff
@@ -44,6 +44,17 @@
...
 
+
+(defgroup cider-browse-ns nil
+  "Display contents of namespaces for CIDER."
+  :prefix "cider-browse-ns-"
+  :group 'cider)
+
+(defface cider-browse-ns-info-face
+  '((t (:inherit shadow)))
+  "Face for displaying headers for sections of browse-ns buffer."
```
bbatsov, line 55: I think you meant to write "heading" here.

cider-browse-ns.el
```diff
@@ -44,6 +44,17 @@
 (require 'easymenu)
 (require 'thingatpt)
 
+
+(defgroup cider-browse-ns nil
+  "Display contents of namespaces for CIDER."
+  :prefix "cider-browse-ns-"
+  :group 'cider)
+
+(defface cider-browse-ns-info-face
```
bbatsov, line 53: info-face -> heading-face

bbatsov:
Btw, we currently don't display private items at all, right? If some I guess the config should be different from what I had initially suggested, but I'm still thinking this should be configurable to some extent. The ns browser was something I had big plans for, but very little time and it remained quite basic.

cider-repl.el
zkry: oh yeah, something else I was working on 😅 . Don't know how that got in here

zkry:
Yeah, currently there are no private items being displayed. I agree about the configuration. It would be a nice feature to have it configurable. I can see how the ns browser could be extended with features like grouping by type, and hiding/showing various types. I'd be more than willing to extend the scope of this PR to get it closer to what your longer-term vision is. What do you think would be a good user interface for configuring this? I assume a static configuration (for default layout) would be there and I could also see a more dynamic (magit transient logs like) UI to switch up how the buffer to different criteria on the fly.

zkry:
I see. Yeah, I think such an interface would work well for for this too. Tests could be another category that could be toggled. Any other toggleable category that you can think of? specs as well maybe?
I'll try my hand at this and see what I can do.

bbatsov:
I guess we can also have functions, "value vars" and types. I don't mind starting small, though - we don't need every possible category right away and we can always add more as we go.

zkry:
So I took inspiration from the stacktrace feature you mentioned and got in a first implementation of the browse-ns feature with various filtering and grouping options available. Here's a quick demo to show how it looks.

https://user-images.githubusercontent.com/20138116/175830284-9492de34-c5f9-4b9a-9880-29cf8b7b2bf8.mov

Also with this work, a bug where in cider-browse-ns-all, pressing RET on a namespace that doesn't have a '.' in its name doesn't recognize it as a namespace.

bbatsov:
The demo looks great!

bbatsov:
I guess we can merge your work at this point, and you can continue building upon this in small increments from here. Thanks for tackling this! Great work! 🙇‍♂️

PR #3224, opened 2022-07-21
`cider-clojure-cli-aliases` should be a string value that need to be explicitly stated

vimfun:
**I tried several times** until acknowledge that the value of `cider-clojure-cli-aliases` should be surrounded by `"` character.

So I guess that will be better to modify the doc of the custom variable. 

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -158,7 +158,7 @@ default to \"powershell\"."
 (defcustom cider-clojure-cli-aliases
   nil
   "A list of aliases to include when using the clojure cli.
-Alias names should be of the form `:foo:bar`.
+Alias names should be of the form `\":foo:bar\"`.
 Leading \"-A\" \"-M\" \"-T\" or \"-X\" are stripped from aliases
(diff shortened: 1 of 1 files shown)


cider.el
```diff
@@ -158,7 +158,7 @@ default to \"powershell\"."
 (defcustom cider-clojure-cli-aliases
   nil
   "A list of aliases to include when using the clojure cli.
-Alias names should be of the form `:foo:bar`.
+Alias names should be of the form `\":foo:bar\"`.
```
bbatsov, line 161: Those backticks should be removed.

vimfun: Yes. 😄

vimfun: > Those backticks should be removed.

Yes. 😄

vemv:
Thanks much!

vimfun:
> Thanks much!

My pleasure 😄

Issue #3019, opened 2021-05-24
completion at point ignores namespace prefix


jaor:
## Expected behavior

completion works as usual, across imported namespaces

## Actual behavior

when calling completion-at-point, only completions belonging to the current namespace are offered
(after C-c C-k to compile and load the namespace). this started happening very recently.

## Steps to reproduce the problem

for me, it fails with any namespace i try.  just C-c C-k and then M-x completion-at-point with point after
a ':as' prefix.  the prefix seems to be ignored, and only current namespace names are offered.

### CIDER version information

```
;; CIDER 1.2.0snapshot (package: 20210524.832), nREPL 0.8.3
;; Clojure 1.10.1, Java 11.0.11
```

### Lein/Boot version

Lein 2.9.1

### Emacs version

27.2 (but also happens with 28)

### Operating system

Debian unstable

bbatsov:
This code hasn't been changed in years, so I'm struggling to see what might have caused it for you. It'd be useful if you added some troubleshooting info here - e.g. the nREPL message exchange log.

jaor:
On Tue, May 25 2021, Bozhidar Batsov wrote:

> This code hasn't been changed in years, so I'm struggling to see what
> might have caused it for you. It'd be useful if you added some
> troubleshooting info here - e.g. the nREPL message exchange log.

Hi Bozhidar... i'm struggling too.  After discarding emacs 28, i at
first thought it could be related to using corfu instead of company for
region completions, but disabling both doesn't make any difference, and
i cannot think of anything directly related in my config that might have
changed.  I was hoping someone else was having the same problem.  I'll
try to find a short reproducer.  Thanks for chiming in!

jaor:
i think i've found the cause of the problem.  i was playing with completion-styles.  its default is '(basic partial-completion emacs22).  if one changes it to, say, '(partial-completion) (or, in my case, '(orderless)), completion stops working.  it seems basic is needed (or something similar).  fwiw, other capf like eshell's or elisp's don't seem to have this limitation.  things also work if basic is combined with other styles (e.g., i'm using now '(orderless basic) without trouble).

bbatsov:
Might be related to https://github.com/clojure-emacs/cider/issues/3006

vkz:
Just ran into the exact same problem. Confirmed on Emacs 28.0.50; cider 1.2.0
[...]

Issue #3006, opened 2021-04-18
Cider completion style is invalid


minad:
Hi Bozhidar!

Cider adds a custom cider completion style to the `completion-styles-alist`. However the entry that is being added does not follow the required specification of completion-styles. This issue has been found by @manuel-uberti (see https://github.com/minad/corfu/issues/8 for the original issue). I wonder how cider gets away with the current implementation. Does the default `completion-at-point` function work with Cider?

Daniel

----

~~~ elisp
(defun cider-company-unfiltered-candidates (string &rest _)
  "Return CIDER completion candidates for STRING as is, unfiltered."
  (cider-complete string))

(add-to-list 'completion-styles-alist
             '(cider
               cider-company-unfiltered-candidates
               cider-company-unfiltered-candidates
               "CIDER backend-driven completion style."))
~~~

* The first function should follow the calling convention of `completion-try-completion`, it should return t, nil or a pair (newstr.newpoint)
* The second function should follow `completion-all-completions` (I think it follows the spec, but you are not returning a base size in the last cdr which is okay I guess?)

bbatsov:
> Does the default completion-at-point function work with Cider?

Yeah, it does. This code is one of the oldest parts of CIDER and has seen minimal changes in the past 9 years, so I'm assuming that some APIs might have changed in Emacs. Is there some documentation on the topic that I can peruse?

minad:
The docstrings of `completion-try-completion` and `completion-all-completions` should have all the necessary information.

bbatsov:
I see. So basically we need some `cider-completion-try-completion` function alongside the existing one. I guess I'll also have to check what are the benefits/use-cases of all the params we've suppressed for `cider-company-unfiltered-candidates`, as I see the actual definition has a lot of params:

```
(completion-all-completions STRING TABLE PRED POINT &optional METADATA)
```

bbatsov:
Btw, I've also forgotten what's the difference between `completion-at-point-functions` and `completion-styles-alist`. Perhaps @dgutov can help by suggesting what'd be the appropriate changes.

dgutov:
@bbatsov Completions functions return some data from their own set, completion styles filter the data provided by the former.

I haven't written any completion styles myself, though, so I'm not going to help with the exact diff.

bbatsov:
I was mostly wondering if those completion styles have anything to do with `company-mode`, but I guess I'll figure this out myself when I find a bit of time.

bbatsov:
[...]

PR #3226, opened 2022-07-25
Populate completions with metadata

toniz4:
This makes possible to override the cider completion style in `completion-category-defaults` or `completion-category-overrides`.

This will simplify the workarounds needed to make cider work with `orderless` set as the "global" completion style. like seen in https://github.com/clojure-emacs/cider/issues/3019#issuecomment-1024898789. With this pull request, the following will work. 

```elisp
(setq completion-category-defaults '((cider (styles basic)))
```

The code is basicaly the same as the function returned by `completion-table-dynamic`, but simplified since we don't need to check `switch-buffer`, since is always `nil` in this context. And when `(eq action 'metadata)`, return the metadata. 

completion-table-dynamic, in minibuffer.el
```elisp
(defun completion-table-dynamic (fun &optional switch-buffer)
  (lambda (string pred action)
    (if (or (eq (car-safe action) 'boundaries) (eq action 'metadata))
        ;; `fun' is not supposed to return another function but a plain old
        ;; completion table, whose boundaries are always trivial.
        nil
      (with-current-buffer (if (not switch-buffer) (current-buffer)
                             (let ((win (minibuffer-selected-window)))
                               (if (window-live-p win) (window-buffer win)
                                 (current-buffer))))
        (complete-with-action action (funcall fun string) string pred)))))
```
The implementation in the pull request
```elisp
(lambda (string pred action)
              (cond ((eq action 'metadata) `(metadata (category . cider)))
                    ((eq (car-safe action) 'boundaries) nil)
                    (t (with-current-buffer (current-buffer)
                         (complete-with-action action
                                               (cider-complete string) string pred)))))
```

I've been using this implementation for 2 days with `advice-add`, experienced no problems.

Let me know if its needed to add any documentation around this .

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
@dgutov Can you take a look at the proposed changes as an expert in the API in question?

cider-completion.el
```diff
@@ -237,12 +237,18 @@ performed by `cider-annotate-completion-function'."
...
+                    (t (with-current-buffer (current-buffer)
+                         (complete-with-action action
+                                               (cider-complete string) string pred)))))
             :annotation-function #'cider-annotate-symbol
             :company-kind #'cider-company-symbol-kind
             :company-doc-buffer #'cider-create-doc-buffer
             :company-location #'cider-company-location
-            :company-docsig #'cider-company-docsig))))
+            :company-docsig #'cider-company-docsig
+            :exclusive 'no))))
```
bbatsov, line 251: What does this property mean?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New features
 
+- [#3226](https://github.com/clojure-emacs/cider/pull/3226): Populate completions metadata, making it possible to change the style of completion via `completion-category-override` or `completion-category-defaults`.
+
```
bbatsov, line 8: Please kill the blank line.

cider-completion.el
toniz4: as stated in https://www.gnu.org/software/emacs/manual/html_node/elisp/Completion-in-Buffers.html

> :exclusive If the value is no, then if the completion table fails to match the text at point, completion-at-point moves on to the next function in completion-at-point-functions instead of reporting a completion failure.

dgutov: I'm going to quote a smart person here:

>  Better stay far away from ":exclusive 'no".
> It's an ugly hack that's fundamentally broken (not just in its implementation).

https://lists.gnu.org/archive/html/emacs-devel/2022-04/msg00814.html

dgutov:
The `meta -> (categiry . cider)` part seems sound.

Though I don't know if it's a good idea to create a category per completion source. Perhaps @minad or @monnier have some opinion.

cider-completion.el
```diff
@@ -237,12 +237,18 @@ performed by `cider-annotate-completion-function'."
     (when (and (cider-connected-p)
                (not (or (cider-in-string-p) (cider-in-comment-p))))
       (list (car bounds) (cdr bounds)
-            (completion-table-dynamic #'cider-complete)
+            (lambda (string pred action)
+              (cond ((eq action 'metadata) `(metadata (category . cider)))
```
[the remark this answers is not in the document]
toniz4: As the docs says, the completion function should accept 3 args, one of them being a flag describing the complete operation, and that can have some values described at https://www.gnu.org/software/emacs/manual/html_node/elisp/Programmed-Completion.html

In the implementation in master, when action is `'metadata` or `'boundaries` it evaluates to `nil`, and when action is `nil`, `t` or `lambda`, action is passed to `(complete-with-action)`.

cider-completion.el
toniz4: Fair enough, it was working well without :exclusive 'no either way.

monnier:
Dmitry Gutov [2022-07-25 09:30:16] wrote:
> The `meta -> (categiry . cider)` part seems sound.

[ Assuming it really says `category` instead.  ]

> Though I don't know if it's a good idea to create a category per completion
> source.  Perhaps @minad or @monnier have some opinion.

The completion source is probably not a good choice of "category" since
the category should describe the kind of things that are completed
rather than how we found those things.

But I can imagine sound reasons to want to pay attention to the "how",
e.g. because some completion styles work poorly with particular backends.

[ Reminds me that I suspect the notion of "category" will probably need to
  grow some kind of hierarchy (or something) at some point.  ]


        Stefan

bbatsov:
So can someone give me some examples of categories? I get this should be the implementation (e.g. CIDER), so what is it supposed to be?

toniz4:
> So can someone give me some examples of categories? I get this should be the implementation (e.g. CIDER), so what is it supposed to be?

you can see that with describe-variable completion-category-defaults, by default it comes pre-populated with some categories.

I agree with @monnier, but this is the only way that i know to get a finer control of the completion style, i.e. only apply a specific style for completions comming from cider.

Eglot and lsp-mode follows the same idea:

https://github.com/emacs-lsp/lsp-mode/blob/b2dc84128fc6b2cd0460569bae6d0451856d2e0b/lsp-completion.el#L514

https://github.com/joaotavora/eglot/blob/6cc6fcc0fabb51671980e52110ad74f2e691e387/eglot.el#L2613

Eglot sets the category as eglot, and lsp-mode sets it to lsp-capf

cider-completion.el
```diff
@@ -237,7 +237,12 @@ performed by `cider-annotate-completion-function'."
     (when (and (cider-connected-p)
                (not (or (cider-in-string-p) (cider-in-comment-p))))
       (list (car bounds) (cdr bounds)
-            (completion-table-dynamic #'cider-complete)
+            (lambda (string pred action)
```
bbatsov, line 240: I think `prefix` is a better name then `string` here.

cider-completion.el
```diff
@@ -237,7 +237,12 @@ performed by `cider-annotate-completion-function'."
...
+              (cond ((eq action 'metadata) `(metadata (category . cider))) ; metadata
```
bbatsov, line 241: Here I meant some high-level comment explaining the completion API. :-) The current comments don't really explain anything for someone who doesn't know what this lambda is supposed to return.

bbatsov:
Fair enough. It'd be good if you also added something in the completion docs for `(setq completion-category-defaults '((cider (styles basic)))`.

cider-completion.el
minad: I am not as smart as @monnier, but I disagree with him that :exclusive=no is fundamentally broken. The implementation in minibuffer.el is fundamentally broken, the substitute in corfu.el is not. I also claim that there is value in :exclusive=no even if merging of Capfs or completion tables may be preferable in many scenarios.

Anyway there is an easy solution - just declare all the Capf properties as a separate variable, which can be adjusted by the user. See:

https://github.com/minad/cape/blob/main/cape.el#L235-L239
https://github.com/minad/cape/blob/ad13cc9e3a505fc3fa56ac559d15e3eca7bcc731/cape.el#L268

cider-completion.el
toniz4: The current explanation seems more appropriate? Be aware that english is my second language, so grammar mistakes are bound to happen.

minad:
@toniz4 Since you touch the completion code here, it would be great if #3006 could be addressed too, such that the default cider completion style becomes more compliant.

toniz4:
> @toniz4 Since you touch the completion code here, it would be great if #3006 could be addressed too, such that the default cider completion style becomes more compliant.

Yes, I opened this PR following that rabbit role, but I think I will try to adress that problem in another PR. I will have to understand how the cider completion works in a deeper level.

The interesting thing that I found is that the builtin emacs funcions for building capf's are kinda non-complient, like (complete-with-action) or (completion-table-dynamic). Or i'm undersanding stuff wrong, documentation is rare about this stuff.

@minad Where do you find good documentation about capf's? Or the emacs manuals should be enought?

minad:
> The interesting thing that I found is that the builtin emacs funcions for building capf's are kinda non-complient, like (complete-with-action) or (completion-table-dynamic). Or i'm undersanding stuff wrong, documentation is rare about this stuff.

The limitation is that `completion-table-dynamic` requires you to use the `basic`, `emacs21` or `emacs22` completion styles such that the input string is passed to the completion table.

> Where do you find good documentation about capf's? Or the emacs manuals should be enought?

For Capfs the Emacs manual is not really a sufficient resource, but you can grep for example Capfs. My [Cape](https://github.com/minad/cape) package also comes with a bunch of simple Capfs you could use as reference.

I just looked at #3006 again. My recommendation is to remove the cider completion style completely:

https://github.com/clojure-emacs/cider/blob/b47fe53b39184b03b413284139d81b255d559db9/cider-completion.el#L281-L295

Instead configure the `basic` completion style for the `cider` category you introduced above.

monnier:
> I agree with @monnier, but this is the only way that i know to get a finer
> control of the completion style, i.e. only apply a specific style for
> completions comming from cider.

I'd need to know *why* you want the completion style to be different
when you use CIDER to be able to start thinking a bout what's
a better answer.

Is CIDER used only for one particular kind of completion (e.g. is it
only used to complete function names, or also file names, module names,
...)?

toniz4:
> I agree with @monnier, but this is the only way that i know to get a finer control of the completion style, i.e. only apply a specific style for completions comming from cider.
> I'd need to know *why* you want the completion style to be different when you use CIDER to be able to start thinking a bout what's a better answer. Is CIDER used only for one particular kind of completion (e.g. is it only used to complete function names, or also file names, module names, ...)?

 I use orderless as the completion style globaly, and I would like to keep it that way for most of the capf's, but some capf's don't play nicely with that, as #3006 shows. And it's nice to have the option to set a different completion style in this category, orderless gives you the option of making custom completion engines.
 
 I could set `completion-styles` to `basic` localy when using cider, but every other completion will use `basic` too, if not overwriten in `completion-category-defaults` or `completion-category-overrides`, so it would fix one part and break the rest.

minad:
Using a completion category here is valid. But you could also set globally:

~~~ elisp
(setq completion-styles '(orderless basic)
~~~

This is the recommended configuration from the Orderless configuration and will make sure that completion-table-dynamic works properly.

cider-completion.el
toniz4: So how is the current comments looking?

bbatsov: It's fine.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -115,6 +115,19 @@ image::completion-annotations.png[Completion Annotations]
 TIP: Completion annotations can be disabled by setting
 `cider-annotate-completion-candidates` to `nil`.
 
+=== Changing the completion style
+
+Sometimes there will be problems displaying completions when using a different value in
+`completion-styles` that does not include `basic`, like when using
+https://github.com/oantolin/orderless["orderless"]. That can be fixed by setting
+`completion-category-defaults`, overriting the completion style just for the cider
```
bbatsov, line 123: overriding
cider -> CIDER.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -115,6 +115,19 @@ image::completion-annotations.png[Completion Annotations]
 TIP: Completion annotations can be disabled by setting
 `cider-annotate-completion-candidates` to `nil`.
 
+=== Changing the completion style
+
+Sometimes there will be problems displaying completions when using a different value in
```
bbatsov, line 120: That is a bit vague. Perhaps this shouldn't framed in terms of problems (this can go in the troubleshooting section), but in terms of what's the benefit of changing the default completion style in the first place.

bbatsov (changes requested):

doc/modules/ROOT/pages/usage/code_completion.adoc
toniz4: Updated it, how is it now?

bbatsov: It's okay.

monnier:
> The interesting thing that I found is that the builtin emacs funcions for
> building capf's are kinda non-complient, like (complete-with-action) or
> (completion-table-dynamic). Or i'm undersanding stuff wrong, documentation
> is rare about this stuff.

These don't build "capf"s, they build completion tables.

   CAPF =  () -> (POS POS COMPLETION-TABLE)


-- Stefan

PR #3227, opened 2022-07-30
Support the Gradle wrapper for jack-in

ajoberstar:
Enhancing Gradle jack-in support to allow (and use by default) the Gradle wrapper `./gradlew` instead of a globally installed `gradle` command on the `PATH`. See commit messages for further details.

I'm pretty new to elisp, so I'm not sure if that `locate-file` approach I took is the preferred option. I'm open to any feedback/changes you'd like, so just let me know.

Side note: I plan to work on some more Gradle related changes, particularly support for dependency injection, as well. Once I have some of those other changes ready, I can do a more thorough documentation update to flesh out how to use Cider with Gradle.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 * Upgrade injected `cider-nrepl` to [0.28.5](https://github.com/clojure-emacs/cider-nrepl/releases/tag/v0.28.5).
 * [#3200](https://github.com/clojure-emacs/cider/issues/3200): Improve cider-browse-ns interface to allow selective hiding of var types as well as grouping options.  Include private vars in result list.
+* Changed default `cider-gradle-command` to `./gradlew` to use the Gradle wrapper
+* Changed default `cider-gradle-global-options` to `""` (empty, formerly `--no-daemon`)
 
 ## 1.4.1 (2022-05-25)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [#3226](https://github.com/clojure-emacs/cider/pull/3226): Populate completions metadata, making it possible to change the style of completion via `completion-category-override` or `completion-category-defaults`.
 - [#2946](https://github.com/clojure-emacs/cider/issues/2946): Add custom var `cider-merge-sessions` to allow combining sessions in two different ways: Setting `cider-merge-sessions` to `'host` will merge all sessions associated with the same host within a project. Setting it to `'project` will combine all sessions of a project irrespective of their host.
+- Support Gradle jack-in via the Gradle wrapper, instead of just a globally installed `gradle` on the `PATH`.
 
 ## Changes
 
--- a/cider.el
+++ b/cider.el
@@ -189,7 +189,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :package-version '(cider . "0.17.0"))
 
 (defcustom cider-gradle-command
-  "gradle"
+  "./gradlew"
   "The command used to execute Gradle."
   :type 'string
   :safe #'stringp
@@ -352,7 +352,11 @@ Throws an error if PROJECT-TYPE is unknown."
     ('shadow-cljs (let ((parts (split-string cider-shadow-cljs-command)))
                     (when-let* ((command (cider--resolve-command (car parts))))
                       (mapconcat #'identity (cons command (cdr parts)) " "))))
-    ('gradle (cider--resolve-command cider-gradle-command))
+    ;; here we have to account for use of the Gradle wrapper which is
+    ;; a shell script within their project, so if they have a clearly
+    ;; relative path like "./gradlew" use locate file instead of checking
+    ;; the exec-path
+    ('gradle (cider--resolve-project-command cider-gradle-command))
     (_ (user-error "Unsupported project type `%S'" project-type))))
 
 (defun cider-jack-in-global-options (project-type)
@@ -1651,6 +1655,14 @@ assume the command is available."
                            (executable-find command)
                            (executable-find (concat command ".bat")))))
     (shell-quote-argument command)))
+
+(defun cider--resolve-project-command (command)
+  "Find COMMAND in project dir or exec path (see variable `exec-path').
+If COMMAND starts with ./ or ../ resolve relative to `clojure-project-dir',
+otherwise resolve via `cider--resolve-command'."
+  (if (string-match-p "\\`\\.\\{1,2\\}/" command)
+      (locate-file command (list (clojure-project-dir)) '("" ".bat") 'executable)
+    (cider--resolve-command command)))
 
 (defcustom cider-connection-message-fn #'cider-random-words-of-inspiration
   "The function to use to generate the message displayed on connect.
--- a/cider.el
+++ b/cider.el
@@ -196,7 +196,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :package-version '(cider . "0.10.0"))
 
 (defcustom cider-gradle-global-options
-  "--no-daemon"
+  ""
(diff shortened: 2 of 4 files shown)


vemv (commented):
Hi! Thanks for the PR.

I'm not sure we want any sort of breaking change - it's not a good habit.

As the name implies a `defcustom` is made for customization if the default doesn't suit you.

Is there a strong, uncontroversial case for changing the default? Have you considered crafting the PR the opposite way (keep the default but make the code handle _possible_ custom values)?

ajoberstar:
Appreciate the quick feedback! Completely understand the concerns about breaking changes, I was focused primarily on best out-of-box experience for new users which understandably wasn't enough care taken.

I can break up the PR into multiple or drop pieces, as desired. Here's more detail on the parts that could be considered breaking.

## cider-gradle-command default change

The rationale here was that idiomatic use of Gradle (not saying its universal) is to use [the wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) to ensure you're using the version of Gradle the project was authored against. Not all projects have the wrapper, but any open source project using Gradle I've seen does and I'd expect most others do as well. 

Obviously anecdotal, but I've used Gradle for 10+ years (and maintain the [Gradle plugin for Clojure](https://github.com/clojurephant/clojurephant)) and have rarely had a global install of gradle on my device, so the existing default of `gradle` would always fail on my machines.

Intuitively (from a Gradle user perspective), the best default experience would probably be to prefer the gradle wrapper (if it exists in the project) and fall back to a global Gradle install, if it doesn't. 

Which would imply changes like this:

- change `defcustom cider-gradle-command` back to `gradle`
- introduce `defcustom cider-gradlew-command` defaulting to `./gradlew`
- introduce either:
  - `defcustom cider-prefer-gradlew-to-gradle` (naming?) defaulting to `t` (or `nil` if that's considered too breaking)
  - `defcustom cider-preferred-gradle-command` (naming?) defaulting to `'(cider-gradlew-command cider-gradle-command)` (or vice-versa if that's considered too breaking)

_However_, if the preference is to just keep the defaults the same, then I'd probably just add more to the docs, suggesting users customize the `cider-gradle-command` variable to `./gradlew` if they want the wrapper to be used. (Which is definitely simpler, so not opposed to this.)

## cider-gradle-global-options default change

As for dropping `--no-daemon`, the rationale is:

1. No one's project will fail by removing this option. It just controls a performance-related setting in Gradle.
    Gradle would fallback to the users' preference (if set in a user or project-level `gradle.properties` file) or Gradle's default of enabling it.
1. The closest to a _break_ in the user experience is that the Gradle daemon would stay running after `cider-quit` (ready for the next Gradle invocation on the device). While this is atypical compared to other tools, this is the default experience of Gradle and something users either like, made their peace with, or likely disabled at a user-level.
1. Disabling the daemon means slower startup in many cases, and goes against the default experience of Gradle.
1. I'm not aware of any good rationale of why Cider would want to disable the daemon by default.

bbatsov:
Given the low usage of Gradle in the community and the super limited support for Gradle in CIDER right now, I'm fine with breaking changes for the greater good (getting this ideally to feature parity with Lein/tools.deps). For a while I actually considered just removing support for Gradle, as it seemed like a lost cause.

vemv:
Thanks much for the greater context! Sounding good.

> The rationale here was that idiomatic use of Gradle (not saying its universal) is to use [the wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) to ensure you're using the version of Gradle the project was authored against. Not all projects have the wrapper, but any open source project using Gradle I've seen does and I'd expect most others do as well.

So, say, an average enterprise project would use the wrapper as well?

ajoberstar:
> So, say, an average enterprise project would use the wrapper as well?

Speaking from my experience (one company), yes we exclusively use the wrapper.

vemv (approved):

vemv:
LGTM!

It seems clear by now that if anything, we're nudging people to do the right thing.

It also would be counterproductive to introduce further defcustom, in face of what @bbatsov said.

CHANGELOG.md
```diff
@@ -6,11 +6,14 @@
...
 - [#3226](https://github.com/clojure-emacs/cider/pull/3226): Populate completions metadata, making it possible to change the style of completion via `completion-category-override` or `completion-category-defaults`.
 - [#2946](https://github.com/clojure-emacs/cider/issues/2946): Add custom var `cider-merge-sessions` to allow combining sessions in two different ways: Setting `cider-merge-sessions` to `'host` will merge all sessions associated with the same host within a project. Setting it to `'project` will combine all sessions of a project irrespective of their host.
+- Support Gradle jack-in via the Gradle wrapper, instead of just a globally installed `gradle` on the `PATH`.
 
 ## Changes
 
 * Upgrade injected `cider-nrepl` to [0.28.5](https://github.com/clojure-emacs/cider-nrepl/releases/tag/v0.28.5).
 * [#3200](https://github.com/clojure-emacs/cider/issues/3200): Improve cider-browse-ns interface to allow selective hiding of var types as well as grouping options.  Include private vars in result list.
+* Changed default `cider-gradle-command` to `./gradlew` to use the Gradle wrapper
+* Changed default `cider-gradle-global-options` to `` (empty, formerly `--no-daemon`)
```
vemv, line 16:
```suggestion
* Changed default `cider-gradle-global-options` to `""` (empty, formerly `--no-daemon`)
```

For better .md rendering

cider.el
```diff
@@ -352,7 +352,13 @@ Throws an error if PROJECT-TYPE is unknown."
     ('shadow-cljs (let ((parts (split-string cider-shadow-cljs-command)))
                     (when-let* ((command (cider--resolve-command (car parts))))
                       (mapconcat #'identity (cons command (cdr parts)) " "))))
-    ('gradle (cider--resolve-command cider-gradle-command))
+    ;; here we have to account for use of the Gradle wrapper which is
+    ;; a shell script within their project, so if they have a clearly
+    ;; relative path like "./gradlew" use locate file instead of checking
+    ;; the exec-path
+    ('gradle (if (string-match-p "\\`\\.\\{1,2\\}/" cider-gradle-command)
```
vemv, line 359: Please extract this logic to a (pure/ish) defun, and unit-test that one

ajoberstar: Will do

ajoberstar: Wasn't sure how to test the locate-file/executable-find pieces, so tried using the spies I saw in other places. Basically just testing that the regex matches the patterns I was expecting. Let me know if you'd like this in a different form.

test/cider-tests.el
```diff
@@ -525,6 +525,23 @@
               :to-equal
               "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch :client-build) (shadow/watch :other-build) (shadow/nrepl-select :client-build))"))))
 
+(describe "cider--resolve-project-command"
+          (it "if command starts with ./ it resolves relative to clojure-project-dir"
+              (spy-on 'locate-file :and-return-value "/project/command")
+              (spy-on 'executable-find :and-return-value nil)
```
vemv, line 531: Should there be a test case(s) for when neither executable-find or locate-file are nil?

test/cider-tests.el
```diff
@@ -525,6 +525,23 @@
               :to-equal
               "(do (require '[shadow.cljs.devtools.api :as shadow]) (shadow/watch :client-build) (shadow/watch :other-build) (shadow/nrepl-select :client-build))"))))
 
+(describe "cider--resolve-project-command"
```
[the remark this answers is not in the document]
ajoberstar: Indentation should be fixed now (in this one, will go back and do the same on the other PR).

test/cider-tests.el
ajoberstar: I can add those

vemv:
Great work, thank you!

PR #3228, opened 2022-07-30
Support injecting dependencies into Gradle

ajoberstar:
Gradle doesn't have any native support, but as of 0.7.0-alpha.6, Clojurephant supports a Gradle property to list dependencies to be injected into the nREPL server.

This property is backwards compatible for users on older Clojurephant versions, since Gradle ignores unknown properties on the command line, falling back to no injection.

The middleware arguments have been supported in Clojurephant since 0.4.0-beta.4 in 2018, so I think it should be safe to start passing those in.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -577,6 +577,49 @@ and MIDDLEWARES.  PARAMS and MIDDLEWARES are passed on to
           (cider-boot-dependencies (append (cider--jack-in-required-dependencies) dependencies))
           (cider-boot-middleware-task params middlewares)))
 
+(defun cider--gradle-dependency-notation (dependency)
+  "Returns Gradle's GAV dependency syntax.
+For a \"group/artifact\" \"version\") DEPENDENCY list
+return as group:artifact:version notation."
+  (let ((group-artifact (replace-regexp-in-string "/" ":" (car dependency)))
+        (version (cadr dependency)))
+    (format "%s:%s" group-artifact version)))
+
+(defun cider--gradle-jack-in-property (dependencies)
+  "Returns Clojurephant's dependency jack-in property.
+For DEPENDENCIES, translates to Gradle's dependency notation
+using `cider--gradle-dependency-notation`.''"
+  (if (seq-empty-p dependencies)
+      ""
+    (shell-quote-argument
+     (concat "-Pdev.clojurephant.jack-in.nrepl="
+             (mapconcat #'cider--gradle-dependency-notation dependencies ",")))))
+
+(defun cider--gradle-middleware-params (middlewares)
+  "Returns Gradle-formatted middleware params.
+Given a list of MIDDLEWARES symbols, this returns
+the Gradle parameters expected by Clojurephant's
+ClojureNRepl task."
+  (mapconcat (lambda (middleware)
+               (shell-quote-argument (concat "--middleware=" middleware)))
+             middlewares
+             " "))
+
+(defun cider-gradle-jack-in-dependencies (global-opts params dependencies middlewares)
+  "Create gradle jack in dependencies.
+Does so by concatenating GLOBAL-OPTS, DEPENDENCIES,
+and MIDDLEWARES.  GLOBAL-OPTS and PARAMS are taken as-is.
+DEPENDENCIES are translated into Gradle's typical
+group:artifact:version notation and MIDDLEWARES are
+prepared as arguments to Clojurephant's ClojureNRepl task."
+  (concat global-opts
+          (unless (seq-empty-p global-opts) " ")
+          (cider--gradle-jack-in-property (append (cider--jack-in-required-dependencies) dependencies))
+          " "
+          params
+          (unless (seq-empty-p params) " ")
+          (cider--gradle-middleware-params middlewares)))
+
 (defun cider--lein-artifact-exclusions (exclusions)
   "Return an exclusions vector described by the elements of EXCLUSIONS."
   (if exclusions
@@ -717,10 +760,12 @@ dependencies."
                    params
                    (cider-add-clojure-dependencies-maybe
                     cider-jack-in-dependencies)))
-    ('gradle (concat
+    ('gradle (cider-gradle-jack-in-dependencies
               global-opts
-              (unless (seq-empty-p global-opts) " ")
-              params))
+              params
+              (cider-add-clojure-dependencies-maybe
+               cider-jack-in-dependencies)
+              (cider-jack-in-normalized-nrepl-middlewares)))
     (_ (error "Unsupported project type `%S'" project-type))))
 
 
--- a/doc/modules/ROOT/pages/basics/middleware_setup.adoc
+++ b/doc/modules/ROOT/pages/basics/middleware_setup.adoc
@@ -12,7 +12,8 @@ automatically injects this middleware and other dependencies as required.
 NOTE: In the past, if you were setting up CIDER, you might have had to
 modify `profiles.clj` or `profile.boot`. CIDER now handles
 everything automatically and you don't need to add anything
-special to these files. The same is true of your `deps.edn` file.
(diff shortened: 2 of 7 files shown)


ajoberstar:
Need clojurephant/clojurephant#179 to be released also.

ajoberstar:
This seems to be working well for me. Unsure if there are other parity issues with Gradle (besides ClojureScript, I know that will still be a sore spot for a bit). Let me know if there are any changes/discussion needed.

cider.el
```diff
@@ -573,6 +573,49 @@ and MIDDLEWARES.  PARAMS and MIDDLEWARES are passed on to
           (cider-boot-dependencies (append (cider--jack-in-required-dependencies) dependencies))
           (cider-boot-middleware-task params middlewares)))
 
+(defun cider--gradle-dependency-notation (dependency)
```
vemv, line 576: These pure defns are great candidates for unit testing

cider.el
```diff
@@ -573,6 +573,49 @@ and MIDDLEWARES.  PARAMS and MIDDLEWARES are passed on to
...
+  "Returns Gradle-formatted middleware params.
+Given a list of MIDDLEWARES symbols, this returns
+the Gradle parameters expected by Clojurephant's
+ClojureNRepl task."
+  (mapconcat (lambda (middleware)
+               (shell-quote-argument (concat "--middleware=" middleware)))
+             middlewares
+             " "))
+
+(defun cider-gradle-jack-in-dependencies (global-opts params dependencies middlewares)
```
vemv, line 604: I reckon that https://github.com/clojure-emacs/clj-refactor.el/blob/1bd7b09976c05ebd1ff1ee3d5c882af303780b67/clj-refactor.el#L4220-L4238 would have to start handling this variable.

I'll do it after this PR is merged

vemv (commented):
Looking great!

cider.el
ajoberstar: If I'm reading that right, it looks like it might work without anything extra. Haven't used clj-refactor before, but can give it a try after I add the tests you asked about.

cider.el
ajoberstar: Yep, can do

cider.el
vemv: > If I'm reading that right, it looks like it might work without anything extra.  

Correct

test/cider-tests.el
```diff
@@ -105,6 +105,38 @@
       (expect (cider-project-type) :to-equal cider-jack-in-default))))
 
 ;;; cider-jack-in tests
+(describe "cider--gradle-dependency-notation"
+          (it "returns a GAV when given a two-element list"
+              (expect (cider--gradle-dependency-notation '("cider/piggieback" "0.5.3"))
+                      :to-equal "cider:piggieback:0.5.3"))
+          (it "returns a broken AV if the first element is missing a slash"
+              ;; This would be an invalid Gradle dependency
+              (expect (cider--gradle-dependency-notation '("nrepl" "0.9.0"))
```
ajoberstar, line 114: Curious if this is valid syntax in Cider (as it is in lein). If so, I'd want to add logic to expand those out to `nrepl:nrepl:0.9.0`.

cider.el
ajoberstar: Tests added now. Let me know if I missed anything.

test/cider-tests.el
vemv: IIRC `'("nrepl" "0.9.0")` generally works in CIDER because Lein is fine with it.

tools.deps (in its newer versions) isn't, and because CIDER doesn't perform any translation, would result in a broken process.

So, I think CIDER shouldn't perform a translation here either (for consistency), but at the same time adding a test showing broken-ness isn't particularly desirable.

It's better to let it be undefined behavior. You can add an appropiate remark in the docs if there isn't one already (grep for `cider-jack-in-` in doc)

test/cider-tests.el
```diff
@@ -105,6 +105,38 @@
       (expect (cider-project-type) :to-equal cider-jack-in-default))))
 
 ;;; cider-jack-in tests
+(describe "cider--gradle-dependency-notation"
+  (it "returns a GAV when given a two-element list"
+    (expect (cider--gradle-dependency-notation '("cider/piggieback" "0.5.3"))
+            :to-equal "cider:piggieback:0.5.3"))
+  (it "returns a broken AV if the first element is missing a slash"
```
vemv, line 112: See https://github.com/clojure-emacs/cider/pull/3228#discussion_r933916969

test/cider-tests.el
```diff
@@ -167,8 +199,11 @@
                          " repl -s wait")))
 
     (it "can inject dependencies in a gradle project"
-      (expect (cider-inject-jack-in-dependencies "" "--no-daemon clojureRepl" 'gradle)
-              :to-equal "--no-daemon clojureRepl")))
+        (expect (cider-inject-jack-in-dependencies "--no-daemon" ":clojureRepl" 'gradle)
```
vemv, line 202: Indentation

test/cider-tests.el
ajoberstar: Sounds good, will remove the test and see if there's any mention in the docs.

test/cider-tests.el
ajoberstar: Should be fixed now

test/cider-tests.el
ajoberstar: Removed

test/cider-tests.el
ajoberstar: Added a note in the two places in the docs where `cider-jack-in-dependencies` and `cider-jack-in-cljs-dependencies` are shown.

vemv (approved):
Excellent!

doc/modules/ROOT/pages/basics/middleware_setup.adoc
```diff
@@ -95,7 +96,24 @@ run `cider-connect` or `cider-connect-cljs`.
...
 
-NOTE: This section is currently a stub. Contributions welcome!
+NOTE: Make sure you're using https://github.com/clojurephant/clojurephant[Clojurephant] 0.4.0 or newer.
+
+.build.gradle
+[source, groovy]
+----
+dependencies {
+  devImplementation 'nrepl:nrepl:0.9.0'
+  devImplementation 'cider:cider-nrepl:0.28.4'
```
vemv, line 106: 0.28.4 should be 0.28.5 throughout this PR

ajoberstar: Should I just update the docs or should I do a full find/replace? There are 23 references to 0.28.4 (most of which not part of my PR) and just 1 reference to 0.28.5.

vemv: Let's go full, thank you!

ajoberstar: Should be all up to date now

vemv:
Thanks again.

Are there more PRs incoming or should we start thinking of a new CIDER release?

ajoberstar:
I think this is a good stopping point for now. Feel free to ping me if there are issues with the new Gradle support.

vemv:
btw, the build failed on Windows :(

(For some reason, sometimes CircleCI doesn't build PRs)

Could you please add your cider fork to your CircleCI account, enable builds to it and create a fix?

ajoberstar:
Working on it now. Wondering if it was those tests where I manually did the shell escaping.

bbatsov:
@ajoberstar Btw, you might want to include some of the Gradle instructions for nREPL in nREPL's docs as well. I guess that'd be good for Clojurephant's visibility as well.

ajoberstar:
@bbatsov Good idea, will get a PR sent over to nREPL's docs.

PR #3229, opened 2022-07-31
Fix shell quoting in Gradle tests

ajoberstar:
Needed to pass Windows

-----------------

Thanks!

manual][1] extremely useful.*


ajoberstar:
@vemv This should fix those failing Windows tests

vemv:
Indeed, it's green now!

@bbatsov , you might want to create a CIDER release

bbatsov:
@vemv Let's do it after @ajoberstar does all the Gradle changes he has in mind.

vemv:
They're done! https://github.com/clojure-emacs/cider/pull/3228#issuecomment-1200340807

PR #3234, opened 2022-08-04
Locate multiple ports for a single project

### LINKED ISSUE #3140: Auto detect multiple nrepl connection options, if they exist
## Expected behavior

I often need to connect to both a clj nrepl server as well as a shadow-cljs nrepl server for full stack dev. Because the clj nrepl server prints its port to `.nrepl-port`, and shadow prints its port to `.shadow-nrepl/nrepl.port`, ideally their respective ports should both be automatically discoverable when I try to connect via `cider-connect`.

## Actual behavior

If a clj nrepl server has already printed its port to `.nrepl-port`, `cider-connect` will fail to auto-detect the shadow nrepl server port in `.shadow-nrepl/nrepl.port`. With only a shadow nrepl server running, the shadow nrepl port is automatically detected.

Make sure you have both a clj nrepl server started as well as a shadow cljs nrepl server started before you try to connect to the shadow-cljs server via `cider-connect`.

I'm not super familiar with the code-base, but as best I can tell the conditional in [`nrepl-extract-port` in `nrepl-client.el`](https://github.com/clojure-emacs/cider/blob/master/nrepl-client.el#L231-L236) will always return a single result and cause `.nrepl-port` to shadow `.shadow-nrepl/nrepl.port`. `nrepl-extract-port` might need to return a list.

### CIDER version information

```
;; CIDER 1.1.0 (package: 1.1.0.1) (Plovdiv), nREPL 0.8.3
;; Clojure 1.10.3, Java 1.8.0_302
```

### Emacs version

GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0)

### Operating system

OSX 11.6.2

danmidwood:
Searching for a nrepl port to connect to would return a maximum of one
port per project, ignoring any others that you might want to connect
to. This creates a case where when connecting to a shadow-cljs nrepl
server the port is not presented to you as a completion when you
already have another nrepl server running.

This commit changes nrepl-extract-port to nrepl-extract-ports and the
return type from a single port to a list of ports (including nils
where a specific project type nrepl port file doesn't exist) to
provide the full view of nrepl servers that are available in the
project, and then the cider-locate-running-nrepl-ports fn is changed
to accommodate that.

This fixes #3140.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1625,10 +1625,11 @@ of remote SSH hosts."
 When DIR is non-nil also look for nREPL port files in DIR.  Return a list
 of list of the form (project-dir port)."
   (let* ((paths (cider--running-nrepl-paths))
-         (proj-ports (mapcar (lambda (d)
-                               (when-let* ((port (and d (nrepl-extract-port (cider--file-path d)))))
-                                 (list (file-name-nondirectory (directory-file-name d)) port)))
-                             (cons (clojure-project-dir dir) paths))))
+         (proj-ports (apply #'append
+                            (mapcar (lambda (d)
+                                      (mapcar (lambda (p) (list (file-name-nondirectory (directory-file-name d)) p))
+                                              (and d (nrepl-extract-ports (cider--file-path d)))))
+                                    (cons (clojure-project-dir dir) paths)))))
     (seq-uniq (delq nil proj-ports))))
 
 (defun cider--running-nrepl-paths ()
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -237,6 +237,16 @@ PARAMS is as in `nrepl-make-buffer-name'."
       (nrepl--port-from-file (expand-file-name "target/repl-port" dir))
       (nrepl--port-from-file (expand-file-name ".shadow-cljs/nrepl.port" dir))))
 
+(defun nrepl-extract-ports (dir)
+  "Read ports from applicable repl-port files in directory DIR."
+  (delq nil
+        (list (nrepl--port-from-file (expand-file-name "repl-port" dir))
+              (nrepl--port-from-file (expand-file-name ".nrepl-port" dir))
+              (nrepl--port-from-file (expand-file-name "target/repl-port" dir))
+              (nrepl--port-from-file (expand-file-name ".shadow-cljs/nrepl.port" dir)))))
+
+(make-obsolete 'nrepl-extract-port 'nrepl-extract-ports "1.4.2")
+
 (defun nrepl--port-from-file (file)
   "Attempts to read port from a file named by FILE."
   (when (file-exists-p file)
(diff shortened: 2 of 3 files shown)


nrepl-client.el
```diff
@@ -230,12 +230,12 @@ PARAMS is as in `nrepl-make-buffer-name'."
   "Return t if HOST is local."
   (string-match-p tramp-local-host-regexp host))
 
-(defun nrepl-extract-port (dir)
-  "Read port from applicable repl-port file in directory DIR."
-  (or (nrepl--port-from-file (expand-file-name "repl-port" dir))
-      (nrepl--port-from-file (expand-file-name ".nrepl-port" dir))
-      (nrepl--port-from-file (expand-file-name "target/repl-port" dir))
-      (nrepl--port-from-file (expand-file-name ".shadow-cljs/nrepl.port" dir))))
+(defun nrepl-extract-ports (dir)
```
bbatsov, line 233: That's a public function, so renaming it would be a breaking change and we want to avoid those. You can just introduce a few function instead and perhaps mark the old function as obsolete.

nrepl-client.el
```diff
@@ -230,12 +230,12 @@ PARAMS is as in `nrepl-make-buffer-name'."
...
+  "Read ports from applicable repl-port files in directory DIR."
+  (list (nrepl--port-from-file (expand-file-name "repl-port" dir))
```
bbatsov, line 235: I'd probably filter out the `nil` values here.

cider.el
```diff
@@ -1625,10 +1625,14 @@ of remote SSH hosts."
...
   (let* ((paths (cider--running-nrepl-paths))
-         (proj-ports (mapcar (lambda (d)
-                               (when-let* ((port (and d (nrepl-extract-port (cider--file-path d)))))
-                                 (list (file-name-nondirectory (directory-file-name d)) port)))
-                             (cons (clojure-project-dir dir) paths))))
+         (proj-ports (apply #'append
+                            (mapcar (lambda (d)
+                                      (when-let* ((ports (and d (nrepl-extract-ports (cider--file-path d)))))
+                                        (mapcar (lambda (p)
+                                                  (when-let* ((p p))
```
bbatsov, line 1632: If you apply the filter to the returned ports this `when-let` will become redundant and the code would read better.

bbatsov:
Looks good to me overall. This need a changelog entry and perhaps we should explain somewhere in the user docs how CIDER finds nREPL ports automatically in a couple of sentences.

bbatsov:
@danmidwood Ping :-)

danmidwood:
> @danmidwood Ping :-)

I agree with all of the feedback, I will schedule some time this week to make the changes to the PR.

bbatsov:
Thanks!

Issue #3239, opened 2022-08-22
Add command for eval and tap sexp at point


skjoenberg:
This request is a bit related to #3094, where the issue creator wants to be able to `tap>` every form that is evaluated.
In my workflow, I only use `tap>` when I need to inspect a large or complex piece of data (in portal or something similar).
Therefore, I propose a command that evaluates the sexp-at-point and taps the result.

I don't know whether such a feature is general enough to be an enhancement, or whether it should live in my own config.
I've found a simple solution, that seems to work for me and I'll open a related merge request.
Feel free to reject it, if it's too niche.

PR #3240, opened 2022-08-22
Added evaluate and tap commands last-sexp and sexp-at-point

skjoenberg:
This PR solves #3239 by adding two commands that evaluate and tap either the last-sexp or the sexp-at-point. 
This is simply done by wrapping the form in a `(doto *form* (tap>))`.

I've not added any tests, since I couldn't find much in the `cider-eval-tests.el`. I'll happily do it if anybody can point me in the right direction.

I've not chosen any keybinding for this. I don't have any clue on how to pick them.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1016,6 +1016,25 @@ If invoked with OUTPUT-TO-CURRENT-BUFFER, output the result to current buffer."
     (goto-char (cadr (cider-sexp-at-point 'bounds)))
     (cider-eval-last-sexp output-to-current-buffer)))
 
+(defun cider-tap-last-sexp (&optional output-to-current-buffer)
+  "Evaluate and tap the expression preceding point.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current
+buffer."
+  (interactive "P")
+  (let ((tapped-form (concat "(clojure.core/doto " (cider-last-sexp) " (clojure.core/tap>))")))
+    (cider-interactive-eval tapped-form
+                            (when output-to-current-buffer (cider-eval-print-handler))
+                            nil
+                            (cider--nrepl-pr-request-map))))
+
+(defun cider-tap-sexp-at-point (&optional output-to-current-buffer)
+  "Evaluate and tap the expression around point.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, output the result to current buffer."
+  (interactive "P")
+  (save-excursion
+    (goto-char (cadr (cider-sexp-at-point 'bounds)))
+    (cider-tap-last-sexp output-to-current-buffer)))
+
 (defvar-local cider-previous-eval-context nil
   "The previous evaluation context if any.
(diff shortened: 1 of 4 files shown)


cider-eval.el
```diff
@@ -1016,6 +1016,26 @@ If invoked with OUTPUT-TO-CURRENT-BUFFER, output the result to current buffer."
     (goto-char (cadr (cider-sexp-at-point 'bounds)))
     (cider-eval-last-sexp output-to-current-buffer)))
 
+(defun cider-tap-last-sexp (&optional output-to-current-buffer)
+  "Evaluate and tap the expression preceding point.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current
+buffer."
+  (interactive "P")
+  (let* ((form (apply #'buffer-substring-no-properties (cider-last-sexp 'bounds)))
+         (tapped-form (concat "(doto " form " (tap>))")))
```
vemv, line 1025:
```suggestion
         (tapped-form (concat "(clojure.core/doto " form " (clojure.core/tap>))")))
```

We never know if something is shadowing them

vemv (commented):
This sounds a reasonable addition to me, that might be handy to others

cider-eval.el
skjoenberg: Ah, yes! Haven't thought of that, thanks.

cider-eval.el
```diff
@@ -1016,6 +1016,26 @@ If invoked with OUTPUT-TO-CURRENT-BUFFER, output the result to current buffer."
     (goto-char (cadr (cider-sexp-at-point 'bounds)))
     (cider-eval-last-sexp output-to-current-buffer)))
 
+(defun cider-tap-last-sexp (&optional output-to-current-buffer)
+  "Evaluate and tap the expression preceding point.
+If invoked with OUTPUT-TO-CURRENT-BUFFER, print the result in the current
+buffer."
+  (interactive "P")
+  (let* ((form (apply #'buffer-substring-no-properties (cider-last-sexp 'bounds)))
```
bbatsov, line 1024: I'm surprised `cider-last-sexp` would returns something with unstripped properties. Probably something like this should be done there to avoid repetition in the client code.

bbatsov:
New eval commands should be added to both the menu and the eval keymap to ease their discovery.

cider-eval.el
skjoenberg: I found that `cider-last-exp` actually did this, so I just used it directly without `'bounds`. I hope this fixes your comments, otherwise, I might need a bit more explanation.

skjoenberg:
> New eval commands should be added to both the menu and the eval keymap to ease their discovery.

I've added two keybindings and added the commands to `cider-mode-eval-menu`. If this was not the menu you were referring to, I'll need a hint to a file. :)

bbatsov:
Looks good. Thanks!

PR #3243, opened 2022-08-30
Remove needless quote from the choices

syohex:
This fixes the following byte-compile warning too on newer Emacs.

```
cider.el:443:12: Warning: defcustom for ‘cider-jack-in-auto-inject-clojure’
    has syntactically odd type ‘'(choice (const :tag None nil) (const :tag
    Latest 'latest) (const :tag Minimal 'minimal) (string :tag Specific
    Version) (list :tag Artifact ID and Version (string :tag Artifact ID)
    (string :tag Version)))’
```

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Bump the injected nREPL version to 1.0.
 
+### Bugs fixed
+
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #3247, opened 2022-09-18
Add cider-stacktrace-analyze-in-region

r0man:
Add command to analyze and inspect the stacktrace in region

Please see:
- https://github.com/clojure-emacs/orchard/pull/164
- https://github.com/clojure-emacs/cider-nrepl/pull/758
-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 - [#3249](https://github.com/clojure-emacs/cider/pull/3249): Add support for Clojure Spec 2.
+- [#3247](https://github.com/clojure-emacs/cider/pull/3247) Add the `cider-stacktrace-analyze-at-point` and `cider-stacktrace-analyze-in-region` commands to view printed exceptions in the stacktrace inspector.
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


vemv:
Another thing that you may be interested in to do is making a stacktrace 'clickable'.

Often when a repl command emits a stacktrace, each element of that stacktrace is clickable, taking you to the right file/line. (I think if you grep for `fontify` you will find the related mechanism we use?)

That seems super handy as it's an "inline UI". For instance I might paste a stacktrace from Datadog into `*scratch*` and just leave it there for a while.

If it turns out to be disproportionately hard, then it's no big deal.

bbatsov:
> Another thing that you may be interested in to do is making a stacktrace 'clickable'.

Clickable in what way? Even now you can click on stack frames and go to the revelant source code.

vemv:
I mean an arbitrary one e.g. pasted from the clipboard

bbatsov:
The new command should be mentioned somewhere in the mode menu and in the "dealing with stacktraces" docs.

bbatsov:
I guess we're mostly ready here and we'll move forward with a proper review after the stacktrace-related code is moved to haystack and haystack is integrated into cider-nrepl.

r0man:
Nice, also thanks to you and @vemv and @benedekfazekas for all the help. See you next time!

PR #3249, opened 2022-09-19
Spec2

r0man:
This PR add support for rendering the following Clojure Spec 2 forms:
- s/schema
- s/select
- s/union 

Support for Clojure Spec 2 has been added to the Orchard here:
https://github.com/clojure-emacs/orchard/pull/163

-----------------

### DIFF
--- a/cider-browse-spec.el
+++ b/cider-browse-spec.el
@@ -142,7 +142,79 @@ Display TITLE at the top and SPECS are indented underneath."
 
 (defun cider--spec-fn-p (value fn-name)
   "Return non nil if VALUE is clojure.spec.[alpha]/FN-NAME."
-  (string-match-p (concat "^\\(clojure.spec\\|clojure.spec.alpha\\)/" fn-name "$") value))
+  (string-match-p (concat "^\\(clojure.spec\\|clojure.spec.alpha\\|clojure.alpha.spec\\)/" fn-name "$") value))
+
+(defun cider-browse-spec--render-schema-map (spec-form)
+  "Render the s/schema map declaration SPEC-FORM."
+  (let ((name-spec-pairs (seq-partition (cdaadr spec-form) 2)))
+    (format "(s/schema\n {%s})"
+            (string-join
+             (thread-last
+               (seq-sort-by #'car #'string< name-spec-pairs)
+               (mapcar (lambda (s) (concat (cl-first s) " " (cider-browse-spec--pprint (cl-second s))))))
+             "\n  "))))
+
+(defun cider-browse-spec--render-schema-vector (spec-form)
+  "Render the s/schema vector declaration SPEC-FORM."
+  (format "(s/schema\n [%s])"
+          (string-join
+           (thread-last
+             (cl-second spec-form)
+             (mapcar (lambda (s) (cider-browse-spec--pprint s))))
+           "\n  ")))
+
+(defun cider-browse-spec--render-schema (spec-form)
+  "Render the s/schema SPEC-FORM."
+  (let ((schema-args (cl-second spec-form)))
+    (if (and (listp schema-args)
+             (nrepl-dict-p (cl-first schema-args)))
+        (cider-browse-spec--render-schema-map spec-form)
+      (cider-browse-spec--render-schema-vector spec-form))))
+
+(defun cider-browse-spec--render-select (spec-form)
+  "Render the s/select SPEC-FORM."
+  (let ((keyset (cl-second spec-form))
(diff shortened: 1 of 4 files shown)


bbatsov:
You might consider also extending the docs here a little bit https://docs.cider.mx/cider/usage/misc_features.html#browsing-the-clojure-spec-registry and mention that Spec 2 is now also supported, plus some of the pitfalls of mixing spec versions. 

Potentially we can move the Spec stuff to a dedicated page if we end up having more spec-relation functionality.

cider-browse-spec.el
```diff
@@ -142,7 +142,46 @@ Display TITLE at the top and SPECS are indented underneath."
...
+  (string-match-p (concat "^\\(clojure.spec\\|clojure.spec.alpha\\|clojure.alpha.spec\\)/" fn-name "$") value))
+
+(defun cider-browse-spec--render-schema-map (spec-form)
+  "Render the s/schema map declaration SPEC-FORM."
+  (let ((name-spec-pairs (seq-partition (cdaadr spec-form) 2)))
+    (format "(s/schema\n {%s})"
+            (string-join
+             (thread-last
+               (seq-sort-by #'car #'string< name-spec-pairs)
+               (mapcar (lambda (s) (concat (cl-first s) " " (cider-browse-spec--pprint (cl-second s))))))
```
bbatsov, line 154: `cl-first` is basically `car`. :-)

cider-browse-spec.el
```diff
@@ -142,7 +142,46 @@ Display TITLE at the top and SPECS are indented underneath."
...
+  (format "(s/schema\n [%s])"
+          (string-join
+           (thread-last
+             (cl-second spec-form)
+             (mapcar (lambda (s) (cider-browse-spec--pprint s))))
+           "\n  ")))
+
+(defun cider-browse-spec--render-schema (spec-form)
+  "Render the s/schema SPEC-FORM."
+  (let ((schema-args (cl-second spec-form)))
```
bbatsov, line 168: I prefer the use of `nth` over forms like `cl-second`.

cider-browse-spec.el
```diff
@@ -142,7 +142,46 @@ Display TITLE at the top and SPECS are indented underneath."
...
+    (if (and (listp schema-args)
+             (nrepl-dict-p (cl-first schema-args)))
+        (cider-browse-spec--render-schema-map spec-form)
+      (cider-browse-spec--render-schema-vector spec-form))))
+
+(defun cider-browse-spec--render-select (spec-form)
+  "Render the s/select SPEC-FORM."
+  (let ((keyset (cl-second spec-form))
+        (selection (cl-third spec-form)))
```
bbatsov, line 177: Ditto.

cider-browse-spec.el
r0man: Let's leave all those `cl-` vs `nth` for now? I think it is more consistent with the rest of the code in this namespace, since `cl-first` to `cl-third` are actually used a LOT in this namespace. ;) WDYT?

r0man:
@bbatsov I added a section about Clojure Spec versions to the docs. Regarding the use of the `cl-` functions, I don't mind, but let's do it in another PR? It's not only used in this new code, but in a lot of other places. And changing all of them might be a bit out of scope for this PR. Wdyt?

cider-browse-spec.el
r0man: I don't really mind, but one advantage I see in using the `cl-` functions is, your mind does not need to care about the additional index parameter at all. And maybe you save a cons cell :)

cider-browse-spec.el
bbatsov: Fair enough, I guess I didn't pay enough attention to this bit of code when it was submitted initially.

bbatsov:
Yeah, that's fine.

cider-browse-spec.el
bbatsov: I just don't like mixing APIs. For me `cl-lib` is mostly a remnant of the past and its rarely needed these days, given many advancements that happened in the core Elisp APIs. But this is definitely not a big deal and I didn't spend any time on a focused effort to expunge `cl-lib` from any project.

r0man:
@bbatsov This one would be ready from my side. Would you like to take another look please?

bbatsov:
I think the code is in a good shape, but we need to cut new orchard and cider-nrepl releases before I can merge this PR. Now that we've decided how to proceed we'll `haystack` I'll do this soon.

PR #3259, opened 2022-10-16
Consecutive result overlays can now seen

pestctrl:
This is a minor bug that has just bothered me for years, and I've finally decided to fix it.

When I run cider-eval-last-sexp twice, the second overlay never shows up. This is because we're letting the hook that's supposed to kill overlay No. 1 also kill overlay No. 2. When creating a second overlay right after the first one, we already remove overlay No. 1 in #'cider--make-result-overlay, and thus we can safely remove #'cider--remove-result-overlay from the post-command-hook.

-----------------
Previously, if multiple overlays were created one after another that
are meant to last for one command, every second overlay wouldn't show
up.

This is because there's some bad hook management causing the second
overlay in a sequence to be deleted. Whenever we're creating a second
overlay, the first one is automatically removed. But,
#'cider--remove-result-overlay' is still in the 'post-command-hook,
and thus, right after the second overlay is created, it's immediately
deleetd.

Instead, when we're creating a new overlay, we should call remove any
instances of #'cider--remove-result-overlay in the
'post-command-hook, so that the newly created overlay doesn't get
hosed.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Remove needless quotes from the choices of `cider-jack-in-auto-inject-clojure`.
 - [#2561](https://github.com/clojure-emacs/cider/issues/2561): Disable undo in `*cider-test-report*` buffers.
 - [#3251](https://github.com/clojure-emacs/cider/pull/3251): Disable undo in `*cider-stacktrace*` buffers.
+- Consecutive overlays will not be spuriously deleted. 
 
 ## 1.5.0 (2022-08-24)
 
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -231,6 +231,8 @@ overlay."
                ;; inherit the face of the following text.
                (display-string (format (propertize format 'face 'default) value))
                (o nil))
+          ;; Remove any overlay at the position we're creating a new one, if it
+          ;; exists.
           (remove-overlays beg end 'category type)
           (funcall (if cider-overlays-use-font-lock
                        #'font-lock-prepend-text-property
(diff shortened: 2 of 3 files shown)


bbatsov:
Nice fix! Thanks! 🙇‍♂️

ikappaki:
Hi @pestctrl, this PR appears to have broken the macos tests. I've had a quick lookeat the change but it doesn't look easy to figure this out. do you think you could either have a look and try to fix it or explain why this might be happening and suggest possible ways to move forward?

```
cider--make-result-overlay when overlay duration is given in secs duration overlays are currently the only overlays that can be deleted independently from the other types

https://github.com/clojure-emacs/cider/actions/runs/3344697974/jobs/5539355929
  buttercup-fail("%s" "Expected `(overlay-count)' to be `equal' to `2', but instead it was `3' which does not match because: (different-atoms (3 \"#x3\" \"?\") (2 \"#x2\" \"?\")).")
  signal(buttercup-failed "Expected `(overlay-count)' to be `equal' to `2', but instead it was `3' which does not match because: (different-atoms (3 \"#x3\" \"?\") (2 \"#x2\" \"?\")).")
FAILED: Expected `(overlay-count)' to be `equal' to `2', but instead it was `3' which does not match because: (different-atoms (3 "#x3" "?") (2 "#x2" "?")).
```
I could not reproduce this on my macos computer so I can debug this locally, but consistently fails on GH action, and get the same result on circleci (I've tried this as an experiment).

Thanks

pestctrl:
Hmmm, thanks for letting me know. I'll have a look and see what I can do.

pestctrl:
I have a couple ideas for how to fix this, though I'm not 100% sure on what the root cause is.

Should I do anything from a git perspective? Open a new pull request? 

The fix I think is to add darwin to the list of system-types that are unsafe to speed up the unit tests, or to change the value of cider-overlay--scale-time to 0.1 instead.

bbatsov:
@pestctrl Yeah, a PR would be great.

ikappaki:
ping @pestctrl :)

pestctrl:
@ikappaki I have opened #3289, though none of the unit tests are being run from the workflow file. Could you try out those changes?

ikappaki:
ping @pestctrl :)

> @ikappaki I have opened #3289, though none of the unit tests are being run from the workflow file. Could you try out those changes?

Hi @pestctrl, thanks! I've moved the macos regression tests to circleci, they seem to be passing now, let me post the details in #3829.

PR #3260, opened 2022-10-17
Scroll repl buffer in other frame

pdbrown:
This is a little fix to make the repl buffer scroll to the bottom after `cider-insert-in-repl` when it's in another frame. It doesn't currently scroll in other frames because `get-buffer-window` only considers the current one by default. Setting the opt arg `t` means to consider all windows on all existing frames (and seems to start with the selected frame).


pdbrown:
Yep! Added a note to the changelog.

bbatsov:
Thanks!

Issue #3261, opened 2022-10-24
Parse remote user name from currently open TRAMP buffer


OknoLombarda:
**Is your feature request related to a problem? Please describe.**

Now, when I try to jack-in from TRAMP buffer, CIDER tries to open SSH tunnel using current local user name, even though TRAMP buffer uses a different user to connect

**Describe the solution you'd like**

Instead of manually formatting path here
https://github.com/clojure-emacs/cider/blob/master/nrepl-client.el#L573
we could use `(buffer-file-name)` to get current buffer's name to get all necessary properties for connection

**Additional context**

I'm trying to connect to a remote PC with username that is different from the one on my local machine. I could, of course, just create user with this name on a remote PC, but I don't really want to create a separate user just to be able to connect from CIDER. Also, there is [this](https://github.com/clojure-emacs/cider/issues/2822) issue, that could also be resolved by applying proposed solution

Issue #2822, opened 2020-03-17
cider-jack-in on TRAMP buffer always tries SSH connection with port 22


xorgy:
## Expected behavior
It should attempt the nREPL SSH tunnel on the port used to connect to the TRAMP buffer, rather than the default port.

## Actual behavior
Regardless of what SSH port you've used to connect to your TRAMP buffer, cider will attempt to tunnel the nREPL through SSH on port 22.

## Steps to reproduce the problem
1) Open a TRAMP buffer with ssh through a port other than 22.
2) <kbd>M-x</kbd><kbd>cider-jack-in</kbd>

### CIDER version information
```
;; CIDER 0.25.0snapshot (package: 20200303.1131), nREPL 0.6.0
;; Clojure 1.10.1, Java 1.8.0_242
```

### Emacs version
GNU Emacs 26.3

### Operating system

Arch Linux

xorgy:
In the TRAMP path, the SSH port is provided after a `#` symbol in the host part.

PR #3264, opened 2022-11-05
If current buffer is a TRAMP buffer, use its filename for SSH tunnel

OknoLombarda:
> https://github.com/clojure-emacs/cider/issues/3261
https://github.com/clojure-emacs/cider/issues/2822

If current buffer is a TRAMP buffer, use its filename for SSH tunnel. This way, username and SSH port will be parsed from existing SSH connection, instead of using current username and default SSH port.

I'll edit changelog and manual (if needed) after review

-----------------

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -570,7 +570,18 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 (defun nrepl--ssh-tunnel-connect (host port)
   "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
   (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
-  (let* ((remote-dir (if host (format "/ssh:%s:" host) default-directory))
+  (let* ((current-buf (buffer-file-name))
+         (tramp-file-regexp "/ssh:\\(.+@\\)?\\(.+?\\)\\(:\\|#\\).+")
+         (remote-dir (cond
+                      ;; If current buffer is a TRAMP buffer and its host is
+                      ;; the same as HOST, reuse its connection parameters for
+                      ;; SSH tunnel.
+                      ((and (string-match tramp-file-regexp current-buf)
+                            (string= host (match-string 2 current-buf))) current-buf)
+                      ;; Otherwise, if HOST was provided, use it for connection.
+                      (host (format "/ssh:%s:" host))
+                      ;; Use default directory as fallback.
+                      (t default-directory)))
          (ssh (or (executable-find "ssh")
                   (error "[nREPL] Cannot locate 'ssh' executable")))
          (cmd (nrepl--ssh-tunnel-command ssh remote-dir port))
@@ -598,11 +609,12 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
     ;; forwarding is set up, which is used to synchronise on, so that
     ;; the port forwarding is up when we try to connect.
     (format-spec
-     "%s -v -N -L %p:localhost:%p %u'%h'"
+     "%s -v -N -L %p:localhost:%p %u'%h' %n"
      `((?s . ,ssh)
        (?p . ,port)
        (?h . ,v-host)
-       (?u . ,(if v-user (format "-l '%s' " v-user) ""))))))
+       (?u . ,(if v-user (format "-l '%s' " v-user) ""))
+       (?n . ,(if v-port (format "-p '%s' " v-port) ""))))))
 
 (autoload 'comint-watch-for-password-prompt "comint"  "(autoload).")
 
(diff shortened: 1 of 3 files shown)


bbatsov:
The proposed change seems reasonable. Probably we should also document this behavior as it might not be obvious to people.

nrepl-client.el
```diff
@@ -569,14 +569,21 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 
 (defun nrepl--ssh-tunnel-connect (host port)
   "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
-  (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
-  (let* ((remote-dir (if host (format "/ssh:%s:" host) default-directory))
+  (let* ((current-buf (buffer-file-name))
+         (tramp-file-regexp "/ssh:\\(.+@\\)?\\(.+?\\)\\(:\\|#\\).+")
+         (remote-dir (cond
```
bbatsov, line 574: You might also want to sprinkle a few comments here to make it clearer what we're trying to do here.

bbatsov:
Ping :-)

OknoLombarda:
Yeah, I didn't forget :D. Didn't have much time lately. I added comments, but doesn't this line in docs already imply that CIDER takes care of SSH tunneling if user works with remote files over TRAMP? Or should I describe what exactly it does?
image
https://docs.cider.mx/cider/basics/up_and_running.html#working-with-remote-hosts

bbatsov:
I was thinking it might be useful to explain in a bit more details what the TRAMP integration does and what exactly people can do to influence what SSH connection gets established. Maybe cover some edge cases, etc. The current docs are definitely very basic, and I myself never use TRAMP, so I didn't bother to play with this and document it better.

OknoLombarda:
While checking some cases I found that this change would make `cider-connect-*` ignore host that was selected by user and still connect to the one of current TRAMP buffer. I guess TRAMP connection should only be taken into account if its host is the same as the one selected by user. Like so (diff with master)
```diff
 (defun nrepl--ssh-tunnel-connect (host port)
   "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
   (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
-  (let* ((remote-dir (if host (format "/ssh:%s:" host) default-directory))
+  (let* ((current-buf (buffer-file-name))
+         (tramp-file-regexp "/ssh:\\(.+@\\)?\\(.+?\\)\\(:\\|#\\).+")
+         (remote-dir (cond
+                      ;; If current buffer is a TRAMP buffer and its host is
+                      ;; the same as HOST, reuse its connection parameters for
+                      ;; SSH tunnel.
+                      ((and (string-match tramp-file-regexp current-buf)
+                            (string= host (match-string 2 current-buf))) current-buf)
+                      ;; Otherwise, if HOST was provided, use it for connection.
+                      (host (format "/ssh:%s:" host))
+                      ;; Use default directory as fallback.
+                      (t default-directory)))
          (ssh (or (executable-find "ssh")
                   (error "[nREPL] Cannot locate 'ssh' executable")))
          (cmd (nrepl--ssh-tunnel-command ssh remote-dir port))
@@ -598,11 +609,12 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
     ;; forwarding is set up, which is used to synchronise on, so that
     ;; the port forwarding is up when we try to connect.
     (format-spec
-     "%s -v -N -L %p:localhost:%p %u'%h'"
+     "%s -v -N -L %p:localhost:%p %u'%h' %n"
      `((?s . ,ssh)
        (?p . ,port)
        (?h . ,v-host)
-       (?u . ,(if v-user (format "-l '%s' " v-user) ""))))))
+       (?u . ,(if v-user (format "-l '%s' " v-user) ""))
+       (?n . ,(if v-port (format "-p '%s' " v-port) ""))))))
 
 (autoload 'comint-watch-for-password-prompt "comint"  "(autoload).")
```

I'd then change the documentation like this:
```diff
 You can safely run `cider-jack-in-*` while working with remote files over TRAMP. CIDER
-will handle this use-case transparently for you.
+will reuse existing SSH connection's parameters (like port and username) for establishing SSH tunnel.
+The same will happen if you try to `cider-connect-*` to a host that matches the one you're currently
+connected to.
 
 == Connecting via unix domain file socket
```

To be honest, I don't know what else can be written about it. I don't use TRAMP that much, too. I just decided to try it out because REPL starts too slow on my laptop :D

bbatsov:
> I guess TRAMP connection should only be taken into account if its host is the same as the one selected by user.

> To be honest, I don't know what else can be written about it. I don't use TRAMP that much, too. I just decided to try it out because REPL starts too slow on my laptop :D

Yeah, I get it. I mostly meant to outline what are the assumptions we're making to figure out which SSH host to target. It's fine to keep this short.

OknoLombarda:
Well, I think that's all

bbatsov:
Can you just rebase on top of `master` to address the merge conflict?

OknoLombarda:
Done

Issue #3061, opened 2021-10-01
nbb nrepl-server: can't send expressions from .cljs file


borkdude:
## Expected behavior

I expect to be able to send expressions from a `.cljs` file in `clojurescript-mode` to an nbb nrepl server.

## Actual behavior

Nothing happens until I activate `clojure-mode` in a `.cljs` file.

## Steps to reproduce the problem

```
npm install -g nbb
nbb nrepl-server :port 1338
```

Then `cider-connect` from emacs to that server and try to evaluate some expressions from a `.cljs` file:

foo.cljs:
``` clojure
(defn version []
  js/process.version)

(version)

(require '["fs" :as fs])

(fs/existsSync "README.md")
(fs/existsSync "README2.md")

(require '[clojure.string :as str])

(first (str/split js/process.env.PATH ":"))
``` 

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; Connected to nREPL server - nrepl://localhost:1337
;; CIDER 1.2.0snapshot (package: 20210914.1315)
```

### Emacs version

27.2

### Operating system

macOS

bbatsov:
Do you have the same problem with `cider-connect-cljs`? I'm guess "yes" as this still assumes the Clojure nREPL server, but I'm asking just in case.

borkdude:
With `cider-connect-cljs` it's the same + the `clojure-mode` workaround doesn't work there.

bbatsov:
Got it.

borkdude:
Thanks for looking into this.

benjamin-asdf:
with cider `318fe6878d8bedf5db9dfa649dedb45d72b2e7ee`
I try 
- run `nbb nrepl-server :port 1338`
- `cider-connect-clj`
- then cider tries to call some classpath functions or something that are not there
falls back to reading from minibuffer 

output from nbb server: 
```

Connection accepted
request {:op "clone", :id "1"}
response {"new-session" "a70ec1ab-299e-494f-a6d9-c83c54e36511", "status" ["done"], 
"id" "1"}
request {:op "clone", :id "2"}
response {"new-session" "e4b9cc25-91a2-48b5-bd24-b13dbccc6035", "status" ["done"], 
"id" "2"}
request {:op "describe", :session "a70ec1ab-299e-494f-a6d9-c83c54e36511", :id "3"}
[...]

Issue #3255, opened 2022-10-13
Smoother repl start for sci based repls


benjamin-asdf:
I would like to do this: 

```elisp
(cider-register-cljs-repl-type 'nbb-or-scittle-or-joyride "(+ 1 2 3)")

(defun mm/cider-connected-hook ()
  (when (eq 'nbb-or-scittle-or-joyride cider-cljs-repl-type)
    (setq-local cider-show-error-buffer nil)
    (cider-set-repl-type 'cljs)))

(add-hook 'cider-connected-hook #'mm/cider-connected-hook)

(defun mm/cider-jack-in-nbb ()
  (interactive)
  (cider-jack-in-cljs
   '(:jack-in-cmd
     "nbb nrepl-server"
     :cljs-repl-type
     nbb-or-scittle-or-joyride)))
```
This almost works except that `cider--check-cljs` ends up throwing because "require 'cljs.core" in sci based repls throws an error. 
Here is a proposal: 

```elisp
(defun cider-verify-clojurescript-is-present ()
  "Check whether ClojureScript is present."
  (unless
      (nrepl-dict-get (cider-sync-tooling-eval "cljs.core/inc") "value")
    (user-error "ClojureScript is not available.  See https://docs.cider.mx/cider/basics/clojurescript for details")))
```

checking cljs.core/inc should work everywhere. I already checked my nbb command and connecting to a shadow project repl. 
Slack conversation: 
https://clojurians.slack.com/archives/C029PTWD3HR/p1665687978090599

benjamin-asdf:
I realized another hurdle for a command like this is that `cider-jack-in-resolve-command` and similar commands have a predefined set of project types. Irrc there is another discussion around this already? Maybe a really big alist is being called for? 
Here is my current command: 

```elisp
(defun mm/cider-jack-in-nbb ()
  "Start a nbb nrepl process and connect."
  (interactive)
  (let* ((cider-allow-jack-in-without-project t)
	 (orig-buffer (current-buffer))
	 (params '(:jack-in-cmd "nbb nrepl-server :port 0"
				:cljs-repl-type nbb-or-scittle-or-joyride))
	 (params (cider--update-project-dir
		  params)))
    (nrepl-start-server-process
     (plist-get params :project-dir)
     (plist-get params :jack-in-cmd)
     (lambda (server-buffer)
       (with-current-buffer
	   orig-buffer
	 (cider-connect-sibling-cljs
	  params
	  server-buffer))))))
```

[...]

PR #3272, opened 2022-12-01
Allow connecting to nbb via connect-clj (#3061)

benjamin-asdf:
This is for nbb, scittle, joyride.
Better support for "pain" repls, support "cljs" without setup.

we want to

1. connect with a plain client
2. make any assumptions explicit
3. connect cljs buffers with those plain repls

1. Check for cider middleware being present, before using it (already being done)
2. Check for cider-library-present-p, before relying on anything in the runtime
3. Make assumptions about the runtime explicit 

My suggestion is =cider-connection-capabilities=.
Currently, there was an implicit assumption about the compilation error format.

=cider-connection-capabilities=
Is my suggestion for 2.

=cider-repls=
Now returns cljs, if the repl capabilities include 'cljs This way we can make a "plain" clj client, upgrade on connect with cljs capability and have it be connected in cljs buffers.

This is more a concession / workaround the current repl-type setup.

In the future we might get rid of repl-type?
The only reason we have it, and creaating issue 3., is because we want to be ergonomic about which buffer to use inside a given source buffer.
I found that I am able to juggle multiple clients by swapping to the buffer I want, thereby setting it active.

This can be a user command. Can also make a modeline element that swaps between clj/cljs like calva.

slack discussion:
https://clojurians.slack.com/archives/C04CAKAGADU

--- 

This pr also includes a few less consequential fixes that made nbb, scittle and joyride work.
(there was an error in the default repl init form "no such var, repl requires" )

- [x] The commits are consistent with our [contribution guideglines](../blob/master/.github/CONTRIBUTING.md)
  - [byte-compilation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Byte-Compilation.html),
    [`checkdoc`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips.html),
    [check-declare](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declaring-Functions.html),
    packaging metadata, indentation, and trailing whitespace checks.

	I ran it but it is currently cluttered. I am still 90% certain I
       fixed the ones I would be introducing.

	If some warnings are ignored, please explain how to ignore.

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -453,13 +453,13 @@ about this buffer (like variable `cider-repl-type')."
 (defvar-local cider-connection-capabilities '()
   "A list of some of the capabilites of this connection buffer.
 Aka what assumptions we make about the runtime.
-This is more genaral than
+This is more general than
 `cider-nrepl-op-supported-p' and `cider-library-present-p'.
 But does not need to replace them.")
 
-(defun cider-connection-has-capability-p (capability)
+(defun cider-connection-has-capability-p (connection-buffer capability)
   "Return non nil when the cider connection has CAPABILITY."
-  (with-current-buffer (cider-current-repl)
+  (with-current-buffer connection-buffer
     (member capability cider-connection-capabilities)))
 
 
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -343,6 +343,19 @@ buffer."
        (when cider-auto-mode
          (cider-enable-on-existing-clojure-buffers))
 
+       (setf cider-connection-capabilities
+             (append
+              (pcase (cider-runtime)
+                ('clojure '(clojure jvm-compilation-errors))
+                ('babashka '(babashka jvm-compilation-errors))
+                (_ '()))
+              (when
+                  ;; see `cider-sync-tooling-eval', but it is defined on a higher layer
+                  (nrepl-dict-get
+                   (nrepl-sync-request:eval "cljs.core/demunge" (current-buffer) nil 'tooling)
+                   "value")
+                '(cljs))))
+
        (run-hooks 'cider-connected-hook)))))
 
 (defun cider--disconnected-handler ()
@@ -436,6 +449,18 @@ about this buffer (like variable `cider-repl-type')."
               (or (cider--project-name nrepl-project-dir) "<no project>")
               (plist-get nrepl-endpoint :host)
               (plist-get nrepl-endpoint :port))))))
+
+(defvar-local cider-connection-capabilities '()
+  "A list of some of the capabilites of this connection buffer.
+Aka what assumptions we make about the runtime.
+This is more genaral than
+`cider-nrepl-op-supported-p' and `cider-library-present-p'.
+But does not need to replace them.")
+
+(defun cider-connection-has-capability-p (capability)
+  "Return non nil when the cider connection has CAPABILITY."
+  (with-current-buffer (cider-current-repl)
+    (member capability cider-connection-capabilities)))
 
 
 ;;; Connection Management Commands
@@ -885,7 +910,13 @@ no linked session or there is no REPL of TYPE within the current session."
     (cond ((null buffer-repl-type) nil)
           ((or (null type) (eq type 'multi) (eq type 'any)) t)
           ((listp type) (member buffer-repl-type type))
-          (t (string= type buffer-repl-type)))))
+          (t
+           (or (string= type buffer-repl-type)
+               (let ((capabilities
+                      (buffer-local-value 'cider-connection-capabilities buffer)))
+                 (cond ((listp type)
+                        (cl-some (lambda (it) (member it capabilities)) type))
+                       (t (member type capabilities)))))))))
 
 (defun cider--get-host-from-session (session)
   "Returns the host associated with SESSION."
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -457,9 +457,10 @@ This is more general than
 `cider-nrepl-op-supported-p' and `cider-library-present-p'.
 But does not need to replace them.")
 
-(defun cider-connection-has-capability-p (connection-buffer capability)
-  "Return non nil when the cider connection has CAPABILITY."
-  (with-current-buffer connection-buffer
+(defun cider-connection-has-capability-p (capability :optional connection-buffer)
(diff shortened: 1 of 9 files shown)


benjamin-asdf:
Don't merge I need to figure out a better way for checking "cljs present".

bbatsov:
> Don't merge I need to figure out a better way for checking "cljs present".

You can also mark PRs as "Draft" to make this visually clear, btw.

cider-eval.el
```diff
@@ -498,9 +505,10 @@ into a new error buffer."
 (defun cider-default-err-handler ()
   "This function determines how the error buffer is shown.
 It delegates the actual error content to the eval or op handler."
-  (if (cider-nrepl-op-supported-p "stacktrace")
-      (cider-default-err-op-handler)
-    (cider-default-err-eval-handler)))
+  (cond ((cider-nrepl-op-supported-p "stacktrace") (cider-default-err-op-handler))
```
bbatsov, line 508: Seems to me this change won't result in anything different.

bbatsov: Ah, actually the similar names confused me. Now I see the last branch is different.

cider-connection.el
```diff
@@ -437,6 +450,18 @@ about this buffer (like variable `cider-repl-type')."
               (plist-get nrepl-endpoint :host)
               (plist-get nrepl-endpoint :port))))))
 
+(defvar-local cider-connection-capabilities '()
+  "A list of some of the capabilites of this connection buffer.
+Aka what assumptions we make about the runtime.
+This is more genaral than
```
bbatsov, line 456: general

cider-connection.el
```diff
@@ -437,6 +450,18 @@ about this buffer (like variable `cider-repl-type')."
...
+`cider-nrepl-op-supported-p' and `cider-library-present-p'.
+But does not need to replace them.")
+
+(defun cider-connection-has-capability-p (capability)
```
bbatsov, line 460: Shouldn't this operate only on the current buffer? Probably it'd be nice to be able to pass a connection to it explicitly.

cider-overlays.el
```diff
@@ -303,6 +303,7 @@ focused."
   (let* ((font-value (if cider-result-use-clojure-font-lock
                          (cider-font-lock-as-clojure value)
                        value))
+         (font-value (string-trim-right font-value))
```
bbatsov, line 306: I don't quite understand the purpose of this change.

benjamin-asdf: nbb errors always had a trailing newline and the I think the visual of the overlays is better without the newline.

cider.el
```diff
@@ -778,9 +778,13 @@ Generally you should not disable this unless you run into some faulty check."
   :safe #'booleanp
   :package-version '(cider . "0.17.0"))
 
+(defun cider-clojurescript-present-p ()
+  "Return non nil when ClojureScript is present."
+  (nrepl-dict-get (cider-sync-tooling-eval "cljs.core/demunge") "value"))
```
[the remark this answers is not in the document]
benjamin-asdf: #3255 Basically `(require 'cljs.core)` did not work for nbb.

cider-connection.el
benjamin-asdf: :+1:

cider-connection.el
benjamin-asdf: :+1:

cider-overlays.el
bbatsov: Got it. This name `font-value` confused me, as that's just value (which is potentially using font-locking). I was wondering why someone would be trimming the font-value. :D

cider-repl.el
```diff
@@ -153,7 +153,8 @@ you'd like to use the default Emacs behavior use
 (make-obsolete-variable 'cider-repl-print-level 'cider-print-options "0.21")
 
 (defvar cider-repl-require-repl-utils-code
-  '((clj . "(clojure.core/apply clojure.core/require clojure.main/repl-requires)")
+  '((clj . "(when-let [requires (resolve 'clojure.main/repl-requires)]
```
bbatsov, line 156: You'll have to explain me the need for this change.

cider-connection.el
```diff
@@ -437,6 +450,18 @@ about this buffer (like variable `cider-repl-type')."
...
               (plist-get nrepl-endpoint :port))))))
 
+(defvar-local cider-connection-capabilities '()
+  "A list of some of the capabilites of this connection buffer.
+Aka what assumptions we make about the runtime.
+This is more general than
+`cider-nrepl-op-supported-p' and `cider-library-present-p'.
+But does not need to replace them.")
+
+(defun cider-connection-has-capability-p (connection-buffer capability)
```
bbatsov, line 460: You can also swap the params if you want to make the connection optional (e.g. the current buffer or current-connection buffer by default).

bbatsov:
Overall I like how this is shaping up. Apart from my small remarks I think it'd be nice to document somewhere that you can use `cider-connect-clj` for nbb & friends. Of course, after #3275 you'll be able to connect to native ClojureScript repls normally as well, but in this case I think more is better. :-) I guess we'll quickly make a lot of progress on that front.

cider-repl.el
benjamin-asdf: When you currently connect to joyride it throws "cannot resolve symbol repl-requires". It was the same for an earlier version of nbb.

cider-connection.el
benjamin-asdf: ok!

cider-repl.el
bbatsov: Might be good to mention such things in comments, so those things are clearer someone doesn't optimize this away. As @ikappaki mentioned in #3274 we don't really have integration tests, so it's not hard to imagine someone doing a change that would break this again for other runtimes.

bbatsov: We should also document somewhere that such REPL automatic util requires are limited to Clojure and the hosted ClojureScript.

benjamin-asdf:
I added a doc in the docs `scittle_and_friends.adoc`

bbatsov:
Thanks for tackling this! 🙇‍♂️

mk:
@benjamin-asdf @bbatsov great to see this, thank you both!

jackrusher:
@benjamin-asdf @bbatsov :D

Issue #3274, opened 2022-12-01
Feature req: Add integration tests


ikappaki:
Change issues do not usually manifest themselves unitl tested at a hight level with they tools they are used with.

There's no integration tests currently available in CIDER so that any changes can be tested at a high level, e.g. jacking in to a clojure cli deps.clj project. They may test localy, but only for certain featuers and certain architecture, thus perhaps introducing bugs in the codebase without use being aware. 

**Describe the solution you'd like**

Introduce a new test command that will invoke the integration tests. The first integration test can be jacking-in to all supported project types. The tests are added to CI.

**Describe alternatives you've considered**

Running integration manually for one architecture with what I think is suffucinent for the problem at hand. This is very subjective and might miss crucial points changed with the commit.

**Additional context**

I'm about to introduce support for `nbb` and would like to ensure that I haven't broken anything else at the jack in level.

I have created a prototype that appears to work:

1. A new test options is introduced, `eldev test --test-type integration` that will invoke the integration tests, everything else works exactly as before.
2. An integration test has been written to jack in to clojure tools project and test basic interaction. 
3. GH worfklow action to install project tools and test across macos, ubuntu and windows.

Happy to take this forward, please let me know of your thoughts.

bbatsov:
I'd welcome work in this direction - the main reason this didn't happen so far was lack of time and plenty of gaps in the test coverage in general. Even at the unit level most of the code base is untested.

bbatsov:
I think we can close this now.

PR #3275, opened 2022-12-02
Support nbb as native cljs repl

ikappaki:
Hi,

(while I was opening this PR, I've realised that @benjamin-asdf has arleady submitted #3272 to address the same using a more general methodology through the clj REPL. I don't think there is a source code conflict between the two. His PR appears to be allowing the clj REPL to upgrade to cljs and thus support a variaty of cljs cases, this PR is specific to nbb to support it as a native cljs REPL and goes through the established root of declaring a new project tool.  Thus I provisionally open this PR for deliberation only and should be declined in favor of benjamin's if it conflicts in any way or is considered overlapping/obsolete)

Could you please consider PR to support nbb as a native cljs repl when jacking in or connecting to cljs. Addresses #3272 

The logic will consider a project as nbb based on whether a `package.json` is present. A user can jack in using `cider-jack-in-cljs` or `cider-connect-*-cljs`

The previous logic assumed that all cljs REPLs are clojure repls that required some form to switch them over to cljs. With this update a particular project type can be declared as native cljs by not declaring any repl init forms.

I've included some tests to test the above.

It will be nice if we have integration tests to confirm we haven't broken anything when having such high level changes that cannot be unit tested (e.g. the jack in command), thus raised #3274 to discuss whether we can bring this forward.

I will also need to add nbb documentation if this is accepted. Thanks

-----------------

Thanks!

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -235,21 +235,21 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   "The command used to execute nbb."
   :type 'string
   :safe #'stringp
-  :package-version '(cider . "1.2.0"))
+  :package-version '(cider . "1.3.0"))
 
 (defcustom cider-nbb-global-options
   nil
   "Command line options used to execute nbb."
   :type 'string
   :safe #'stringp
-  :package-version '(cider . "1.2.0"))
+  :package-version '(cider . "1.3.0"))
 
 (defcustom cider-nbb-parameters
   "nrepl-server"
   "Params passed to nbb to start an nREPL server via `cider-jack-in'."
   :type 'string
   :safe #'stringp
-  :package-version '(cider . "1.2.0"))
+  :package-version '(cider . "1.3.0"))
 
 (defcustom cider-jack-in-default (if (executable-find "clojure") 'clojure-cli 'lein)
   "The default tool to use when doing `cider-jack-in' outside a project.
--- a/cider.el
+++ b/cider.el
@@ -1715,7 +1715,7 @@ PROJECT-DIR defaults to current project."
                         (shadow-cljs . "shadow-cljs.edn")
                         (gradle      . "build.gradle")
                         (gradle      . "build.gradle.kts")
-                        (nbb         . "package.json"))))
+                        (nbb         . "nbb.edn"))))
     (delq nil
           (mapcar (lambda (candidate)
                     (when (file-exists-p (cdr candidate))
--- a/cider.el
+++ b/cider.el
@@ -12,7 +12,7 @@
 ;; Maintainer: Bozhidar Batsov <bozhidar@batsov.dev>
 ;; URL: http://www.github.com/clojure-emacs/cider
 ;; Version: 1.6.0-snapshot
-;; Package-Requires: ((emacs "26") (clojure-mode "5.15.1") (parseedn "1.0.6") (queue "0.2") (spinner "1.7") (seq "2.22") (sesman "0.3.2"))
+;; Package-Requires: ((emacs "26") (clojure-mode "5.16.0") (parseedn "1.0.6") (queue "0.2") (spinner "1.7") (seq "2.22") (sesman "0.3.2"))
 ;; Keywords: languages, clojure, cider
 
 ;; This program is free software: you can redistribute it and/or modify
@@ -1296,7 +1296,7 @@ server buffer, in which case a new session for that server is created."
        (cider--update-cljs-type)
        (cider--update-cljs-init-function)
        (plist-put :session-name ses-name)
-       (plist-put :cider-repl-cljs-upgrade-pending t)))))
+       (plist-put :repl-type 'cljs)))))
 
 ;;;###autoload
 (defun cider-connect-clj (&optional params)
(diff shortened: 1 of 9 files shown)


bbatsov:
I like your proposal, as it's quite in line with how things are structured today in CIDER and it making it possible to use `cider-jack-in-cljs` without the hardcoded connection upgrade logic. I'll provide a bit more inline feedback, but I can tell from now that I'll accept this PR.

cider.el
```diff
@@ -230,6 +230,27 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.2.0"))
 
+(defcustom cider-nbb-command
+  "nbb"
+  "The command used to execute nbb."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.2.0"))
```
bbatsov, line 238: The next release will be 1.3.0.

bbatsov:
Actually the code is so good that I don't even have much feedback on it. :D Well done!

Now we need just a bit of docs and maybe those integration tests that you've suggested.

cider.el
ikappaki: Done.

ikappaki:
> Now we need just a bit of docs and maybe those integration tests that you've suggested.

I would also like some feedback from @benjamin-asdf whether our patches overlap, I dont' think so, because his is more like a general approach to upgrade a clj to cljs which addresses a wider set of problems. If they overlap, I'd rather support his, since he was submitted the PR first.

@borkdude, I've used `project.json` as the project file to suggest this is likely an `nbb` project so that CIDER will either suggest using nbb or jack in to nbb without confirmation in abasense of any other project files. Is it a good choice of nbb "project file"? Also would be grateful if you look at the new nbb doc section and suggest if you need anything else to go into this (I've just referenced what was there already for babashka).

There is some change in general behaviour that I've noticed that I think it's good to mention for completion:
1. When jacking into a `shadow-cljs` with `shadow-cljs.edn` project (a primarly shadow project), CIDER will now ask the user whether they want to now use `shadow` or `nbb` (because of the `package.json`), instead of defaulting to shadow.
2. For every cljs project, including `nbb`, CIDER will by default ask the user which repl type the user wants to use and present them with the whole list of supported repls (shadow-select, nbb etc etc). It has been always like this, but now it looks silly for CIDER to ask a user who they are just selecting `nbb` (either directly or indirectly), which cljs repl they want to use from the full cljs repl list, given that there is only one cljs type that works, and this is `nbb`. Perhaps a further general improvement would be to reduce the cljs repl types available to only those support by the project tool.

I have clojure cli, shadow cljs and nbb included in the regression testing, perhaps I can commit the integration tests first followed by this patch? I think i only need to add one more, leiningen,. in the list and that would be sufficient.

Thanks

borkdude:
@ikappaki `nbb.edn` would be a good choice. I've never heard of `project.json`, do you mean `package.json`?

ikappaki:
> @ikappaki `nbb.edn` would be a good choice. I've never heard of `project.json`, do you mean `package.json`?

Yes sorry, `package.json`, is `nbb.edn` currently a thing being used?

borkdude:
@ikappaki Yes :) You can declare deps in it, just like in `bb.edn`. No `:tasks` though. It requires `bb` to download the deps.

benjamin-asdf:
@ikappaki It doesn't overlap I think. As long have the build tool commands it makes sense to add new ones I guess. So thanks! I still like my pr, or something like it, because I can connect to nbb, scittle, joyride with 1 command and I feel like that is a good direction to go.

ikappaki:
> @ikappaki It doesn't overlap I think. As long have the build tool commands it makes sense to add new ones I guess. So thanks! I still like my pr, or something like it, because I can connect to nbb, scittle, joyride with 1 command and I feel like that is a good direction to go.

@thanks @benjamin-asdf, is good to know :)

ikappaki:
> I have clojure cli, shadow cljs and nbb included in the regression testing, perhaps I can commit the integration tests first followed by this patch? I think i only need to add one more, leiningen,. in the list and that would be sufficient.

Hi @bbatsov, now that the integration tests have been merged, can we have clojure-mode with [nbb support](https://github.com/clojure-emacs/clojure-mode/issues/641) released so I can resume work on this PR please? (as a reminder, we needed nbb support in clojure-mode to locate the root dir of the project)

bbatsov:
I'm on the road right now without a computer. I can cut a new release tomorrow evening or on Monday.

ikappaki:
> I'm on the road right now without a computer. I can cut a new release tomorrow evening or on Monday. 

No worries, thanks, we are not in a rush :)

bbatsov:
@ikappaki Done - https://github.com/clojure-emacs/clojure-mode/releases/tag/v5.16.0

ikappaki:
Hi @bbatsov,

I think PR is complete with latest commit ready for review
1. Upgraded `clojure-mode` to 5.16.0 to pick up nbb support
2. Incorporated latest  pending cljs changes from mainline and updated tests
3. Introduced nbb integration test

Thanks

bbatsov:
Great work! 🙇‍♂️

Issue #3277, opened 2022-12-04
Error in process filter: ClojureScript is not available when jack'ing in to shadow


ikappaki:
## Expected behavior

Able to jackin to shadow 

## Actual behavior

```
error in process filter: cider-verify-clojurescript-is-present: ClojureScript is not available.  See https://docs.cider.mx/cider/basics/clojurescript for details
```

## Steps to reproduce the problem

1. Install `nodejs`, and `shadow-cljs` with `npm install shadow-cljs -g`
2. Create a an empty `shadow-cljs.edn` and a basic `package.json` in a new directory
_shadow-cljs.edn_
``` clojure
{}
```
_package.json_
```json
{"dependencies":
 {"shadow-cljs": "^2.20.13"}}
```
2. Run `npm install`.
3. Open `shadow-cljs.edn` in Emacs, and `cider-jack-in-cljs`, select `shadow` for ClojureScript REPL type and `node-repl` for the shadow-cljs build.
4. The above error is returned.

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; CIDER 1.6.0-snapshot (package: 1.6.0-snapshot), nREPL 1.0.0
;; Clojure 1.11.1, Java 17.0.5
```

### Lein / Clojure CLI version

shadow-cljs@2.16.8

### Emacs version

28.1

### Operating system

MS-Windows

### JDK distribution

Oracle

This is most likely due to https://github.com/clojure-emacs/cider/pull/3272 cc @benjamin-asdf

bbatsov:
In the mean time you can set `cider-check-cljs-repl-requirements` to `nil` as a workaround for this.

bbatsov:
I'm also wondering if we should not kill this functionality at some point. It was mostly meant to to prevent people from trying to run something without the necessary deps and provide them with more meaningful error messages, but a few years down the road I'm somewhat doubtful that this really added much value.

benjamin-asdf:
@bbatsov My suggested fix for now is to check both, demunge for nbb and ```(cider-library-present-p "cljs.core")``` for "normal".

bbatsov:
@benjamin-asdf Your proposed fix seems reasonable to me.

bbatsov:
@benjamin-asdf Are you going to submit this as a PR?

pieterbreed:
For what it's worth; the suggested work-around does not work for me.

benjamin-asdf:
[...]

PR #3278, opened 2022-12-05
Introduce integration tests

ikappaki:
Hi, 

can you please consider patch to introduce integration tests to CIDER. It addresses #3274.

Other than adding integration tests for jack-in covering the major project tools, it also fixes a long standing issue with nREPL leaving orphaned child processes on MS-Windows after being killed,. The issue is fixed by contracting the external `taskkill` MS-Windows program to do the job.

The integration tests appear to work consistently across all platforms, and have already discovered a bug with shadow reported under #3277, thus it will currently show as failing. 

I haven't updated the changelog yet, waiting for some review feedback.

Detailed changelog in order of appears as follows:

1. Migrated macos regression tests to circle ci, but with a reduced set to only cover Emacs latest version.
2. Replaced existing macos only testing github action to run the integration tests across macos, ubuntu and windows latest on Emacs 28, 27 and 26. It was nearly impossible to do so on circleci across all platforms, thus a github action was chosen.
3. Updated the `Eldev` project file to support integration test via a new `--test-type` command line option (the template was taken from the `Eldev` tool's project file itself).
4. The cider server is given now a 0.5 sec opportunity to execute the `close` op before the server is killed.
5. The hacking section in documentation has been updated to mention integration tests.
6. The `nrepl--kill-process` has been updated to address the orphaned child process menace under MS-Windows, whereby the nREPL child processes were most likely to be left orphaned after the parent process was killed. The kill is now done by contracting the external `taskkill` windows program to do the job if available, and falls back to the previous logic otherwise.
7. The `nrepl-server-sentinel` has been simplified to only report an error if the nREPL process couldn't bootstrap itself, or send an exit message otherwise. The old code only closed client connections on `hangup` (i.e. HUP signal), but the new logic will close clients on any fatal signal. Not sure why the old code only did this on HUP, but it seems sensible to me that any open client connection should close when the server has exited.
8. Fixed an issue in `cider-tests.el` that a couple of buffer local shadow-cljs variables were set as global by mistake, thus affecting other tests (and in particular the integration tests).
9. Updated nrepl-client-tests.el to test the new nrepl plist property to indicate that the nREPL server has been successfully brought up, and also test that the new process-status `exit` returned on MS-Windows with `taskkill` is set.
10. Introduced integration tests covering jack-in for bb, clojure cli, lein and shadow.

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/test/integration/integration-tests.el
+++ b/test/integration/integration-tests.el
@@ -29,41 +29,18 @@
 (require 'cider)
 (require 'nrepl-dict)
 (require 'nrepl-tests-utils "test/utils/nrepl-tests-utils")
-
-;; https://emacs.stackexchange.com/a/55031
-(defmacro with-temp-dir (temp-dir &rest body)
-  "Create a temporary directory and bind it to TEMP-DIR while evaluating BODY.
-Remove the temp directory at the end of evaluation."
-  `(let ((,temp-dir (make-temp-file "" t)))
-    (unwind-protect
-      (progn
-        ,@body)
-      (condition-case err
-          (delete-directory ,temp-dir t)
-        (error
-         (message ":with-temp-dir-error :cannot-remove-temp-dir %S" err))))))
-
-(defun nrepl-client-connected?-ref-make! ()
-  "Return a reference to indicate when the client is connected to nREPL server.
-This is done by adding a hook to `cider-connected-hook` and is only active
-in the scope of the current buffer."
-  (let (connected?)
-    (add-hook 'cider-connected-hook
-              (lambda ()
-                (setq connected? t))
-              nil
-              ;; only set in the current buffer scope.
-              t)))
-
+(require 'integration-test-utils)
 
 (describe "jack in"
-  ;; See "bb" case for basic commentary
+  ;; See "babashka" case for commentary on the base template used by all other
+  ;; tests.
   ;;
   ;; It has been observed that some REPLs (Clojure cli, shadow) might take a
   ;; very long time to bring up/respond/shutdown, and thus sleep duration values
   ;; are set rather high.
 
   (it "to babashka"
+    (with-cider-test-sandbox
       (with-temp-dir temp-dir
         ;; set up a project directory in temp
         (let* ((project-dir temp-dir)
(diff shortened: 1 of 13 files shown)


bbatsov:
At first glance, everything looks pretty good to me. Might be a good idea to split the Windows fix to a separate commit, but whatever other feedback I have will likely be of cosmetic nature.

ikappaki:
> At first glance, everything looks pretty good to me. Might be a good idea to split the Windows fix to a separate commit, but whatever other feedback I have will likely be of cosmetic nature.

Yes, sorry, I've failed to mention that the windows child process fix is essential, otherwise the integration tests are intermittedly failing (especially the clojure cli jack in test) due to the incomplete kill.

ikappaki:
Hi @bbatsov,

now that https://github.com/clojure-emacs/cider/issues/3277 is fixed, I've added the following with the latest commit.
1. A changelog entry
2. An additional check to the each jack in test to wait for the client to connect to the server before proceeding further. This is needed because it has been observed that the client now sends asynchronous commands to the server to enquire about its capabilities. If the server response does not arrive before the test is compete, then the client handler raises an error. Waiting for the client to connect (and thus receive the server capabilities) aliviates this issue.

All integration tests now pass, are you happy to proceed with the review please?

Thanks

test/integration/integration-tests.el
```diff
@@ -43,6 +43,19 @@ Remove the temp directory at the end of evaluation."
         (error
          (message ":with-temp-dir-error :cannot-remove-temp-dir %S" err))))))
 
+(defun nrepl-client-connected?-ref-make! ()
```
bbatsov, line 46: Using `?` is not common in Elisp - here it'd be `is-connected` for a variable and `connected-p` for a predicate.

bbatsov:
Other than the small naming remark I made, everything looks reasonably good to me.

bbatsov:
I love the most recent set of changes! Well done! While there's always some room for improvement I think we can merge them at this point and re-iterate on them as we go. 

I'm curious to see them in action and in particular how stable will they be, as with integration tests of such complex software there's always the risk for some flakiness if we didn't account properly for something. Let's see!

ikappaki:
> I love the most recent set of changes! Well done! While there's always some room for improvement I think we can merge them at this point and re-iterate on them as we go.
> 
> I'm curious to see them in action and in particular how stable will they be, as with integration tests of such complex software there's always the risk for some flakiness if we didn't account properly for something. Let's see!

Thanks! happy to support and look after any potential issues that might come out of it, just let me know if something comes up :)

PR #3280, opened 2022-12-06
Fix cider-clojurescript-present-p #3277

benjamin-asdf:
I checked and first had the bug as described. With this change I don't have the bug.


bbatsov:
Thanks!

PR #3281, opened 2022-12-06
Replace newline chars with newlines in test report

expez:
When the library / app under test returns long error messages on test
failures it will often include newlines.  These are today displayed as
`\\n` and we get an error message on one very long line.

This commit turns these newline characters into actual newlines, which
makes the content of the `*cider-test-report*` buffer much more pleasant
to read.

Before:

image

After:
image

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -478,6 +478,12 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
                       (cider-test-render-assertion buffer test)))))
               vars))
            results)))
+      ;; Replace any newline chars with actual newlines to make long error
+      ;; messages more readable
+      (goto-char (point-min))
+      (while (search-forward "\\n" nil t)
+        (replace-match "
+"))
       (goto-char (point-min))
       (current-buffer))))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## master (unreleased)
 - Allow using `npx nbb` as `cider-nbb-command`.
 
+### Changes
+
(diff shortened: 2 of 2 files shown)


cider-test.el
```diff
@@ -478,6 +478,12 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
                       (cider-test-render-assertion buffer test)))))
               vars))
            results)))
+      ;; Replace any newline chars with actual newlines to make long error
+      ;; messages more readable
+      (goto-char (point-min))
+      (while (search-forward "\\n" nil t)
+        (replace-match "
```
bbatsov, line 485: Hmm, is there something else you can use that's visually clearer? I'm guessing an `\n` doesn't work, right?

bbatsov:
I'm okay with the fix, but it also needs a changelog entry. A regression test that verifies that this conversation works would be appreciated as well.

cider-test.el
expez: > I'm guessing an \n doesn't work, right?

Correct

expez:
> I'm okay with the fix, but it also needs a changelog entry.

Pushed 

>  A regression test  [..] would be appreciated as well.

Is there an existing file for these tests, where I can insert a few choice lines?

bbatsov:
Not sure - I don't see a `cider-test.el` in the tests, but it might be because it would have had the same name as the implementation file.

bbatsov:
You'll also need to rebase on top of `master`.

bbatsov:
@expez Can you at least rebase this on top of `master`? Even without a test it's a fix that I'd like to merge.

expez:
@bbatsov done.

Sorry about the slow response, travel and Christmas holiday got in the way.

Happy new year!

bbatsov:
Happy new year!

PR #3282, opened 2022-12-08
Fix cljs type not set correctly

benjamin-asdf:
I introduced a bug with 24b9891e0b4603dca7519427d02ff4af2cfac57a
that made cljs repls not upgrade automatically anymore to cljs (repl-type).
What actually happened was that after the init function, it was set an
ultimate time, because of the "demunge" eval call I added.

1. Move the connection capability block in fron of
=cider-nrepl-init-function=, this would by itself already fix it
because then we would eval as clj repl and in the end once as cljs
repl (which sets the repl-type see cider-repl--state-handler)

2. do not make the "demunge" eval call at all, when we already know
the repl is going to be cljs (cljs or pending-cljs).
(This would by itself also be a fix).

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -301,6 +301,27 @@ See command `cider-mode'."
   (unless (cider-sessions)
     (cider-disable-on-existing-clojure-buffers)))
 
+
+(defun cider--set-connection-capabilities (&optional conn-buffer)
+  "Set `cider-connection-capabilities' for CONN-BUFFER during repl init.
+See `cider-connection-capabilities'."
+  (with-current-buffer (or conn-buf (current-buffer))
+    (setf cider-connection-capabilities
+          (append
+           (pcase (cider-runtime)
+             ('clojure '(clojure jvm-compilation-errors))
+             ('babashka '(babashka jvm-compilation-errors))
+             (_ '()))
+           (when
+               (or
+                (member cider-repl-type '(cljs pending-cljs))
+                ;; This check is currently basically for nbb.
+                ;; See `cider-sync-tooling-eval', but it is defined on a higher layer
+                (nrepl-dict-get
+                 (nrepl-sync-request:eval "cljs.core/demunge" (current-buffer) (cider-current-ns) 'tooling)
+                 "value"))
+             '(cljs))))))
+
 (declare-function cider--debug-init-connection "cider-debug")
 (declare-function cider-repl-init "cider-repl")
 (declare-function cider-nrepl-op-supported-p "cider-client")
@@ -311,8 +332,7 @@ buffer."
   ;; `nrepl-connected-hook' is run in the connection buffer
   ;; `cider-enlighten-mode' changes eval to include the debugger, so we inhibit
   ;; it here as the debugger isn't necessarily initialized yet
-  (let ((cider-enlighten-mode nil)
-        (cljsp (member cider-repl-type '(cljs pending-cljs))))
+  (let ((cider-enlighten-mode nil))
     ;; after initialization, set mode-line and buffer name.
     (cider-set-repl-type cider-repl-type)
     (cider-repl-init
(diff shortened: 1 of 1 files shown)


cider-connection.el
```diff
@@ -311,7 +311,8 @@ buffer."
   ;; `nrepl-connected-hook' is run in the connection buffer
   ;; `cider-enlighten-mode' changes eval to include the debugger, so we inhibit
   ;; it here as the debugger isn't necessarily initialized yet
-  (let ((cider-enlighten-mode nil))
+  (let ((cider-enlighten-mode nil)
+        (cljsp (member cider-repl-type '(cljs pending-cljs))))
```
bbatsov, line 315: I don't see the point in binding this so far from it's only usage.

cider-connection.el
```diff
@@ -337,25 +338,28 @@ buffer."
          ;; first.
          (cider--debug-init-connection))
 
-       (when cider-repl-init-function
-         (funcall cider-repl-init-function))
-
-       (when cider-auto-mode
-         (cider-enable-on-existing-clojure-buffers))
-
        (setf cider-connection-capabilities
```
bbatsov, line 341: I think for the sake of readability the capabilities setting should be moved to their own function.

cider-connection.el
```diff
@@ -337,25 +338,28 @@ buffer."
...
                 ('clojure '(clojure jvm-compilation-errors))
                 ('babashka '(babashka jvm-compilation-errors))
                 (_ '()))
               (when
-                  ;; see `cider-sync-tooling-eval', but it is defined on a higher layer
-                  (nrepl-dict-get
-                   (nrepl-sync-request:eval "cljs.core/demunge" (current-buffer) nil 'tooling)
-                   "value")
+                  (or
+                   cljsp
```
bbatsov, line 349: I think for variables it's better to use a name like `is-cljs` as `p` stands for predicate, which by definition is a function.

bbatsov:
Apart from my remarks the fix seems reasonable.

bbatsov:
The lint CI check is failing because you've used tabs instead of spaces for indentation - please address this as well.

cider-connection.el
benjamin-asdf: ah nice

benjamin-asdf:
wasn't sure if I should amend my commit, made a commit with the same message.

benjamin-asdf:
@bbatsov  damm I accidentally have a lisp error int the function (conn-buff, conn-buffer) 
8aaf5585d4fc5ed6157897a1d9923a1ef9cd064c

ikappaki:
Hi @benjamin-asdf :)

I believe this change has caused a lint error, as per circle ci report:

=> ci/circleci: test-lint — Your tests failed on CircleCI 

```
eldev -dtT compile --warnings-as-errors
# ...
[00:00.603]  ELC      nrepl-client.el
[00:00.603]  Byte-compiling file `nrepl-client.el' early as `require'd from another file...
[00:00.603]  Loading file `nrepl-client.el' before byte-compiling it...
[00:00.710]  Loading the byte-compiled file `nrepl-client.elc'...

[00:00.740]  
             In cider--set-connection-capabilities:
             cider-connection.el:304:54: Error: Unused lexical argument `conn-buffer'
[00:00.741]  cider-connection.el:307:28: Error: reference to free variable `conn-buf'
[00:00.773]  
             In end of data:
             cider-connection.el:320:81: Error: the function `cider-current-ns' is not
                 known to be defined.
[00:00.774]  Saving target dependencies to file `.eldev/28.2/target-dependencies.build'...

[00:00.774]  Failed to byte-compile `cider-connection.el'
[00:00.774]  Finished erroneously on Thu Dec 8 15:55:14 2022

```

bbatsov:
@ikappaki Indeed. Some functions will need to be shuffled around if we really need to use `cider-current-ns` in this file, as `cider-client` depends on `cider-connection` and the function is actually defined in `cider-client`.

benjamin-asdf:
I have a new version of the function at a8b5c40de46a125c39acee0cbb5767dd8563fc2a it doesn't need to user `cider-current-ns` turns out. We can pass nil, seems to be fine.

PR #3289, opened 2022-12-11
Don't scale down time for overlay tests on MacOS

pestctrl:
Overlay tests involve testing the timer functionality of overlays, and require the use of sleep. The tests scale all time-related arguments by a certain factor to avoid overly inflating test-running times.

The scaling factor was initially set at 1 order of magnitude slower than the consistently succeeding value on my machine, but the value is apparently still too fast for the MacOS test runner, so I have added MacOS to the list of platforms unsafe to do this optimization for.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/test/cider-overlay-tests.el
+++ b/test/cider-overlay-tests.el
@@ -67,7 +67,7 @@ being set that way"
 (defun cider-overlay--safe-to-speed-up-tests ()
   (and (<= 28 emacs-major-version)
        (not (member system-type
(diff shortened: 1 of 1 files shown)


ikappaki:
Hi, just to clarify, the macos tests have moved to circleci under `ci/circle: test-macos-emacs-latest` with #3278 and it is green now with this PR:

image

So I think this PR is good to go.

bbatsov:
Thanks!

Issue #3287, opened 2022-12-10
When jacking in to shadow, the cider-repl-type is prematurely being reported as `cljs` while it is not (yet)


ikappaki:
Hi,

It appears that the latest logic in #3282 is prematurely reporting that a shadow nREPL session, created with `cider-jack-in-cljs`, has switched to `cljs` while it is still `clj`, which is wrong. 

E.g. calling `(cider-repls 'cljs nil)` just a little while after  a `cider-jack-in-cljs` for a node repl, returns a cljs cider REPL even though the shadow repl has not yet received the commands from the client to switch to cljs.

This misbehavior is caught by the cljs `jack in` `to-shadow` integration test.

The test does the following to test that jacking in to a shadow cljs REPL works (high level description):

1. Sets up an empty shadow-cljs project
2. Jacks in to it
3. Waits until a cljs REPL is available, by periodically calling `(cider-repls 'cljs nil)` until it returns a cider cljs repl.
4. Evals some clojurescript command on the REPL to check that is working. In this case the command is `(print :cljs? (some? *clojurescript-version*))`

The expectation here is that the cljs REPL will print out the clojurescript version, but the REPL returns the following error (requires the improved integration anlalytics https://github.com/clojure-emacs/cider/pull/3285 to see it)

```
"Syntax error compiling at (*cider-repl tmp/NiSeGW:localhost:44593(pending-cljs)*:1:15).\nUnable to resolve symbol: *clojurescript-version* in this context\n\"
```
This is because the REPL evaluated the code in Clojure but not in ClojureScript as expected. The issue is with step 3 above that is misreported that there is a cider cljs repl available, while there is none (yet) available.

This used to work as expected before #3282. Looking at the code, I've traced the error to be caused by the new connection capabilities as follows (high level description):

1. When jacking in to cljs, the `cider-repl-type` on the client buffer is set to `pending-cljs`.
2. When the client is conneted to the nREPL server, the new capabilities logic kicks in, and because the `cider-repl-type` is set to `pending-cljs`, the connection capability is set to `cljs`:
```elisp
(defun cider--set-connection-capabilities (&optional conn-buffer)
  "Set `cider-connection-capabilities' for CONN-BUFFER during repl init.
See `cider-connection-capabilities'."
  (with-current-buffer (or conn-buffer (current-buffer))
    (setf cider-connection-capabilities
          (append
           (pcase (cider-runtime)
             ('clojure '(clojure jvm-compilation-errors))
[...]

PR #3291, opened 2022-12-14
Reify cider-cljs-pending separate from repl-type

benjamin-asdf:
This pull request addresses https://github.com/clojure-emacs/cider/issues/3287 by separating the concept of a repl type from the concept of pending status. The pending variable is now set when the repl is initialized, and it is set to `nil` when we are sure that we have a cljs repl in `cider-repl--state-handler.`

In `cider-repls`, we no longer return pending repls, which is the same semantic as before (calling `(cider-repls 'cljs) or (cider-repls 'clj) or (cider-repls nil)` does not return pending repls. The only potential use case that is broken with this change is if a user has written an emacs lisp function that calls `(cider-repls 'pending-cljs)`. We could add a legacy special case for this, where calling cider-repls with 'pending-cljs returns the repls that are `cider-cljs-pending-p`.

Overall, this change will prevent future headaches by clearly separating the concepts of repl type and pending status.


bbatsov:
> The only potential use case that is broken with this change is if a user has written an emacs lisp function that calls (cider-repls 'pending-cljs). We could add a legacy special case for this, where calling cider-repls with 'pending-cljs returns the repls that are cider-cljs-pending-p.

Pretty sure that's not an use-case we should worry about, as I can't see why someone would be doing this. Your proposed solution is reasonable. I agree that the pending upgrade status is not really a REPL type, but rather an internal implementation detail.

cider-repl.el
```diff
@@ -224,6 +224,8 @@ This cache is stored in the connection buffer.")
       (nrepl-dbind-response response (repl-type changed-namespaces)
         (when (and repl-type cider-repl-auto-detect-type)
           (cider-set-repl-type repl-type))
+        (when (string= repl-type 'cljs)
```
bbatsov, line 227: You should use `string=` to compare something with a symbol - `eq` is what you're looking for here.

cider-repl.el
```diff
@@ -224,6 +224,8 @@ This cache is stored in the connection buffer.")
...
+          (setf cider-cljs-repl-pending nil))
```
bbatsov, line 228: It's better to use `setq` here.

cider-connection.el
```diff
@@ -762,10 +762,17 @@ Session name can be customized with `cider-session-name-template'."
 (defvar-local cider-repl-type nil
   "The type of this REPL buffer, usually either clj or cljs.")
 
+(defvar-local cider-cljs-repl-pending nil
```
bbatsov, line 765: I think this might be a good time to change the terminology to the more descriptive `cider-repl-cljs-upgrade-pending`. This will make the code more readable across the board.

bbatsov:
The proposed change needs a changelog entry as well, as we're removing a REPL type (even if it was kind of fake).

cider-repl.el
benjamin-asdf: I tried this with `eq` first. `repl-type` is a string at that point. Could instead call `cider-maybe-intern` on the type and say `eq` ?

cider-connection.el
benjamin-asdf: :+1:

cider-repl.el
benjamin-asdf: (updated version seems better)

cider-repl.el
```diff
@@ -224,6 +224,8 @@ This cache is stored in the connection buffer.")
       (nrepl-dbind-response response (repl-type changed-namespaces)
         (when (and repl-type cider-repl-auto-detect-type)
           (cider-set-repl-type repl-type))
+        (when (eq (cider-maybe-intern repl-type) 'cljs)
```
bbatsov, line 227: Ah, if that's a string than the comparison should be `(string= repl-type "cljs")`. I just saw `'cljs` was a symbol and I assumed that `repl-type` was a symbol as well.

bbatsov:
I'll merge this as is and I'll make a couple of small tweaks myself, so I won't waste your time with them.

PR #3294, opened 2022-12-16
Start bb nREPL server at a random port

### LINKED ISSUE #3293: Can't jack in to more than one bb projects
## Expected behavior

Should be able to jack in to more than one bb projects

## Actual behavior

```
You can run the command ‘cider-jack-in-clj’ with C-c M-j
error in process sentinel: Could not start nREPL server: Exception in thread "main" java.net.BindException: Address already in use (Bind failed)
```

1. Create two  `bb.edn` files with `{}` contents in separate new directories
2. Jack in to the first directory, a bb session is established.
3. Try to jack in to the second directory, the above error is thrown.

This is because the bb nREPL server is always start up on the same port at 1667.

any

PR to follow.

ikappaki:
Hi,

could you please consider PR to start bb nREPL server at a random port. It fixes #3293.

The `:0` format eligibility to start the server has been confirmed by @borkdude on chat.

I did not bother to include a test for such a small change, but here is the output from the bb integration test that shows port is now not 1667 but random:

```
[00:11.202]  [nREPL] Starting server via /opt/hostedtoolcache/Babashka/1.0.165/x64/bb nrepl-server :0
[00:11.209]  [nREPL] server started on 33711
[00:11.209]  [nREPL] Establishing direct connection to 127.0.0.1:33711 ...
[00:11.209]  [nREPL] Direct connection to 127.0.0.1:33711 established
[00:11.393]  Connected! In the absence of parentheses, chaos prevails.
[00:12.145]  [nREPL] Connection closed
[00:12.146]  nREPL server exited.
jack in
  to babashka
  to babashka (1.15s)
```

I've also piggybacked an update to the circleci job names to include `ubuntu` so names are aligned with the rest of the jobs, if you don't mind.

Thanks

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -224,7 +224,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :package-version '(cider . "1.2.0"))
 
 (defcustom cider-babashka-parameters
-  "nrepl-server"
+  "nrepl-server :0"
   "Params passed to babashka to start an nREPL server via `cider-jack-in'."
   :type 'string
   :safe #'stringp
--- a/cider.el
+++ b/cider.el
@@ -224,7 +224,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :package-version '(cider . "1.2.0"))
 
 (defcustom cider-babashka-parameters
-  "nrepl-server :0"
+  "nrepl-server localhost:0"
   "Params passed to babashka to start an nREPL server via `cider-jack-in'."
   :type 'string
   :safe #'stringp
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ commands:
           command: eldev -dtT compile --warnings-as-errors
 
 jobs:
-  test-emacs-26:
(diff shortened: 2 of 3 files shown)


cider.el
```diff
@@ -224,7 +224,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :package-version '(cider . "1.2.0"))
 
 (defcustom cider-babashka-parameters
-  "nrepl-server"
+  "nrepl-server :0"
```
borkdude, line 227: it might be more readable to write 127.0.0.1:0 here maybe?

ikappaki: I personally find `:0` intuitive, but I've updated it to `localhost:0` nevertheless so that we can also cover IPv6 as well.

benjamin-asdf:
ah lol you beat me to it. But `0` also works.

bbatsov:
Looks good to me. Thanks!

P.S. This PR reminded how useless the `*-global-params` variables are, as we never used them for anything. Probably we should just remove them across the board to simplify a bit the addition of new jack-in backends.

Issue #3179, opened 2022-04-09
Jack in to babashka repl without Clojure project


didibus:
Sometimes I like to start ephemeral REPL sessions to play around with some idea and I don't need to persist files or have a project for it.

Right now you can cider-jack-in without a project or a file, in some empty clojure buffer, and Cider asks you if you want to run cider-jack-in without a project, which will start a Clojure REPL.

I was wondering if we could have the same functionality but for a bb repl as well?

bbatsov:
That's just a matter of setting `cider-jack-in-default` to `babashka`. Perhaps we can also add option to prompt for the REPL type to use in such cases.

didibus:
Having an option prompt would be great. I don't always want it to be Babashka, I'd want to pick which one.

bbatsov:
Got it.

vlnn:
Instead of having extra-configuration or extra user input every time you start cider-jack-in why not to create something like **cider-jack-in-bb**?

bbatsov:
@vlnn That's a simple solution as well, but it's a slippery slope to have a jack-in command for every possible Clojure runtime. I'm not saying we won't do it, but I'm definitely cautious about this path.

vlnn:
@bbatsov I agree, but making jack-in process longer and requiring user input is a bit worrying. Actually worrying me a lot! And it wouldn't be easy to pre-configure destination, as sometimes user may want to jack-in "unusual" environment. So it would be nice to have some kind of agreement here (maybe temporary!).

didibus:
There could be a `cider-jack-in-dialect` or `cider-jack-in-prompt` (or whatever better name you can think of) that prompts for the dialect to jack-into. That way the standard jack-in can remain ClojureJVM. So there'd be at most one more jack-in for a multiple of runtimes.

bbatsov:
@didibus Technically speaking there's no "standard" `cider-jack-in` anymore. It's just an alias for `cider-jack-in-clj`. I renamed it when we originally introduced support for ClojureScript. 

But yeah - in general I was thinking of some command that prompts you for the backend to run, which might not even be a Clojure dialect down the road.

ikappaki:
May I suggest a way to jack in to any nREPL without a project by still using `cider-jack-in` but using numeric prefix arguments, something like:

1. `M-1 M-x cider-jack-in`: jack in to clojure cli
2. `M-2 M-x cider-jack-in`: >> lein
[...]

PR #3300, opened 2022-12-25
Add numeric prefix arg to `cider-jack-in' for non-projects

ikappaki:
Hi @bbatsov,

as per #3179, could you please consider a draft PoC  to  upgrade the `cider-jack-in` alias to a command that maintains its current behavior (i.e. calling `cider-jack-in-clj`) but in addition

1. When there is no current project, prompts user to select one to jack in from a list of `babashka`, `clojure-cli`, `lein` or `nbb`.
2. Accepts numeric prefix arguments to jack in to a project, even if there is project in the current dir (`M-1 M-X cider-jack-in` for `clojure-cli`, `M-2` for `lein`, `M-3` for `babashka` and `M-4` for `nbb`).

this PoC is open for discussion, so I haven't added a changelog entry or updated the documentation yet.

Thanks

Thanks!

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -86,6 +86,7 @@
 (require 'cider-debug)
 (require 'cider-util)
 
+(require 'cl-lib)
 (require 'tramp-sh)
 (require 'subr-x)
 (require 'seq)
@@ -345,6 +346,26 @@ Sub-match 1 must be the project path.")
 
 (defvar cider-host-history nil
   "Completion history for connection hosts.")
+
+(defvar cider-jack-in-universal-options
+  '((clojure-cli (:prefix-arg 1 :cmd (:jack-in-type clj  :project-type babashka :edit-project-dir t)))
+    (lein        (:prefix-arg 2 :cmd (:jack-in-type clj  :project-type clojure-cli :edit-project-dir t)))
+    (babashka    (:prefix-arg 3 :cmd (:jack-in-type clj  :project-type lein :edit-project-dir t)))
+    (nbb         (:prefix-arg 4 :cmd (:jack-in-type cljs :project-type nbb :cljs-repl-type nbb :edit-project-dir t))))
+  "The list of project tools that are supported by the universal jack in command.
+
+Each item in the list consists of the tool name and its plist options.
+
+The plist supports the following keys
+
+- :prefix-arg the numerical prefix arg to use to jack in to the tool.
+
+- :cmd a plist of instructions how to invoke the jack in command, with keys
+
+  - :jack-in-type 'clj to start a clj repl and 'cljs for a cljs repl.
+
+  - &rest the same set of params supported by the `cider-jack-in-clj' and
+    `cider-jack-in-cljs' commands.")
 
 ;;;###autoload
 (defun cider-version ()
@@ -1393,9 +1414,15 @@ non-nil, don't start if ClojureScript requirements are not met."
         (t params)))
 
 (defun cider--update-project-dir (params)
-  "Update :project-dir in PARAMS."
+  "Update :project-dir in PARAMS.
+
+Params is a plist with the following keys (non-exhaustive)
+
+ :edit-project-dir prompt (optional) ask user to confirm the project root
+ directory."
   (let* ((params (cider--update-do-prompt params))
-         (proj-dir (if (plist-get params :do-prompt)
+         (proj-dir (if (or (plist-get params :do-prompt)
+                           (plist-get params :edit-project-dir))
                        (read-directory-name "Project: "
                                             (clojure-project-dir (cider-current-dir)))
                      (plist-get params :project-dir)))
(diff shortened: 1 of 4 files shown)


bbatsov:
I'm okay with the general direction, but one thing I want us to do here is get jack-in to a point where it can work with a non-Clojure projects, so it'd warrant a different name (different from `cider-jack-in-clj`). I'm thinking this can be the "generic" jack-in that was discussed a few times in the past. 

On the other hand, this opens up the question of where exactly the quick selection of a project type should happen. Might be weird to have every project type in the generic command, might be fine as well.

bbatsov:
> When there is no current project, prompts user to select one to jack in from a list of babashka, clojure-cli, lein or nbb.

I still think this should be some configuration option for the existing command as well. Or in general - and then we can just have the default value be to prompt for the tool to use.

bbatsov:
One small practical problem to solve is also the keybindings - we'll need some new for `cider-jack-in` if we're going to have it as an universal command going forward. I also think it still makes sense to be able to specify at least the folder to use for the command somehow. (e.g. with a universal argument like the other jack-in commands do it)

ikappaki:
Hi @bbatsov,

thanks for the feedback, there are some points that are not clear to me since I was not involved in the previous discussions, if you could please clarify further.

I'll first start by summarising what I think it is happening at the moment and what the new logic is in its current state.

Current logic:
1. `clojure-jack-in` is an alias of `clojure-jack-in-clj`.
2. The user has to either invoke the `clojure-jack-in-clj` or `clojure-jack-in-cljs` commands to jack in to a Clojure or `ClojureScript` project respectively. If there is no current project, the user is warned and asked if they want to start a (clojure cli) nREPL without a project. (there's also `cider-jack-clj&cljs` which I haven't looked into its details and assume it behaves similarly to the other two).
3. They both accept a prefix argument, from the docstrings """With the prefix argument, allow editing of the jack in command; with a double prefix prompt for all these parameters."""

This PR logic:
1. `clojure-jack-in` is a new command, which functionally behaves the same as before (i.e. the same as `cider-jack-in-clj`,) except if a project cannot be found in the current dir, then it prompts the user for which project tool to invoke, or is started with a numeric prefix argument (`M-x-1` for clojure cli etc). Thus this PR extends the existing logic rather than replacing it.

> I'm okay with the general direction, but one thing I want us to do here is get jack-in to a point where it can work with a non-Clojure projects, so it'd warrant a different name (different from `cider-jack-in-clj`). I'm thinking this can be the "generic" jack-in that was discussed a few times in the past.

This PR uses `cider-jack-in`, so I think it matches with the above expectation?

> On the other hand, this opens up the question of where exactly the quick selection of a project type should happen. Might be weird to have every project type in the generic command, might be fine as well.

I assume here you are referring to the hardcoded project values (e.g. `clojure-cli`, `lein` etc) in the `cider-jack-in` and `cider--jack-in-to-project`  fns? I can move them out to something configurable if so, e.g. `defcustoms`.


> > When there is no current project, prompts user to select one to jack in from a list of babashka, clojure-cli, lein or nbb.
> 
> I still think this should be some configuration option for the existing command as well. Or in general - and then we can just have the default value be to prompt for the tool to use.

By existing command you mean `cider-jack-in-clj` and `cider-jack-in-cljs`? 

> One small practical problem to solve is also the keybindings - we'll need some new for `cider-jack-in` if we're going to have it as an universal command going forward. 

So the current key bindings work if cider mode is enabled (e.g. when visiting a clojure file):
- `C-c M-j` for `cider-jack-in-clj`,
- `C-c M-J` for `cider-jack-in-cljs` 

Perhaps `C-c M-i` for jack `i`n? It appears to be available.

> I also think it still makes sense to be able to specify at least the folder to use for the command somehow. (e.g. with a universal argument like the other jack-in commands do it)

We can make this mandatory, it always asks the user for the project dir with the default option being the current dir, so the user in the most common case has just to hit enter to accept.

Thanks

bbatsov:
My main point is that I don't want to confuse the users with too many jack-in commands, so it should be very clear to them what are they supposed to use. Let's be honest - almost nobody reads documentation. :-) Originally there was only `cider-jack-in`. Then we've added `cider-jack-in-cljs` and renamed the first command to `cider-jack-in-clj` fore symmetry, but we kept the alias around, so we wouldn't break anything. Then we've added a command to start both connections with a jack-in. I hope you see my point. 

With a name as generic as `cider-jack-in` it'd be nice to just be able to do pretty much everything (instead of being some slightly different version of let's say `cider-jack-in-clj`, while still have the behavior more or less the same as for the other commands. E.g. I don't like that one command will prompt for a project if it's not there, but the others won't. That's just bad (confusing) UX in my book. I hope that makes sense. 

> By existing command you mean cider-jack-in-clj and cider-jack-in-cljs?

I guess both, as they check the same config about the preferred build tool if I remember correctly. I was thinking there could be some setting that controls whether in the absence of a project commands prompt for some build tool or just use some default. 

> Perhaps C-c M-i for jack in? It appears to be available.

The above bindings are considered legacy and semi-deprecated (I think that's mentioned somewhere in the docs). I don't plan to added any similar keybindings going forward, just `C-c C-x ...`. 

> We can make this mandatory, it always asks the user for the project dir with the default option being the current dir, so the user in the most common case has just to hit enter to accept.

That'd be fine by me. 

> I assume here you are referring to the hardcoded project values (e.g. clojure-cli, lein etc) in the cider-jack-in and cider--jack-in-to-project fns? I can move them out to something configurable if so, e.g. defcustoms.

Yep, I don't think it's a good idea to have the project types that can be started quickly be hardcoded.

ikappaki:
Hi @bbatsov,

Just to summarise, there are two options discussed here:

1. This PR: build on top of what we currently have, repurpose `cider-jack-in`  to allow a user to jack in to a project tool without a current project. `pros`: building on top of current API maintaining existing behavior. `cons`: the current `cider-jack-in-*` API is considered to have overgrown its purpose and needs a redesign, and this change would make it harder to overhaul?
2. The current `cider-jack-in-*` interface has outlived its purpose and needs a redesign, that can better support cases such as jacking in to a project tool without a current project. The new API should only have only one entry point and this should be a  single `cider-jack-in` function. `pros`: a better API that would make it easier to maintain and extend. `cons`: breaking compatibility with previous versions (i.e. because `cider-jack-in-*` are removed)?

For case #2, do you have perhaps a general outline in mind how this could work? I can think of the following (based on ideas from this PR), but please do amend it with how you think the new interface should behave.

1. `cider-jack-in` based on the type of the current project (clj or cljs) either jack-in to the project directly or if there are multiple project types, it asks the user for confirmation which project to jack in to. If 
2. Running the above with single and double prefix argument allows editing fo the jack in command and edit all project parameters respectively.
3. If there is no current project, then the user is asked for a project tool to jack in, and the default directory to use. They can use a numeric prefix argument (e.g. `M-x 1`) to shortcut a jack in to project tool without a project, it will be also asked for the current directory.

Thanks

bbatsov:
Yeah, I think you've got the right idea with 1) and 2). I'm leaning towards 2) for a couple of reasons:

- I don't want us to have dedicated jack-in/connect commands for everything; generally only the hosted ClojureScript warranted some special handling because the "upgrade" procedure for connections and the common setup where people would start a Clojure and ClojureScript REPL right away. But I definitely don't see in the future things like `cider-jack-in-bb`, `cider-jack-in-nbb` or some different `cider-connect` commands. 
- I don't want to commit the name `cider-jack-in` to some temporary solution, given how good of name it is in general. 
- I'd love for us to clean up & simplify the jack-in codebase in general

> breaking compatibility with previous versions (i.e. because cider-jack-in-* are removed)?

Well, we can just remove them outright - if we go down this path the old functions likely stay as deprecated for a while to ensure some smooth transition for all users. 

As I realize that option 2 is going to require a lot more work, I'd be OK to go with 1) for the time being, but with a different name - e.g. something like `cider-jack-in-universal`.

bbatsov:
@ikappaki ping :-)

ikappaki:
Hi @bbatsov 

as discussed, I've moved the logic into `cider-jack-in-universal` and made all the no-project option configurable via the new `cider-jack-in-universal-options` variable. All the no-project configuration lines include an option to prompt for the project directory now. Could you please have another look?

Thanks

bbatsov:
@ikappaki Overall I like how things are shaping up. Now we just have to polish things (add docs, keybindings, etc) and we're good to go.

ikappaki:
> @ikappaki Overall I like how things are shaping up. Now we just have to polish things (add docs, keybindings, etc) and we're good to go.

Hi @bbatsov, I've added a documentation entry and assigned `C-c C-x j u` to the universal fn. Let me know what you think and if there is anything more you would like me to look at. Thanks

cider.el
```diff
@@ -1571,8 +1607,6 @@ canceled the action, signal quit."
 
 ;;; Aliases
 
-;;;###autoload
-(defalias 'cider-jack-in #'cider-jack-in-clj)
```
bbatsov, line 1575: Is there any reason to remove this?

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -118,6 +118,40 @@ but was subsequently switched to `clj`, Clojure's basic startup command.
...
+jack in without a project choosing from a list of pre-configured
+Clojure build tools as setup in
+`cider-jack-in-universal-options`. When this command is called from
+outside of a project, the user is given the option to select to jack
+in with one of the pre-configured tools, as well as to confirm the
+root directory to use as a base. If the command is called from within
+a project directory, it behaves exactly the same as `cider-jack-in`
+does.
+
+It utilizes Emacs'
```
bbatsov, line 133: I think it's Emacs's. See https://emacsredux.com/blog/2021/11/01/the-one-true-possessive-form-emacs-s-vs-emacs/

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -118,6 +118,40 @@ but was subsequently switched to `clj`, Clojure's basic startup command.
...
+
+The following clojure build tools are supported so far
+
+- kbd:[M-1 C-c C-x j u] jack-in using clojure-cli.
+- kbd:[M-2 C-c C-x j u] jack-in using leiningen.
+- kbd:[M-3 C-c C-x j u] jack-in using babashka.
+- kbd:[M-4 C-c C-x j u] jack-in using nbb.
+
+Here is an example of how to bind F12 for quickly bringing up a
+babashka REPL
```
bbatsov, line 146: I'd terminate this line with a `:`.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -118,6 +118,40 @@ but was subsequently switched to `clj`, Clojure's basic startup command.
 TIP: You can set `cider-allow-jack-in-without-project` to `t` if you'd like to
 disable the warning displayed when jacking-in outside a project.
 
+==== Universal jack-in
+
+`cider-jack-in-universal` kbd:[C-c C-x j u] is another way to quickly
+jack in without a project choosing from a list of pre-configured
+Clojure build tools as setup in
+`cider-jack-in-universal-options`. When this command is called from
```
bbatsov, line 126: I think near the end you should add some example how to add more items to the options.

cider.el
ikappaki: Restored.

doc/modules/ROOT/pages/basics/up_and_running.adoc
ikappaki: Updated.

doc/modules/ROOT/pages/basics/up_and_running.adoc
ikappaki: Done.

doc/modules/ROOT/pages/basics/up_and_running.adoc
ikappaki: I don't think it is easy to make up an example, I've explained the `nbb` entry instead, let me know your thoughts.

ikappaki:
Hi @bbatsov, I've addressed the comments, could you please have another look.

bbatsov:
Some tests are failing and you'll need to rebase on top of master.

bbatsov:
I saw #3320 a bit too late! 😄 Thanks for tackling the issue with the images!

ikappaki:
> Some tests are failing and you'll need to rebase on top of master.

> I saw #3320 a bit too late! 😄 Thanks for tackling the issue with the images!

re-based anyway :) (the spurious integration tests failures are under investigation elsewhere)

Issue #3262, opened 2022-10-24
Different hotkeys inside inspect buffer


lucasemmoreira:
Taking the liberty here get off the script to appreciate this awesome package! kudos to everyone involved!

**Is your feature request related to a problem? Please describe.**

Most certainly I would not put as a problem, only a suggestion. But I think it would be interesting to have the `npbf` keys to navigate the inspect buffer. Something similar to how it goes in magit.

**Describe the solution you'd like**

Like I stated before, pressing `n` would go to the next line, `p` to the previous one and so on. I also think it would be interesting to have these keys in "smart" way, meaning that if I press `n` it would go to the next value of the map and not necessarily to the next line. In other words, pressing `n` could have a behavior like `TAB`, and `p` could be like `shift TAB`.

**Describe alternatives you've considered**

If this is an interesting feature, probably a more simple approach would be to have `n` and `p` to behave like `C-n` and `C-p`.

I might be out my league here, but minor mode would be interesting? As far as I understand, it would be easier to have other different hotkeys. For example to take a variable to the repl more easily.

**Additional context**

This is a mere suggestion to have a similar workflow with other packages (just magit comes to mind now) out of the box. Not sure if it is inside the proposition of cider and it is actually valuable or if it would unnecessarily bloat the package.

;; Connected to nREPL server - nrepl://localhost:40561
;; CIDER 1.5.0 (Strasbourg), nREPL 1.0.0
;; Clojure 1.10.1, Java 11.0.15

PR #3301, opened 2022-12-26
Add extra keys for navigation of inspect buffer

gpapadok:
#3262
Added navigation functionality to `npfb` keys in inspect mode. Specifically `np` behave like `Tab` and `Shift-Tab` and `fb` like `C-f/b`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 - [#3262](https://github.com/clojure-emacs/cider/issues/3262) Add navigation functionality to `npfb` keys inside inspect buffer.
+
 ### Changes
 
 - Allow using `npx nbb` as `cider-nbb-command`.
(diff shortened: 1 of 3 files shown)


bbatsov:
Don't forget to mention the new keybindings in the docs https://docs.cider.mx/cider/debugging/inspector.html

bbatsov:
It seems you'll also have to rebase.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+- [#3262](https://github.com/clojure-emacs/cider/issues/3262) Add navigation functionality to `npfb` keys inside inspect buffer.
```
bbatsov, line 7: You forgot a `:` after the ticket link.

bbatsov:
False alarm. Thanks for tackling this!

PR #3302, opened 2022-12-26
Allow "npx nbb" as the cider-nbb-command

byronclark:
Users of nbb may prefer the version in the project versus the global version. Unlike shadow-cljs don't default to the npx command as default, but allow the command to function when set as `cider-nbb-command`.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -377,12 +377,16 @@ Throws an error if PROJECT-TYPE is unknown."
     ('shadow-cljs (let ((parts (split-string cider-shadow-cljs-command)))
                     (when-let* ((command (cider--resolve-command (car parts))))
                       (mapconcat #'identity (cons command (cdr parts)) " "))))
+    ;; here we have to account for the possibility that the command is either
+    ;; "nbb" (default) or "npx nbb".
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -377,12 +377,16 @@ Throws an error if PROJECT-TYPE is unknown."
     ('shadow-cljs (let ((parts (split-string cider-shadow-cljs-command)))
                     (when-let* ((command (cider--resolve-command (car parts))))
                       (mapconcat #'identity (cons command (cdr parts)) " "))))
+    ;; here we have to account for the possibility that the command is either
+    ;; "nbb" (default) or "npx nbb".
+    ('nbb (let ((parts (split-string cider-nbb-command)))
```
byronclark, line 382: This only change from the shadow-cljs handling is the variable name. Happy to extract this to a new defun, but I didn't do it already because it isn't useful outside these two commands.

bbatsov: I'm not surer happy with this duplication, but I can live with this for now.

Issue #3310, opened 2023-01-23
Git dependencies in cider-jack-in-dependencies


ApricotLace:
**Problem**

It's impossible to add git dependency to `'cider-jack-in-dependencies`, because the entries in that list will always be formatted to `<dep> {:mvn/version <something>}`

**Describe the solution you'd like**

Alternative list for git dependencies?

bbatsov:
How do those look? I don't use `tools.deps`, but I guess we can just tweak the command generation logic for them.

ApricotLace:
> How do those look? I don't use `tools.deps`, but I guess we can just tweak the command generation logic for them.

Similar to mvn coordinate declaration, here is a spec for git coordinate: 
https://github.com/clojure/tools.deps/blob/6ae2b6f71773de7549d7f22759e8b09fec27f0d9/src/main/clojure/clojure/tools/deps/specs.clj#L36-L41

An example of a common git dep:
```
{zen-lang/zen {:git/url "https://github.com/zen-lang/zen"
               :sha     "b3324af8de7b6856395cb5e24df61e00424d5d14"}}
```

ApricotLace:
@bbatsov, hi there!

I've implement this functionality like this:
```diff --git a/cider.el b/cider.el
index a703440..7859123 100644
--- a/cider.el
+++ b/cider.el
@@ -662,6 +662,7 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   "Removes the duplicates in DEPS."
   (cl-delete-duplicates deps :test 'equal))

+
 (defun cider-clojure-cli-jack-in-dependencies (global-options _params dependencies)
   "Create Clojure tools.deps jack-in dependencies.
 Does so by concatenating DEPENDENCIES and GLOBAL-OPTIONS into a suitable
@@ -675,7 +676,15 @@ one used."
                      ;; `java.lang.IllegalArgumentException: Duplicate key [...]`:
                      (cider--dedupe-deps)
                      (seq-map (lambda (dep)
-                                (format "%s {:mvn/version \"%s\"}" (car dep) (cadr dep))))))
+                                (if (listp (cadr dep))
+                                    (format "%s {%s}"
+                                            (car dep)
+                                            (seq-reduce
+                                             (lambda (acc v)
+                                               (concat acc (format " :%s \"%s\" " (car v) (cdr v))))
+                                             (cadr dep)
+                                             ""))
[...]

PR #3313, opened 2023-01-27
Custom coordinates in jack-in-dependecies

ApricotLace:
Add ability to use custom coordinates in jack-in-dependecies.
Custom coordinates will be described as an alist, which will later be expanded into a clojure map.

Example:
```elisp
                    "org.clojure/tools.deps"
                    '(("git/sha" . "6ae2b6f71773de7549d7f22759e8b09fec27f0d9")
                      ("git/url" . "https://github.com/clojure/tools.deps/")))
```

Related issue: https://github.com/clojure-emacs/cider/issues/3310

-----------------


bbatsov:
Thanks!

PR #3321, opened 2023-02-08
Fix jack in universal project-type to match project tool

ikappaki:
Hi,

can you please consider follow up fix to match the project-type to the project tool type for `cider-jack-in-universal`. It fixes an options misalignment issue introduced with #3300, where an invocation with prefix argument 1, which is assigned to `clojure-cli` tool, would have started a `babashka` nREPL instead.

Thanks,


bbatsov:
Thanks!

Issue #3298, opened 2022-12-22
some integration tests might intermittently timeout while enquiring server's capabilities


ikappaki:
## Expected behavior

all tests should pass when running on GH

## Actual behavior

From recent (windows-emacs-27.2 GH action run logo)[https://github.com/clojure-emacs/cider/actions/runs/3752449361/jobs/6374620615]:
```
jack in to clojure tools cli

Traceback (most recent call last):
  signal(error ("Sync nREPL request timed out (op eval code cljs.core/demunge)"))
error: (error "Sync nREPL request timed out (op eval code cljs.core/demunge)")
```

From the diagnostics produced by the failure, it appears the nREPL client has timed out waiting for a response to an the `cljs.core/demunge` eval request (id:7):
```
:BUFFER "*nrepl-messages Temp/FcZn6A:localhost:51084(clj)*<2>"
343
[01:28.901]  (-->
344
               id         "1"
345
               op         "clone"
346
               time-stamp "2022-12-21 20:09:12.312433000"
347
             )
348
             (<--
349
               id          "1"
350
               session     "6653bdcd-09a2-4db9-a33e-464162ab64fb"
351
               time-stamp  "2022-12-21 20:09:12.391205000"
352
               new-session "6250556f-33de-4718-aa4f-79f2da42e2d9"
353
               status      ("done")
354
             )
355
             (-->
356
               id         "2"
357
               op         "clone"
358
               time-stamp "2022-12-21 20:09:12.405385000"
359
             )
360
             (<--
361
               id          "2"
362
               session     "4fc4d5f0-7a78-4251-8700-a9b04c7074f6"
363
               time-stamp  "2022-12-21 20:09:12.407864000"
364
               new-session "81991b04-cbe2-4a6c-9fe0-7cb0ba58190a"
365
               status      ("done")
366
             )
367
             (-->
368
               id         "3"
369
               op         "describe"
370
               session    "6250556f-33de-4718-aa4f-79f2da42e2d9"
371
               time-stamp "2022-12-21 20:09:12.426393000"
372
             )
373
             (<--
374
               id         "3"
375
               session    "6250556f-33de-4718-aa4f-79f2da42e2d9"
376
               time-stamp "2022-12-21 20:09:12.443685000"
377
               aux        (dict ...)
378
               ops        (dict ...)
379
               status     ("done")
380
               versions   (dict ...)
381
             )
382
             (-->
383
               id                                 "4"
384
               op                                 "eval"
385
[...]

PR #3322, opened 2023-02-08
spy-on `read-file-name' directly to fake user input

ikappaki:
It appears that the new integration tests introduced with #3300 failing on GH MS-Windows run were legitimate failures.

`spy-on` `completing-read` does not work on GH actions for MS-Windows when prompting user to read a filename, thus we  `spy-on` the `read-file-name` instead to fake user input.

error was
```
  signal(end-of-file ("Error reading from stdin"))
error: (end-of-file "Error reading from stdin")
```

Thanks

### DIFF
--- a/test/integration/integration-tests.el
+++ b/test/integration/integration-tests.el
@@ -410,8 +410,13 @@
                                      ;; select nbb
                                      ("No project found in current dir, select project type to jack in: "
                                       "nbb")
-                                     ;; project src directory, use suggested
-                                     ("Project: " initial-input)
+                                     (_ (error ":integration-test-unsupported-prompt-error %S" prompt)))))
+          (spy-on 'read-file-name
+                  :and-call-fake (lambda (prompt &optional dir _default-filename _mustmatch
+                                                 _initial _predicate)
+                                   (pcase prompt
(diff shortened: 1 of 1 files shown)


bbatsov:
> spy-on completing-read does not work on GH actions for MS-Windows when prompting user to read a filename, thus we spy-on the read-file-name instead to fake user input.

Pretty weird. I wonder why this is behaving differently on Windows.

ikappaki:
> > spy-on completing-read does not work on GH actions for MS-Windows when prompting user to read a filename, thus we spy-on the read-file-name instead to fake user input.
> 
> Pretty weird. I wonder why this is behaving differently on Windows.

and it was only failing on the github windows runner, it worked fine on the desktop. It would have taken some time to figure this out so I thought to go for the obvious solution for the interest of time :)

bbatsov:
Fair enough.

bbatsov:
Seems some integration tests are still failing on Windows.

ikappaki:
> Seems some integration tests are still failing on Windows.

These are the intermittent MS-Windows/macos failures that are under investigation with #3298

PR #3323, opened 2023-02-10
Fix #3307: Incorrect special-form eldocs

### LINKED ISSUE #3307: Documentation highlighting appears to be incorrect for special forms
I've noticed that, when I'm writing special forms (such as `if` or `let`), CIDER seems to be highlighting the element prior to the element that I'm currently on. 

For example, if I write:

```clojure
(if (> 3 5)
           ↑
         point
```

I get 

screenshot of if highlighted 

Notice how `if` is highlighted, even though the point is on the test.

Then, when I write the `then` part of the special form, as follows:

```clojure
(if (> 3 5)
  (println "Three is greater than five?")
                                         ↑
                                       point
```

I see

screenshot of test highlighted

Notice how `test` is highlighted, even though the point is on the `then` part of the special form.

Finally, after I finish the `if` with:

```clojure
(if (> 3 5)
  (println "Three is greater than five?")
  (println "Math works as I'd expect")
                                      ↑
                                    point
```

I see

screenshot of then highlighted

Notice how `then` is highlighted, even though I just finished writing the `else` clause.

I've noticed this "one-behind" behavior on other special forms, such as `let`, as well. However, highlighting appears to work correctly for macros and functions.

I'm running `CIDER 1.6.0 (Buenos Aires)` with `nREPL 1.0.0`, on `Clojure 1.11.1, Java 17.0.5`. My Java environment is:
```
OpenJDK Runtime Environment (Red_Hat-17.0.5.0.8-1.fc37) (build 17.0.5+8)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.5.0.8-1.fc37) (build 17.0.5+8, mixed mode, sharing)
```

quanticle:
Fix #3307: Incorrect special-form eldocs

This fix is the CIDER companion to [this PR](https://github.com/clojure-emacs/orchard/pull/166) in orchard, which lets us know if the symbol we're querying eldocs for is a special form or an ordinary function.

This change uses that information to deduplicate the arglists for the special form. In the process, the change adds a
`cider-eldoc-format-special-form` function, which serves as a home for special form formatting.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Allow using `npx nbb` as `cider-nbb-command`.
 - [#3281](https://github.com/clojure-emacs/cider/pull/3281): Replace newline chars with actual newlines in `*cider-test-report*` buffer, for prettier error messages.
 - Bump the injected `cider-nrepl` to 0.30.
+- [#3307](https://github.com/clojure-emacs/cider/issues/3307): Make eldoc highlighting on emacs special forms better match the location of the point when latest `cider-nrepl` is used.
 
 ## 1.6.0 (2022-12-21)
 
--- a/cider-eldoc.el
+++ b/cider-eldoc.el
@@ -230,6 +230,22 @@ arglists.  ELDOC-INFO is a p-list containing the eldoc information."
             (cider-eldoc-format-thing ns symbol thing 'fn)
             (cider-eldoc-format-arglist arglists pos))))
 
+(defun cider-eldoc-format-special-form (thing pos eldoc-info)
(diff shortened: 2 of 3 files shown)


bbatsov:
You'll also have to rebase on top of `master` to resolve the changelog conflict.

bbatsov:
@quanticle ping :-)

bbatsov:
@quanticle Ping 2 :-)

bbatsov:
@vemv As you had some small suggestion do you want to drive this PR to the finish line? Alternatively I can just merge it as is.

Issue #3295, opened 2022-12-16
Remove cider-*-global-options (part of the jack-in command)


bbatsov:
We've got a ton of those configuration options (`e.g. `cider-lein-, but I think we don't really need them as there's nothing inherently different between `cider-*-parameters` and `cider-*-global-options`. I think that's just legacy from the earliest days of `cider-jack-in` which was modeled for Lein initially. E.g.:

```
(defcustom cider-lein-global-options
  nil
  "Command global options used to execute Leiningen (e.g.: -o for offline)."
  :type 'string
  :safe #'stringp)
```

It seems initially we wanted to separate the REPL task from the rest of the command invocation, but in hindsight this seems pointless and it just adds a point of confusion for people who now have to wonder which config to modify exactly. I think we'll be better off without this extra complexity. 

I'm opening this issue now because #3275 reminded me about this problem.

bbatsov:
To be clear - my plan is to deprecate those right away and remove them in CIDER 2.0.

callmetushar123:
@bbatsov I new to open source can you guide me on what to do here?

bbatsov:
Sure. It's just a matter of marking those as obsolete and folding them with `cider-*-parameters`. I think they are essentially unused anyways (and have like one usage in codebase each), so at the end that'd be mostly a documentation change.  

Now we'll just deprecate them and we'll delete them in CIDER 2.0.

PR #3330, opened 2023-03-23
#3295 Deprecate cider-*-global-options

gpapadok:
Addresses #3295.

Marked all cider-*-global-options as obsolete and replaced with cider-*-parameters.
Also updated the documentation to mark all global-options references as deprecated.

Maybe I am wrong, but the only caveat is for clojure-cli the only configuration variable is `cider-clojure-cli-aliases` which is not intuitive for other types of parameters other than aliases. For example, in the documentation there is a reference to a jvm-option you can pass by setting `(setq cider-clojure-cli-global-options "-J-XX:-OmitStackTraceInFastThrow")`. I tried the equivalent with `cider-clojure-cli-aliases` and the command used to start the repl was this:

```
clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "1.0.0"} cider/cider-nrepl {:mvn/version "0.29.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[cider.nrepl/cider-middleware]"]}}}' -M-J-XX:-OmitStackTraceInFastThrow:cider/nrepl
```
As oposed to this when setting global-options:
```
clojure -J-XX:-OmitStackTraceInFastThrow -Sdeps '{:deps {nrepl/nrepl {:mvn/version "1.0.0"} cider/cider-nrepl {:mvn/version "0.29.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
```

Does this work as expected?

-----------------

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -156,6 +156,13 @@ default to \"powershell\"."
   :safe #'stringp
   :package-version '(cider . "0.17.0"))
 
+(defcustom cider-clojure-cli-parameters
+  nil
+  "Params passed to clojure cli to start an nREPL server via `cider-jack-in'."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.8.0"))
+
 (defcustom cider-clojure-cli-aliases
   nil
   "A list of aliases to include when using the clojure cli.
@@ -254,7 +261,8 @@ By default we favor the project-specific shadow-cljs over the system-wide."
 
 (make-obsolete-variable 'cider-lein-global-options 'cider-lein-parameters "1.8.0")
 (make-obsolete-variable 'cider-boot-global-options 'cider-boot-parameters "1.8.0")
-(make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-aliases "1.8.0")
+(make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-parameters "1.8.0")
+(make-obsolete-variable 'cider-clojure-cli-aliases 'cider-clojure-cli-parameters "1.8.0")
 (make-obsolete-variable 'cider-shadow-cljs-global-options 'cider-shadow-cljs-parameters "1.8.0")
 (make-obsolete-variable 'cider-gradle-global-options 'cider-gradle-parameters "1.8.0")
 (make-obsolete-variable 'cider-babashka-global-options 'cider-babashka-parameters "1.8.0")
@@ -440,7 +448,7 @@ Throws an error if PROJECT-TYPE is unknown."
   (pcase project-type
     ('lein        cider-lein-parameters)
     ('boot        cider-boot-parameters)
-    ('clojure-cli nil)
+    ('clojure-cli cider-clojure-cli-parameters)
     ('babashka    cider-babashka-parameters)
     ('shadow-cljs cider-shadow-cljs-parameters)
     ('gradle      cider-gradle-parameters)
@@ -737,7 +745,7 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   "Removes the duplicates in DEPS."
   (cl-delete-duplicates deps :test 'equal))
 
-(defun cider-clojure-cli-jack-in-dependencies (global-options _params dependencies)
+(defun cider-clojure-cli-jack-in-dependencies (global-options params dependencies)
   "Create Clojure tools.deps jack-in dependencies.
 Does so by concatenating DEPENDENCIES and GLOBAL-OPTIONS into a suitable
 `clojure` invocation.  The main is placed in an inline alias :cider/nrepl
@@ -764,7 +772,7 @@ one used."
                       (cider-jack-in-normalized-nrepl-middlewares)
                       ","))
          (main-opts (format "\"-m\" \"nrepl.cmdline\" \"--middleware\" \"[%s]\"" middleware)))
-    (format "%s-Sdeps '{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s]}}}' -M%s:cider/nrepl"
+    (format "%s-Sdeps '{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s]}}}' -M%s:cider/nrepl%s"
             (if global-options (format "%s " global-options) "")
             (string-join all-deps " ")
             main-opts
@@ -772,7 +780,8 @@ one used."
                 ;; remove exec-opts flags -A -M -T or -X from cider-clojure-cli-aliases
                 ;; concatenated with :cider/nrepl to ensure :cider/nrepl comes last
                 (format "%s" (replace-regexp-in-string "^-\\(A\\|M\\|T\\|X\\)" "" cider-clojure-cli-aliases))
-              ""))))
+              "")
+            (if params (format " %s" params) ""))))
 
 (defun cider-shadow-cljs-jack-in-dependencies (global-opts params dependencies)
   "Create shadow-cljs jack-in deps.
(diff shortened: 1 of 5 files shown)


bbatsov:
> For example, in the documentation there is a reference to a jvm-option you can pass by setting (setq cider-clojure-cli-global-options "-J-XX:-OmitStackTraceInFastThrow").

I was thinking that `cider-*-parameters` should replace the global options in essentially all cases. Not sure if the introduction of `cider-clojure-cli-aliases` made this different for the Clojure CLI.

cider.el
```diff
@@ -252,6 +252,14 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.6.0"))
 
+(make-obsolete-variable 'cider-lein-global-options 'cider-lein-parameters "1.6.0")
```
bbatsov, line 255: 1.8 actually, as 1.7 was just released

doc/modules/ROOT/pages/troubleshooting.adoc
```diff
@@ -293,7 +293,7 @@ In particular, by adding it to a `deps.edn` file under an alias (eg. `:dev`)
 Or by customising the jack-in options.
 [source,lisp]
 ---
-(setq cider-clojure-cli-global-options "-J-XX:-OmitStackTraceInFastThrow")
+(setq cider-clojure-cli-aliases "-J-XX:-OmitStackTraceInFastThrow")
```
bbatsov, line 296: We should aim for some example using `cider-clojure-cli-parameters`.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -201,8 +201,8 @@ so we'll examine them tool by tool.
 
 * `cider-lein-command` - the name of the Leiningen executable (`lein` by default)
 * `cider-lein-parameters` - the command-line params to start a REPL (e.g. `repl :headless`)
-* `cider-lein-global-options` - these are passed to the command directly, in
-first position (e.g., `-o` to `lein` enables offline mode).
+* `cider-lein-global-options ` - these are passed to the command directly, in
```
bbatsov, line 204: I'd add some example showing how to pass additional params here `e.g. -o` as well via `lein-parameters`. You can also remove the note beneath as it's basically an explanation why everything has the `global-options` which in essence came from Lein.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -227,19 +227,19 @@ will likely result in a better overall development experience.
 * `cider-boot-parameters` - these are usually task names and their parameters
 (e.g., `dev` for launching boot's dev task instead of the standard `repl -s
 wait`)
-* `cider-boot-global-options`
+* `cider-boot-global-options` (DEPRECATED)
```
bbatsov, line 230: Those can just be removed everywhere, but it'd be nice if we have more examples for every tool setting more `cider-parameters`, so it's clear for people they can pass whatever they want there.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -227,19 +227,19 @@ will likely result in a better overall development experience.
...
 
 ==== Gradle Options
 
 * `cider-gradle-command` - the name of the Gradle executable (`./gradlew` by default)
 * `cider-gradle-parameters` - the Gradle arguments to invoke the repl task (`clojureRepl` by default)
-* `cider-gradle-global-options` - these are usually global options to gradle, such as `--no-daemon` or `--configuration-cache` (empty by default)
+* `cider-gradle-global-options` - these are usually global options to gradle, such as `--no-daemon` or `--configuration-cache` (empty by default) (DEPRECATED)
```
bbatsov, line 236: Same - this needs to become some example for `cider-gradle-parameters`.

doc/modules/ROOT/pages/troubleshooting.adoc
gpapadok: There is no `cider-clojure-cli-parameters`. It was removed in 1.1 according to the changelog. So if we remove `cider-clojure-cli-global-options` the only configuration option would be `cider-clojure-cli-aliases`.
Should we maybe rename aliases to parameters so that we have uniform configuration for all jack-in types?

bbatsov: Yeah, we should rename it.

gpapadok: Do you want it done on this pr? It is a breaking change. Should we leave it for the 2.0 release?

bbatsov:
Thanks!

gpapadok:
Thank you too!
I am not very savvy with the different build options so I didn't add any more parameter examplea in the docs. Maybe I'll do another PR in the future.

cjohansen:
What was the point of this pull request? Seems like all it does is to rename things, causing everyone's setup to break. Changes like this come with a big cost to users. Clojure has a strong focus on backwards compatibility and stability - it would be cool if the tooling adopted the same mindset.

bbatsov:
@cjohansen See https://github.com/clojure-emacs/cider/issues/3295

Basically the purpose of this change is to simplify the jack-in configuration which was poorly designed initially to account for one peculiarity of Lein (in the beginning it was the only tool that `cider-jack-in` supported). I'm curious what exactly broke for you, as this PR just deprecates stuff and they will be removed only in CIDER 2.0. 

> Clojure has a strong focus on backwards compatibility and stability - it would be cool if the tooling adopted the same mindset.

I'm all for stability myself, but if some bad design decision is causing a lot of pain I'd rather fix it than live with it for all eternity. Those global options were meant to be used only with Leiningen and were copy/pasted accidentally to other tools. Now this is being fixed.

bbatsov:
For extra context - we're not "renaming things", but rather deprecating the redundant global options (they still work, but people get a warning to use the matching parameters variables instead):

```
(make-obsolete-variable 'cider-lein-global-options 'cider-lein-parameters "1.8.0")
(make-obsolete-variable 'cider-boot-global-options 'cider-boot-parameters "1.8.0")
(make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-parameters "1.8.0")
(make-obsolete-variable 'cider-shadow-cljs-global-options 'cider-shadow-cljs-parameters "1.8.0")
(make-obsolete-variable 'cider-gradle-global-options 'cider-gradle-parameters "1.8.0")
(make-obsolete-variable 'cider-babashka-global-options 'cider-babashka-parameters "1.8.0")
(make-obsolete-variable 'cider-nbb-global-options 'cider-nbb-parameters "1.8.0")
```

cjohansen:
I guess I was a little quick - this wasn't the root cause, sorry for making a wrong claim.

I still think that removing these are a grave mistake. It causes nothing but problems. I understand the confusion that they can cause, but I think a better solution is to remove all concerning documentation of the old ones, mark the old ones as legacy in the code (so cider developers don't trip on them), but leave them be, and drop the deprecation.

bbatsov:
> I still think that removing these are a grave mistake. It causes nothing but problems. I understand the confusion that they can cause, but I think a better solution is to remove all concerning documentation of the old ones, mark the old ones as legacy in the code (so cider developers don't trip on them), but leave them be, and drop the deprecation.

I appreciate the input, but I've made up my mind about this. I'm not fond of carting dead code forever, as it complicates my work on CIDER (and that of all other contributors) and impedes potential future improvements. It's a relatively small change and CIDER 2.0 is not coming any time soon. The impact on the end users will be minimal, if any, when we get to the 2.0 release.

PR #3333, opened 2023-04-17
Keep repl buffers alive for reuse on sesman-restart

chopptimus:
Closing connections in the server sentinel breaks sesman-restart because Cider tries to reuse the repl buffer on restart.

I do not know why you would only close connections on hangup or at all - I just restored the behaviour we had before the change to always close connections.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+- Keep repl buffers alive when restarting with sesman-restart
+
 ## 1.7.0 (2023-03-23)
 
 ### New features
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -1212,8 +1212,9 @@ an `error' if the nREPL PROCESS exited because it couldn't start up."
                                   (eq (buffer-local-value 'nrepl-server-buffer b)
                                       server-buffer))
                                 (buffer-list))))
-      ;; close any known open client connections
-      (mapc #'cider--close-connection clients)
+
+      (when (string-match-p "^hangup" event)
+        (mapc #'cider--close-connection clients))
 
       (if (process-get process :cider--nrepl-server-ready)
           (progn


chopptimus:
a4b8ab1e0d98c09b56bffd78a89128b6142424a4 is the culprit as far as I can tell.

iarenaza:
> Hi!
> 
> Thanks for your contribution.
> 
> > just restored the behaviour we had before the change to always close connections.
> 
> For completeness, what was this change? e.g. Please point to the culprit commit or the most recent known-good version.

I think this is the commit that changed the behaviour: https://github.com/clojure-emacs/cider/commit/a4b8ab1e0d98c09b56bffd78a89128b6142424a4#diff-38905f4224283c5805229f55b0cdd0701a8bf803819cc771f39730d9fe9a8cb6L1131-R1200

chopptimus:
> I'm not sure why do we need to treat "HUP" in some special way?

Neither do I - it's my attempt to revert the problematic part of the commit that causes this issue. Is there a way to check for a sesman-restart before closing connections instead?

bbatsov:
I'm guessing you can check the value of `this-command`, but that would be brittle. I'm guessing it'd be better to have some explicit param passed that indicates that connections should be kept or something along those lines.

ikappaki:
> From my side I don't have a particular opinion for/against the revert. However given that we identified the commit, we can ask @ikappaki directly!
> 
> WDYT?

Hi *

I think I removed it during the refactoring because I couldn't see its purpose and the newly introduced integration tests which exercised that path of code were passing fine.

So, it turns out sesman requires to let the server know that it manages the client connections and should not close them on shutdown.  This was done before by having sesman kill the server and have the sentinel check the exit event message, and only if it was the `hangup` it was allowed to close the connection (the logic I removed). i.e. sesman would kill the server in such a manner that the server's termination signal would not be `hangup`.

We can restore than mechanism, but I don't think this is either a portable or reliable method, and we could potentially do better.

I had a quick look at the code, and there is `no-kill` argument that is passed to `cider--close-connection` by the cider sesman methods to indicate the intention the server should not touch the client connection. I think I can use this to communicate to the server via its process plist not to touch the client connections on exit. I think this is a more precise way to communicate the intent from sesman to the server than checking the termination signal type.

If you are not in a hurry, I can have a look implementing this in the week and I also add a test for the same?

Thanks

ikappaki:
Hi,

Upon further examination of the code, it seems that the previous behavior, where the server sentinel closed client connections only upon receiving a SIGHUP signal, was actually an exception rather than the norm.

It is not obvious by looking at the git history why SIGHUP should trigger closing of the connections in the sentinel.

The condition was introduce with the following commit and would call eventually call `cider-quit`.

https://github.com/clojure-emacs/cider/commit/73525f5babfac654b65fb4b3bca540d1d03ce5c9

It has then been upgraded to close the repl buffer with

https://github.com/clojure-emacs/cider/commit/ff9ac74181fca4b368f07d8422fb05de43f0f957

--
In normal operation, when the server process terminates, It triggers a signal to all the active client sentinels indicating its demise. This allows the client connections to react and perform necessary cleanup tasks.

There appears to be no valid reason for the server sentinel to abruptly close the client connections and their associated buffers. Emacs would typically send a SIGKILL or SIGINT signal to gracefully shut down the server.

The [HUP signal](https://en.wikipedia.org/wiki/SIGHUP) handler, which was present prior to my commit and responsible for closing the client connections, is typically sent by the operating system to a process when the process is running in a terminal and the terminal is closed. However, I fail to see how this concept of "the terminal" can be relevant to the nrepl server process spawned by Emacs, as there is no involvement of a terminal in this scenario.

Therefore, instead of our previous behavior of closing client connections only when the server was terminated with SIGHUP, and the current incorrect approach of always forcefully closing all client connections, I suggest in the server sentinel, that we never enforce the closure of any connections.

What if it turns otu there's an important scenario in which using SIGHUP to forcibly close the connections in the server sentinel becomes necessary? If such a situation arises, it will likely be identified as a regression issue, shedding light on the specific reason why it is needed.

To summarise the options
1. Restore the behavior of closing client connections upon receiving a SIGHUP signal (current patch). This option is the safest choice,  even though the underlying reason for this behavior is still  unclear.
2. Do not close client connections within the server sentinel. There seems to be no apparent reason for the server sentinel to close    these connections, although it deviates from the original behavior  associated with SIGHUP.
3. Have the server sentinel forcibly close all client connections, unless explicitly instructed otherwise through the process properties. However, considering the previous discussions, this idea seems to be adding unecessary logic and should be disregarded.

What do you think?

Thanks

ikappaki:
> > Restore the behavior of closing client connections upon receiving a SIGHUP signal (this PR). This option is the safest choice, even though the underlying reason for this behavior is still unclear.
> 
> I would guess that the intended behavior, as OP hints, is supporting sesman-restart?
> 
> From reading the code, the restored code (e.g. what this PR accomplishes) makes sense: only close connections only when the closing appears to be deliberate or unrecoverable, else keep them for a possible restart.
> 
> Still, we would benefit from better docstrings/comments/test around this section.

Yes, the docstring can be amended from

```emacs-lisp
"Handle nREPL server PROCESS EVENT.
On a fatal EVENT, attempt to close any open client connections, and signal
an `error' if the nREPL PROCESS exited because it couldn't start up."
```
to something like

```emacs-lisp
"Handle nREPL server PROCESS EVENT.
On a fatal EVENT, signal an `error' if the nREPL PROCESS exited because it couldn't 
start up. Furthermore, if the EVENT is SIGHUP, then forcibly close any open client connections."
```

I can't think of anything to say as an inline comment as to why we are closing on SIGHUP though :)

I also plan to follow up on `sesman-restart` test once this is in  (not sure I can amend this PR as a simple user)

Thanks

PR #3292, opened 2022-12-14
Document updated paredit ret behaviour

mindbat:
Motivated by: https://github.com/clojure-emacs/cider/issues/3288

Adds documentation for the odd repl behaviour some users (myself included) have seen with newer versions of paredit, along with a fix (hat-tip to @rrudakov).

Issue #3288, opened 2022-12-10
Evaluation hangs in cider-jack-in if paredit-mode enabled in repl buffer


mindbat:
## Expected behavior

After running `cider-jack-in`, and typing `(+ 2 2)` at the prompt, then hitting enter, the expression is evaluated and the buffer displays the return value of `4` before giving a new prompt.

## Actual behavior

Repl buffer hangs indefinitely after hitting enter to evaluate any expression.

## Steps to reproduce the problem

1. Install Emacs 28.2 via homebrew.
2. Install `cider` via MELPA.
3. Install `paredit` via MELPA.
4. Install `lein` and use `lein new` to create a new project.
5. Run `emacs <name of core.clj file in new project>` to open the clj file
6. Run `cider-jack-in`
7. In the repl buffer, run `paredit-mode` to turn on paredit mode.
8. At the repl prompt, enter `(+ 2 2)` and hit enter.

_Note: On my system, if I do not turn on paredit-mode in the repl buffer, everything works normally._

### CIDER version information

```
 ;; Connected to nREPL server - nrepl://localhost:53570
 ;; CIDER 1.5.0 (Strasbourg), nREPL 0.9.0
 ;; Clojure 1.11.1, Java 19.0.1
 ;;     Docs: (doc function-name)
 ;;           (find-doc part-of-name)
 ;;   Source: (source function-name)
 ;;  Javadoc: (javadoc java-object-or-class)
```

### Lein / Clojure CLI version

`Leiningen 2.9.10 on Java 19.0.1 OpenJDK 64-Bit Server VM`

### Emacs version

28.2

### Operating system

MacOS 12.6.1

### JDK distribution

```
java version "19.0.1" 2022-10-18
Java(TM) SE Runtime Environment (build 19.0.1+10-21)
Java HotSpot(TM) 64-Bit Server VM (build 19.0.1+10-21, mixed mode, sharing)
```

rrudakov:
The new version of paredit map RET key to `paredit-RET` command. You can either unset this key in `paredit-mode-map` or use `C-RET` in REPL buffer.

mindbat:
> The new version of paredit map RET key to paredit-RET command

@rrudakov Ah, that explains why I've not had this issue before 😊 

You're right, rebinding the `RET` key via `(define-key paredit-mode-map (kbd "RET") nil)` in my `init.el` file fixed it 👍

bbatsov:
[...]

PR #3335, opened 2023-04-24
Disable paredit-RET binding in cider-repl-mode

yuhan0:
-------

See #3288, #3334
This applies the suggested fix only to the buffer-local map in the repl, such that "RET" continues to be bound to paredit-RET in regular code buffers.

Might negate the need for the doc changes in #3292

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1866,7 +1866,12 @@ constructs."
     (add-hook 'kill-buffer-hook #'cider-repl-history-just-save t t)
     (add-hook 'kill-emacs-hook #'cider-repl-history-just-save))
   (add-hook 'completion-at-point-functions #'cider-complete-at-point nil t)
-  (add-hook 'paredit-mode-hook (lambda () (clojure-paredit-setup cider-repl-mode-map))))
+  (add-hook 'paredit-mode-hook
+            (lambda ()
+              (clojure-paredit-setup cider-repl-mode-map)
+              ;; Disable paredit-RET, see https://github.com/clojure-emacs/cider/issues/3288
+              (make-local-variable 'paredit-mode-map)
+              (define-key paredit-mode-map "RET" nil))))
 
 (provide 'cider-repl)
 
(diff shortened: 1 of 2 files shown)


yuhan0:
It also appears that the function `clojure-paredit-setup` has a minor bug - by defining the `{` and `}` keys in the major mode map it causes those bindings to stick around if Paredit minor mode is disabled at a later point.

cider-repl.el
```diff
@@ -1866,7 +1866,12 @@ constructs."
     (add-hook 'kill-buffer-hook #'cider-repl-history-just-save t t)
     (add-hook 'kill-emacs-hook #'cider-repl-history-just-save))
   (add-hook 'completion-at-point-functions #'cider-complete-at-point nil t)
-  (add-hook 'paredit-mode-hook (lambda () (clojure-paredit-setup cider-repl-mode-map))))
+  (add-hook 'paredit-mode-hook
+            (lambda ()
+              (clojure-paredit-setup cider-repl-mode-map)
+              ;; Disable paredit-RET, see https://github.com/clojure-emacs/cider/issues/3288
```
bbatsov, line 1872: Isn't it better to move this to the `clojure-paredit-setup` function?

yuhan0: The idea is that paredit users might want to have the binding present in clojure-mode source files, where paredit-RET does some sort of electric indentation on the following forms, but not in the cider-repl buffers where RET should evaluate the prompt.

yuhan0: We could also change the meaning of the optional argument over at clojure-mode, there are no other references to the function as far as I can tell.

```diff
-(defun clojure-paredit-setup (&optional keymap)
+(defun clojure-paredit-setup (&optional disable-RET)
   "Make \"paredit-mode\" play nice with `clojure-mode'.
 
 If an optional KEYMAP is passed the changes are applied to it,
 instead of to `clojure-mode-map'.
 Also advice `paredit-convolute-sexp' when used on a let form as drop in
 replacement for `cljr-expand-let`."
   (when (>= paredit-version 21)
-    (let ((keymap (or keymap clojure-mode-map)))
-      (define-key keymap "{" #'paredit-open-curly)
-      (define-key keymap "}" #'paredit-close-curly))
+    (make-local-variable 'paredit-mode-map)
+    (define-key paredit-mode-map "{" #'paredit-open-curly)
+    (define-key paredit-mode-map "}" #'paredit-close-curly)
+    (when disable-RET
+      (define-key paredit-mode-map "RET" nil))
     (make-local-variable 'paredit-space-for-delimiter-predicates)
     (add-to-list 'paredit-space-for-delimiter-predicates
                  #'clojure-space-for-delimiter-p)
     (advice-add 'paredit-convolute-sexp :after #'clojure--replace-let-bindings-and-indent)))
  ```

bbatsov: I guess I'll merge your PR in its current form and you can follow up with something that simply uses a buffer-local keymap. We can remove the optional argument completely given that it won't make sense anymore.

PR #3336, opened 2023-04-26
Fix incorrect paredit binding override in #3335

yuhan0:
I reverted the previous commit and used the proper fix suggested in https://github.com/clojure-emacs/cider/commit/0fa58acd93d7305ddc823d5bd670a2e3362bacd2#commitcomment-110663006
 

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Yeah, I agree it's better to handle this out-of-the-box.

PR #3337, opened 2023-04-26
Debug on exception

bo-tato:
Add support on the cider side to go with https://github.com/clojure-emacs/cider-nrepl/pull/769
passing a negative argument to `cider-eval-defun-at-point` will instrument it with #dbgexn
also added some documentation to the debugger docs describing the new feature

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1287,7 +1287,9 @@ command `cider-debug-defun-at-point'."
       (when inline-debug
         (cider--prompt-and-insert-inline-dbg)))
     (cider-interactive-eval (when (and debug-it (not inline-debug))
-                              (concat "#dbg\n" (cider-defun-at-point)))
+                              (if (eq '- debug-it)
+                                  (concat "#dbgexn\n" (cider-defun-at-point))
+                                (concat "#dbg\n" (cider-defun-at-point))))
                             nil
                             (cider-defun-at-point 'bounds)
                             (cider--nrepl-pr-request-map))))
--- a/doc/modules/ROOT/pages/debugging/debugger.adoc
+++ b/doc/modules/ROOT/pages/debugging/debugger.adoc
@@ -200,6 +200,19 @@ plus metadata annotation in your code. Note that you'll have to delete
 this annotation by hand; you cannot simply use kbd:[C-M-x] as you
 can to un-instrument kbd:[C-u C-M-x].
 
+== Debug on Exception
+
+There is basic experimental support for entering the debugger when an exception
+is thrown. When you instrument a function by passing a negative argument to
+`cider-eval-defun-at-point` kbd:[C-u - C-M-x] then you will enter the debugger
+whenever any form in that function throws an exception. You can also instrument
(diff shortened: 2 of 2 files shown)


bbatsov:
@yuhan0 I guess you'll take this from here, right?

yuhan0:
How do you feel about this proposal of overloading cider-eval-defun with a negative argument? 
I find that sort of API quite bad for discoverability - in my personal config I have these custom commands instead, along with similar ones for various other temporary debugging forms - basically generalizing the existing `cider--prompt-and-insert-inline-dbg` interface.

```clj
;; * Insert #break / #dbg reader forms
(defun +cider-insert-dbg (ex)
  "Insert #dbg reader.
If EX is non nil or called with prefix arg, insert #dbg! to break on exception.
Automatically reevaluates form if inside a defn."
  (interactive "P")
  (unless (looking-at-p "(")
    (backward-up-list))
  (cider-insert-fragile-reader-macro
   (if ex "#dbg!"
     (concat "#dbg"
             (let ((cond (cider-read-from-minibuffer "Condition: ")))
               (unless (equal "" cond)
                 (format " ^{:break/when %s}" cond))))))
  (when (clojure-top-level-form-p "defn")
    (cider-eval-defun-at-point)))

(defun +cider-insert-breakpoint (ex)
  "Insert #break reader.
If EX is non nil or called with prefix arg, insert #break! to break on exception."
  (interactive "P")
  (cider-insert-fragile-reader-macro
   (if ex "#break!"
     (concat "#break"
             (let ((cond (cider-read-from-minibuffer "Condition: ")))
               (unless (equal "" cond)
                 (format " ^{:break/when %s}" cond)))))))

```

yuhan0:
A quick demo of what that looks like in practice
https://github.com/clojure-emacs/cider/assets/22216124/8e0440e4-da2c-4de0-890d-8d882185719a

bbatsov:
> How do you feel about this proposal of overloading cider-eval-defun with a negative argument?

I'm not a big fan of prefix arguments in general, but from time to time they fit in nicely. Definitely not a big fan of the negative argument, as it's probably the hardest to discover organically by users. I'm ok with adding some special commands and I guess we can have something with prefix arguments as well, for whoever prefers them, given that it adds little complexity.

The demo looks good to me.

bbatsov:
@yuhan0 Any chance you'll have time for this soon? I think it's the only feature remaining for the next CIDER release.

yuhan0:
Yes, I have the working code for this feature but unfortunately don't have much time over the next couple of days to work on testing / refining it into a PR. If it's a blocker for a release maybe I'll push those commits now and see if anyone can take it from there? Might be able to do something over the weekend too.

bbatsov:
@yuhan0 No rush, there's one more blocker anyways (https://github.com/clojure-emacs/cider/issues/3365). If you can spend some time on this over the weekend that'd be great. 

I'll close this PR in favor of #3366 that you've opened just now.

Issue #3341, opened 2023-05-10
Cider will not jack-in but can cider-connect


paulrd:
## Expected behavior
Should be able to run repl from cider-jack-in without error.
## Actual behavior
The command cider-jack-in-clj does not create a repl. Here is the trace: 
```shell
Starting new CIDER session ...
[nREPL] Starting server via "c:/Users/Paul Dumais/scoop/shims/clojure.exe" -Sdeps '{:deps {nrepl/nrepl {:mvn/version "1.0.0"} cider/cider-nrepl {:mvn/version "0.30.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
error in process sentinel: nrepl-server-sentinel: Could not start nREPL server: Error building classpath. Don't know how to create ISeq from: clojure.lang.Symbol

java.lang.IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.Symbol

	at clojure.lang.RT.seqFrom(RT.java:557)

	at clojure.lang.RT.seq(RT.java:537)

	at clojure.core$seq__5467.invokeStatic(core.clj:139)

	at clojure.core$merge_with$merge2__6033.invoke(core.clj:3083)

	at clojure.lang.ArrayChunk.reduce(ArrayChunk.java:63)

	at clojure.core$reduce1.invokeStatic(core.clj:944)

	at clojure.core$reduce1.invokeStatic(core.clj:936)

	at clojure.core$merge_with.invokeStatic(core.clj:3076)

	at clojure.core$merge_with.doInvoke(core.clj:3068)

	at clojure.lang.RestFn.applyTo(RestFn.java:139)

	at clojure.core$apply.invokeStatic(core.clj:669)

	at clojure.tools.deps$merge_edns.invokeStatic(deps.clj:141)

	at clojure.tools.deps.script.make_classpath2$run_core.invokeStatic(make_classpath2.clj:86)

	at clojure.tools.deps.script.make_classpath2$run.invokeStatic(make_classpath2.clj:153)

	at clojure.tools.deps.script.make_classpath2$_main.invokeStatic(make_classpath2.clj:203)

	at clojure.tools.deps.script.make_classpath2$_main.doInvoke(make_classpath2.clj:174)

	at clojure.lang.RestFn.applyTo(RestFn.java:137)

	at clojure.lang.Var.applyTo(Var.java:705)

[...]

PR #3342, opened 2023-05-11
Escape clojure cli command params on MS-win for native commands

ikappaki:
Hi,

could you please review patch to encode clojure cli native command args on MS-Windows. It addresses #3341.

The clojure cli command line args for clojure cli invocations other than powershell also need to be escaped. 

I've relocated the escaping logic into `cider-clojure-cli-jack-in-dependencies`, assuming that powershell invocations are only relevant to clojure-cli, as is the new native escaping.

I've also added a windows-nt only clojure-cli `deps.exe` integration test to test the same. For this, I need to install the deps.clj tools in the workflow for windows-nt.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs
+
+- [#3341](https://github.com/clojure-emacs/cider/issues/3341): Escape clojure-cli args on MS-Windows on non powershell invocations.
+
 ## 1.7.0 (2023-03-23)
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master (unreleased)
 
-### Bugs
+### Bugs fixed
 
 - [#3341](https://github.com/clojure-emacs/cider/issues/3341): Escape clojure-cli args on MS-Windows on non powershell invocations.
 
--- a/cider.el
+++ b/cider.el
@@ -767,12 +767,12 @@ rules to quote it."
          (utf-16le-command (encode-coding-string command 'utf-16le)))
     (format "-encodedCommand %s" (base64-encode-string utf-16le-command t))))
 
-(defun cider-clojure-cli-jack-in-dependencies (global-options params dependencies command)
+(defun cider-clojure-cli-jack-in-dependencies (global-options params dependencies &optional command)
   "Create Clojure tools.deps jack-in dependencies.
 Does so by concatenating DEPENDENCIES, PARAMS and GLOBAL-OPTIONS into a
-suitable `clojure` invocation and quoting suitable for COMMAND invocation.
-The main is placed in an inline alias :cider/nrepl so that if your aliases
-contain any mains, the cider/nrepl one will be the one used."
+suitable `clojure` invocation and quoting, also accounting for COMMAND if
+provided.  The main is placed in an inline alias :cider/nrepl so that if
+your aliases contain any mains, the cider/nrepl one will be the one used."
   (let* ((all-deps (thread-last
                      dependencies
                      (append (cider--jack-in-required-dependencies))
@@ -838,12 +838,12 @@ See also `cider-jack-in-auto-inject-clojure'."
               dependencies))
     dependencies))
 
-(defun cider-inject-jack-in-dependencies (global-opts params project-type command)
+(defun cider-inject-jack-in-dependencies (global-opts params project-type &optional command)
   "Return GLOBAL-OPTS and PARAMS with injected REPL dependencies.
 These are set in `cider-jack-in-dependencies', `cider-jack-in-lein-plugins'
 and `cider-jack-in-nrepl-middlewares' are injected from the CLI according
-to the used PROJECT-TYPE and COMMAND.  Eliminates the need for hacking
-profiles.clj or the boot script for supporting CIDER with its nREPL
+to the used PROJECT-TYPE, and COMMAND if provided.  Eliminates the need for
+hacking profiles.clj or the boot script for supporting CIDER with its nREPL
 middleware and dependencies."
   (pcase project-type
     ('lein (cider-lein-jack-in-dependencies
--- a/cider.el
+++ b/cider.el
@@ -744,12 +744,19 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   "Removes the duplicates in DEPS."
   (cl-delete-duplicates deps :test 'equal))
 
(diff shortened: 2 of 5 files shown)


ikappaki:
> lgtm - great work!

Thanks @vemv, I've  just released I need to rebase some tests that escaped my attention. I'm also thinking of some further improvements to the code to treat powershell quoting a special `shell-quote-args` case rather than a hack.

I'll post an update soon.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs
```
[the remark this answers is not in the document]
ikappaki: Done.

ikappaki:
Hi, 

I've updated the code to treat PowerShell quoting similarly to `shell-quote-argument` via the new `cider--shell-quote-argument`, rebased the tests that were failing due to the previous commit, and extended an existing test to cover a PowerShell jack in invocation. 

all tests are passing, I think we are good to go for the final review.

Thanks

cider.el
```diff
@@ -813,13 +838,13 @@ See also `cider-jack-in-auto-inject-clojure'."
               dependencies))
     dependencies))
 
-(defun cider-inject-jack-in-dependencies (global-opts params project-type)
+(defun cider-inject-jack-in-dependencies (global-opts params project-type command)
```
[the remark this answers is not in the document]
bbatsov: I was going to mention this as well. Right now it's a breaking API change. Alternatively (and probably my preference) we can introduce a new private function with a similar name (e.g. `cider--inject-jack-in-dependencies`) and there the params can be in their natural order - starting with `command`. We should check at some point which of the functions are used in `clj-refactor` to reduce the surface of the public API here, as I think many functions are public for no particular reason.

test/cider-tests.el
```diff
@@ -268,7 +268,7 @@
                                 (shell-quote-argument "cider.nrepl/cider-middleware")
                                 " repl -s wait")))
     (it "can concat in a gradle project"
-      (expect (cider-inject-jack-in-dependencies "--no-daemon" ":clojureRepl" 'gradle)
+      (expect (cider-inject-jack-in-dependencies "--no-daemon" ":clojureRepl" 'gradle "grandle")
```
bbatsov, line 271: gradle :-)

cider.el
```diff
@@ -744,12 +744,35 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
...
+    (shell-quote-argument argument)))
+
+(defun cider--powershell-encode-command (cmd-params)
+  "Base64 encode the powershell command and jack-in CMD-PARAMS for clojure-cli."
+  (let* ((quoted-params cmd-params)
+         (command (format "clojure %s" quoted-params))
+         (utf-16le-command (encode-coding-string command 'utf-16le)))
+    (format "-encodedCommand %s" (base64-encode-string utf-16le-command t))))
+
+(defun cider-clojure-cli-jack-in-dependencies (global-options params dependencies command)
```
bbatsov, line 770: This also looks like a breaking API change.

cider.el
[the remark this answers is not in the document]
ikappaki: Thanks both.

> I was going to mention this as well. Right now it's a breaking API change. Alternatively (and probably my preference) we can introduce a new private function with a similar name (e.g. `cider--inject-jack-in-dependencies`) and there the params can be in their natural order - starting with `command`. We should check at some point which of the functions are used in `clj-refactor` to reduce the surface of the public API here, as I think many functions are public for no particular reason.

I provisionally set the new argument to `&optional` since it appears to have a couple of advantages over creating a new function: 

1. The new feature/bug fix is made available on the API for clients to consume.
2. The duplicate fn requires to duplicate most of the comments, i.e. introduce more boilerplate.

Let me know what you think, I don't mind going for the private fn option.

test/cider-tests.el
ikappaki: I've updated the API and this argument is now optional, thus I removed the new argument altogether since it only effective in clojure-cli

cider.el
ikappaki: I've provisionally set the argument as `&optional` pending further discussion.

ikappaki:
Hi, I've provisionally set the new jack-in-dependencies COMMAND arg to `&optional` pending further discussion whether this is satisfactory or go with the private fns solution. Let me know what you think.

I've also added a test for `cider--shell-quote-argument` which I've missed with my previous commit.

thanks

bbatsov:
I think we're good to go now. Thanks!

ikappaki:
> Unfortunately the build is red now:
> image

Hi @vemv,

could this have been a GH transient issue GH accessing ... GH via PowerShell?  Can you please rerun the action and see if it passes?

I've just updated my master CIDER fork with the latest updates and that particular bit worked fine:

> Run iwr -Uri https://raw.githubusercontent.com/borkdude/deps.clj/master/install.ps1 -outfile install_clojure.ps1
>   iwr -Uri https://raw.githubusercontent.com/borkdude/deps.clj/master/install.ps1 -outfile install_clojure.ps1
>   .\install_clojure.ps1
>   get-command deps.exe | split-path -parent | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
>   shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
>   env:
>     PATH: C:\Program Files\MongoDB\Server\5.0\bin;C:\aliyun-cli;C:\vcpkg;C:\Program Files (x86)\NSIS\;C:\tools\zstd;C:\Program Files\Mercurial\;C:\hostedtoolcache\windows\stack\2.9.3\x64;C:\cabal\bin; ...
> Downloading...
> Extracting...
> Installing deps.exe to C:\Users\runneradmin\deps.clj...
> Adding ~\deps.clj to your path.
> Cleaning up...
> Successfully installed deps.exe.
> Restart cmd.exe for changes to the path to take effect.

Looking at the `deps.clj` install_clojure.clj script at https://raw.githubusercontent.com/borkdude/deps.clj/master/install.ps1, and specifically the lines 30 and 31 (it fails at L31)

```
L30: $releast_version_url = "https://raw.githubusercontent.com/borkdude/deps.clj/master/resources/DEPS_CLJ_RELEASED_VERSION"
L31: $latest_release = (Invoke-WebRequest -Uri $releast_version_url -UseBasicParsing)
```

it  tries to download `https://raw.githubusercontent.com/borkdude/deps.clj/master/resources/DEPS_CLJ_RELEASED_VERSION`, that shouldn't fail, unless there was an transient networking issue perhaps?

Thanks

PR #3349, opened 2023-05-31
Modified `xref-backend-references` to use buffer locations instead of filenames

### LINKED ISSUE #3112: cider--xref-backend returns incorrect relative filenames for references
## Expected behavior
xref-find-references  with cider backend must return either absolute file name or the name relative to the file of the current buffer.

## Actual behavior
the file name returned is realtive to src directory and, as a result, selecting a reference in xref buffer doesn't work.

Put a cursor to some function in a clojure buffer, invoke xref-find-references. Try to select some of the candidates in the xref window.

### CIDER version information

```
;; CIDER 1.2.0snapshot (package: 20211215.925), nREPL 0.9.0-beta5
;; Clojure 1.10.3, Java 11.0.13
```

### Emacs version

GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4)
 of 2021-03-26

### Operating system

Arch Linux

jeeger:
The filenames returned by `cider-sync-request:fn-refs` were not in the format expected by uxref,
probably because they are relative to the project root and not relative to the current buffer (which
is what xref needs, from my understanding). I've implemented a somewhat dirty fix by using identical
code to `xref-backend-definitions`, which uses `cider--find-buffer-for-filename`. This function
seems to work around the issue, with the drawback of potentially opening a large amount of buffers.

The cleaner way to fix this bug would be computing absolute file names, but I am unsure what the filenames returned by `cider-sync-request:fn-refs` are relative to.

Fixes #3112.


bbatsov:
> The cleaner way to fix this bug would be computing absolute file names, but I am unsure what the filenames returned by cider-sync-request:fn-refs are relative to.

They are relative to `clojure-project-dir` (from `clojure-mode`).

jeeger:
Am I correct in thinking that I should convert filenames using that information instead of the proposed fix in the PR?

jeeger:
I'm looking at contributing to cider-nrepl, but I'm having problems loading a locally installed version of `cider-nrepl` into leiningen. Is there something special besides `PROJECT_VERSION=… make install` I need to do?

iarenaza:
> Am I correct in thinking that I should convert filenames using that information instead of the proposed fix in the PR? Update: And I think that would cause `cider` to have a dependency on `clojure-mode`, which it currently doesn't have?

I would rather use the `cider--find-buffer-for-filename` way. That function takes care of translating the file paths as seen by the REPL process, to the local paths seen by Emacs. These two might be different (are usually different) if you are using Tramp or if you are running your REPL process in another machine or inside a container (or if you are using a couple of other special cases also handled by `cider-find-file`). `cider--find-buffer-for-filename` takes care of all of those details, by calling `cider-find-file`.

bbatsov:
Fine by me.

bbatsov:
Yep. Probably we should eventually have the middleware return absolute paths, but I think it's preferable short-term to go with any fix instead of leaving this functionality broken. I've been super busy in the past few weeks, otherwise I would have tackled this myself.

cider-find.el
```diff
@@ -280,7 +280,13 @@ thing at point."
               (let* ((filename (nrepl-dict-get info "file"))
                      (column (nrepl-dict-get info "column"))
                      (line (nrepl-dict-get info "line"))
-                     (loc (xref-make-file-location filename line column)))
+                     (buf (cider--find-buffer-for-file filename))
+                     (loc (xref-make-buffer-location buf (with-current-buffer buf
```
bbatsov, line 284: I'd add some comment here explaining the need for this and some todo to eventually use `xref-make-file-location` once the nREPL api has been modified.

bbatsov:
@vemv Did you see @iarenaza's concerns?

bbatsov:
I have to admit I still use grep instead of xref, so for me that's not a big issue. 😆

jeeger:
I was looking into orchard/cider-nrepl, but I didn't get anywhere last week — turns out that the `:file` meta property is either already absolute (when something is evaluated in the REPL) or relative to the classpath (when loaded from somewhere else). Since references can also be inside JAR files, I got the feeling that reusing the already existing logic was the most expedient way forward.

bbatsov:
No objections from me.

vemv:
The final fix took some inspiration from here, with nuances, and also some changes at cider-nrepl level.

Things should now be pretty for all things xref:

* references
* deps
* jump to definition

Feel free to try them out in CIDER master.

Thanks again!

PR #3355, opened 2023-06-16
Fix issues with cider-auto-mode

### LINKED ISSUE #3346: Deprecate or change default setting for `cider-auto-mode`
The default `t` setting for `cider-auto-mode` causes performance and other issues detailed in the above issue, and only benefits a (presumably small) subset of Cider users with workflows involving inf-clojure and not adding cider-mode to their major mode hooks before until the first Cider-initiated connection.

I suggest the default config should at least be changed to nil, solving the downsides for most users. Users of `inf-clojure` can re-enable it in their own configs if needed.
If the feature is deprecated or removed entirely, it is also relatively simple to implement in one's own config, adding custom functions to `cider-connected-hook` and `cider-disconnected-hook`.

## Possible issues:

New and inexperienced users of Emacs/Cider might unknowingly depend on this behaviour, manually calling `M-x cider-mode` in a clojure buffer the first time they start Emacs, jacking in and then magically not having to do it in subsequent buffers. (Perhaps treating it like some sort of global minor mode)

I think the docs should encourage the usual Emacs convention of adding minor modes to a major mode hook, instead of supporting incorrect mindsets or habits which may cause future confusion for users.

https://docs.cider.mx/cider/config/basic_config.html#disable-automatic-cider-mode-in-clojure-mode-buffers

### CIDER version information

```
;; CIDER 1.7.0 (Côte d'Azur), nREPL 1.0.0
;; Clojure 1.11.1, Java 20
```

Clojure CLI version 1.11.1.1273

### Emacs version

GNU Emacs 29.0.90 

### Operating system

macOS 12.5.1

### JDK distribution

Temurin jdk-20.0.1+9

yuhan0:
The first two commits fix the cider-mode disabling behavior and performance issues mentioned in #3346.
The latter ones introduce a technically breaking change by changing the default setting of cider-auto-mode, and rewording the getting started section of the user manual to something hopefully less misleading and more beginner-friendly.

-----------------


yuhan0:
Sorry for the noisy commits, I'm not sure why these warnings only show up in the CI and not when I run `eldev lint / test` locally.

bbatsov:
@yuhan0 Let's pull the change to the default to a separate PR as that's a breaking change likely to affect a lot of users and I want to discuss the implications better.

Generally, I'm worried that a lot of people will be surprised that cider-mode doesn't get enabled automatically. I've gotten a lot of flak in the past for such changes, so I try to be extra careful with them. :-)

yuhan0:
Ok, I've rebased the non-breaking commits and added on the extra changes from #3354 (going with the double prompt to kill REPLs after choosing not to reuse them)

yuhan0:
Updated the PR description so #3346 doesn't get auto-closed, discussion can be continued there :)

bbatsov:
Thanks!

PR #3366, opened 2023-07-05
Display debug-on-exception messages

yuhan0:
Support for https://github.com/clojure-emacs/cider-nrepl/pull/772

A quick PR without the proper changelogs / updates to manual, etc. Changes untested after being rebased on master branch, but I've been dogfooding them on a messy local branch with many (hopefully unrelated) changes for the last month or two. Might be able to clean things up a bit this weekend.

Does not provide feature for auto-inserting `#dbg!` macros or instrumenting defns (eg. with negative argument) - see #3337

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-debug.el
+++ b/cider-debug.el
@@ -132,7 +132,10 @@ configure `cider-debug-prompt' instead."
               (member "stack" status))
       ;; TODO: Make the error buffer a bit friendlier when we're just printing
       ;; the stack.
-      (cider--render-stacktrace-causes causes))
+      (nrepl-dbind-response response (causes caught-msg)
+        (if cider-show-error-buffer
+            (cider--render-stacktrace-causes causes)
+          (cider--debug-display-result-overlay nil caught-msg))))
     (when (member "need-debug-input" status)
       (cider--handle-debug response))
     (when (member "done" status)
@@ -154,15 +157,17 @@ configure `cider-debug-prompt' instead."
   #("." 0 1 (display (left-fringe right-triangle)))
   "Used as an overlay's before-string prop to place a fringe arrow.")
 
-(defun cider--debug-display-result-overlay (value)
+(defun cider--debug-display-result-overlay (value caught)
   "Place an overlay at point displaying VALUE."
   (when cider-debug-use-overlays
     ;; This is cosmetic, let's ensure it doesn't break the session no matter what.
     (ignore-errors
       ;; Result
-      (cider--make-result-overlay (cider-font-lock-as-clojure value)
+      (cider--make-result-overlay (or caught (cider-font-lock-as-clojure value))
         :where (point-marker)
         :type 'debug-result
+        :prepend-face (if caught 'cider-error-overlay-face
+                        'cider-result-overlay-face)
         'before-string cider--fringe-arrow-string)
       ;; Code
       (cider--make-overlay (save-excursion (clojure-backward-logical-sexp 1) (point))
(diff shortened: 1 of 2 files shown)


bbatsov:
> many (hopefully unrelated) changes

Looking forward to seeing some of those unrelated changes become PRs as well. ;-)

cider-debug.el
```diff
@@ -154,15 +157,17 @@ configure `cider-debug-prompt' instead."
   #("." 0 1 (display (left-fringe right-triangle)))
   "Used as an overlay's before-string prop to place a fringe arrow.")
 
-(defun cider--debug-display-result-overlay (value)
+(defun cider--debug-display-result-overlay (value caught)
```
bbatsov, line 160: You'll need to mention this in the docstring to make checkdoc happy.

bbatsov:
> Might be able to clean things up a bit this weekend.

No rush. CIDER 1.8 is still in development in probably won't be released for a week or two more. If this PR can make there it'd be great, but if not - not a big deal.

vemv:
Hey @yuhan0 - do you have the chance to revisit this one?

Else it seems quite minor - could give it a shot myself today/tomorrow.

yuhan0:
> Else it seems quite minor - could give it a shot myself today/tomorrow.

@vemv That would be great if you could 🙏  
Sorry for the long absence but I've been away from OSS / Clojure-related work for the past month or so.. Glad to see such activity around Cider 1.8 happening in the meantime, but realistically I won't have time to pull these new changes and integrate/test against them in next couple of days. 
Feel free to adapt this PR however you like!

vemv:
@yuhan0 if you'd like to re-take this PR siib let us know, I plan to work on misc stuff for the following few days

yuhan0:
Thanks - I'm tentatively back from that long-ish absence, currently in the midst of piecing together a broken Emacs config from the 29.1 update 😅  

But yes, I'll be able to revisit this PR as soon as I re-familiarize myself with the issue and the recent updates to Cider and friends :)

vemv:
Cheers. Much better like that 🍻

bo-tato:
great to see yuhan0 has put a lot of work on this :smile: 
I was reading some about jvm and [Debug Wire Protocol](https://docs.oracle.com/en/java/javase/20/docs/specs/jdwp/jdwp-protocol.html), and it seems you are able to have it break on uncaught exceptions. The only catch being that if there is native code in the call stack, it will sometimes break on an "uncaught exception" that really will be caught by the native code but the jvm can't know that. There is also the `PopFrames` command that allows to restart execution from some other frame higher up in the stack. It seems a near Common Lisp level experience is possible with clojure on the JVM, to break on uncaught exceptions and inspect local variables in all stack frames and fix code and then restart from some stack frame.
There is [this project](https://github.com/GeorgeJahad/cdt/) that uses jvm debug wire protocol to implement a clojure debugger but it is long since abandoned, and cider and flowstorm debugger took a different approach of instrumenting code, which is great but has limitations that you have to first replicate the issue, and then instrument the code and rerun it, rather than having the debugger already present when you first run into the issue. It seems jdwt has almost no overhead when you don't have breakpoints set, so it could even be possible to leave running in production, and log local variables and function arguments and everything at times of exception to easily reproduce whatever exceptions your application throws in production. It seems there is some [commercial](https://www.harness.io/products/continuous-error-tracking) product for the jvm that does this but I couldn't find anything open source.
It seems it would be promising to build a clojure debugger on top of jdwp that would allow much better experience regarding breaking on exceptions. Does anyone know why cdt was discontinued and the clojure debuggers that came after took a different approach, are there limitations or difficulties to debugging clojure code with jdwp?

bbatsov:
@bo-tato I can answer your questions, but I'd suggest to move this to a discussion topic first as I don't think it belongs here.

bo-tato:
you're right it's getting off topic from the PR, I opened a discussion [here](https://github.com/clojure-emacs/cider/discussions/3452)

yuhan0:
Sorry for delaying this feature this past 6 months..! I must have completely forgot about this PR 👀 

Rebased onto master and the error overlays seem to be working as expected, also fixed the lint warnings from last time. 

image

bbatsov:
Thanks!

PR #3367, opened 2023-07-06
Don't use keybindings reserved for users

r0man:
I initially picked the keybinding prefix `C-c l <key>` for CIDER log mode not being aware that `C-c` is reserved for users. I now changed this to `C-x M-l <key>`.

-----------------

Thanks!

manual][1] extremely useful.*


cider-log.el
```diff
@@ -797,12 +797,12 @@ The KEYS are used to lookup the values and are joined by SEPARATOR."
 (defvar cider-log-mode-map
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map logview-mode-map)
-    (define-key map (kbd "C-c l a") #'cider-log-appender)
-    (define-key map (kbd "C-c l c") #'cider-log-consumer)
-    (define-key map (kbd "C-c l e") #'cider-log-event)
-    (define-key map (kbd "C-c l f") #'cider-log-framework)
-    (define-key map (kbd "C-c l i") #'cider-log-info)
-    (define-key map (kbd "C-c l l") #'cider-log)
+    (define-key map (kbd "C-x l a") #'cider-log-appender)
```
bbatsov, line 800: We can't use `C-x l` directly for various reasons (e.g. a built-in command is using it and it's not a mode map prefix), it should be something like `C-c C-l` or `C-x C-l` (provided we're not already using them for something else).

r0man: I changed it to `C-x C-l` now.

bbatsov:
Damn, just saw that `C-x C-l` is bound to `downcase-region` in Emacs, so it's a no go as well. `C-c C-l` is `cider-load-file`. You'll have to get more creative. :-)

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+- [#3367](https://github.com/clojure-emacs/cider/pull/3367) Change CIDER Log Mode keybindings to use `C-c M-l <key>` instead of user reserved bindings.
```
bbatsov, line 7: That's not needed, as `cider-log-mode` hasn't been released yet. :-)

r0man: I removed it.

bbatsov:
Thanks!

PR #3368, opened 2023-07-06
Make logview optional

r0man:
This PR makes logview optional. If it is available (fboundp) it will be used otherwise not. 
-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-log.el
+++ b/cider-log.el
@@ -28,7 +28,7 @@
 (require 'cider-inspector)
 (require 'cider-stacktrace)
 (require 'cl-lib)
-(require 'logview)
+(require 'logview nil t)
 (require 'org)
 (require 'seq)
 (require 'transient)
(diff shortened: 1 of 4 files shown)


cider-log.el
```diff
@@ -82,6 +82,13 @@
   :safe #'integerp
   :type 'integer)
 
+(defcustom cider-log-logview-p (fboundp 'logview-mode)
```
bbatsov, line 85: I'd name this `cider-log-use-logview`.

cider-log.el
```diff
@@ -829,6 +830,22 @@ the CIDER Inspector and the CIDER stacktrace mode.
   (when (fboundp 'evil-set-initial-state)
     (evil-set-initial-state 'cider-log-mode 'emacs)))
 
+(defvar cider-log--mode-doc
```
bbatsov, line 833: Why the `-doc` suffix?

r0man: This is just the doc string for the mode, which I define below in a conditional (depending if log mode is avalable or not).

r0man: I didn't find a better way to derive a mode depending on a condition (except for using `eval` maybe). Do you have a better idea?

bbatsov: This will do.

r0man:
@bbatsov  I now added `(eldev-add-extra-dependencies 'runtime '(:package logview :optional t))` to the Eldev file  to fix the byte compile warnings. But I wonder, do we run the test also on NonGNU ELPA? And would it fail there? I see that I could also specify a `:archive` field in the specification.

bbatsov:
@r0man I don't think they run our test suite.

r0man:
@bbatsov Ok, perfect. The other failing test runs are because of some rate limit I believe.

bbatsov:
Okay, thanks for tackling this!

r0man:
Thanks for the advice!

PR #3373, opened 2023-07-20
`cider-test`: add timing information

vemv:
> See also: https://github.com/clojure-emacs/cider-nrepl/issues/755

The results can be seen in these annotated screenshots:

Screen Shot 2023-07-20 at 21 40 56

Screen Shot 2023-07-20 at 21 41 46

Cheers - V

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -526,7 +526,7 @@ The optional arg TEST denotes an individual test name."
       (let* ((ms (nrepl-dict-get elapsed-time "ms"))
              (ms (if ms
                      (propertize (concat " in " (prin1-to-string ms) "ms") 'face 'font-lock-comment-face)
-                   "")))
+                   ".")))
       (message (propertize
                 "%sRan %d assertions, in %d test functions. %d failures, %d errors%s"
                 'face (cond ((not (zerop error)) 'cider-test-error-face)


cider-test.el
```diff
@@ -410,6 +413,10 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
                 (bg `(:background ,cider-test-items-background-color :extend t)))
             (cider-insert (capitalize type) type-face nil " in ")
             (cider-insert var 'font-lock-function-name-face t)
+            (when elapsed-time
+              (let ((humanized (nrepl-dict-get elapsed-time "humanized")))
+                (when humanized
```
bbatsov, line 418: You use `when-let` to make this a bit more compact.

bbatsov:
I'm not sure I like the position of the timing info in the minibuffer message, but that's not a big deal. Apart from my small remark about `when-let` the PR looks good to me.

bbatsov:
You can just use the comment face for them and put this text in parentheses, so it stands more apart from the rest of the information in the message.

cider-test.el
```diff
@@ -524,20 +523,20 @@ The optional arg TEST denotes an individual test name."
     (if (nrepl-dict-empty-p results)
         (message (concat (propertize "No assertions (or no tests) were run." 'face 'cider-test-error-face)
                          "Did you forget to use `is' in your tests?"))
-      (let* ((humanized (nrepl-dict-get elapsed-time "humanized"))
-             (humanized (if humanized
-                            (propertize (concat " " humanized) 'face 'default)
-                          "")))
+      (let* ((ms (nrepl-dict-get elapsed-time "ms"))
+             (ms (if ms
+                     (propertize (concat " in " (prin1-to-string ms) "ms") 'face 'font-lock-comment-face)
```
bbatsov, line 528: It's better to use `format` than `concat` here. It would remove the need for `prin1-to-string`.

PR #3374, opened 2023-07-21
Introduce fail-fast functionality

vemv:
Adds a defcustom for controlling `"fail-fast"` as introduced in https://github.com/clojure-emacs/cider-nrepl/pull/783

Commits:

* Prefer `format`
  * https://github.com/clojure-emacs/cider/pull/3373#discussion_r1270531001
* Document indentation inference
* Bump cider-nrepl
  * https://github.com/clojure-emacs/cider-nrepl/blob/39c39e3f57a79dad7333ad0220093556feb87a9a/CHANGELOG.md#master-unreleased
* Introduce `cider-test-fail-fast`

I've QA'd this and works as intended. Can be easiy toggled with `(setq cider-test-fail-fast <nil | t>)`

I also used the chance to QA that `diff:` is now only rendered when it makes sense:

image

Cheers - V

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -389,6 +389,9 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
         (cider-insert (format "%d errors" error) 'cider-test-error-face t))
       (when (zerop (+ fail error))
         (cider-insert (format "%d passed" pass) 'cider-test-success-face t))
+      (when cider-test-fail-fast
+        (cider-insert "cider-test-fail-fast: " 'font-lock-comment-face nil)
+        (cider-insert "t" 'cider-test-constant-face t))
       (insert "\n\n"))))
 
 (defun cider-test-render-assertion (buffer test)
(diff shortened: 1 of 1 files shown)


cider-test.el
```diff
@@ -653,6 +653,12 @@ The selectors can be either keywords or strings."
    (split-string
     (cider-read-from-minibuffer message))))
 
+(defcustom cider-test-fail-fast t
+  "If `t', test will be considered complete as soon as the first given test
```
bbatsov, line 657: This will break the Elisp linting - the first line has to be a sentence on its own. E.g. "Controls whether to stop a test run on failure/error.".

bbatsov:
Apart from my small remark - everything looks good to me.

cider-test.el
```diff
@@ -653,6 +653,12 @@ The selectors can be either keywords or strings."
    (split-string
     (cider-read-from-minibuffer message))))
 
+(defcustom cider-test-fail-fast t
```
bbatsov, line 656: I'm wondering a bit if we shouldn't keep this as `nil` by default and change it only in CIDER 2.0, as the new behavior might surprise some users. Alternatively we can display somewhere in the report whether fast fail is enabled or not, so no one would be surprised.

PR #3375, opened 2023-07-21
`cider-test`: don't render a newline between expected and actual

vemv:
Before:

Screen Shot 2023-07-21 at 17 57 03

After:

Screen Shot 2023-07-21 at 17 56 30

Personally I just didn't like this newline. For large reports it can make quite a difference.

Cheers - V

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -401,12 +401,10 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
       (insert "\n\n"))))
 
 (defun cider-test--string-contains-newline (input-string)
-  "Returns whether INPUT-STRING contains a newline.
-Ignores any newlines at the end of the string."
+  "Returns whether INPUT-STRING contains an escaped newline."
   (when (stringp input-string)
-    (let ((trimmed-string (replace-regexp-in-string "\n\\'" "" input-string)))
-      (and (string-match-p "\n" trimmed-string)
-           t))))
+    (and (string-match-p "\\n" input-string)
+         t)))
 
 (defun cider-test-render-assertion (buffer test)
   "Emit into BUFFER report detail for the TEST assertion."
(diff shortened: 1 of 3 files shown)


bbatsov:
For short results it looks better without the newline, but I wondering if it helps readability for multi-line results. Let me think a bit about this.

bbatsov:
I meant the same @aisamu, but I don't feel strongly about this.

vemv:
Makes sense @aisamu - implemented!

image


image

image

image

cider-test.el
```diff
@@ -401,12 +401,10 @@ With the actual value, the outermost '(not ...)' s-expression is removed."
...
 
 (defun cider-test--string-contains-newline (input-string)
-  "Returns whether INPUT-STRING contains a newline.
-Ignores any newlines at the end of the string."
+  "Returns whether INPUT-STRING contains an escaped newline."
   (when (stringp input-string)
-    (let ((trimmed-string (replace-regexp-in-string "\n\\'" "" input-string)))
-      (and (string-match-p "\n" trimmed-string)
-           t))))
+    (and (string-match-p "\\n" input-string)
```
mattiase, line 406: What does this regexp _really_ match?

vemv: Escaped newlines as coming from Clojure / cider-nrepl. These aren't vanilla elisp newlines.

mattiase: It matches the letter "n" in a string.

vemv: Thanks! will fix

vemv: https://github.com/clojure-emacs/cider/commit/605e253776b76fdc6ec821626842ed9d566f88df

mattiase: Thank you!

Issue #3236, opened 2022-08-06
cider-repl-set-ns changes repl session from cljs:shadow to clj


kakaLQY:
## Expected behavior

`cider-repl-set-ns` should not change the `cider-repl` session type.

## Actual behavior

`cider-repl-set-ns` in a `cljs:shadow` session will change the session type to `clj`. 

## Steps to reproduce the problem

In a shadow-cljs project, run `cider-repl-set-ns`. The `cider-repl` session will be changed from `cljs:shadow` to `clj` by checking with `sesman-browser`.

### CIDER version information

```
;; CIDER 1.5.0-snapshot (package: 1.5.0-snapshot), nREPL 0.9.0
;; Clojure 1.10.3, Java 11.0.15
```

### Lein / Clojure CLI version

1.11.1 

### Emacs version

28.1

### Operating system

NixOS

### JDK distribution

openjdk version "11.0.15" 2022-04-19

## Hints

It seems the bug is introduced by https://github.com/clojure-emacs/cider/commit/5f705b0a3a0266e4e64bc37016135a6284d2321f, as after reverted, the bug will be gone.

caumond:
+1, on CIDER 1.5.0-snapshot, clj 1.10.0, emacs 28.1, ubuntu, jdk 11.0.16+8-post-Ubuntu-0ubuntu122.04

jellelicht:
This behaviour seems to somehow be triggered through the `cider-repl--state-handler` function, and specifically the following snippet:
```elisp
(when (and repl-type cider-repl-auto-detect-type)
    (cider-set-repl-type repl-type))
```

For whatever reason, the nrepl response seems to indicate a `repl-type` of `clj` in my shadow-cljs repl.

jellelicht:
Using `cider-tooling-eval` passes along `'tooling` to `nrepl-request:eval`. I'm guessing this leads to using the 'wrong' session somehow, one where the repl-type is (still) `clj` instead of `cljs`.

bbatsov:
That's an interesting edge case as there's just a single tooling session and it's a Clojure session. Perhaps for ClojureScript we can run this in the ClojureScript eval session instead, although it's going to mess a bit with the eval results vars (e.g. `*1`).

jellelicht:
@bbatsov Perhaps a stupid idea, but maybe we can temporarily bind `cider-repl-auto-detect-type` to be falsey?

tvirolai:
[...]

PR #3376, opened 2023-07-21
Fix: `cider-repl-set-ns` no longer changes the repl session type from `cljs:shadow` to `clj`

vemv:
> Closes https://github.com/clojure-emacs/cider/issues/3236

I could both repro the original issue and verify that this proposed fix works.

The minimal repro was:

```clj
(nrepl-sync-request:eval "42"
                         (cider-current-repl nil 'ensure)
                         "my.ns"
                         'tooling)
```

As soon as one evaluated that code, the cljs repl type would be change to `clj`.

Thanks much to @jellelicht for his analysis: https://github.com/clojure-emacs/cider/issues/3236#issuecomment-1246044723

Cheers - V


bbatsov:
Looks good to me.

PR #3380, opened 2023-07-26
Show per-var test timing

vemv:
* Show per-var test timing in the summary
* Don't show per-testing-context timing anymore
  * It's generally not as reliable
* Add `cider-test-toggle-fail-fast` interactive command and keybindings

More details: https://github.com/clojure-emacs/cider-nrepl/pull/784

Result:

image

Worth noting, in this particular screenshot, there are quite a few vars, however with the `cider-test-fail-fast` default, typically just one will be displayed, which is cleaner.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - [#3354](https://github.com/clojure-emacs/cider/issues/3354): Add new customization variable `cider-reuse-dead-repls` to control how dead REPL buffers are reused on new connections.
 - `cider-test`: add timing information.
 - `cider-test`: only show diffs for collections.
-- `cider-test`: fail-fast by default, as controlled by the new `cider-test-fail-fast` defcustom.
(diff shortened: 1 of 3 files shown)


bbatsov:
Perhaps `C-c C-t f` & `C-c C-t C-f`?

bbatsov:
> Do these mean 'toggle', or one for enabling, one for disabling?

`toggle`. We simply have duplicated keybindings for almost everything as some people don't like letting go of the Control key.

bbatsov:
One more thing - it might be nice to make the list of vars expandable (make the ns a clickable item), so people can view it only if they want to.

bbatsov:
I think the stacktrace code has some code about hiding/showing elements. There's also some code to interact with elements in the inspector, namespace browser and apropos. Might be something similar elsewhere, but I don't remember at this point.

vemv:
Added the keybindings, doc included.

If you don't mind, I'd like to leave the UI toggling for some other occasion, so that I can focus on bigger topics atm.

Issue #3331, opened 2023-04-05
Inconsistent cursor behaviour after evaluation of clojure code resulting in error


practicalli-john:
## Expected behavior

Consistent behaviour of the cursor after evaluating an expression in the buffer that causes an error.

The behaviour should happen irrespective of if `cider-show-error-buffer` is set to `t` or `nil` (except when `t` then `q` should be pressed to quit the error buffer)

## Actual behavior

Cursor inconsistent movement after evaluating an expression in the buffer that causes an error (incorrect clojure code)

## Steps to reproduce the problem

Open a Clojure .clj file from a project which has a namespace definition:

1. Call a function that is not defined and the cursor will jump to the opening bracket of the call expression, `|(foo)`
2. Call an evaluated function with incorrect number of arguments, cursor does not move
3. Evaluate a non-existent symbol, e.g `blahblah` and the cursor jumps to the top of the namespace

It is assumed that 1. and 2. would behave the same, the cursor jumping to the start of the calling expression (or not moving)

Behaviour 3 feels quite strange, I dont know what value is provided by the cursor jumping away (I appreciate this situation would not happen very often though)

Is there a way to customise the cursor movement on error? Or is it simply the way its been implemented?

### CIDER version information

```
CIDER 1.7.0-snapshot (package: 20230226.1413)
nrepl 1.0.0
```

### Lein / Clojure CLI version

Clojure CLI version 1.11.1.1252

### Emacs version

Emacs 28

### Operating system

Ubuntu 22.04

### JDK distribution
```
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment (build 17.0.6+10-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.6+10-Ubuntu-0ubuntu122.04, mixed mode, sharing)
```

PR #3382, opened 2023-07-26
`cider-eval`: never jump to spurious locations

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3331

Maybe not the best possible fix, since in #3331 my analysis suggests fixing this at nrepl level.

Anyway, this fix seems fairly universal:

* We fix the current issue, allowing us to deprioritize the nrepl one which seems complex.
* We fix the issue for users of older nrepls
* We might also prevent unforeseen cases by which column info might also be spurious
  * e.g. `eval` coming from runtime-powered tools like refactor-nrepl, eastwood, etc. that might live in the same JVM.

I verified locally that it works as intended.

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -626,7 +626,7 @@ If location could not be found, return nil."
                     (goto-char (point-min))
                     (forward-line (1- line))
                     (move-to-column col)
-                    ;; if this condition is true, it means that `col` was a spuriously large value,
+                    ;; if this condition is false, it means that `col` was a spuriously large value,
                     ;; therefore the whole calculation should be discarded:
                     (when (or (not col) ;; if there's no col info, we cannot judge if it's spurious/not
                               ;; (current-column) never goes past the last column in the actual file,
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -625,12 +625,19 @@ If location could not be found, return nil."
                     (widen)
                     (goto-char (point-min))
                     (forward-line (1- line))
-                    (move-to-column (or col 0))
-                    (let ((begin (progn (if col (cider--goto-expression-start) (back-to-indentation))
-                                        (point)))
-                          (end (progn (if col (forward-list) (move-end-of-line nil))
-                                      (point))))
-                      (list begin end buffer))))))))))))
+                    (move-to-column col)
+                    ;; if this condition is true, it means that `col` was a spuriously large value,
+                    ;; therefore the whole calculation should be discarded:
+                    (when (or (not col) ;; if there's no col info, we cannot judge if it's spurious/not
+                              ;; (current-column) never goes past the last column in the actual file,
+                              ;; so if it's <, then the message had spurious info:
+                              (= (+ 1 (current-column))
+                                 col))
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -625,12 +625,19 @@ If location could not be found, return nil."
...
-                    (move-to-column (or col 0))
-                    (let ((begin (progn (if col (cider--goto-expression-start) (back-to-indentation))
-                                        (point)))
-                          (end (progn (if col (forward-list) (move-end-of-line nil))
-                                      (point))))
-                      (list begin end buffer))))))))))))
+                    (move-to-column col)
+                    ;; if this condition is false, it means that `col` was a spuriously large value,
+                    ;; therefore the whole calculation should be discarded:
+                    (when (or (not col) ;; if there's no col info, we cannot judge if it's spurious/not
```
bbatsov, line 631: I'd rewrite this as `when (and col (...)`. Reads better IMO.

cider-eval.el
```diff
@@ -625,12 +625,19 @@ If location could not be found, return nil."
...
+                              ;; (current-column) never goes past the last column in the actual file,
+                              ;; so if it's <, then the message had spurious info:
+                              (= (+ 1 (current-column))
```
bbatsov, line 634: Why `=`? I thought you want to check it's not out of bounds?

cider-eval.el
```diff
@@ -625,12 +625,19 @@ If location could not be found, return nil."
...
-                    (let ((begin (progn (if col (cider--goto-expression-start) (back-to-indentation))
-                                        (point)))
-                          (end (progn (if col (forward-list) (move-end-of-line nil))
-                                      (point))))
-                      (list begin end buffer))))))))))))
+                    (move-to-column col)
+                    ;; if this condition is false, it means that `col` was a spuriously large value,
+                    ;; therefore the whole calculation should be discarded:
+                    (when (or (not col) ;; if there's no col info, we cannot judge if it's spurious/not
+                              ;; (current-column) never goes past the last column in the actual file,
```
bbatsov, line 632: file or buffer?

cider-eval.el
```diff
@@ -625,12 +625,19 @@ If location could not be found, return nil."
                     (widen)
                     (goto-char (point-min))
                     (forward-line (1- line))
-                    (move-to-column (or col 0))
```
bbatsov, line 628: I recall that the intention here was to highlight errors without a column normally are at the namespace level, but I can see why some people find this confusing.

cider-eval.el
bbatsov: Well, we can keep it as is but it has to be explained better.

cider-eval.el
bbatsov: My point was simply that I always try to avoid negations in conditions to make them easier to read, but in this case it's ok.

cider-eval.el
vemv: Finally I left it as >= to be on the safe side

vemv:
QAed again, keeps working.

Issue #3385, opened 2023-07-29
Core functions are not highlighted for ClojureScript


nikolavojicic:
## Expected behavior

Core functions should be highlighted like `map` and `range` here. This is how it is for **Clojure**.

![Screenshot 2023-07-29 121408](https://github.com/clojure-emacs/cider/assets/10941031/3143bc86-acec-4f4a-a334-415ce5c9eb2e)

## Actual behavior

Core functions ( `map` and `range`) are not highlighted. This is how it is for **ClojureScript**.

![Screenshot 2023-07-29 121224](https://github.com/clojure-emacs/cider/assets/10941031/dfcac6f7-83c7-4eb9-b189-fadcc70d2af6)

## Steps to reproduce the problem

Content of `shadow-cljs.edn`:
```edn
{:source-paths ["src"]
 :dependencies [[binaryage/devtools "1.0.6"]
                [cider/cider-nrepl "0.32.0"]
                [net.cgrand/xforms "0.19.2"]
                [reagent "1.1.1"]
                [re-frame "1.2.0"]
                [cljs-http "0.1.46"]
                [day8.re-frame/http-fx "0.2.4"]
                [org.clojure/core.async "1.5.648"]]
 :builds       {:app {:target     :browser
                      :output-dir "public/js"
                      :asset-path "/js"
                      :modules    {:app {:entries [fisherman-fe.core]}}
                      :devtools   {:after-load fisherman-fe.core/init!}}}
 :dev-http     {3001 {:root    "public"
                      :handler user/app}}}
```

1. Run `cider-jack-in-cljs`
2. Choose `shadow`
3. Choose `:app`
4. Visit application in browser
5. Write the code above
6. Evaluate that code

## Environment & Version information

- shadow-cljs 2.20.13
- For CIDER and NREPL see screenshots above
- GNU Emacs 28.2 (build 2, x86_64-w64-mingw32) of 2022-09-13
- OpenJDK Runtime Environment Corretto-17.0.2.8.1 (build 17.0.2+8-LTS)
- Windows 10 (also in 11)

bbatsov:
Why are you using `cider-nrepl` 0.32? I'm guessing that's some regression from the recent changes that @vemv has been doing. I guess he'll address this quickly once he sees the ticket.

nikolavojicic:
@vemv advised me to try 0.32... Same behavior is with 0.30

[...]

PR #3379, opened 2023-07-26
Use cider-nrepl 0.34.0

vemv:
https://github.com/clojure-emacs/cider-nrepl/blob/v0.34.0/CHANGELOG.md#0340-2023-08-03

https://clojars.org/cider/cider-nrepl/versions/0.34.0

Cheers - V

PR #3386, opened 2023-07-30
Fix `clojure.core` function highlighting on ClojureScript

vemv:
Part of https://github.com/clojure-emacs/cider/issues/3385

(Needs a trivial cider-nrepl fix too, but I'm adding this as a changelog anyway. Will be upgraded in the following few days)

### DIFF
--- a/cider-resolve.el
+++ b/cider-resolve.el
@@ -107,7 +107,8 @@ This will be clojure.core or cljs.core depending on the return value of the
 function `cider-repl-type'."
   (when-let* ((repl (cider-current-repl)))
     (with-current-buffer repl
-      (cider-resolve--get-in (if (eq cider-repl-type 'cljs)
+      (cider-resolve--get-in (if (or (eq cider-repl-type 'cljs)
+                                     (eq major-mode 'clojurescript-mode))
                                  "cljs.core"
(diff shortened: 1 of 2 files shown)


cider-resolve.el
```diff
@@ -107,7 +107,8 @@ This will be clojure.core or cljs.core depending on the return value of the
 function `cider-repl-type'."
   (when-let* ((repl (cider-current-repl)))
     (with-current-buffer repl
-      (cider-resolve--get-in (if (eq cider-repl-type 'cljs)
+      (cider-resolve--get-in (if (or (eq cider-repl-type 'cljs)
+                                     (derived-mode-p 'clojurescript-mode))
```
bbatsov, line 111: Looking at the code this shouldn't be needed - notice the use of `with-current-buffer`, which means that will always come from the REPL and you can never really have `clojurescript-mode` there.

vemv: Good eye. Then it should be a matter of simply updating cider-nrepl.

bbatsov: Yep.

vemv:
I'll just move the changelog entry to https://github.com/clojure-emacs/cider/pull/3379

Issue #3383, opened 2023-07-27
Don't pass cljs-repl-type to jvm repls


vemv:
When running `cider-jack-in-clj&cljs` / `cider-connect-clj&cljs`, a cljs-repl-type will be passed to both jvm and cljs repls.

That results in the following code being run for the jvm repl:

```
(when cljs-repl-type
          (emit-comment (concat "ClojureScript REPL type: " (symbol-name cljs-repl-type))))
```

...which is noisy/confusing.

PR #3389, opened 2023-08-01
`cider-connect-clj&cljs`: don't render `"ClojureScript REPL type:" for JVM repls

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3383

I verified this one locally.

Cheers - V

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1453,8 +1453,9 @@ non-nil, don't start if ClojureScript requirements are not met."
                        copy-sequence
                        (map-delete :cljs-repl-type)))
          (clj-repl (cider-connect-clj clj-params)))
-    (when (or (not soft-cljs-start)
-              (cider--check-cljs (plist-get params :cljs-repl-type) 'no-error))
+    (when (if soft-cljs-start
+              (cider--check-cljs (plist-get params :cljs-repl-type) 'no-error)
+            t)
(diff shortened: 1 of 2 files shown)


cider.el
```diff
@@ -1448,10 +1448,13 @@ non-nil, don't start if ClojureScript requirements are not met."
...
-         (clj-repl (cider-connect-clj params)))
-    (if soft-cljs-start
-        (when (cider--check-cljs (plist-get params :cljs-repl-type) 'no-error)
-          (cider-connect-sibling-cljs params clj-repl))
+         (clj-params (thread-first
+                       params
+                       copy-sequence
+                       (map-delete :cljs-repl-type)))
+         (clj-repl (cider-connect-clj clj-params)))
+    (when (or (not soft-cljs-start)
```
vemv, line 1456: Also simplified this condition, since `(cider-connect-sibling-cljs params clj-repl)` was duplicated.

bbatsov: Not a big fan of this negation in the condition, though.

vemv: Maybe?

```lisp
(if soft-cljs-start
    (cider--check-cljs (plist-get params :cljs-repl-type) 'no-error)
  t)
```

bbatsov: Much better IMO.

bbatsov (approved):

Issue #3390, opened 2023-08-01
When using cider-connect with localhost nrepls running, those ports are not shown among the available ones


ag91:
## Expected behavior

For a running nrepl on localhost, I expected `(cider-select-endpoint)` to allow me to select ports on which nrepl is running.
(I use this to connect ob-clojure buffers to cider when doing literate programming and a bit of completion comes handy)

Fixes https://github.com/clojure-emacs/cider/issues/3390

## Actual behavior

It requires me to input the port manually.

## Steps to reproduce the problem

1. cider-jack-in without a project
2. cider-connect from another clj buffer
3. follow instructions to get to the cider session you opened before
4. see that there are no available ports

or more specifically:

2. eval in emacs `(cider--running-nrepl-paths)`
5. see an empty output sigh!

## My solution

Just for curiosity, I redefined `cider--running-nrepl-paths` (which seems to be the culprit) as follows:

``` elisp
(defun cider--running-nrepl-paths ()
  "Retrieve project paths of running nREPL servers.
Use `cider-ps-running-nrepls-command' and
`cider-ps-running-nrepl-path-regexp-list'."
  (seq-uniq
   (seq-map
    (lambda (b)
      (with-current-buffer b default-directory))
    (seq-filter
     (lambda (b) (string-prefix-p "*cider-repl" (buffer-name b)))
     (buffer-list)))))
```
That way things seem to work fine (for localhost).
I opened an issue because I think `cider-ps-running-nrepls-command` and `cider-ps-running-nrepl-path-regexp-list` are somewhat outdated. I replaced the behavior going via Emacs active buffers.
Any chance you can clarify why cider is looking for active nrepl via a terminal command (greping for leiningen -- I didn't find anything using that command manually)?

If this solution seems useful, I could look into opening a PR :)
 
### CIDER version information

1.8.0-snapshot

### Lein / Clojure CLI version

lein 2.9.1 / 1.10.3

### Emacs version

28.1

### Operating system

Ubuntu 22.04

vemv:
Hi!

Thanks for the detailed issue.

> Any chance you can clarify why cider is looking for active nrepl via a terminal command (greping for leiningen -- I didn't find anything using that command manually)?

This very likely aims to make it possible to connect to repls that were created outside of CIDER, in the terminal, especially if their pwd is not the `current-directory` of any Emacs buffer.

[...]

PR #3392, opened 2023-08-03
Show available nrepl ports when connecting

ag91:
Improve cider-connect to show all available ports of running nREPLs.
Fixes https://github.com/clojure-emacs/cider/issues/3390
See recording for a peek :)

https://github.com/clojure-emacs/cider/assets/6580039/02105147-a71e-43f5-a7d0-ddfe53ad7c67

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - [#3236](https://github.com/clojure-emacs/cider/issues/3236): `cider-repl-set-ns` no longer changes the repl session type from `cljs:shadow` to `clj`.
 - [#3383](https://github.com/clojure-emacs/cider/issues/3383): `cider-connect-clj&cljs`: don't render `"ClojureScript REPL type:" for JVM repls. 
 - [#3331](https://github.com/clojure-emacs/cider/issues/3331): `cider-eval`: never jump to spurious locations, as sometimes conveyed by nREPL.  
-- [#3390](https://github.com/clojure-emacs/cider/issues/3390) Enhance cider-connect to show all nREPLs available ports, instead of only Leinengen ones.
+- [#3390](https://github.com/clojure-emacs/cider/issues/3390): Enhance cider-connect to show all nREPLs available ports, instead of only Leiningen ones.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - [#3236](https://github.com/clojure-emacs/cider/issues/3236): `cider-repl-set-ns` no longer changes the repl session type from `cljs:shadow` to `clj`.
 - [#3383](https://github.com/clojure-emacs/cider/issues/3383): `cider-connect-clj&cljs`: don't render `"ClojureScript REPL type:" for JVM repls. 
 - [#3331](https://github.com/clojure-emacs/cider/issues/3331): `cider-eval`: never jump to spurious locations, as sometimes conveyed by nREPL.  
+- [#3390](https://github.com/clojure-emacs/cider/issues/3390) Enhance cider-connect to show all nREPLs available ports, instead of only Leinengen ones.
 
 ### Changes
 
--- a/cider.el
+++ b/cider.el
@@ -1826,7 +1826,7 @@ Use `cider-ps-running-nrepls-command' and
 (defun cider--running-other-nrepl-paths ()
   "Retrieve project paths of running nREPL servers other than Lein ones."
   (let* ((take-non-lein-nrepl-pids
-          "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | tr '\012' ,")
+          "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | paste -s -d,")
          (take-external-paths
           (concat
            "lsof -a -d cwd -p $("
--- a/cider.el
+++ b/cider.el
@@ -1837,13 +1837,14 @@ Use `cider-ps-running-nrepls-command' and
 (defun cider--running-local-nrepl-paths ()
   "Retrieve project paths of running nREPL servers.
 Do it by looping over the open REPL buffers."
-  (seq-map
-   (lambda (b)
-     (with-current-buffer b
-       (plist-get (cider--gather-connect-params) :project-dir)))
-   (seq-filter
-    (lambda (b) (string-prefix-p "*cider-repl" (buffer-name b)))
-    (buffer-list))))
+  (thread-last
+    (buffer-list)
+    (seq-filter
+     (lambda (b) (string-prefix-p "*cider-repl" (buffer-name b))))
+    (seq-map
+     (lambda (b)
+       (with-current-buffer b
+         (plist-get (cider--gather-connect-params) :project-dir))))))
 
 (defun cider--running-nrepl-paths ()
   "Retrieve project paths of running nREPL servers.
--- a/cider.el
+++ b/cider.el
@@ -1823,6 +1823,17 @@ Use `cider-ps-running-nrepls-command' and
           (setq paths (cons (match-string 1) paths)))))
     paths))
 
+(defun cider--running-other-nrepl-paths ()
+  "Retrieve project paths of other running nREPL servers."
+  (let* ((take-non-lein-nrepl-pids
+          "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | tr '\012' ,")
+         (take-external-paths
+          (concat
+           "lsof -a -d cwd -p $("
+           take-non-lein-nrepl-pids
+           ") -n -Fn | awk '/^n/ {print substr($0,2)}'")) )
+    (string-split (shell-command-to-string take-external-paths))))
+
 (defun cider--running-local-nrepl-paths ()
   "Retrieve project paths of running nREPL servers.
 Do it by looping over the open REPL buffers."
@@ -1839,7 +1850,8 @@ Do it by looping over the open REPL buffers."
 Search for lein or java processes including nrepl.command nREPL"
   (let ((paths (append
                 (cider--running-lein-nrepl-paths)
-                (cider--running-local-nrepl-paths))))
+                (cider--running-local-nrepl-paths)
+                (cider--running-other-nrepl-paths))))
     (seq-uniq paths)))
 
 (defun cider--identify-buildtools-present (&optional project-dir)
--- a/cider.el
+++ b/cider.el
@@ -1824,7 +1824,7 @@ Use `cider-ps-running-nrepls-command' and
     paths))
 
 (defun cider--running-other-nrepl-paths ()
-  "Retrieve project paths of other running nREPL servers."
+  "Retrieve project paths of running nREPL servers other than Lein ones."
   (let* ((take-non-lein-nrepl-pids
           "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | tr '\012' ,")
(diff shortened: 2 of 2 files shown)


ag91:
@vemv I seem to have managed to make things work via `cider--running-nrepl-paths` because `nrepl-extract-ports` seemed pretty good at finding the correct port.
I could have done through lsof, but things seemed to work fine.
The shell command for finding external nREPL came out a bit long (see https://github.com/clojure-emacs/cider/pull/3392/commits/4da151c5e1b4ec593fca6522c1666323fe0e14f9):

I had to select the right process id via ps + grep + awk and then feed all the ids separated by comma to lsof to extract the current working directory of the processes. It is a mouthful to write down and a pretty long shell command: I would be grateful for an opinion about how to make it more readable :)

cider.el
```diff
@@ -1821,6 +1821,37 @@ Use `cider-ps-running-nrepls-command' and
         (goto-char 1)
         (while (re-search-forward regexp nil t)
           (setq paths (cons (match-string 1) paths)))))
+    paths))
+
+(defun cider--running-other-nrepl-paths ()
+  "Retrieve project paths of other running nREPL servers."
```
vemv, line 1827:
```suggestion
  "Retrieve project paths of running nREPL servers other than Lein ones."
```

cider.el
```diff
@@ -1821,6 +1821,37 @@ Use `cider-ps-running-nrepls-command' and
...
+  (let* ((take-non-lein-nrepl-pids
+          "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | tr '\012' ,")
```
vemv, line 1829: what does `| tr '\012'` accomplish?

* This works on my macos machine `ps u | grep java | grep -v leiningen | awk '{print $2}'`
* adding ` | tr '\012'` doesn't:

```
$ ps u | grep java | grep -v leiningen | awk '{print $2}'  | tr '\012'
usage: tr [-Ccsu] string1 string2
       tr [-Ccu] -d string1
       tr [-Ccu] -s string1
       tr [-Ccu] -ds string1 string2
```

cider.el
```diff
@@ -1821,6 +1821,37 @@ Use `cider-ps-running-nrepls-command' and
...
+(defun cider--running-local-nrepl-paths ()
+  "Retrieve project paths of running nREPL servers.
+Do it by looping over the open REPL buffers."
+  (seq-map
+   (lambda (b)
+     (with-current-buffer b
+       (plist-get (cider--gather-connect-params) :project-dir)))
+   (seq-filter
+    (lambda (b) (string-prefix-p "*cider-repl" (buffer-name b)))
+    (buffer-list))))
```
vemv, line 1846: You can use thread-last here

cider.el
```diff
@@ -1821,6 +1821,37 @@ Use `cider-ps-running-nrepls-command' and
...
+           ") -n -Fn | awk '/^n/ {print substr($0,2)}'")) )
+    (string-split (shell-command-to-string take-external-paths))))
+
+(defun cider--running-local-nrepl-paths ()
+  "Retrieve project paths of running nREPL servers.
+Do it by looping over the open REPL buffers."
+  (seq-map
+   (lambda (b)
+     (with-current-buffer b
+       (plist-get (cider--gather-connect-params) :project-dir)))
```
vemv, line 1843: Maybe guard for nil return values, just in case: `(thread-last (buffer-list) (seq-filter ,,,) (seq-map ,,,) (seq-filter #'identity))`

cider.el
```diff
@@ -1821,6 +1821,37 @@ Use `cider-ps-running-nrepls-command' and
...
+   (seq-filter
+    (lambda (b) (string-prefix-p "*cider-repl" (buffer-name b)))
+    (buffer-list))))
+
+(defun cider--running-nrepl-paths ()
```
vemv, line 1848: `nrepl-extract-ports` will be called over each element of `cider--running-nrepl-paths`.

`defun nrepl-extract-ports` is based on `(expand-file-name "repl-port" dir)`. That won't work over all possible invocations. It also represents redundant work, to some degree.

I'd suggest that when possible, return a `(list dir port)` instead of just a `dir`.

So that we can make this change:

```diff
(defun cider-locate-running-nrepl-ports (&optional dir)
  "Locate ports of running nREPL servers.
When DIR is non-nil also look for nREPL port files in DIR.  Return a list
of list of the form (project-dir port)."
  (let* ((paths (cider--running-nrepl-paths))
         (proj-ports (apply #'append
                            (mapcar (lambda (d)
+                                     (if (listp d)
+                                         d ;; (directory port) pair was passed, nothing to do
+                                       ;; directory was passed, use old behavior:
                                        (mapcar (lambda (p)
                                                  (list (file-name-nondirectory (directory-file-name d)) p))
                                                (and d (nrepl-extract-ports (cider--file-path d))))))
                                    (cons (clojure-project-dir dir) paths)))))
    (seq-uniq (delq nil proj-ports))))
```

vemv (commented):
Looks like a great start to me!

Added some suggestions.

> I would be grateful for an opinion about how to make it more readable

It looks reasonable to me, perhaps I'm not a fan of awk but I don't have have an immediate suggestion. Can always be refined later.

cider.el
ag91: `tr '\012' ,` is replacing newlines with commas
in linux, running this in the terminal

``` sh
 echo "hi                    
hello" | tr '\012' ,
```
produces this

```
hi,hello,
```
It seems that `lsof` can take a comma separated list of process ids in the -p option, which was nice for our use case.
Maybe on MacOS the character for new line is not parsed as in Linux, I am substituting that with `paste -s -d,`, which seems also less cryptic (that changes both newlines or spaces in commas)

cider.el
ag91: uhm. I see the optimization here. A note though:  we have to return a `(list d)` and we still have to apply `(file-name-nondirectory (directory-file-name d))`, to get just the directory name.
This also makes the seq-uniq needing a predicate for handling equality between paths and pairs (path-port).
So it feels we are trading too much on readability for what the optimization gives to us, no?

To get more, I could make also the other nREPL search return ports and I indeed had a thought while implementing that if I use lsof, I could get ports as well but it seemed less readable to further complicate the shell command.
Should this maybe be tackled in another issue to avoid the added code complexity for now?

ag91: (here the patch with the change if you want to see how the code would look like
[opt.txt](https://github.com/clojure-emacs/cider/files/12261516/opt.txt)
)

cider.el
```diff
@@ -1821,6 +1821,39 @@ Use `cider-ps-running-nrepls-command' and
         (goto-char 1)
         (while (re-search-forward regexp nil t)
           (setq paths (cons (match-string 1) paths)))))
+    paths))
+
+(defun cider--running-other-nrepl-paths ()
+  "Retrieve project paths of running nREPL servers other than Lein ones."
+  (let* ((take-non-lein-nrepl-pids
+          "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | paste -s -d,")
```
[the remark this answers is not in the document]
ag91: uhm I made the redirection explicit following this https://www.reddit.com/r/linuxquestions/comments/y96jbk/why_cant_i_pipe_awk_into_paste_macos/?rdt=44305

sorry for the multiple attempts

cider.el
```diff
@@ -1810,17 +1810,56 @@ of list of the form (project-dir port)."
                                     (cons (clojure-project-dir dir) paths)))))
     (seq-uniq (delq nil proj-ports))))
 
-(defun cider--running-nrepl-paths ()
-  "Retrieve project paths of running nREPL servers.
+(defun cider--windows-p ()
```
bbatsov, line 1813: Do we even need this helper? I mean - nothing modern is going to return `ms-dos` and I think `cygwin` provides the common Unix commands.

cider.el
```diff
@@ -1810,17 +1810,56 @@ of list of the form (project-dir port)."
...
+    (let (paths)
+      (with-temp-buffer
+        (insert (shell-command-to-string cider-ps-running-nrepls-command))
+        (dolist (regexp cider-ps-running-nrepl-path-regexp-list)
+          (goto-char 1)
+          (while (re-search-forward regexp nil t)
+            (setq paths (cons (match-string 1) paths)))))
+      paths)))
+
+(defun cider--running-other-nrepl-paths ()
```
bbatsov, line 1831: I think "other" in this name might be a big confusing, as you won't understand without the docstring that it means "other than Lein". Perhaps "non-lein"?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - [#3236](https://github.com/clojure-emacs/cider/issues/3236): `cider-repl-set-ns` no longer changes the repl session type from `cljs:shadow` to `clj`.
 - [#3383](https://github.com/clojure-emacs/cider/issues/3383): `cider-connect-clj&cljs`: don't render `"ClojureScript REPL type:" for JVM repls. 
 - [#3331](https://github.com/clojure-emacs/cider/issues/3331): `cider-eval`: never jump to spurious locations, as sometimes conveyed by nREPL.  
+- [#3390](https://github.com/clojure-emacs/cider/issues/3390) Enhance cider-connect to show all nREPLs available ports, instead of only Leinengen ones.
```
bbatsov, line 24: Don't forget the `:` here. Also - Lein**i**ngen. :-)

cider.el
ag91: I am not sure myself, I wanted to add the guard because we didn't test there and there may be chance the functions don't work there. I can definitely remove ms-dos, I added it only for completeness.

cider.el
```diff
@@ -1810,17 +1810,56 @@ of list of the form (project-dir port)."
...
+          (goto-char 1)
+          (while (re-search-forward regexp nil t)
+            (setq paths (cons (match-string 1) paths)))))
+      paths)))
+
+(defun cider--running-non-lein-nrepl-paths ()
+  "Retrieve project paths of running nREPL servers other than Lein ones."
+  (unless (cider--windows-p)
+    (let* ((take-non-lein-nrepl-pids
+            "ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | paste -s -d, -")
```
vemv, line 1835: I verified that `lsof -a -d cwd -p $(ps u | grep java | grep -v leiningen | grep nrepl.cmdline | awk '{print $2}' | paste -s -d, -) -n -Fn | awk '/^n/ {print substr($0,2)}'` can run OK on macOS - it can return either nothing or ports.

However, your code should handle the case where nothing is returned. Else the `lsof` call will fail.

For that, I'd suggest running `take-non-lein-nrepl-pids` as an individual shell-command-to-string-call. Abort the defun if it returned nothing / empty string. Else, it's safe to proceed.

Under this approach, `$()` is no longer needed.

vemv (commented):
One last detail, I believe 🙌

vemv:
Self-reviewed and green, merging

vemv:
(I wrote something intended for another PR, deleted. Please ignore)

vemv:
I'll reopen this one from scratch, preserving attribution.

* After more QAing, I noticed some code fails to handle more than one available port
* I'm finding it's simpler / more portable to do things in Elisp (vs awk, etc)

I would not want to create a frustrating "feedback loop" - we all know what it's to be there!

Thanks - V

ag91:
> I would not want to create a frustrating "feedback loop" - we all know what it's to be there!

it was alright, I suspect it would have made sense for me to just add the local dir paths and leave the other bit for a second PR. 
Although I hope my work helped threading ideas together to get to your final solution ;)

PR #3395, opened 2023-08-09
Use cider-nrepl 0.35.0

vemv:
https://github.com/clojure-emacs/cider-nrepl/blob/v0.35.0/CHANGELOG.md#0350-2023-08-09 https://clojars.org/cider/cider-nrepl/versions/0.35.0


vemv:
Self-reviewed and green, merging

Issue #3393, opened 2023-08-05
ClojureScript dynamic font lock not working for new functions (`var`)


nikolavojicic:
First go through this issue https://github.com/clojure-emacs/cider/issues/3385 and use cider-nrepl version 0.34.0.

## Expected behavior

User defined functions should be dynamically highlighted.

## Actual behavior

In ClojureScript, new user defined functions are not dynamically highlighted but old ones are (there is no such problem in Clojure).

<img width="212" alt="258242216-97a9316e-a64a-4efd-881e-99b61a3842c4" src="https://github.com/clojure-emacs/cider/assets/10941031/0d5095e4-77e5-4e62-b801-adc2056b4e63">

## Steps to reproduce the problem

1. `(setq cider-font-lock-dynamically '(macro core function var))` in your init
2. Run cider-jack-in-cljs
3. Choose shadow
4. Choose :app
5. Visit application in browser
6. Write the code from above
7. Evaluate that code

There is no such problem if you write `(setq cider-font-lock-dynamically '(macro core function deprecated))` in your config, i.e. the problem is probably with `var`.

vemv:
Thanks for the new issue, much easier to manage things that way 👍

vemv:
I can repro (note to self: the key to reproing is creating a new function).

A few observations:

* from the nrepl log, I can see that the changed-namespaces key sometimes doesn't come for cljs
  * e.g. for a jvm repl, it always comes after a simple repl interaction (eval op), for cljs, never
  * a way to cause changed-namespaces to be included: change between .cljs namespaces, causing a `in-ns ,,,` eval

It's as if `cider.nrepl.middleware.track-state` sometimes failed to send anything at all, for cljs. Worth noting that it sends stuff async, with an `agent`. Maybe errors are being swallowed, although error handling seems ok.

---

Fixing this should be pretty important beyond the syntax highlighting, because the `track-state` middleware supports all type of functionality.

vemv:
I have the impression that the track-state middleware runs just once for cljs, then never again.

Maybe it has to do with middleware order?

This is shadow's order:

https://github.com/thheller/shadow-cljs/blob/bb586a0fab475c13c6d090164c4122f2782308e3/src/main/shadow/cljs/devtools/server/nrepl.clj#L98-L105

...emphasis on `nrepl.middleware.interruptible-eval/interruptible-eval` being placed after `cider.nrepl/cider-middleware` - it might be usually the opposite?

[...]

PR #3396, opened 2023-08-09
Recompute namespace info on each shadow-cljs recompilation or evaluation

vemv:
Fixes https://github.com/clojure-emacs/cider/issues/3393

Works thanks to https://github.com/clojure-emacs/cider-nrepl/pull/799

I tested this one locally.

The defcustoms are bit of a POC - feedback welcome. Initially I directly inlined the code, later I figured that it would pollute the main logic.

Cheers - V

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1007,34 +1007,22 @@ This is particularly necessary for shadow-cljs because:
         (when (nrepl-op-supported-p "cider/get-state" conn)
           (nrepl-send-request '("op" "cider/get-state") nil conn))))))
 
-(defun cider--shadow-cljs-stderr-hook (buffer err)
+(defun cider--shadow-cljs-handle-stderr (buffer err)
   "Refresh the changed namespaces metadata given BUFFER and ERR."
   (cider--maybe-get-state-for-shadow-cljs buffer err))
 
-(defun cider--shadow-cljs-done-hook (buffer)
+(defun cider--shadow-cljs-handle-done (buffer)
   "Refresh the changed namespaces metadata given BUFFER."
   (cider--maybe-get-state-for-shadow-cljs buffer))
 
-(defcustom cider-repl-stdout-hooks nil
-  "Hooks to be invoked each time new stdout is received on a repl buffer.
+(defvar cider--repl-stderr-functions (list #'cider--shadow-cljs-handle-stderr)
+  "Functions to be invoked each time new stderr is received on a repl buffer.
 
 Good for, for instance, monitoring specific strings that may be logged,
-and responding to them."
-  :type '(repeat function)
-  :package-version '(cider . "1.8.0"))
+and responding to them.")
 
-(defcustom cider-repl-stderr-hooks (list #'cider--shadow-cljs-stderr-hook)
-  "Hooks to be invoked each time new stderr is received on a repl buffer.
-
-Good for, for instance, monitoring specific strings that may be logged,
-and responding to them."
-  :type '(repeat function)
-  :package-version '(cider . "1.8.0"))
-
-(defcustom cider-repl-done-hooks (list #'cider--shadow-cljs-done-hook)
-  "Hooks to be invoked each time a given REPL interaction is complete."
-  :type '(repeat function)
-  :package-version '(cider . "1.8.0"))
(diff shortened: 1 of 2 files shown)


bbatsov:
Generally, it's good to mention all hooks in the user docs as well.

cider-repl.el
```diff
@@ -986,6 +986,56 @@ t, as the content-type response is (currently) an alternative to the
...
+               (eq cider-cljs-repl-type 'shadow)
+               (not cider-repl-cljs-upgrade-pending)
+               (if err
+                   (string-match-p "Build completed\\." err)
+                 t))
+      (when-let ((conn (cider-current-repl 'cljs)))
+        (when (nrepl-op-supported-p "cider/get-state" conn)
+          (nrepl-send-request '("op" "cider/get-state") nil conn))))))
+
+(defun cider--shadow-cljs-stderr-hook (buffer err)
```
bbatsov, line 1010: The list of functions to call is called a "hook", not the functions in it.

cider-repl.el
```diff
@@ -986,6 +986,56 @@ t, as the content-type response is (currently) an alternative to the
...
+  "Refresh the changed namespaces metadata given BUFFER and ERR."
+  (cider--maybe-get-state-for-shadow-cljs buffer err))
+
+(defun cider--shadow-cljs-done-hook (buffer)
+  "Refresh the changed namespaces metadata given BUFFER."
+  (cider--maybe-get-state-for-shadow-cljs buffer))
+
+(defcustom cider-repl-stdout-hooks nil
```
bbatsov, line 1018: hooks -> hook

Refer to other hooks in the codebase.

cider-repl.el
```diff
@@ -994,14 +1044,20 @@ nREPL ops, it may be convenient to prevent inserting a prompt.")
      (lambda (buffer value)
        (cider-repl-emit-result buffer value t))
      (lambda (buffer out)
+       (dolist (f cider-repl-stdout-hooks)
```
bbatsov, line 1047: You should see if you can use `run-hooks` here. I don't recall how it played for hooks with arguments.

vemv: `run-hooks` docstring says that no arguments will be passed.

In face of that, maybe we can refrain from using the `hook` terminology here?

e.g. `defcustom cider-repl-stdout-callbacks`/ `defcustom cider-repl-stdout-functions`.

(`callbacks` would seem more expressive to me here)

bbatsov: It depends on whether you expect users to actually do something here or not - if you do I'd stick with `hooks`, just because it's the established term in the community. If not you can just makes so `defvars` (perhaps private) and then the names won't matter as much.

vemv: I'd be wary of creating 'hooks' that couldn't be run with `run-hooks`

Defvars make sense :) we can always promote them later.

vemv:
Ready, QAed again as well.

cider-repl.el
```diff
@@ -996,12 +1034,16 @@ nREPL ops, it may be convenient to prevent inserting a prompt.")
      (lambda (buffer out)
        (cider-repl-emit-stdout buffer out))
      (lambda (buffer err)
+       (dolist (f cider--repl-stderr-functions)
```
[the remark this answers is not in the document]
vemv: Note that the guards are carefully ordered:

```lisp
    (when (and (eq cider-repl-type 'cljs)
               (eq cider-cljs-repl-type 'shadow)
               (not cider-repl-cljs-upgrade-pending)
               (if err
                   (string-match-p "Build completed\\." err)
                 t))
```

no matter how large the output we're receiving is, for most cases we're not even inspecting it.

WIth that said, yesterday I created https://github.com/thheller/shadow-cljs/issues/1139 which is going in a good direction. A few features could be built on top of it, and then this could be discarded.

bbatsov:
Looks good. Just rebase and squash before merging.

cider-repl.el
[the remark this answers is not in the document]
bbatsov: Yeah, yeah - I'm aware of this. Just a point of concern, as we have to be careful about this. Hopefully we'll come up with a better solution down the road.

PR #3397, opened 2023-08-09
xref fixes and improvements

vemv:
Commits:

* Fix the CIDER `xref-find-references` backend to return correct filenames
  * Fixes https://github.com/clojure-emacs/cider/issues/3112
* Fix the `cider-xref-fn-deps` buttons to direct to the right file
  * This fixed the (unrepored) issue of clicking e.g. `clojure/core.clj` and being directed to an unrelate file like `refactor_nrepl/core.clj`.
* Fix the `xref-find-definitions` CIDER backend to return correct filenames
* Improve the presentation of `xref` data
  * Shorten .jar filenames
  * render the cider-xref-fn-deps UI better

It's all tested out locally.

Cheers - V


cider-find.el
```diff
@@ -275,14 +301,42 @@ thing at point."
...
+                                  (column (nrepl-dict-get info "column"))
+                                  (line (nrepl-dict-get info "line"))
+                                  (friendly-name (cider--xref-extract-friendly-file-name info))
+                                  ;; translate .jar urls and such:
+                                  (buf (cider--find-buffer-for-file filename))
+                                  (bfn (and buf (buffer-file-name buf)))
+                                  (loc (when buf
+                                         ;; favor `xref-make-file-location' when possible, since that way, we can close their buffers.
+                                         (if bfn
+                                             (xref-make-file-location bfn line (or column 0))
```
vemv, line 318: From what I could see, one can basically make either an xref location based on a filename (which is the best choice, as it doesn't "leak" buffers by opening them) or in a buffer (which is the secondary choice, needed for "special" buffers like those for .jar files).

Checking out xref's source code, it looks as if locations were some sort of implementable/inheritable thing.

So one could create a new xref location type that was lazy, e.g. was a `"jar:file:..."` url that was only open if the user clicked on it.

Could be done some other day. xref.el seemed a bit dense, and I'm not fluent in `cl-lib`.

cider-find.el
```diff
@@ -28,8 +28,10 @@
 
 (require 'cider-client)
 (require 'cider-common)
+(require 'cider-doc)
```
bbatsov, line 31: Why is this needed here?

cider-find.el
```diff
@@ -241,11 +243,35 @@ thing at point."
   "Return the relevant identifier at point."
   (cider-symbol-at-point 'look-back))
 
+(defun cider--xref-extract-file (dict)
```
bbatsov, line 246: I wonder if all the `xref` related stuff belong to `cider-find`, which is more focused on definition lookup. Perhaps the `xref-backend` should go somewhere else?

cider-find.el
```diff
@@ -275,14 +301,42 @@ thing at point."
...
+                                         ;; favor `xref-make-file-location' when possible, since that way, we can close their buffers.
+                                         (if bfn
+                                             (xref-make-file-location bfn line (or column 0))
+                                           (xref-make-buffer-location buf (with-current-buffer buf
+                                                                            (save-excursion
+                                                                              (goto-char 0)
+                                                                              (forward-line line)
+                                                                              (move-to-column (or column 0))
+                                                                              (point)))))))
+                                  (should-be-closed? (and
```
bbatsov, line 325: Names with `?` are not common in Elisp.

cider-find.el
vemv: I agree that the existing layout was a mixed bag, I'll see if I can move it

vemv:
All feedback satisfied in individual commits.

bbatsov:
Looks great now. Thanks!

PR #3399, opened 2023-08-10
Enhance `cider-connect` to show all nREPLs available ports, instead of only Leiningen ones

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3390

Rehash of https://github.com/clojure-emacs/cider/pull/3392 with some improvements directly applied, while preserving attribution:

* Do more processing in Elisp (vs `awk`)
* For each 'stragegy' (lein processes, non-lein-process, local, etc), a list of `<dir, port>` pairs is returned, instead of just `dir`
  * this avoids redundantly re-computing the ports later.
* avoid permission errors, e.g. trying to read .nrepl-port on macOS `/Users/$USER/Desktop`

And a minor CI improvement expressed in a separate commit.

I've QA'd all code paths - it works as intended:

Screen Shot 2023-08-10 at 16 29 05

Cheers - V

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1798,7 +1798,7 @@ of remote SSH hosts."
                       (or (eq ?s filetype)
                           (eq ?d filetype))))))
 
-(defun cider--path->path-ports-pair (path)
+(defun cider--path->path-port-pairs (path)
   "Given PATH, returns all the possible <path, port> pairs."
   (mapcar (lambda (port)
             (list path port))
@@ -1824,7 +1824,7 @@ When DIR is non-nil also look for nREPL port files in DIR.  Return a list
 of list of the form (project-dir port)."
   (let* ((pairs (cider--running-nrepl-paths))
          (pairs (if-let (c (and dir (clojure-project-dir dir)))
-                    (cons (cider--path->path-ports-pair c) pairs)
+                    (cons (cider--path->path-port-pairs c) pairs)
                   pairs)))
     (thread-last pairs
                  (delq nil)
@@ -1846,7 +1846,7 @@ Use `cider-ps-running-lein-nrepls-command' and
           (while (re-search-forward regexp nil t)
             (setq paths (cons (match-string 1) paths)))))
       (seq-mapcat (lambda (path)
-                    (cider--path->path-ports-pair path))
+                    (cider--path->path-port-pairs path))
                   paths))))
 
 (defun cider--running-non-lein-nrepl-paths ()
--- a/cider.el
+++ b/cider.el
@@ -1895,7 +1895,7 @@ Do it by looping over the open REPL buffers."
 
 (defun cider--running-nrepl-paths ()
   "Retrieve project paths of running nREPL servers.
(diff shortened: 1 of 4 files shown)


vemv:
@ag91 feel free to try this one locally.

`curl https://patch-diff.githubusercontent.com/raw/clojure-emacs/cider/pull/3399.diff | git apply` may be handy

nrepl-client.el
```diff
@@ -232,18 +232,24 @@ PARAMS is as in `nrepl-make-buffer-name'."
 
 (defun nrepl-extract-port (dir)
   "Read port from applicable repl-port file in directory DIR."
-  (or (nrepl--port-from-file (expand-file-name "repl-port" dir))
-      (nrepl--port-from-file (expand-file-name ".nrepl-port" dir))
-      (nrepl--port-from-file (expand-file-name "target/repl-port" dir))
-      (nrepl--port-from-file (expand-file-name ".shadow-cljs/nrepl.port" dir))))
+  (condition-case nil
```
ag91, line 235: I was thinking to tackle this with `with-demoted-errors` (although also `ignore-errors` would do here if you don't want to let the user know about what caused the ignored error): maybe it saves some code?

vemv: Thanks! Sounds reasonable.

Now that I think it, we may want to catch errors for each individual directory entry, instead of for the whole call, so that one bad dir doesn't cause all its neighbors to be excluded.

cider.el
```diff
@@ -1798,30 +1798,107 @@ of remote SSH hosts."
                       (or (eq ?s filetype)
                           (eq ?d filetype))))))
 
+(defun cider--path->path-ports-pair (path)
```
bbatsov, line 1801: path-port-pairs maybe? (looking at what it actually does)

cider.el
```diff
@@ -1798,30 +1798,107 @@ of remote SSH hosts."
...
-`cider-ps-running-nrepl-path-regexp-list'."
-  (let (paths)
-    (with-temp-buffer
-      (insert (shell-command-to-string cider-ps-running-nrepls-command))
-      (dolist (regexp cider-ps-running-nrepl-path-regexp-list)
-        (goto-char 1)
-        (while (re-search-forward regexp nil t)
-          (setq paths (cons (match-string 1) paths)))))
-    (seq-uniq paths)))
+Search for lein or java processes including nrepl.command nREPL"
```
bbatsov, line 1898: You forgot the final `.` here.

cider.el
```diff
@@ -1798,30 +1798,107 @@ of remote SSH hosts."
...
+  (append (cider--invoke-running-nrepl-path #'cider--running-lein-nrepl-paths)
```
bbatsov, line 1899: We don't expect duplicates here?

vemv: > We don't expect duplicates here?

They're handled in the `cider--running-nrepl-paths` callsite. Which seems better since `cider-locate-running-nrepl-ports` can `cons` an element to it, so you'd have to run seq-uniq _again_

Issue #3402, opened 2023-08-12
[Emacs 29.1] Error when starting Cider (wrong-number-of-arguments (1 . 1) 0)


cordelia04m:
## Expected behavior
Cider should start without problem.

## Actual behavior
Cider doesn't start. I got the following backtrack:
```
Debugger entered--Lisp error: (wrong-number-of-arguments (1 . 1) 0)
  clj()
  format-spec("*cider-repl %s(%r:%S)*" ((115 . "ClojureProjects/PPL:localhost:36453") (104 . "localhost") (72 . "") (112 . 36453) (106 . "PPL") (74 . "ClojureProjects/PPL") (114 . clj) (83 . "")))
  (replace-regexp-in-string "\\([:-]\\)[:-]+" "\\1" (format-spec template specs))
  (replace-regexp-in-string "\\(^[:-]\\)\\|\\([:-]$\\)" "" (replace-regexp-in-string "\\([:-]\\)[:-]+" "\\1" (format-spec template specs)))
  (replace-regexp-in-string "[:-]\\([])*]\\)" "\\1" (replace-regexp-in-string "\\(^[:-]\\)\\|\\([:-]$\\)" "" (replace-regexp-in-string "\\([:-]\\)[:-]+" "\\1" (format-spec template specs))))
[...]

PR #3403, opened 2023-08-12
Clean up test suite

vemv:
Commits:

* Remove unused `provide`s in tests
  * These can be confusing, since Buttercup doesn't use them
* Use `it` more
  * This ensures their execution is displayed on CI logs.
* Settle on `*tests.el` naming
* Add a standard header suggesting `it` blocks
* Bump some copyright years
* Run test suite on Emacs 29
  * We already were (via `master`), but now we will run 29 stable and 29 master (which will become 30 at some point, I guess)
* Cover `cider-format-connection-params`
  * Part of https://github.com/clojure-emacs/cider/issues/3402
* Fix `cider-format-connection-params` edge case for Emacs 29
  * Closes https://github.com/clojure-emacs/cider/issues/3402

Cheers - V


vemv:
Finally I added a fix for https://github.com/clojure-emacs/cider/issues/3402 . It's a bit unrelated to this PR's theme, but small enough to let it slip (hopefully)

cider-connection.el
```diff
@@ -720,6 +720,19 @@ Fallback on `cider' command."
...
+  "Ensures SPEC cannot be invoked as a function.
+
+Invokeable specs are an Emacs 29 feature
+that we don't intend to use in this context."
+  (let ((spec-char (car spec))
+        (spec-value (cdr spec)))
+    `(,spec-char
+      .
+      ,(if (symbolp spec-value)
+           (prin1-to-string spec-value)
```
vemv, line 733: Avoids this code path https://github.com/emacs-mirror/emacs/blame/d3dae88e6cc8118c875957ba0347be9599014b34/lisp/format-spec.el#L110

e.g. if the user had a function named `clj`, we change it from symbol to string, so that it's not `functionp` anymore.

I was able to reproduce this scenario on CI, verifying that the tests would fail if we remove this fix.

bbatsov:
Nice improvements!

PR #3404, opened 2023-08-12
Recompute namespace info on each fighweel-main recompilation

vemv:
* Bump the injected Piggieback to 0.5.3
  * We never did after releasing it quite some time ago.
  * I've tested it locally - working OK.
* Recompute namespace info on each fighweel-main recompilation
  * Similar to https://github.com/clojure-emacs/cider/pull/3396, except that it's only necessary on recompilation - not on evaluations.
  * QAd locally, works as intended.

Cheers - V


bbatsov:
The changes look good to me, but you'll need to address the conflict in the changelog. Might also be good at some point to update a bit https://docs.cider.mx/cider/basics/clojurescript.html to reflect the current state of affairs.

vemv:
Thanks!

The docs look fairly reasonable. There are some changes in master not deployed yet.

Perhaps we can nuke the Boot docs at some point.

...Or maybe hide them under a collapsible widget, if that's possible (over github we have `<details>...</details>`)

bbatsov:
Probably in CIDER 2.0 we should nuke the support for Boot: 
image

It's completely dead at this point.

Issue #3406, opened 2023-08-13
cider-connect-clj fails with wrong-type-argument stringp


ChrisHacker:
## Expected behavior
cider-connect-clj connects to specified localhost:port repl
## Actual behavior
fails with error `wrong-type-argument stringp ("/home/PATH/REDACTED/" "40000")
## Steps to reproduce the problem
While in an open clj buffer, call M-x cider-connect, provide "localhost" as host value, and hit enter

### CIDER version information

cider doesn't launch - from package-manager 20230811.1023

### Lein / Clojure CLI version

Leiningen 2.10.0 on Java 17.0.5 OpenJDK 64-Bit Server VM

### Emacs version

27.1

### Operating system

Ubuntu 22.04

### JDK distribution

Java 17.0.5 OpenJDK

```
Debugger entered--Lisp error: (wrong-type-argument stringp ("/home/PATH/REDACTED/" "40000"))                                                                                                              
   directory-file-name(("/home/PATH/REDACTED/" "40000"))                                                                                                                                                   
   #f(compiled-function (x) #<bytecode 0x1563d5a7a4c5>)((("/home/PATH/REDACTED/" "40000") ("/home/PATH/REDACTED/" "40000")))                                                                          
   mapcar(#f(compiled-function (x) #<bytecode 0x1563d5a7a4c5>) ((("/home/PATH/REDACTED/" "40000") ("/home/PATH/REDACTED/" "40000"))))                                                                 
   cider-locate-running-nrepl-ports("/home/PATH/REDACTED/")                                                                                                                                         
   cider--infer-ports(#("localhost" 0 9 (idx 0 ivy-index 0)) (("blits") (#("chris" 0 6 (tramp-default t)))))                                                                                                   
   cider-select-endpoint()                                                                                                                                                                                      
   cider--update-host-port((:project-dir "/home/PATH/REDACTED/"))                                                                                                                                          
   cider-connect-clj(nil)                                                                                                                                                                                       
   funcall-interactively(cider-connect-clj nil)                                                                                                                                                                 
   #<subr call-interactively>(cider-connect-clj record nil)                                                                                                                                                     
   apply(#<subr call-interactively> cider-connect-clj (record nil))                                                                                                                                             
   call-interactively@ido-cr+-record-current-command(#<subr call-interactively> cider-connect-clj record nil)                                                                                                   
[...]

PR #3407, opened 2023-08-13
Fix a bug recently introduced into `cider-locate-running-nrepl-ports`

vemv:
Closes https://github.com/clojure-emacs/cider/issues/3406

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -258,7 +258,8 @@ PARAMS is as in `nrepl-make-buffer-name'."
   (when (file-exists-p file)
     (with-temp-buffer
       (insert-file-contents file)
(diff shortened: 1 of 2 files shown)


nrepl-client.el
```diff
@@ -258,7 +258,8 @@ PARAMS is as in `nrepl-make-buffer-name'."
   (when (file-exists-p file)
     (with-temp-buffer
       (insert-file-contents file)
-      (buffer-string))))
+      (replace-regexp-in-string "\n\\'" "" ;; remove any trailing newline, so that our UIs look better.
```
bbatsov, line 261: Btw, there's also `string-trim-right` for cases like this.

Issue #3408, opened 2023-08-13
`cider-connect` improvements


vemv:
* `nrepl--port-from-file` can return a file that exists but doesn't reflect an actually active port
  * connections to it will fail, so it's better to avoid displaying them to begin with
* Add test coverage for https://github.com/clojure-emacs/cider/pull/3407/commits/b923537da930574a13cdd09a92961ce22aafa0a4

PR #3410, opened 2023-08-14
`cider-connect` improvements

vemv:
* Check `.nrepl-port`-like files for liveness
  * Fixes https://github.com/clojure-emacs/cider/issues/3408
* Describe `cider-locate-running-nrepl-ports`

Cheers - V


nrepl-client.el
```diff
@@ -254,12 +254,18 @@ PARAMS is as in `nrepl-make-buffer-name'."
...
-      (insert-file-contents file)
-      (replace-regexp-in-string "\n\\'" "" ;; remove any trailing newline, so that our UIs look better.
-                                (buffer-string)))))
+    (let ((port-string (with-temp-buffer
+                         (insert-file-contents file)
+                         (string-trim-right (buffer-string)))))
+      (if (eq system-type 'windows-nt)
+          port-string
+        (when (not (equal ""
+                          (shell-command-to-string (concat "lsof -i:" port-string))))
```
iarenaza, line 267: I know the possibility of not having `lsof` installed on a Linux machine is low (maybe it comes by default in OSX, but that is not the case in every Linux distribution). But if it isn't installed, `shell-command-to-string` will return a non-empty string with an error similar to:

```
"/bin/bash: line 1: lsof: command not found
"
```

which will make the above code assume that `port-string` is a "alive", even if we couldn't check that it is.

vemv: Thanks for the review!

I actually considered that.

If there's no `lsof` installed, the only reasonable option we have is to allow the port, instead of excluding it.

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts.

vemv:
Thanks 🍻

nrepl-client.el
yuhan0: Wouldn't splicing arbitrary file contents into a shell command in this manner introduce some sort of vulnerability? 

Not that I have a specific threat model in mind, but it might at least help to sanitize the input (eg. matching against a basic `"^\\d+$"` regexp)

vemv: Let's do that, thanks!

Most of all it would seem a good way to ensure the command being run is reasonable (e.g. not an empty string or non-numeric contents)

Issue #3250, opened 2022-09-21
CIDER session is lost when navigating to definition


mmontone:
Hi,

when navigating to a definition via `xref-find-definition` (M-.), if the target buffer is not part of the Clojure project I'm in, the buffer is 'unlinked', and so I cannot continue navigating or evaluate Clojure code in the new buffer, unless I manually link it with sesman-link-buffer.

So, I've added this to my .emacs, that propagates the current CIDER session to the definition buffer:

```lisp
(defun cider--xref-show-definitions-buffer (fetcher alist)
  "Propagate the CIDER session to the definitions buffer."
  (let ((session (sesman-current-session 'CIDER)))
    (xref-show-definitions-buffer fetcher alist)
    (sesman-link-with-buffer nil session)))

(add-hook 'cider-mode-hook
	  (lambda ()
	    (setq-local xref-show-definitions-function 'cider--xref-show-definitions-buffer)))
```

Not sure if this is the greatest fix, but it does the job for me.

vemv:
I cannot reproduce this although I think I've seen it happen. Also, definitely have heard of it happening to other people (e.g. [today](https://github.com/clojure-emacs/cider/issues/3409#issuecomment-1677431452))

@bbatsov , do you recall what the mechanism is for making sure a Sesman session is 'inherited' from one buffer to a new one (the new one being that `M-.` brought you to)?

I thought it had to do with `default-directory` but it's not necessarily the case, from what I've tried.

vemv:
Ok, this summarizes it https://docs.cider.mx/cider/usage/managing_connections.html#friendly-sessions

A way to debug this is by

* ensuring the current buffer is the problematic one (e.g. place the cursor over the buffer that `M-.` just brought you to)
* running the same code that `cl-defmethod sesman-friendly-session-p` would, e.g.:

```lisp
(mapcar (lambda (session)
          (sesman-friendly-session-p 'CIDER session))
        (sesman--all-system-sessions 'CIDER))
```

The resulting list should include at least one truthy item.

If not, one has to debug why. Get started by inlining the `sesman-friendly-session-p` into the lambda:

```lisp
(mapcar (lambda (session)
          (setcdr session (seq-filter #'buffer-live-p (cdr session)))
          (when-let* ((repl (cadr session))
                      (proc (get-buffer-process repl))
                      (file (file-truename (or (buffer-file-name) default-directory))))
[...]

PR #3412, opened 2023-08-15
TRAMP improvements

vemv:
* Add a Dockerfile for TRAMP development purposes
  * Includes a README.
* Make TRAMP functionality work when using non-standard ports
  * e.g. when we connect from a SSH port other than `22`.
* Don't lose the CIDER session over TRAMP files
  * Makes `M-.` work recursively (instead of just once) when using TRAMP.
  * Fixes https://github.com/clojure-emacs/cider/issues/3250


bbatsov:
The changes look good overall. Do we also need to update/extend https://docs.cider.mx/cider/basics/up_and_running.html#working-with-remote-hosts ?

vemv:
Hey @mmontone @igrishaev ,

`No linked CIDER sessions` should be far more rare to encounter now (as of CIDER master, as released already into MELPA non- `Stable`)

* For vanilla CIDER connections, it generally shouldn't happen
* For TRAMP connections, we fixed the issue
* For Docker-backed connections, we also fixed the issue

It would be much helpful if you could try removing the workaround and checking if it works now.

If it didn't, you now can:

* Try jumping to definition,
* See it fail with `No linked CIDER sessions`
* Then, use the new command `M-x cider-debug-sesman-friendly-session-p`
* Copy its output (as visible in the `*messages*` buffer) and paste it here.

Thanks much - V

PR #3414, opened 2023-08-16
Make jump-to-definition work in projects needing `cider-path-translations`

### LINKED ISSUE #3413: `M-.` - possible Docker issue
A nuisance that I recall from a couple past jobs is that if one `cider-connect`s to a Docker machine, `M-.` would work once, but then not recursively.

I fixed this for tramp [today](https://github.com/clojure-emacs/cider/pull/3412). Maybe Docker needs a similar fix?

vemv:
> Fixes #3413

The gist of it is:

* during friendly session calculation, we translate paths in one direction; and
* during jump-to-definition (i.e. the `cider--file-path`, as invoked from `cider-find-file`), we translate paths in the opposite direction.

Cheers - V


dev/docker-sample-project/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM clojure:temurin-17-lein-bullseye
```
vemv, line 1: (This is a different Dockerfile from that of https://github.com/clojure-emacs/cider/pull/3412)

vemv:
@bbatsov rebased and green now

cider-connection.el
```diff
@@ -644,7 +645,33 @@ REPL defaults to the current REPL."
...
+                                          (string-prefix-p path translated-path))
+                                        classpath)
+                              (seq-find (lambda (path)
+                                          (string-prefix-p path translated-path))
+                                        classpath-roots)))
+                        translated))
+            (when debug
+              (list file "was not determined to belong to classpath:" classpath "or classpath-roots:" classpath-roots)))))))
+
+(defun cider-debug-sesman-friendly-session-p ()
```
bbatsov, line 662: Why is this named as a predicate?

You might want to mention it somewhere in the troubleshooting/sesman docs.

vemv: > Why is this named as a predicate?

Funny you ask, I intended it to read `debug(sesman-friendly-session-p)` i.e. the -p is for the thing being debugged.

Alternative welcome.

Good call on adding it to the manual!

cider-common.el
```diff
@@ -286,22 +286,38 @@ in the container, the alist would be `((\"/src\" \"~/projects/foo/src\"))."
   :group 'cider
   :package-version '(cider . "0.23.0"))
 
-(defun cider--translate-path (path direction)
-  "Attempt to translate the PATH in the given DIRECTION.
+(defun cider--translate-path (path direction &optional return-all)
```
bbatsov, line 289: Can you remind me why we can have multiple mappings for the same path? I never used TRAMP, so I don't remember very well how this was supposed to work.

vemv: Generally we don't expect that to be the case, however I'm using the new `cider--all-path-translations` function (it has a comment justifying its existence) which gets the buffer-local `cider-path-translations` value from _all_ buffers.

I imagine it's possible that some unrelated buffer's cider-path-translations can unexpectedly work for a given path.

So by specifying `return-all`, we get all possible matches - good and bad ones. Later, in the 'friendly session' computation, we compare them against the classpath. The bad ones won't matter - they won't be found in the classpath, so nothing is changed.

bbatsov: Seems overly complicated to me, but given that I don't use or care about this functionality I'll assume you know what you're doing. :-)

vemv: I wish it could be less complicated. 😄

vemv:
Thanks!

PR #3417, opened 2023-08-19
Introduce `cider-stacktrace-navigate-to-other-window` defcustom

vemv:
By default, when you jump to the source of a given stack frame, an Emacs window other than that of `*cider-error*` will be chosen.

`cider-stacktrace-navigate-to-other-window` offers to remain in the same window instead.

### DIFF
--- a/cider-stacktrace.el
+++ b/cider-stacktrace.el
@@ -56,6 +56,12 @@ If nil, messages will not be wrapped.  If truthy but non-numeric,
   :type 'list
   :package-version '(cider . "0.6.0"))
 
+(defcustom cider-stacktrace-navigate-to-other-window t
+  "If truthy, navigating from a stack frame will use other window.
+Pick nil if you prefer the same window as *cider-error*."
+  :type 'boolean
+  :package-version '(cider . "1.8.0"))
+
 (make-obsolete 'cider-stacktrace-print-length 'cider-stacktrace-print-options "0.20")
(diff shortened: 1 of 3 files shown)


cider-stacktrace.el
```diff
@@ -529,6 +529,12 @@ Achieved by destructively manipulating `cider-stacktrace-suppressed-errors'."
         (button-put button 'help-echo "Click to promote these stacktraces."))
       (button-put button 'suppressed (not suppressed)))))
 
+(defcustom cider-stacktrace-navigate-to-other-window t
```
bbatsov, line 532: I'd probably move this next to the other defcustoms.

doc/modules/ROOT/pages/usage/dealing_with_errors.adoc
```diff
@@ -58,6 +58,15 @@ To disable auto-selection of the error buffer when it's displayed:
...
 ----
 
+By default, when you jump to the source of a given stack frame,
+an Emacs window other than that of `*cider-error*` will be chosen.
+If you wish to reuse `*cider-error*`'s window instead, please configure:
+
+[source,lisp]
+----
+(setq cider-stacktrace-navigate-to-other-window nil)
+----
```
bbatsov, line 68: Might be useful to throw some tip to use `C-x arrow` if they pick this setting. (see https://stackoverflow.com/questions/71248133/going-back-to-a-previous-buffer-in-emacs)

I'm assuming that unless people know this they won't understand why this setting makes sense.

vemv:
All done. Thanks!

PR #3420, opened 2023-08-20
Don't possibly pass garbage to lsof

vemv:
Tried locally.

Thanks, @yuhan0

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -77,7 +77,7 @@
 (require 'sesman)
 (require 'tramp)
 
-
+
 ;;; Custom
 
 (defgroup nrepl nil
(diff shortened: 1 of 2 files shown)


nrepl-client.el
```diff
@@ -126,7 +126,6 @@ Setting this to nil disables the timeout functionality."
 When true some special buffers like the server buffer will be hidden."
   :type 'boolean)
 
-
```
vemv, line 129: (these were some odd `^L` characters)

bbatsov: Those are special page breaks that Emacs uses for page navigation. There were there for a reason - to make it easy to jump between meaningful sections of the code. :-) See https://stackoverflow.com/questions/1576887/what-does-l-in-emacs-lisp-source-code-mean

vemv: Ahhh sorry for that, I'm used to see odd whitespace (especially windows-related) from time to time.

Had never heard of this feature.

Issue #3418, opened 2023-08-19
`cider-show-error-buffer`/`cider-auto-jump-to-error`: honor `:clojure.error/phase`


vemv:
> See also: https://github.com/clojure-emacs/cider/discussions/3338

`cider-show-error-buffer` and `cider-auto-jump-to-error` are customizable, but their semantics are a little too coarse-grained: one may not necessarily `'always` or `'never` want them to work.

A commonly desirable thing would be that one is presented detailed error reports on application-level exceptions. Compile-time exceptions would not deserve them, by default.

Therefore, we could make those two defcustoms have default value of `lambda (exception-phase)`, and have the appropiate logic there.

---

**NOTE:** this GH issue is only concerned with not presenting stacktrace buffers when it's not particularly helpful.

Presenting a friendly, non-stacktrace-centric error report (e.g., turn `clojure.core.specs.alpha/even-number-of-forms?` into a pretty string / UI) would be done separately.

PR #3423, opened 2023-08-21
Introduce `cider-ignored-error-phases`

vemv:
Commits:

* Use cider-nrepl 0.36.0
  * Supports this PR, and also [broadens](https://github.com/clojure-emacs/haystack/pull/11) the semantics of the `tooling` filter:

image

* Introduce `cider-ignored-error-phases`
  * Fixes https://github.com/clojure-emacs/cider/issues/3418
* `cider-auto-jump-to-error`: also recognise macroexpansion errors
  * Mostly unrelated, but also very small.

### QA

This PR's main focus is easy to QA:

* `(setq cider-show-error-buffer 'always)`
* enter an invalid token like `::::a` in the REPL
  * no stacktrace will show up, a concise message will be printed at the repl
* enter an invalid macroexpansion like `(let [1])` in the REPL
  * no stacktrace will show up, a concise message will be printed at the repl
* enter an expr that will cause a runtime exception like `(/ 2 0)`
  * the `*cider-error*` buffer will show up
  * if the `tooling` filter is on, notably few stack frames will be visible
  * the usual output will also be visible at the REPL output.

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -70,7 +70,15 @@ not to be automatically shown.
 
 Irrespective of the value of this variable, the `cider-error-buffer' is
 always generated in the background.  Use `cider-selector' to
-navigate to this buffer."
+navigate to this buffer.
+
+Please note, if the error phase belongs to
+one of the `cider-clojure-compilation-error-phases',
+then no stacktrace showing will happen.
+That defcustom takes precedence over this one.
+
+See its doc for understanding its rationale.  You can also customize it to nil
+in order to void its effect."
   :type '(choice (const :tag "always" t)
                  (const except-in-repl)
                  (const only-in-repl)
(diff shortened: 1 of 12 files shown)


vemv:
@yuhan0 : feel free to review this one, I reckon it might interest you

cider-eval.el
```diff
@@ -478,32 +485,63 @@ op/situation that originated this error."
     (let ((error-buffer (cider-new-error-buffer #'cider-stacktrace-mode error-types)))
       (cider-stacktrace-render error-buffer (reverse causes) error-types))))
 
-(defun cider--handle-stacktrace-response (response causes)
-  "Handle stacktrace op RESPONSE, aggregating the result into CAUSES.
+(defcustom cider-ignored-error-phases '("read-source"
```
bbatsov, line 488: I think this name is very confusing, as it doesn't indicate any relationship with the stacktrace display functionality. I also have issue with the term "ignored", as I don't think it reflects what we are really doing here.

cider-eval.el
```diff
@@ -478,32 +485,63 @@ op/situation that originated this error."
...
+                                        "macroexpansion"
+                                        "compile-syntax-check"
+                                        "compilation"
+                                        ;; "execution" is certainly not to be included here.
+                                        ;; "read-eval-result" and "print-eval-result" are not to be included here,
+                                        ;; because they mean that the code has been successfully executed.
+                                        )
+  "Clojure error phases which will not trigger a UI to become visible.
+
+Those UIs include, at the moment:
```
bbatsov, line 499: Given there's just one UI, I'm guessing you can simplify the wording here. :-)

cider-eval.el
bbatsov: Perhaps just name this `cider-clojure-compilation-error-phases` or something like this? I doubt anyone would be changing this one much.

bbatsov:
You also need to update the "dealing with errors" docs.

vemv:
Thanks for the review!

I've amended the last commit applying all feedback.

vemv:
Amended again 

The only change:

> In the docs I call defcustom "configuration option, as I think it's more newcomer friendly.

bbatsov:
Thanks!

Issue #3419, opened 2023-08-19
Also match friendly sessions based on the buffer's ns


vemv:
Cider's sesman 'friendly session' calculation is based on the buffer filename.

But we can also observe the buffer's ns form.

This allows us to match copies of files that _are_ in the classpath.

For instance:

* `foo.clj` (with a specifc file/jar prefix) is in the classpath
* the current buffer's filename does not belong to the classpath
* but the current buffer ns name (`foo`) matches with that of the file that does belong to the classpath.

This way, we avoid 'No linked session' errors, for certain cases.

This approach is a little lax, since a given copy of a ns might be different to the original, but most times it doesn't matter - all that we want is a valid Clojure connection.

PR #3424, opened 2023-08-22
Also match friendly sessions based on the buffer's ns form

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3419

Sample use case:

* Start a REPL on the Haystack project
  * REPL / interactions work, as usual
* Switch to the Orchard project, visit a ns file
  * REPL / interactions work, because the ns form matches with Haystack's ns list
  * (Orchard is a direct dependency of Haystack)
* Switch to an unrelated project
  * REPL / interactions rightfully won't work (`No linked CIDER sessions`), because the ns form won't match with Haystack's ns list

The "ns list" is basically:

* `(all-ns)`, computed once
* `cider-repl-ns-cache`, i.e. the changed namespaces as conveyed by `track-state`.

Cheers - V

### DIFF
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -615,64 +615,6 @@ REPL defaults to the current REPL."
 
 (declare-function cider-classpath-entries "cider-client")
 
-(defun cider--sesman-friendly-session-p (session &optional debug)
-  "Check if SESSION is a friendly session, DEBUG optionally."
-  (setcdr session (seq-filter #'buffer-live-p (cdr session)))
-  (when-let* ((repl (cadr session))
-              (proc (get-buffer-process repl))
-              (file (file-truename (or (buffer-file-name) default-directory))))
-    ;; With avfs paths look like /path/to/.avfs/path/to/some.jar#uzip/path/to/file.clj
-    (when (string-match-p "#uzip" file)
-      (let ((avfs-path (directory-file-name (expand-file-name (or (getenv "AVFSBASE")  "~/.avfs/")))))
-        (setq file (replace-regexp-in-string avfs-path "" file t t))))
-    (when-let ((tp (cider-tramp-prefix (current-buffer))))
-      (setq file (string-remove-prefix tp file)))
-    (when (process-live-p proc)
-      (let* ((classpath (or (process-get proc :cached-classpath)
-                            (let ((cp (with-current-buffer repl
-                                        (cider-classpath-entries))))
-                              (process-put proc :cached-classpath cp)
-                              cp)))
-             (classpath-roots (or (process-get proc :cached-classpath-roots)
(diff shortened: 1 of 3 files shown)


cider-connection.el
```diff
@@ -615,64 +615,6 @@ REPL defaults to the current REPL."
 
 (declare-function cider-classpath-entries "cider-client")
 
-(defun cider--sesman-friendly-session-p (session &optional debug)
```
vemv, line 618: Because the new impl uses misc variables/defuns like cider-repl-ns-cache, cider-sync-request:ns-list, I had to move it, keeping the linter happy.

tbh it makes sense to me that if this code now uses more repl-based capabilities, it's defined in cider-repl.el.

An alternative possibly being `declare-function`?

cider-repl.el
```diff
@@ -1746,6 +1746,79 @@ constructs."
...
+      (let* ((classpath (or (process-get proc :cached-classpath)
+                            (let ((cp (with-current-buffer repl
+                                        (cider-classpath-entries))))
+                              (process-put proc :cached-classpath cp)
+                              cp)))
+             (ns-list (or (process-get proc :all-namespaces)
+                          (let ((ns-list (with-current-buffer repl
+                                           (cider-sync-request:ns-list))))
+                            (process-put proc :all-namespaces ns-list)
+                            ns-list)))
```
vemv, line 1772: This section is new.

cider-repl.el
```diff
@@ -1746,6 +1746,79 @@ constructs."
...
+                        translated))
+            (when-let ((ns (condition-case nil
+                               (substring-no-properties (cider-current-ns :no-default))
+                             (error nil))))
+              ;; if the ns form matches with a ns of all runtime namespaces, we can consider the buffer to match
+              ;; (this is a bit lax, but also quite useful)
+              (with-current-buffer repl
+                (or (when cider-repl-ns-cache ;; may be nil on repl startup
+                      (member ns (nrepl-dict-keys cider-repl-ns-cache)))
+                    (member ns ns-list))))
```
vemv, line 1804: This section is new.

vemv (commented):
The diff might be a little confusing, but luckily, it boils to two additions that I've marked as multi-line comments.

cider-repl.el
bbatsov: Looks good to me. At some point it'd be nice to comment all the cases for a friendly session as the function is now huge and it's kind of hard to follow.

cider-connection.el
bbatsov: I'm not a big fan of `declare-function`, so let's keep it moved.

bbatsov:
Looks good to me now.

PR #3425, opened 2023-08-22
Mention Cider Storm, paren-face in the docs

vemv:


doc/modules/ROOT/pages/debugging/debugger.adoc
```diff
@@ -6,9 +6,8 @@ http://www.gnu.org/software/emacs/manual/html_node/elisp/Edebug.html[Edebug]. Yo
 
 image::cider_debugger.gif[CIDER Debugger]
 
-WARNING: The debugger currently **does not** support ClojureScript. See
-https://github.com/clojure-emacs/cider/issues/1416[this issue]
```
bbatsov, line 10: Why did you remove the link to the ticket?

vemv: My impression was that showing the ticket made sense when there was some sort of intention to implement this, so that one could subscribe to that in-progress discussion.

Now it would be all archeology

PR #3426, opened 2023-08-22
Ensure there's a leading `:` when using `cider-clojure-cli-aliases`

vemv:
In words of Alex Miller:

> aliases should always use `:`
> in some cases, aliases without colons are tolerated (but don't do that)

This PR is a preventive measure, fixing no issue in particular.

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -804,7 +804,10 @@ your aliases contain any mains, the cider/nrepl one will be the one used."
             (if cider-clojure-cli-aliases
                 ;; remove exec-opts flags -A -M -T or -X from cider-clojure-cli-aliases
                 ;; concatenated with :cider/nrepl to ensure :cider/nrepl comes last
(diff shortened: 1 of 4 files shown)


cider.el
```diff
@@ -804,7 +804,10 @@ your aliases contain any mains, the cider/nrepl one will be the one used."
             (if cider-clojure-cli-aliases
                 ;; remove exec-opts flags -A -M -T or -X from cider-clojure-cli-aliases
                 ;; concatenated with :cider/nrepl to ensure :cider/nrepl comes last
-                (format "%s" (replace-regexp-in-string "^-\\(A\\|M\\|T\\|X\\)" "" cider-clojure-cli-aliases))
+                (let ((stripped (format "%s" (replace-regexp-in-string "^-\\(A\\|M\\|T\\|X\\)" "" cider-clojure-cli-aliases))))
```
bbatsov, line 807: I'd just name this `alias`.

Issue #2849, opened 2020-05-24
repl stays in pending state after clojurescript jack in if there's some buffer with an invalid ns declaration


Andre0991:
This error occurs when there's some buffer in `clojure-mode` with an invalid ns declaration (such as `(ns )`.

It seems that the culprit is `clojure-find-ns` (see the stacktrace below), which does not fail gracefully if it can't identify the ns in a buffer. It's not clear from its docstring if throwing an error is its expected behaviour when the ns form is invalid. That's why I'm opening the bug in this repo - I'm not sure if this situation should be handled by `cider` or `clojure-mode`, but it does affect cider.

I use Spacemacs, if that's relevant.

## Expected behavior
Cider would jack in without errors.

## Actual behavior
I got this error (I'm using `toggle-debug-on-error`):

```
Debugger entered--Lisp error: (scan-error "Containing expression ends prematurely" 5 6)
  scan-sexps(5 1)
  forward-sexp(1)
  clojure-forward-logical-sexp()
  clojure--find-ns-in-direction(backward)
  clojure-find-ns()
  cider-current-ns()
  cider-refresh-dynamic-font-lock()
  cider-mode(1)
  cider-enable-on-existing-clojure-buffers()
  #f(compiled-function () #<bytecode 0x477bb459>)()
  #f(compiled-function (buffer) #<bytecode 0x4770c08d>)(#<buffer *cider-repl nu/timeline:localhost:63972(pending-cljs)*>)
  #f(compiled-function (response) #<bytecode 0x4770c109>)((dict "id" "4" "session" "f0733cfc-31b2-4759-bb10-2416c787fa6f" "status" ("done")))
  #f(compiled-function (response) #<bytecode 0x4770c151>)((dict "id" "4" "session" "f0733cfc-31b2-4759-bb10-2416c787fa6f" "status" ("done")))
  nrepl--dispatch-response((dict "id" "4" "session" "f0733cfc-31b2-4759-bb10-2416c787fa6f" "status" ("done")))
  nrepl-client-filter(#<process nrepl-connection> "d2:id1:47:session36:f0733cfc-31b2-4759-bb10-2416c787fa6f5:value3:niled2:id1:42:ns11:shadow.user7:session36:f0733cfc-31b2-4759-bb10-2416c787fa6fed2:id1:47:session36:f0733cfc-31b2-4759-bb10-2416c787fa6f6:statusl4:doneee")
```

## Steps to reproduce the problem
[...]

PR #3428, opened 2023-08-23
Use `clojure-find-ns` safely

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/2849

Bumps clojure-mode, introducing a `cider-get-ns-name` wrapper and using it everywhere.

Callsites handle nil when due.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -114,9 +114,15 @@ instead."
         (clojure-backward-logical-sexp 1)
         (cider--text-or-limits bounds (point) end)))))
 
+(defun cider--clojure-find-ns ()
+  "Calls `clojure-find-ns', supressing any errors.
+Therefore, possibly returns nil,
+so please handle that value from any callsites."
+  (clojure-find-ns :supress-errors))
+
 (defun cider-ns-form ()
   "Retrieve the ns form."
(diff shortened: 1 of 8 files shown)


cider-util.el
```diff
@@ -114,9 +114,15 @@ instead."
         (clojure-backward-logical-sexp 1)
         (cider--text-or-limits bounds (point) end)))))
 
+(defun cider--clojure-find-ns ()
```
bbatsov, line 117: Or just `cider--find-ns`. I guess this one can even be considered public.

vemv: There happens to be a `defun cider--find-ns` already, which does something else

bbatsov: Ah, yes - the one that goes to the ns. I'm guessing it'd still be confusing to have `cider--find-ns` and `cider--clojure-find-ns`, as it's confusing for me even knowing how we ended up here. :-) Let's name the new function `cider-get-ns`, `cider-current-ns` or something like this.

vemv: `cider-get-ns-name`?

Aligned with `clojure.core/ns-name`

bbatsov: Fine by me.

Issue #2436, opened 2018-09-05
Invalid cider-repl-history file prevents cider-jack-in from working


TimoFreiberg:
## Expected behavior

If the cider-repl-history file is invalid, CIDER should still work.
If CIDER could clean up invalid parts automatically, I would prefer that, but if that's impossible (or too hard), I would be fine with getting a prompt to clean it up manually.

## Actual behavior

If the cider-repl-history file is invalid, CIDER doesn't start with the relatively cryptic error message `error in process filter: cider-repl--history-read: End of file during parsing` (although Bozhidar recognized the error immediately :smile: )
To get it working again, you have to clean up/delete the file manually.

## Steps to reproduce the problem

Use [broken-cider-repl-history.txt](https://github.com/clojure-emacs/cider/files/2351750/broken-cider-repl-history.txt) as your `cider-repl-history-file`, which has a malformed expression at the end:
```
(:require [clojure.java.io :as io]
          [datascript.core :as d]
          [mdm-utils
            [service :
```
Run `cider-jack-in-clj`

### CIDER version information

CIDER 0.18.1snapshot (package: 20180903.2111)

### Emacs version

GNU Emacs 26.1

### Operating system

Arch Linux

vspinu:
Should be easy to fix, but how the broken entries get into the history in the first place?

TimoFreiberg:
Unfortunately I have no idea how it happened.

bbatsov:
I'll add here what's needed to be done - we just need to catch that error and display some message prompting the user to reset their history file.

sovelten:
I was having this problem and solved it by removing the nrepl-history file inside `.local/cache`

PR #3429, opened 2023-08-24
Prevent malformed `cider-repl-history-file`s from failing `cider-jack-in`

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/2436

I could repro the original bug and verify it behaves nicely now.

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -2018,7 +2018,10 @@ in an unexpected place."
   ;; apply dir-local variables to REPL buffers
   (hack-dir-local-variables-non-file-buffer)
   (when cider-repl-history-file
-    (cider-repl-history-load cider-repl-history-file)
+    (condition-case nil
+        (cider-repl-history-load cider-repl-history-file)
+      (error
+       (message "Malformed cider-repl-history-file: %s" cider-repl-history-file)))
     (add-hook 'kill-buffer-hook #'cider-repl-history-just-save t t)
     (add-hook 'kill-emacs-hook #'cider-repl-history-just-save))
   (add-hook 'completion-at-point-functions #'cider-complete-at-point nil t)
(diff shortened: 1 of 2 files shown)


cider-repl.el
```diff
@@ -2018,7 +2018,10 @@ in an unexpected place."
   ;; apply dir-local variables to REPL buffers
   (hack-dir-local-variables-non-file-buffer)
   (when cider-repl-history-file
-    (cider-repl-history-load cider-repl-history-file)
+    (condition-case nil
+        (cider-repl-history-load cider-repl-history-file)
+      (error
+       (message (format "Failed to load cider-repl-history-file: %s" cider-repl-history-file))))
```
bbatsov, line 2024: You don't need to embed `format` in `message` - it works in the same way. Maybe mention here that it's probably malformed?

vemv:
Amended!

cider-repl.el
```diff
@@ -2018,7 +2018,10 @@ in an unexpected place."
   ;; apply dir-local variables to REPL buffers
   (hack-dir-local-variables-non-file-buffer)
   (when cider-repl-history-file
-    (cider-repl-history-load cider-repl-history-file)
+    (condition-case nil
+        (cider-repl-history-load cider-repl-history-file)
+      (error
+       (message "Malformed cider-repl-history-file: %s" cider-repl-history-file)))
```
bbatsov, line 2024: Actually you don't need "message" here either, as `error` supports the messaging directly. :-)

vemv: Not sure. From a quick attempt, it looks like `(error ...` formats the string, but doesn't echo it. So the string would become a discarded value.

bbatsov: Hmmm, I can't imagine how this can be the case. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Signaling-Errors.html and other usages of `error` in CIDER.

vemv: But this error is part of the `condition-case` construct, right?

At least that's how I've always understood it.

bbatsov: Hmm, yeah it seems it's different here https://emacs.stackexchange.com/questions/28277/what-is-the-difference-between-the-error-handler-in-condition-case-and-the-error

PR #3432, opened 2023-08-25
Always match friendly sessions for special buffers

vemv:
`*cider-error*`, `*cider-result*`, and possibly others, always deserve to get an associated repl, if possible.

This normally already is the case, because their `default-directory` matches that of the current repl.

However, for some reason, occasionally their default-directory will be something like `~/.emacs.d` which matches no repl's dir.

So this PR fixes that heisenbug.

But it also is an overall improvement, because conceptually, `*cider-error*` / `*cider-result*` belong to no project in particular, so if a repl is available, it should be considered 'friendly' to it.

Cheers - V

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -1750,75 +1750,79 @@ constructs."
                      (mapconcat #'identity (cider-repl--available-shortcuts) ", "))))
         (error "No command selected")))))
 
+(defvar cider--project-agnostic-buffer-names
+  '("*cider-error*" "*cider-result*")
+  "The buffer names known to be project- agnostic/reusable, and from CIDER.")
 
 (defun cider--sesman-friendly-session-p (session &optional debug)
   "Check if SESSION is a friendly session, DEBUG optionally.
 
 The checking is done as follows:
 
+* Consider if the buffer belongs to `cider--project-agnostic-buffer-names`
 * Consider the buffer's filename, strip any Docker/TRAMP details from it
 * Check if that filename belongs to the classpath,
   or to the classpath roots (e.g. the project root dir)
 * As a fallback, check if the buffer's ns form
   matches any of the loaded namespaces."
   (setcdr session (seq-filter #'buffer-live-p (cdr session)))
-  (when-let* ((repl (cadr session))
-              (proc (get-buffer-process repl))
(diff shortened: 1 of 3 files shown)


cider-repl.el
```diff
@@ -1746,72 +1746,76 @@ constructs."
                      (mapconcat #'identity (cider-repl--available-shortcuts) ", "))))
         (error "No command selected")))))
 
+(defvar cider--project-agnostic-buffer-names
```
bbatsov, line 1749: Just check that some of those names are not configurable.

bbatsov: Btw, it seems to me this is more or less `cider-ancillary-buffers`, so I guess you should use it instead.

vemv: Great catch!

Pushed

vemv:
I pushed some unrelated changes, please ignore for now

vemv:
Reviewable again!

bbatsov:
Some tests are failing (other than the Windows tests).

vemv:
The tests pass again if I remove the second commit that observed `cider-ancillary-buffers`. We observe again a hardcoded list again.

Perhaps, it had some nuance?

...For completeness, here's the removed commit: https://github.com/clojure-emacs/cider/commit/7d52b87508515bd4ed101a09701dc0ed9270034b

bbatsov:
Perhaps the list is not full at the time the test is being run? (you'll see that buffers add themselves there after their definitions) Maybe you can just set it to some explicit value for the purpose of the test?

test/cider-connection-tests.el
```diff
@@ -60,6 +60,12 @@
   (before-each
     (setq sesman-sessions-hashmap (make-hash-table :test #'equal)
           sesman-links-alist nil
+          cider-ancillary-buffers (seq-filter (lambda (s)
```
vemv, line 63: Adding this fixed the tests.

vemv (commented):
Ready again!

vemv:
❤️🍻!

We now also have this troubleshooting page for session linking: https://docs.cider.mx/cider/troubleshooting.html#interactions-fail-with-no-linked-cider-sessions

Issue #3437, opened 2023-08-27
`cider--shadow-get-builds`: handle unexpected code


vemv:
> See also https://clojurians.slack.com/archives/C0617A8PQ/p1692587466423359

`cider--shadow-get-builds` is one of the few places where we directly use parseedn, which has [some limitations](https://github.com/clojure-emacs/parseedn/issues/16).

Therefore it seems cautious to add a `condition-case` when getting the build.

PR #3438, opened 2023-08-27
`cider--shadow-get-builds`: handle unexpected code

vemv:
This call can fail due to unbalanced parentheses, unexpected reader literals, etc.

No upstream logic is affected.

Fixes https://github.com/clojure-emacs/cider/issues/3437

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -1029,11 +1029,13 @@ The default options of `browser-repl' and `node-repl' are also included."
 (defun cider--shadow-get-builds ()
   "Extract build names from the shadow-cljs.edn config file in the project root."
   (let ((shadow-edn (concat (clojure-project-dir) "shadow-cljs.edn")))
-    (when (file-exists-p shadow-edn)
-      (with-temp-buffer
-        (insert-file-contents shadow-edn)
-        (let ((hash (car (parseedn-read '((shadow/env . identity))))))
-          (cider--shadow-parse-builds hash))))))
+    (when (file-readable-p shadow-edn)
+      (with-demoted-errors "Error while reading shadow-cljs.edn: %S"
+        (with-temp-buffer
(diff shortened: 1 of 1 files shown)


bbatsov:
Won't the result be more or less the same before as well? I guess the parser fails for invalid EDN anyways.

vemv:
We simply return nil and proceed to ask the user for his preferred build.

It's plausible that parseedn can have bugs that Clojure doesn't, so if we proceed building, the process may succeed (or at least the user will get a more familiar error than `parseclj: Syntax error: "At position 617, unmatched :rbrace"`)

bbatsov:
> We simply return nil and proceed to ask the user for his preferred build.

I prefer for an error to be raised, otherwise a person is less likely to notice the demoted error - after the build prompt will essentially make it disappear. IMO it's a bad idea to suppress errors that require the user to fix something in their setup.

vemv:
The whole piece of code is trying to do something relatively unimportant - extracting an id so that the user doesn't have to type it.

The error was so opaque to an end user that he never got it debugged on its own.

> IMO it's a bad idea to suppress errors that require the user to fix something in their setup.

I'd normally agree, but parseedn isn't as robust as the Clojure reader itself, so errors here are to be expected.

For instance we expect a couple reader literals (`(shadow/env . identity)`, `(env . identity)`), but can't predict if even more will be needed in a future.

bbatsov:
I understand your concerns, but I still don't want to make the suggested change, I don't think this change will improve anything for anyone.  If the build id extraction logic is flawed it should be either improved or disabled. 

I'd rather not waste any time on this right now.

vemv:
We'll finally accomplish this via https://github.com/clojure-emacs/parseedn/pull/17

PR #3439, opened 2023-08-27
Don't pass a nil callback when calling `cider/get-state`

vemv:
* Don't pass a nil callback when calling `cider/get-state`
  * Was causing a warning.
* Improve `No response handler` message

### DIFF
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -992,7 +992,7 @@ nREPL ops, it may be convenient to prevent inserting a prompt.")
     (when (nrepl-op-supported-p "cider/get-state" conn)
       (nrepl-send-request '("op" "cider/get-state")
                           (lambda (_response)
-                            ;; No action is necessary
+                            ;; No action is necessary: this request results in `cider-repl--state-handler` being called.
                             )
                           conn))))
 
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -990,7 +990,11 @@ nREPL ops, it may be convenient to prevent inserting a prompt.")
   "Invokes `cider/get-state' when it's possible to do so."
   (when-let ((conn (cider-current-repl 'cljs)))
     (when (nrepl-op-supported-p "cider/get-state" conn)
-      (nrepl-send-request '("op" "cider/get-state") nil conn))))
+      (nrepl-send-request '("op" "cider/get-state")
+                          (lambda (_response)
+                            ;; No action is necessary
+                            )
+                          conn))))
 
 (defun cider--maybe-get-state-for-shadow-cljs (buffer &optional err)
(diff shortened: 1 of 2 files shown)


cider-repl.el
```diff
@@ -990,7 +990,11 @@ nREPL ops, it may be convenient to prevent inserting a prompt.")
   "Invokes `cider/get-state' when it's possible to do so."
   (when-let ((conn (cider-current-repl 'cljs)))
     (when (nrepl-op-supported-p "cider/get-state" conn)
-      (nrepl-send-request '("op" "cider/get-state") nil conn))))
+      (nrepl-send-request '("op" "cider/get-state")
```
vemv, line 993: I considering changing the `nrepl-send-request` implementation to fail-fast if callback was nil, however that would be a breaking change.

Would be nice if we could have some `assert`s that only ran in CI.

nrepl-client.el
```diff
@@ -480,7 +480,7 @@ older requests with \"done\" status."
                         (gethash id nrepl-completed-requests))))
       (if callback
           (funcall callback response)
-        (error "[nREPL] No response handler with id %s found" id)))))
+        (error "[nREPL] No response handler with id %s found in %s" id (buffer-name))))))
```
vemv, line 483: Note that `nrepl-pending-requests` is buffer-local, so this extra info helps.

cider-repl.el
bbatsov: Won't it be better for this to be sync request and just fetch the state directly? I'm assuming it now leaves the handled to whatever code we used to have in the past for pushing `state`, but that looks kind of weird for an explicit call to `get-state`. (I always find empty handlers to be bad for code readability)

nrepl-client.el
bbatsov: Probably it should say "for" instead of "in", given that it's buffer-local for the connection/REPL buffer only.

cider-repl.el
vemv: `cider-repl--state-handler` is special. `"cider/get-state"` simply triggers that on demand. In that way, it's not a normal request/callback cycle.

Making it sync would decrease user-facing performance.

bbatsov: How so? Emacs is single-treaded, so nothing is really async in CIDER. (it's async only in nREPL)

That out of the way - I'm just concerned about how weird this code looks and how people hacking on CIDER will be wasting time trying to figure out why is this working without a handler. At least expand the comment to explain how this is special.

bbatsov: I saw that that this is not really a state handler, but rather a hook applied to each nREPL message received by CIDER, so we're kind of stuck with the async handler that does nothing. (otherwise the message would be applied twice) I had forgotten how weird this part of the code was.

PR #3440, opened 2023-08-27
Rework `cider-find-keyword`

vemv:
* Bump cider-nrepl
  * necessary for this PR.
* Make `cider-find-keyword` properly work in ClojureScript
  * Thanks to https://github.com/clojure-emacs/cider-nrepl/pull/808, we can open any files (vs. only those that already happened to be open)
* Don't match non-symbols
  * e.g. `"::foo"`, `;; ::foo`, etc
* Don't match keywords that are a superset of the current one
  * e.g. `"::foobar"` is a superset of `::foo`
* Don't leave buffers open if the keyword wasn't found

Cheers - V

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -732,12 +732,27 @@ returned."
                 (cider-nrepl-send-sync-request)
                 (nrepl-dict-get "ns-vars")))
 
-(defun cider-sync-request:ns-path (ns)
-  "Get the path to the file containing NS."
-  (thread-first `("op" "ns-path"
-                  "ns" ,ns)
-                (cider-nrepl-send-sync-request)
-                (nrepl-dict-get "path")))
+(defun cider-sync-request:ns-path (ns &optional favor-url)
+  "Get the path to the file containing NS, FAVOR-URL if specified.
+
+FAVOR-URL ensures a Java URL is returned.
+
+* This always is the case if the underlying runtime is JVM Clojure.
+* For ClojureScript, the default is a resource name.
+  * This often cannot be open by `cider-find-file'
+    (unless there was already a buffer opening that file)
+
+Generally, you always want to FAVOR-URL.
+The option is kept for backwards compatibility.
+
+Note that even when favoring a url, the url itself might be nil,
+in which case we'll fall back to the resource name."
+  (let ((response (cider-nrepl-send-sync-request `("op" "ns-path"
+                                                   "ns" ,ns))))
+    (nrepl-dbind-response response (path url)
+      (if (and favor-url url)
+          url
+        path))))
 
 (defun cider-sync-request:ns-vars-with-meta (ns)
   "Get a map of the vars in NS to its metadata information."
--- a/cider-find.el
+++ b/cider-find.el
@@ -192,6 +192,52 @@ the results to be displayed in a different window."
            (ns (completing-read "Find namespace: " namespaces)))
       (cider--find-ns ns (cider--open-other-window-p arg)))))
 
+(defun cider--find-keyword-in-buffer (buffer kw)
+  "Returns the point where `KW' is found in `BUFFER'.
+Returns nil of no matching keyword is found.
+Occurrences of `KW' as (or within) strings, comments, #_ forms, symbols, etc
+are disregarded."
+  (with-current-buffer buffer
+    (save-excursion
+      (goto-char (point-min))
+      (font-lock-ensure) ;; make the forthcoming `text-properties-at` call useful
+      (let ((found nil)
+            (continue t)
+            (current-point (point)))
+        (while continue
+          (setq found (and (search-forward-regexp kw nil 'noerror)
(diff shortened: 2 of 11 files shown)


vemv:
PR ready. Codespell is red but I'll amend it all in one go after feedback.

cider-client.el
```diff
@@ -732,12 +732,14 @@ returned."
                 (cider-nrepl-send-sync-request)
                 (nrepl-dict-get "ns-vars")))
 
-(defun cider-sync-request:ns-path (ns)
-  "Get the path to the file containing NS."
-  (thread-first `("op" "ns-path"
-                  "ns" ,ns)
-                (cider-nrepl-send-sync-request)
-                (nrepl-dict-get "path")))
+(defun cider-sync-request:ns-path (ns &optional favor-url)
```
bbatsov, line 735: Or you can have a `ns-url` function as well, which would probably be a better API.

cider-find.el
```diff
@@ -192,6 +192,53 @@ the results to be displayed in a different window."
...
+          (setq current-point (point)))
+        (when found
+          current-point)))))
+
+(defun cider--find-keyword-loc (kw)
+  "Given `KW', returns an nrepl-dict with url, dest, dest-point.
+
+Returns the dict in all cases.  `dest-point' indicates success:
+integer on successful finds, nil otherwise."
+  (cider-ensure-connected)
```
bbatsov, line 222: It doesn't make much sense to check for this here as well - it's meant mostly for the commands that would use nREPL APIs.

cider-find.el
```diff
@@ -192,6 +192,53 @@ the results to be displayed in a different window."
...
 
+(defun cider--find-keyword-in-buffer (buffer kw)
+  "Returns the point where `KW' is found in `BUFFER'.
+Returns nil of no matching keyword is found.
+Ocurrences of `KW' as (or within) strings, comments, #_ forms, symbols, etc
+are disregarded."
+  (with-current-buffer buffer
+    (save-excursion
+      (goto-char (point-min))
+      (font-lock-ensure) ;; make the forthcoming `text-properties-at` call useful
```
bbatsov, line 203: I think instead of using font-locking to understand what you're looking at you can use something like https://www.gnu.org/software/emacs/manual/html_node/elisp/Position-Parse.html `syntax-ppss`. I think we used it here and there in clojure-mode.

vemv: Here are two `(syntax-ppss)` results (per clojure-mode's approach), obtained when looking at a symbol, and at a keyword, respectively:

```
(1 2623 2624 nil nil nil 0 nil nil (2623) nil) 
(1 2623 2632 nil nil nil 0 nil nil (2623) nil)
```

...the code I'm looking at being:

```clj
(reg-fx ;; <------- this symbol
 :set-url ;; <------- this keyword
 (fn [{:keys [url]}]
   (router/set-token! url)))

```

Because I don't see distinctive data in those two lists, I wouldn't know how to tell when something is a keyword.

Furthermore, I see CIDER's approach is:

```lisp
(defun cider-in-comment-p ()
  "Return non-nil if point is in a comment."
  (let ((beg (save-excursion (beginning-of-defun) (point))))
    (nth 4 (parse-partial-sexp beg (point)))))
```

...which sounds expensive for this use case, because we repeatedly need to query whether `POS` is looking at at keyword.

With the current approach, we parse exactly once, instead of repeatedly.

cider-client.el
vemv: Per the https://github.com/clojure-emacs/cider-nrepl/pull/808 documentation, in cljs `:url` can be nil, and most likely `:path` has a present value.

Having a single API helps us abstract over that

cider-find.el
vemv: Good catch!

cider-client.el
bbatsov: I get the concern but I'd rather check the return values, and make it explicit what some function is doing. I find it weird that the same function can returning different type of things with different params. If the `ns-path` was written today it'd just return a dictionary, so clients of the function can choose what they want, but obviously the ship has sailed. 

Anyways, I don't feel very strongly about this.

cider-find.el
bbatsov: I meant you could use this to figure out if something is in a string or a comment or not. I don't recall very well the internals of `syntax-ppss`, and I'm super busy today, so we can go with your approach.

cider-client.el
vemv: > If the ns-path was written today it'd just return a dictionary

Yeah, there's a mixture of legacy stuff that prevents one from doing the cleanest thing.

I'll expand the docstring so that intended usage is clearer.

cider-find.el
vemv: I saw some strings and comment helpers but they seemed to not particularly help with the case of `#_`.

Anyway, I appreciate the scrutiny - good to have some thoughts that we can revisit later if needed.

PR #3445, opened 2023-08-29
Improve the reliability of `cider-find-keyword`

vemv:
* `cider-sync-request:ns-path`: add a precondition
  * This would have saved me a confusing stacktrace.
* Improve the reliability of `cider--find-keyword-loc`
  * Allow to jump to a qualified keyword within the same ns
  * Avoid hitting `cider-sync-request:ns-path` when nothing was parsed
  * Explicitly forbid querying non-qualified keywords

Cheers - V

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -747,6 +747,8 @@ The option is kept for backwards compatibility.
 
 Note that even when favoring a url, the url itself might be nil,
 in which case we'll fall back to the resource name."
+  (unless ns
+    (error "No ns provided"))
   (let ((response (cider-nrepl-send-sync-request `("op" "ns-path"
                                                    "ns" ,ns))))
     (nrepl-dbind-response response (path url)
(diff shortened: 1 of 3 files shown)


vemv:
I'll merge this as the test suite has been increased, so there's little chance for error.

cider-client.el
```diff
@@ -747,6 +747,8 @@ The option is kept for backwards compatibility.
 
 Note that even when favoring a url, the url itself might be nil,
 in which case we'll fall back to the resource name."
+  (unless ns
```
bbatsov, line 750: Explicit `nil` checks for params seem a bit weird to me, because this can happen in almost any function.

vemv: Yeah, I thought a bit about it. What made a difference for me here is that this is a nrepl op wrapper.

Error messages are 10x (100x?) more confusing when we add nrepl, jvm, whatever middleware stack is, etc to the equation.

PR #3457, opened 2023-09-08
Fix a flaky test

p4v4n:
This test was failing on my local.
Looks like the exact output generated by `cider-format-connection-params` with nil params depends on where this project is cloned.


vemv:
Perfect! I had noticed this too.

Thanks much!

PR #3460, opened 2023-09-11
Restore xref-based jump-to-definition in Babashka

### LINKED ISSUE #3456: Jump-to-definition doesn't work in Babashka
The [CIDER for Babashka](https://docs.cider.mx/cider/platforms/babashka.html#differences-with-clojure) docs imply that vars defined by the running program should support jump-to-definition:

>Built-in vars (e.g. `clojure.core/map`) don’t have definition location metadata. In practice this means you can’t navigate to their definitions in CIDER.

However, this appears not to be the case. [Clojurians thread](https://clojurians.slack.com/archives/C0617A8PQ/p1694108621229699)

## Expected behavior
I expect to be able to define a var and then jump to its definition from any of its usage sites.

## Actual behavior
Jump to definition does not work for any var, whether built-in or defined locally.

Screenshot 2023-09-08 at 09 32 31

1. [terminal] `bb nrepl-server`
2. [emacs] <kbd>M-x</kbd> `cider-connect`
3. 
```clojure
(ns reproduction)
(def foo 1) ;; => #'reproduction/foo
foo ;; => 1
```
4. with cursor on the latter `foo`, <kbd>M-.</kbd> --> instead of jumping to definition, minibuffer displays "Visit tags table..." prompt

### CIDER version information

```
;; CIDER 1.7.0 (Côte d'Azur), babashka.nrepl 0.0.6-SNAPSHOT
;; Babashka 1.3.181
```

n/a

### Emacs version

GNU Emacs 28.2 (build 1, aarch64-apple-darwin22.1.0, Carbon Version 169 AppKit 2299) of 2022-12-10

### Operating system

macOS Ventura/13.5.1 (22G90)

### JDK distribution
openjdk 17.0.2 2022-01-18
OpenJDK Runtime Environment (build 17.0.2+8-86)
OpenJDK 64-Bit Server VM (build 17.0.2+8-86, mixed mode, sharing)

vemv:
Fixes #3456

I tested this one locally by `cider-connect`ing to a `lein repl` session without cider-nrepl in it.

`M-x xref-find-definitions` now works (and does not work without this patch applied).

Cheers - V

### DIFF
--- a/cider-xref-backend.el
+++ b/cider-xref-backend.el
@@ -42,7 +42,11 @@
   "Used for xref integration."
   ;; Check if `cider-nrepl` middleware is loaded. Allows fallback to other xref
   ;; backends, if cider-nrepl is not loaded.
-  (when (cider-nrepl-op-supported-p "ns-path" nil 'skip-ensure)
+  (when (or
+         ;; the main requirement:
+         (cider-nrepl-op-supported-p "ns-path" nil 'skip-ensure)
+         ;; the fallback, used for bare nrepl or babashka integrations:
+         (cider-nrepl-op-supported-p "lookup" nil 'skip-ensure))
     'cider))
 
 (cl-defmethod xref-backend-identifier-at-point ((_backend (eql cider)))
@@ -91,7 +95,6 @@ These are used for presentation purposes."
 (cl-defmethod xref-backend-definitions ((_backend (eql cider)) var)
   "Find definitions of VAR."
   (cider-ensure-connected)
-  (cider-ensure-op-supported "ns-path")
   (when-let* ((loc (cider--var-to-xref-location var)))
     (list (xref-make var loc))))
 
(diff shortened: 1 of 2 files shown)


cider-xref-backend.el
```diff
@@ -91,7 +95,6 @@ These are used for presentation purposes."
 (cl-defmethod xref-backend-definitions ((_backend (eql cider)) var)
   "Find definitions of VAR."
   (cider-ensure-connected)
-  (cider-ensure-op-supported "ns-path")
```
vemv, line 94: We weren't actually using ns-path under this code path.

We only hit cider-var-info, which looks like this:

```lisp
(defun cider-var-info (var &optional all)
  "Return VAR's info as an alist with list cdrs.
When multiple matching vars are returned you'll be prompted to select one,
unless ALL is truthy."
  (when (and var (not (string= var "")))
    (let ((var-info (cond
                     ((cider-nrepl-op-supported-p "info") (cider-sync-request:info var))
                     ((cider-nrepl-op-supported-p "lookup") (cider-sync-request:lookup var))
                     (t (cider-fallback-eval:info var)))))
      (if all var-info (cider--var-choice var-info)))))
```

So the `lookup` fallback already was there.

bbatsov (approved):

Issue #3207, opened 2022-05-21
Suggestion to migrate CI tests to GitHub 


ikappaki:
Hi,

I believe there is some merit migrating the CI tests to a GitHub workflow, is there any appetite for it?

1. Easy access via the `Actions` menu in GitHub.
2. Currently macos tests are running on GitHub while the rest run on CircleCi, we can get test results under a single badge. 
3. GitHub appears to give unlimited credit for running jobs, thus we can extend the test coverage to include all supported versions across all architectures.

I have created a simple demo workflow to showcase the outcome https://github.com/ikappaki/cider/blob/a4de8824d8e3cdbc8756660fab78f883eea60bfd/.github/workflows/test.yml

It is configured to test what I believe are CIDER supported Emacs versions across macos, ubuntu and windows. There is also Emacs master setup only for ubuntu. 

![image](https://user-images.githubusercontent.com/34983288/169642740-381fcfcf-842c-474d-b7fb-f38c77541369.png)

Thanks

bbatsov:
I'm totally fine with the idea of moving to GHA. (mostly because I don't like mixing CI tools across my projects and in general I've been slowly moving in the direction of GHA)

vemv:
A tdlr would be to look at the big picture - what do we want for clojure-emacs (which most likely benefit from homogeneity), what do we have now, who is actively maintaining their CIs?

ikappaki:
> A tdlr would be to look at the big picture - what do we want for clojure-emacs (which most likely benefit from homogeneity), what do we have now, who is actively maintaining their CIs?
> 
> From that point of view, I wouldn't want to introduce problems while our current situation is non-problematic. Churn is expensive, and our time is limited. I'd love to spend the little "OSS time" I have on actual feature development.

HI @vemv, no worries, this was only a suggestion anyway that I brought up for discussion. I thought of it as an improvement over the current CIDER CircleCI setup -- not that CCI is problematic, but GHA feels better IMHO. I've created a GHA setup with the last three version (26.3, 27.2 and 28.1) running all tests/linters/byte compilers accross macos, ubuntu, win and a snapshot run for Ubuntu. I will keep the branch up (that I was just about to PR) for a while in case it comes up at a later while considering the `bigger picture` or there is a change of sentiment: https://github.com/ikappaki/cider/tree/tech/github-ci-all

Thanks

ikappaki:
[...]

PR #3461, opened 2023-09-11
Add preliminary support for clojure-ts-mode

dannyfreeman:
This does NOT intended to break the dependency from cider to clojure-mode. It is intended to make CIDER work with clojure-ts-mode.

Some functionality like clojure-find-ns, clojure-find-def, etc will still require clojure-mode in order for CIDER to get it's work done.

~I'm at a loss for how to run the tests for this project locally.~ If there are any tests I could write that make sense for these changes I am open to suggestions.

PR is in draft mode because I still need to see where it makes sense to include references to clojure-ts-mode in the documentation.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - [#3352](https://github.com/clojure-emacs/cider/pull/3496): Introduce [`cider-eval-dwim`](https://docs.cider.mx/cider/usage/cider_mode.html#key-reference).
 - Add new customization variable [`cider-clojurec-eval-destination`](https://docs.cider.mx/cider/cljs/up_and_running.html#working-with-cljc-files) to allow specifying which REPL .cljc evals are sent to.
 - [#3354](https://github.com/clojure-emacs/cider/issues/3354): Add new customization variable [`cider-reuse-dead-repls`](https://docs.cider.mx/cider/usage/managing_connections.html#reusing-dead-repls) to control how dead REPL buffers are reused on new connections.
+- Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
 
 ### Bugs fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@
 - [#3354](https://github.com/clojure-emacs/cider/issues/3354): Add new customization variable [`cider-reuse-dead-repls`](https://docs.cider.mx/cider/usage/managing_connections.html#reusing-dead-repls) to control how dead REPL buffers are reused on new connections.
 - Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
 - [#3461](https://github.com/clojure-emacs/cider/pull/3461) Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
+- [#3461](https://github.com/clojure-emacs/cider/pull/3461) Basic support for using CIDER with [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode)
   - The clojure-mode dependency is still required for CIDER to function
   - some features like `cider-dynamic-indentation` and `cider-font-lock-dynamically` do not work with clojure-ts-mode (yet).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@
 - Add new customization variable [`cider-clojurec-eval-destination`](https://docs.cider.mx/cider/cljs/up_and_running.html#working-with-cljc-files) to allow specifying which REPL .cljc evals are sent to.
 - [#3354](https://github.com/clojure-emacs/cider/issues/3354): Add new customization variable [`cider-reuse-dead-repls`](https://docs.cider.mx/cider/usage/managing_connections.html#reusing-dead-repls) to control how dead REPL buffers are reused on new connections.
 - Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
+- [#3461](https://github.com/clojure-emacs/cider/pull/3461) Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
+  - The clojure-mode dependency is still required for CIDER to function
+  - some features like `cider-dynamic-indentation` and `cider-font-lock-dynamically` do not work with clojure-ts-mode (yet).
 
 ### Bugs fixed
 
--- a/Eldev
+++ b/Eldev
@@ -22,24 +22,33 @@
 (setf eldev-standard-excludes `(:or ,eldev-standard-excludes
                                     ;; Avoid including files in test "projects".
                                     (eldev-pcase-exhaustive cider-test-type
-                                                            (`main        "./test/*/")
-                                                            (`integration '("./test/"   "!./test/integration"))
-                                                            (`enrich      '("./test/"   "!./test/enrich"))
-                                                            (`all         '("./test/*/" "!./test/integration")))
+                                                            (`main            "./test/*/")
+                                                            (`integration     '("./test/"   "!./test/integration"))
+                                                            (`enrich          '("./test/"   "!./test/enrich"))
+                                                            (`clojure-ts-mode '("./test/*/" "!./test/clojure-ts-mode"))
+                                                            (`all             '("./test/*/" "!./test/integration")))
                                     "test/integration/projects"
                                     ;; This file is _supposed_ to be excluded
                                     ;; from automated testing.
                                     "test/cider-tests--no-auto.el"))
 
 (eldev-defoption cider-test-selection (type)
-  "Select tests to run; type can be `main', `integration', `enrich' or `all'"
+  "Select tests to run; type can be `main', `integration', `enrich', `clojure-ts-mode' or `all'"
   :options        (-T --test-type)
   :for-command    test
   :value          TYPE
   :default-value  cider-test-type
-  (unless (memq (intern type) '(main integration enrich all))
+  (unless (memq (intern type) '(main integration enrich clojure-ts-mode all))
     (signal 'eldev-wrong-option-usage `("unknown test type `%s'" ,type)))
   (setf cider-test-type (intern type)))
+
+(add-hook 'eldev-load-dependencies-hook
+          (lambda (type additional-sets)
+            (when (and (eq cider-test-type 'clojure-ts-mode)
+                       (member 'runtime additional-sets))
+              (message "Installing tree-sitter grammars")
+              (require 'clojure-ts-mode)
+              (clojure-ts--ensure-grammars))))
 
 (add-hook 'eldev-test-hook
           (lambda ()
--- a/Eldev
+++ b/Eldev
@@ -10,6 +10,7 @@
 
 (eldev-add-loading-roots 'test "test/utils")
 (eldev-add-extra-dependencies 'runtime '(:package logview :optional t))
+(eldev-add-extra-dependencies 'runtime '(:package clojure-ts-mode :optional t))
 
 ;; slightly increase the maximum (applies to checkdoc and the byte compiler alike)
 (setq byte-compile-docstring-max-column 100)
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -549,14 +549,16 @@ higher precedence."
                            (cider--menu-add-help-strings (symbol-value variable))))
     map))
 
-;; This menu works as an easy entry-point into CIDER.  Even if cider.el isn't
-;; loaded yet, this will be shown in Clojure buffers next to the "Clojure"
-;; menu.
 ;;;###autoload
-(with-eval-after-load 'clojure-mode
-  (easy-menu-define cider-clojure-mode-menu-open clojure-mode-map
+(defun cider--setup-menu-for-clojure-major-mode (mode-map)
+  "Setup a Cider menu for a Clojure major mode's MODE-MAP.
+
+This menu works as an easy entry-point into CIDER.  Even if cider.el isn't
+loaded yet, this will be shown in Clojure buffers next to the Clojure menu."
+  (easy-menu-define cider-clojure-mode-menu-open mode-map
     "Menu for Clojure mode.
-  This is displayed in `clojure-mode' buffers, if `cider-mode' is not active."
+  This is displayed in `clojure-mode' and `clojure-ts-mode' buffers,
+  if `cider-mode' is not active."
     `("CIDER" :visible (not cider-mode)
       ["Start a Clojure REPL" cider-jack-in-clj
        :help "Starts an nREPL server and connects a Clojure REPL to it."]
@@ -570,6 +572,14 @@ higher precedence."
        :help "Starts an nREPL server, connects a Clojure REPL to it, and then a ClojureScript REPL."]
       "--"
       ["View user manual" cider-view-manual])))
+
+;;;###autoload
+(with-eval-after-load 'clojure-mode
+  (cider--setup-menu-for-clojure-major-mode clojure-mode-map))
+
+;;;###autoload
+(with-eval-after-load 'clojure-ts-mode
+  (cider--setup-menu-for-clojure-major-mode clojure-ts-mode-map))
 
 ;;; Dynamic indentation
 (defcustom cider-dynamic-indentation t
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -579,7 +579,9 @@ loaded yet, this will be shown in Clojure buffers next to the Clojure menu."
 
 ;;;###autoload
 (with-eval-after-load 'clojure-ts-mode
-  (cider--setup-menu-for-clojure-major-mode clojure-ts-mode-map))
+  (cider--setup-menu-for-clojure-major-mode
+   (with-suppressed-warnings ((free-vars clojure-ts-mode-map))
+     clojure-ts-mode-map)))
 
 ;;; Dynamic indentation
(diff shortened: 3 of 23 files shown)


vemv:
> I'm at a loss for how to run the tests for this project locally

`make test`

Will have a look tomorrow!

bbatsov:
There's some docs about the tests here https://docs.cider.mx/cider/contributing/hacking.html#testing-the-code Feel free to improve it if something is missing.

cider.el
```diff
@@ -2090,18 +2090,31 @@ alternative to the default is `cider-random-tip'."
 
 (add-hook 'cider-connected-hook #'cider--maybe-inspire-on-connect)
 
+;;;###autoload
+(defun cider--setup-clojure-major-mode (mode)
+  "Setup Cider key bindings and hooks for a Clojure major MODE."
+  (cl-flet ((concat-symbol (symbol suffix)
+              (intern (concat (symbol-name mode) suffix))))
```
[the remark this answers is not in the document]
bbatsov: Not sure why this break exactly, but I think it'd be easier to just pass the `mode-map` and the `hook` as a parameter as well. This makes the code simpler IMO. It's not like we're going to need this helper function anywhere else, so it doesn't need to be perfect.

bbatsov:
Overall the changes look fine to me, I'm mostly concerned what happens when you call out to the clojure-mode functions in a buffer with a different mode. I'm guessing they can't work properly without the locals and the syntax table.

vemv:
It's looking good to me as well!

cider-completion.el
```diff
@@ -152,7 +152,7 @@ form, with symbol at point replaced by __prefix__."
                           ;; Important because `beginning-of-defun' and
                           ;; `ending-of-defun' work incorrectly in the REPL
                           ;; buffer, so context extraction fails there.
-                          (derived-mode-p 'clojure-mode))
+                          (derived-mode-p 'clojure-mode 'clojure-ts-mode))
```
vemv, line 155: Maybe it's a good chance to introduce some centralized helpers, e.g. `clojure-mode-like-p`?

Otherwise, it can be easy for contributors to forget that clojure-ts-mode exists, when adding new features.

bbatsov: Yeah, that's fair. Although the names should be some be something like `cider-clojure-(major-)mode-p` or whatever. I'd suggest `cider-util.el` as a good place for such helpers.

bbatsov: Yeah, I think we should have similar helpers for all modes.

vemv:
> There are some other clojure functions to be tested as well. After work I will see about hunting them down and testing them all. This seems like a good candidate for an automated test.

Perhaps you have played as well with the idea of making clojure-mode's test suite succeed with clojure-ts-mode?

It might be not too hard to hack it for accomplishing so. One can always exclude certain tests (our [Eldev](https://github.com/clojure-emacs/cider/blob/ad9225f298aceb40a59906c50ce65211e8fba807/Eldev) shows a test selector setup) if some of them don't make sense for clojure-ts.

bbatsov:
> Perhaps you have played as well with the idea of making clojure-mode's test suite succeed with clojure-ts-mode?

I was just about to ask @dannyfreeman about this on the clojure-ts-mode's issue tracker. Now that semantic indentation has been implemented he can just copy the test suite and disable whatever has not been implemented yet.

bbatsov:
> I've tested clojure-find-ns and clojure-find-def , both of which work in clojure-ts-mode buffers. It could be they work because clojure-ts-mode uses the same syntax-table as clojure-mode (See https://github.com/clojure-emacs/clojure-ts-mode/pull/14). Cider functionality like cider-repl-set-ns that I assume is built on top of clojure-find-ns works as well.

If the syntax-table is the same, they should work.

cider-mode.el
```diff
@@ -568,6 +570,15 @@ higher precedence."
...
       ["View user manual" cider-view-manual])))
 
+;;;###autoload
+(with-eval-after-load 'clojure-mode
+  (cider--setup-menu-for-clojure-major-mode clojure-mode-map))
+
+;;;###autoload
+(with-eval-after-load 'clojure-ts-mode
+  (with-suppressed-warnings ((free-vars clojure-ts-mode-map))
+    (cider--setup-menu-for-clojure-major-mode clojure-ts-mode-map)))
```
vemv, line 580: Would `(symbol-value 'clojure-ts-mode-map)` be an alternative way of fetching this value while avoiding warnings?

This way, we don't suppress any warnings (which might also affect code executed under this code path)

bbatsov: Given how granular `with-suppressed-warnings` is, I don't think it can cause any real issues.

vemv: That's perfect 🍻

bbatsov:
@dannyfreeman What's your timeline for this PR? Do you want to get it in time for CIDER 1.8 or for the next release? I'm asking mostly because I hope that we'll finally release CIDER 1.8 in the next few days. (I know I've said this a few times already :D )

vemv:
> Emacs has facilities for automating this, but it also requires a C compiler and an internet connection.

You are free to add extra steps to https://github.com/clojure-emacs/cider/blob/master/.github/workflows/test.yml which similarly, downloads and installs misc stuff.

(Note that adding it to GH Actions is enough; CircleCI will likely be dropped due to its erratic performance)

> What is the best way to make sure clojure-ts-mode installs only on Emacs29 on the test machine? 

Probably our `Eldev` file can host such logic.

> Similarly, how should certain tests be run conditionally?

By creating a new Eldev group. We have e.g. `integration`. In this PR I added a group named `enrich` (note the Eldev part of the diff): https://github.com/clojure-emacs/cider/pull/3472/files

One can also have conditions at the GH Actions level, see `if: "!startsWith(matrix.os, 'windows')"` in the same PR.

> When clojure-ts-mode is installed it effectively "hijacks" clojure-mode using the new major-mode-remap-alist functionality in Emacs. That could also be troublesome.

I didn't get this point. Would it be troublesome for the existing tests, or for end users?

Cheers - V

vemv:
Nice commit!

> So if clojure-ts-mode was installed for regular test runs, all the tests would start running against clojure-ts-mode instead of clojure-mode like they do now. 

This has an upside. Consider, for example the `describe "cider-defun-at-point"` test. Ideally we would run it once for clojure-mode, and once for clojure-ts-mode.

This way we ensure that CIDER stuff works no matter the `clojure.*mode`.

bbatsov:
> I think the only thing left is to get tests running against clojure-ts-mode. I can try to get them working Saturday, but won't have much time before then. It would be nice to get it in 1.8, but I don't want to hold up the release either.

It seems we won't cut CIDER 1.8 before the end of this week, so you still have time. I'm hoping to finally ship something by the middle of next week.

bbatsov:
Before we merge this PR you can squash this to fewer commits - e.g. one for the changes needed in CIDER and a second one for the CI updates. (or something along those lines)

CHANGELOG.md
```diff
@@ -16,7 +16,9 @@
   * It will be progressively refined and documented, please consider this alpha software.
 - [#2958](https://github.com/clojure-emacs/cider/issues/2958), [#3279](https://github.com/clojure-emacs/cider/issues/3279): `cider-test-run-test`: support arbitrary deftest-like forms, defns with :test metadata, and search for a `-test` counterpart for a given defn (following `cider-test-infer-test-ns` logic).
   - This also makes obsolete the `cider-test-defining-forms` customization variable.
-- Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode) 
+- [#3461](https://github.com/clojure-emacs/cider/pull/3461) Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode)
```
bbatsov, line 19: from -> with?

doc/modules/ROOT/pages/usage/cider_mode.adoc
```diff
@@ -18,7 +18,14 @@ it, so you can be productive.
 (add-hook 'clojure-mode-hook #'cider-mode)
 ----
 
-NOTE: There's no need to enable it explicitly for modes derived from `clojure-mode` like `clojurescript-mode`.
+or if you are using `clojure-ts-mode`
```
bbatsov, line 21: Pleased add `:` at the end of the line.

vemv:
The `cider-expected-ns returns the namespace matching the given string path` test seems rather critical.

Fixing it may be easier than it sounds, since it doesn't appear to hit clojure-mode.

Eldev
```diff
@@ -10,29 +10,43 @@
...
                                     (eldev-pcase-exhaustive cider-test-type
-                                                            (`main        "./test/*/")
-                                                            (`integration '("./test/"   "!./test/integration"))
-                                                            (`all         '("./test/*/" "!./test/integration")))
+                                                            (`main            "./test/*/")
+                                                            (`integration     '("./test/"   "!./test/integration"))
+                                                            (`clojure-ts-mode '("./test/*/" "!./test/clojure-ts-mode"
+                                                                                ;; 1 test in each of the below files fails
+                                                                                ;; when running clojure-ts-mode tests on windows.
+                                                                                "./test/cider-client-tests.el"
```
vemv, line 24: Please only do this if `(member system-type '(ms-dos windows-nt cygwin))`

test/clojure-ts-mode/cider-connection-ts-tests.el
```diff
@@ -0,0 +1,60 @@
...
+    (with-temp-buffer
+      (clojurescript-ts-mode)
+      (expect (cider-repl-type-for-buffer) :to-be 'cljs))
+    (with-temp-buffer
+      (clojurec-ts-mode)
+      (expect (cider-repl-type-for-buffer) :to-be 'multi))))
+
+(provide 'cider-connection-ts-tests)
+
+;;; clojure-connection-ts-tests.el ends here
```
vemv, line 60: Good to remove

(same for `integration-tests.el` - the only test that still has this)

test/clojure-ts-mode/cider-selector-ts-tests.el
```diff
@@ -0,0 +1,40 @@
...
+(require 'clojure-ts-mode)
+(require 'cider-selector-test-utils "test/cider-selector-tests")
+
+(describe "cider-seletor-method-c"
+  (it "switches to most recently visited clojure-ts-mode buffer"
+    (cider-test-selector-method ?c 'clojure-ts-mode "*treesitter-test*.clj")))
+
+(provide 'cider-selector-ts-tests)
+
+;;; clojure-selector-ts-tests.el ends here
```
vemv, line 40: rm

vemv:
The GHA matrix is looking good to me, except for the concrete issue around Windows. It's great that all CIDER tests are passing!

However, clojure-mode tests should also be visibly passing. I've created https://github.com/clojure-emacs/clojure-ts-mode/issues/25 as a suggestion - before we forget, given that this PR's talk has been quite extensive.

For that reason I'll tentatively mark this PR as a draft to prevent accidental merging.

Otherwise merging this too early can introduce unnecessary risks. I could imagine how getting a Treesitter experience could be exciting for users, but if that means bugs for them and troubleshooting time for us, it's of no gain.

Thanks - V

bbatsov:
> Otherwise merging this too early can introduce unnecessary risks. I could imagine how getting a Treesitter experience could be exciting for users, but if that means bugs for them and troubleshooting time for us, it's of no gain.

We can always mark this as experimental "use on your own risk" feature for the time being. Otherwise I doubt we can get it merged in time for CIDER 1.8 (assuming we still aim to release it around the middle of next week).

vemv:
>  Maybe we could consider a smaller release after 1.8 for this to go out with?

Something like this SGTM. 1.8.0 should be released in the middle of next week.

Note that a sizeable chunk of installations use melpa snapshots or Git clones, so clojure-ts-mode would still get plenty of exposure very soon.

I'd really appreciate if https://github.com/clojure-emacs/clojure-ts-mode/issues/25 was given a shot in the meantime, even if many tests had to be disabled. The value I see is visibility and reproducibility - i.e. we all can see in numbers how much trust we can deposite the code at a given moment.

bbatsov:
> Maybe we could consider a smaller release after 1.8 for this to go out with?

That would be fine by me. We can do a 1.9 with a focus on `clojure-ts-mode` and fixes for whatever we manage to break in 1.8, before focusing our attention on 2.0.

Eldev
vemv: Thanks much!

In case it helps debugging, buttercup appears to be bad at getting custom stuff printed (e.g. a `(message "foo")` won't show up in the CI logs), but if you use `(cl-assert ,,, t)`, when it fails, the error will show up.

bbatsov:
Tagging here @ikappaki in case he's willing to help with the Windows tests.

bbatsov:
Now that CIDER 1.8 is finally out, we should attempt to drive this to the finishing line.

ikappaki:
> Tagging here @ikappaki in case he's willing to help with the Windows tests.

Hi all, sorry for the late reply.

Both the aforementioned tests are expected to fail when run on GitHub CI, even without these changes, due to incidental environment issues. They run fine on circleci.

If you happen to remember from some time ago, I've made a suggestion to migrate the circleci jobs to GHA with https://github.com/clojure-emacs/cider/issues/3207, though we've decided we are fine with the current setup.

Although I did not mention it at the time, I've noticed and corrected these two failures in one of my unpublished branches at https://github.com/clojure-emacs/cider/compare/master...ikappaki:cider:tech/github-ci-all .

The first issue with `cider-expected-ns` that @dannyfreeman hacked, has to do with the root directory `d:\a` existing on the GH windows machine by default, and when we ask in the test it to resolve `/a` in the test it resolves it to `d:\a`, which is correct. The fix did in the above branch is to use different names `/cider--a` instead etc

The second issue, eval, is caused by a hardcoded `/tmp` directory (if I remember well by looking at the patch), which is not available on the github machine (fix is to use `tmp-dir` instead).

Are we planning to run the `clojure-ts-mode` tests on GH though? They are akin to the unit test suit already running on circleci and I think they should be bundled with them.

Nevertheless, I can open a PR with just the fixes for these two tests.

Thanks

vemv:
Thanks much for the accurate analysis @ikappaki !

I think it would be fine to follow the current strategy (GH is a superset of Circle because it includes integration tests).

From my side I don't like CCI as much as I once did, but they stilll have some value (especially after https://github.com/clojure-emacs/cider/pull/3500 - now only CCI builds on every commit, which is useful for fast feedback).

Looking forward to the PR.

ikappaki:
> Looking forward to the PR.

Done with https://github.com/clojure-emacs/cider/pull/3532.

Eldev
```diff
@@ -10,29 +10,48 @@
 
 (eldev-add-loading-roots 'test "test/utils")
 (eldev-add-extra-dependencies 'runtime '(:package logview :optional t))
+(eldev-add-extra-dependencies 'runtime '(:package clojure-ts-mode :optional t))
+
+(defvar clojure-ts-failing-tests
+  (when (member system-type '(ms-dos windows-nt cygwin))
+    '("./test/cider-eval-tests.el"))
```
vemv, line 17: These tests no longer fail, after merging master in

cider-client.el
```diff
@@ -154,6 +154,7 @@ nREPL connection."
              (relpath (thread-last
                         (cider-classpath-entries)
                         (seq-filter #'file-directory-p)
+                        (seq-map #'file-truename)
```
ikappaki, line 157: Could you please also revert this change, the test failure should be fixed with #3532

yenda:
@dannyfreeman what's left to do? this looks awesome thanks for doing this great work!

purcell:
Looking forward to this PR landing! (Not a useful comment, certainly, but an opportunity to thank @dannyfreeman for clojure-ts-mode and this work, plus those giving feedback on the PR.)

bitti:
font-lock got broken for me in `clojure-mode` a while ago, but I didn't have the nerve to debug it, so I switched to `clojure-ts-mode` since then. Currently Cider is working fine for me though, just by adding ```(add-hook 'clojure-ts-mode-hook #'cider-mode)``` and patching `cider-repl-type-for-buffer` to include `clojure-ts-mode`.

But it's annoying having to do the patch after each update. If this PR still takes a while, it would be nice if just this one change can be done first.

bbatsov:
@bitti Well, it's perfectly fine for someone else to try to push the necessary changes faster. As I wrote earlier in the discussion here for me the tests were kind of optional at this point and I don't mind cutting a few corners in the interest of making the lives of end users better faster. 

I think it's still a good idea to report a bug for whatever went wrong with `clojure-mode` for you, though. Might be an easy fix.

bbatsov:
@p4v4n @kommen Will one of you be interested in picking up this PR and getting it to mergeable state? (we can tackle porting the tests to `clojure-ts-mode` separately)

kommen:
@bbatsov yep, I'm interested. I might even have some time this weekend to hack on it. @p4v4n if you are interested as well, feel free to hit me up in the Clojurians Slack to coordinate.

kommen:
@bbatsov I have opened https://github.com/clojure-emacs/cider/pull/3622 as a continuation of this PR here, as I couldn't find a way to use this PR. Let me know if you want to use a different approach.

vemv:
SGTM, please only check that there aren't pending items (other than tests)

If there are, please add them as a checklist item in the PR or maybe create an issue in the https://github.com/clojure-emacs/clojure-ts-mode repo?

bbatsov:
> Thanks for picking this up. I didn't have any other pending things in this PR other than tests on the clojure-ts-mode side. I have not had enough free time to be able to work on those.

No worries. Let's just open an issue in the other repo to track this and we'll get it done, when we can. For now it was important to make it easier for people to use CIDER with `clojure-ts-mode` and I think we made some good progress in this direction.

Issue #3344, opened 2023-05-19
Expensive calls to`cider-current-repl` in latency sensitive contexts


yuhan0:
## Expected behavior

Several features in Cider run on short idle timers or hooks that run synchronously while the user is typing/editing. 

- Eldoc hints
- Company etc. auto completion 
- electric / aggressive indentation modes
- Auto formatting

Any latency or momentary freezes in these contexts make for a unpleasant editing experience, so they should be made as optimized as possible.

## Actual behavior

Many of these features call the harmless-looking `(cider-current-repl)` under the hood, and profiling shows it is in fact an expensive operation involving various heuristics and filesystem lookups.

The performance impact grows ~linearly with the number of open REPL connections, See #3343 for more details and a sample backtrace.

- `cider-eldoc` and `cider-complete` via. `(cider-connected-p)`
- `clojure-indent-line` via `cider--get-symbol-indent` ->  `cider-resolve-var`.

These are the main ones that showed up when profiling my own editing usage - eg. I don't use `cider-format` but that seems like it should be affected too. Searching the codebase also finds various `cider-nrepl-sync-request` functions taking an optional `connection` arg which forwards to a call to `(cider-current-repl)`.

## Steps to reproduce the problem
Set the eldoc timer to a short delay to magnify the issues:
`(setq eldoc-idle-delay 0.01)`
then jack-in to many projects simultaneously.

Navigating or typing in a connected clj source buffer should get increasingly laggy with the number of open connections.

### CIDER version information

```
;; CIDER 1.7.0 (Côte d'Azur), nREPL 1.0.0
;; Clojure 1.11.1, Java 20
```

### Lein / Clojure CLI version
Clojure CLI version 1.11.1.1273

### Emacs version

GNU Emacs 29.0.90 

### Operating system

macOS 12.5.1

### JDK distribution

Temurin jdk-20.0.1+9

yuhan0:
Thanks for the feedback! These findings were prompted by my direct experience of stuttering and latency while typing,  so the problem definitely exists to some extent. I can't draw a direct link to eg. Eldoc but the profiler shows a large percentage of CPU and memory being taken up by its idle-timer, specifically by the call to cider-current-repl.

My `eldoc-idle-delay` is set to the default value of 0.5s, but I suggested pushing it to an extreme value to magnify the effects on more powerful machines where they might be hard to observe.

> Eldoc, presumably, is only invoked when one is done typing 
[...]

PR #3463, opened 2023-09-13
Add buffer-local caching to friendly-session calculation

vemv:
Friendly session calculation has always be at least a touch slow ([ref](https://github.com/clojure-emacs/cider/issues/3344), [ref](https://github.com/clojure-emacs/clojure-mode/issues/665#issuecomment-1716440552)), and even more so following recent additions to it.

All slow paths have in common that they run `seq-find` over a series of sequences (the classpath, classpath roots, the ns list). Those can be arbitrarily large.

We may do relatively expensive stuff for each element, like trying to match a string.

So, this PR adds buffer-local caching.

The buffer-local caching is essentially this map:

`<repl buffer>` -> `<is this buffer considered friendly to it?>`

There are timestamps to ensure that we're not basing the caching on stale data.

---

From a quick test, it appears to work. Anyway I'll keep this as a draft to give it at least a couple days of QAing. Review welcome!

### DIFF
--- a/cider-common.el
+++ b/cider-common.el
@@ -309,11 +309,8 @@ whether DIRECTION is 'from-nrepl or 'to-nrepl."
   "Returns the Unix time."
   (float-time))
 
-(defvar cider--all-path-translations-calculated-at nil)
-
 (defun cider--all-path-translations ()
   "Returns `cider-path-translations' if non-empty, else seeks a present value."
-  (setq cider--all-path-translations-calculated-at (cider--unix-time))
   (or cider-path-translations
       ;; cider-path-translations often is defined as a directory-local variable,
       ;; so after jumping to a .jar file, its value can be lost,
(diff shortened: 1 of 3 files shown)


cider-common.el
```diff
@@ -305,8 +305,15 @@ whether DIRECTION is 'from-nrepl or 'to-nrepl."
           (seq-filter #'identity (mapcar f cider-path-translations))
         (seq-some f cider-path-translations)))))
 
+(defun cider--unix-time ()
+  "Returns the Unix time."
+  (float-time))
```
vemv, line 310: Ideally I'd get the time in nanoseconds, as integer. floats make me nervous :)

bbatsov: What's wrong with using `(current-time)`? Sure, it uses seconds, but I'm guessing that'd be fine for your use case.

vemv: I found out that seconds are too coarse-grained for timestamp comparisons.

Say that it was all computed in less than second, we'd get the same timestamp so `>` wouldn't succeed (and `>=` may not be as optimal to use)

vemv:
I'll retake this PR after we get to address https://github.com/clojure-emacs/cider/issues/3468 - otherwise not cool to further touch this area with a release around the corner

bbatsov:
Might be a good time to revisit this for the next minor release.

PR #3464, opened 2023-09-13
middleware_setup.adoc: add instructions for refactor-nrepl

vemv:
* Upgrade refactor-nrepl references in tests
* middleware_setup.adoc: add instructions for refactor-nrepl

...I don't feel too strongly about this one, but seems a sensible improvement (aligned with https://docs.cider.mx/cider/cljs/shadow-cljs.html which also suggests refactor-nrepl).

Cheers - V

### DIFF
--- a/doc/modules/ROOT/pages/basics/middleware_setup.adoc
+++ b/doc/modules/ROOT/pages/basics/middleware_setup.adoc
@@ -32,14 +32,18 @@ Use the convenient plugin for defaults, either in your project's
 
 [source,clojure]
 ----
-:plugins [[cider/cider-nrepl "0.37.0"]]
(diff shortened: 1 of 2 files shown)


bbatsov:
Hmmm, I'll have to think about this. I've intentionally avoided refactor-nrepl in most of the docs, just because so many issues historically came from people trying clj-refactor.el without them knowing what they're are doing. Given that people are often new to both Emacs and Clojure a more complex setup is a recipe for issues IMO.

vemv:
That's reasonable.

Maybe I can include the code preceded by `#_`?

bbatsov:
I think it's best to just keep this out of CIDER's docs.

vemv:
No issue!

Issue #3466, opened 2023-09-16
Simplify nbb-related check 


vemv:
## Background

I noticed some time ago that even a `cider-connect` (which should be very fast) was a touch slower than usual.

## Diagnostic

I found out that this piece of code:

https://github.com/clojure-emacs/cider/blob/09d72192a4734653ca0b1b4b227c38760880fc63/cider-connection.el#L356-L360

Adds 2 seconds to the `cider--connected-handler` sequence.

(Those 2s will only be reproducible on a 'cold' JVM - subsequent calls will be faster. However please note that basically all JVMs are cold when you first start them)

## Task

Run the check only when it's relevant.

I reckon it might be tricky since it might not be clear in advance whether we're looking at a nbb connection.

vemv:
Hi @benjamin-asdf!

I noticed this issue and that you gracefully added nbb support to CIDER.

I wonder if you could check it out?

Perhaps it's easy if `(cider-runtime)` returns `'babashka` for nbb connections. Is it the case?

If so, we could only run the check when the runtime is 'babashka - that would probably zap 90% of the problem.

Cheers - V

vemv:
...while we're there, `cljs.core/demunge` could become `(and (find-ns 'cljs.core) (resolve 'cljs.core/demunge))` - returning values is always at least a fast toucher than throwing exceptions (plus, we make sure we're not hitting anything like an exception-related nrepl middleware)

benjamin-asdf:
Nice find @vemv See pull request. Turns out we can rely on the describe op (runtime..) instead and shave of those 1-2 seconds for jvm startup! 
I fear the penalty is then pushed down to the next first eval call though. Because as you mentionend, only the first takes this long. But I agree having the startup quick is nicer. 

This is a 36% improvement on my end!

PR #3467, opened 2023-09-16
Discover nbb through nrepl describe #3466

benjamin-asdf:
> Fixes https://github.com/clojure-emacs/cider/issues/3466

See issue.

The main idea for the dynamic discovery of the nrepl server capabilities was that you start out with a blank repl
and it augments itself by discovering the server.

The eval during startup was a hit of 1.6 seconds on my machine and 2 for vemv.

1. This check was currently only done for nbb connections (when connecting with cider-connect-clj)
2. For nbb, we do have a version from the describe op, so we already can say if it is nbb, or not

So the current change is to rely on the nrepl descibe op.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
   * It will be progressively refined and documented, please consider this alpha software.
 
 ### Bugs fixed
-
+- [#3466](https://github.com/clojure-emacs/cider/issues/3466): Fix jvm nrepl startup performance (36% improvement).
 - [#3341](https://github.com/clojure-emacs/cider/issues/3341): Escape clojure-cli args on MS-Windows on non powershell invocations.
 - [#3353](https://github.com/clojure-emacs/cider/issues/3353): Fix regression which caused new connections to prompt for reusing dead REPLs.
 - [#3355](https://github.com/clojure-emacs/cider/pull/3355): Fix `cider-mode` disabling itself after a disconnect when `cider-auto-mode` is set to nil.
--- a/cider-connection.el
+++ b/cider-connection.el
@@ -453,7 +453,10 @@ process buffer."
       (nrepl-dict-get nrepl-versions "babashka.nrepl"))))
 
 (defun cider--nbb-nrepl-version ()
-  "Retrieve the underlying connection's babashka.nrepl version."
+  "Retrieve the underlying connection's nbb version.
+
+Note that this is currently not a real version number.
+But helps us know if this is a nbb repl, or not."
   (with-current-buffer (cider-current-repl)
     (when nrepl-versions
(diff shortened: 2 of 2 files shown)


cider-connection.el
```diff
@@ -349,15 +349,10 @@ See `cider-connection-capabilities'."
            (pcase (cider-runtime)
              ('clojure '(clojure jvm-compilation-errors))
              ('babashka '(babashka jvm-compilation-errors))
+             ('nbb '(cljs))
              (_ '()))
            (when
-               (or
```
vemv, line 354: Perhaps, in face of:

> I fear the penalty is then pushed down to the next first eval call though.

(https://github.com/clojure-emacs/cider/issues/3466#issuecomment-1722181337)

...we can issue an async request with a dummy value, e.g. eval the string `"warm-up"`.

That way, we warm up the JVM, and most specifically, we force the loading of the lazily-loaded cider-nrepl middleware.

The user won't perceive a penalty for this forced requires, since it's async.

As a result:

* the repl UI will pop up quickly
* the user's actual first eval will perform more quickly than otherwise

This call is good to add near the bottom of `defun cider--connected-handler`.

cider-connection.el
```diff
@@ -457,11 +452,21 @@ process buffer."
     (when nrepl-versions
       (nrepl-dict-get nrepl-versions "babashka.nrepl"))))
 
+(defun cider--nbb-nrepl-version ()
+  "Retrieve the underlying connection's nbb version.
+
+Note that this is currently not a real version number.
```
vemv, line 458: Is there an issue in nbb for making it so?

vemv (commented):
Thanks much for responding so quickly!

A couple suggestions. Also, please also include a Changelog entry.

cider-connection.el
benjamin-asdf: https://github.com/babashka/nbb/issues/337

cider-connection.el
benjamin-asdf: Ok. Sounds reasonable

vemv (approved):
LGTM!

Please make sure to make the linter pass.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
   * It will be progressively refined and documented, please consider this alpha software.
 
 ### Bugs fixed
-
+- [#3466](https://github.com/clojure-emacs/cider/issues/3466): Fix jvm nrepl startup performance (36% improvement).
```
vemv, line 19:
```suggestion
- [#3466](https://github.com/clojure-emacs/cider/issues/3466): Fix jvm `cider--connected-handler` performance (36% improvement).
```

vemv:
Thanks much!

I've cherry-picked this, preserving attribution.

PR #3471, opened 2023-09-21
Use cider-nrepl 0.38.0

vemv:
https://github.com/clojure-emacs/cider-nrepl/blob/v0.38.0/CHANGELOG.md#0380-2023-09-21
https://clojars.org/cider/cider-nrepl/versions/0.38.0


vemv:
I'll merge this after trying it out (ELPA is down atm and I deleted my cache)

PR #3472, opened 2023-09-21
Introduce Java doc comment rendering

vemv:
Introduces `cider-docstring.el` which is able to render "fragments" (a mixture of plain text and html) into a propertized string.

Starts using such rendered strings in:

* `Company` (and with it, Corfu, since internally it queries Company stuff)
  * For both, the output is always trimmed to 20 lines (customizable)
  * The trimming is smart as it can perform several attempts based on 3 sources: The first sentence, the full body, and the block tags (i.e. the returns/throws/params metadata)
* Eldoc
  * Here we always favor only rendering the first sentence only
* `cider-doc` anciliary buffers
  * as shown e.g. when pressing `f1` while using Company
  * Here, we favor short rendering only if using Company
* Mouse tooltips.

For best results, we also start sending  a Compliment context to `eldoc` and `info` requests.

The context is slightly different than usual: instead of `(.foo__prefix__ bar)`, we have to send (`__prefix__ bar)`, because we're not requesting a completion, but for an [inferred class](https://github.com/clojure-emacs/cider-nrepl/pull/815).

Therefore a `cider-completion-get-info-context-at-point` defun is introduced.

Two new .el files are introduced:

* cider-context, for extracting the context
* cider-docstring, for docstring rendering

This way we avoid cyclical deps.


bbatsov:
> Company (and with it, Corfu, since internally it queries Company stuff) 

I don't get this - Corfu doesn't use anything company-specific to my knowledge.  

> cider-context, for extracting the context

I have some doubts about the extraction of this. I get why you did it, but it looks relatively weird on its own, given the significant connection to completion. Perhaps at least we can name the file `cider-completion-context` so it's clearer what this is?

vemv:
> I don't get this - Corfu doesn't use anything company-specific to my knowledge.

It queries these for showing docstrings:

https://github.com/minad/corfu/blob/cc244c54b392ce4d27ed166730efde4f4d01a07f/extensions/corfu-echo.el#L88C41-L88C52

https://github.com/minad/corfu/blob/cc244c54b392ce4d27ed166730efde4f4d01a07f/extensions/corfu-popupinfo.el#L216C13-L216C62

> Perhaps at least we can name the file cider-completion-context so it's clearer what this is?

SGTM!

> I think we should also expand the completion/docs documentation to mention some of those changes.

Will do 👍

vemv:
@bbatsov ready again.

Besides from Company, cider-doc and eldoc, I've been trying these docstrings over the mouse tooltips as well.

Looking correct and concise in all of those.

Eldev
```diff
@@ -11,6 +11,12 @@
 (eldev-add-loading-roots 'test "test/utils")
 (eldev-add-extra-dependencies 'runtime '(:package logview :optional t))
 
+;; slightly increase the maximum (applies to checkdoc and the byte compiler alike)
+(setq byte-compile-docstring-max-column 100)
```
bbatsov, line 15: The Emacs maintainers advised against doing this, that's why I had changed it. Normally it's best to have development settings in `.dir-locals.el` (you can have them nested in various folders)

test/cider-docstring-tests.el
```diff
@@ -0,0 +1,78 @@
...
+
+(defun cider-render-docstring-test--convert-fragments (fs)
+  (mapcar (lambda (x)
+            (nrepl-dict "type" (gethash :type x)
+                        "content" (gethash :content x)))
+          fs))
+
+(describe "cider--render-docstring"
+  (it "A large corpus of fragments (as produced by Orchard) can be rendered using `shr' without raising errors"
+    (dolist (class '("Thread" "Object" "File" "String"))
```
bbatsov, line 42: I'm thinking this setup code can be extracted and you can have several more granular tests, because now it's a bit harder to follow what exactly is the purpose of the test given it has a bunch of asserts and expects in it. I think it's usually better to have several tests that check for fewer things each than fewer tests that check for more things.

Eldev
bbatsov: Hmm, I can't imagine how - after all `.dir-locals` are applied to the entire directory. Unless Eldev is suppressing those somehow.

vemv: I don't know

There's no mention of .dir-locals in the Eldev repo, which might be a signal that it isn't considered as a config mechanism (it does mentions `four levels of configuration` in the readme)

.github/workflows/test.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         emacs_version: ['26.3', '27.2', '28.2', '29.1']
+        java_version: ['11', '17']
```
vemv, line 17: I introduced a JVM matrix (soon to be expanded with JDK 21)

.github/workflows/test.yml
```diff
@@ -82,3 +83,11 @@ jobs:
         # be GH connectivity runner issues. We attempt to address this
         # problem by rerunning the tests more than once.
         eldev -p -dtTC test --test-type integration || eldev -p -dtTC test --test-type integration
+        
+    - name: Run tests that need enrich-classpath
+      if: "!startsWith(matrix.os, 'windows')"
+      run: |
+        cd dev; ../clojure.sh clojure -M:gen; cd -
```
vemv, line 90: now the `test/File.edn` are generated programatically.

As a nice side-effect, we get to exercise our clojure.sh script, and with it, enrich-classpath.

Both get an extra exercise given the JVM matrix.

We could build more tests like this in a future - nice middle ground between reliable and realistic tests (since there's no nrepl connection - only data).

Eldev
bbatsov: @doublep Does Eldev ignore dir-locals intentionally? If so, I guess we should have the config in both places.

cider-doc.el
```diff
@@ -265,10 +266,17 @@ opposite of what that option dictates."
...
-(defun cider-create-doc-buffer (symbol)
-  "Populates *cider-doc* with the documentation for SYMBOL."
+(defun cider-create-doc-buffer (symbol &optional shorter)
+  "Populates *cider-doc* with the documentation for SYMBOL,
+favoring a SHORTER format if specified."
   (when-let* ((info (cider-var-info symbol)))
-    (cider-docview-render (cider-make-popup-buffer cider-doc-buffer nil 'ancillary) symbol info)))
+    (cider-docview-render (cider-make-popup-buffer cider-doc-buffer nil 'ancillary) symbol info shorter)))
+
+(defun cider-create-shorter-doc-buffer (symbol)
```
bbatsov, line 275: I wonder if "compact" or "abridged" don't reflect better what we're doing here. "shorter" sounds a bit confusing to me.

cider-doc.el
```diff
@@ -265,10 +266,17 @@ opposite of what that option dictates."
 
 (defconst cider-doc-buffer "*cider-doc*")
 
-(defun cider-create-doc-buffer (symbol)
-  "Populates *cider-doc* with the documentation for SYMBOL."
+(defun cider-create-doc-buffer (symbol &optional shorter)
+  "Populates *cider-doc* with the documentation for SYMBOL,
+favoring a SHORTER format if specified."
```
bbatsov, line 271: This can be "compact" format or something similar IMO.

test/cider-docstring-tests.el
bbatsov: I thought you can just generate the large corpus of fragments upfront and test the different things you wanted to verify in it separately in individual tests so it's clearer what's considered correct behavior. The test is not particularly complex, so that's not a big deal. 

My main concern is the usage of `cl-assert` in the test, which I don't quite get, given that you already have `expect`.

Eldev
doublep: No, it simply doesn't do anything special. Which means they are not activated through some automatic magic.

> it does mention four levels of configuration in the readme

Those don't include `.dir-locals`. All ways of configuration are Eldev-specific, but let you evaluate arbitrary Elisp code, not just set variables to constant values.

test/cider-docstring-tests.el
vemv: I use `cl-assert` <-> `expect` in Elisp just like how I use `assert` <-> `is` in clojure: an assert ensures the sanity of the test itself, but does is not concerned with the SUT.

This separation can help seeing which if the two intents one is expressing.

Finally, `cl-assert` appears to fail more clearly under buttercup.

PR #3473, opened 2023-09-21
Don't require `cider-log` eagerly

vemv:
The now-removed `require` was adding some startup performance overhead (as it has transitive dependencies), and most importantly, its https://github.com/magit/transient dep appears to have an issue under Emacs 29 (I'm currently investigating it prior to reporting it).

cider-log.el has autoloads in place - only one was missing so the `require` removal seems safe. Still, I'd be much thankful you could try it out - I still haven't gotten to try out cider-log as I'm busy with misc features.

Cheers - V


bbatsov:
Looks like a safe change to me.

PR #3474, opened 2023-09-21
Adapt if-let/when-let forms for Emacs 29

vemv:
I found out that cider-connect wasn't working for me on Emacs 29. I had heard a similar report a few weeks ago.

It turned out to be the usage of if-let/when-let. Here's an explanation:

https://chat.openai.com/share/eb1d27f2-91a9-4188-97db-1d469bd690d2

I have tested these changes by performing a `cider-connect` and then evaluating some Clojure expressions in Emacs 26 and 29.


vemv:
May be a false alarm for contrived reasons, will check out again tomorrow

vemv:
It turns out, I had a custom if-let definition in my .emacs.d. It hadn't given a problem to me in over a decade, so I had forgotten about its existence!

Sorry for the noise.

PR #3475, opened 2023-09-22
Fix keybindings in cider scratch buffer with newer versions of paredit

p4v4n:
- `C-j` is bound to a different function in newer versions of paredit.
 https://github.com/emacsmirror/paredit/commit/5615023023aea50683f5725284fb9bc6cbbd64ec

### DIFF
--- a/cider-scratch.el
+++ b/cider-scratch.el
@@ -48,6 +48,7 @@
     (set-keymap-parent map clojure-mode-map)
     (define-key map (kbd "C-j") #'cider-eval-print-last-sexp)
     (define-key map [remap paredit-newline] #'cider-eval-print-last-sexp)
+    (define-key map [remap paredit-C-j] #'cider-eval-print-last-sexp)
     (easy-menu-define cider-clojure-interaction-mode-menu map
       "Menu for Clojure Interaction mode"
       '("Clojure Interaction"


cider-scratch.el
```diff
@@ -46,8 +46,9 @@
 (defvar cider-clojure-interaction-mode-map
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map clojure-mode-map)
-    (define-key map (kbd "C-j") #'cider-eval-print-last-sexp)
-    (define-key map [remap paredit-newline] #'cider-eval-print-last-sexp)
+    (define-key map (kbd "C-j") #'cider-eval-last-sexp)
```
bbatsov, line 49: Why not `cider-eval-print-last-sexp`? This doesn't seem correct to me.

p4v4n: The `cider-scratch-initial-message` says
```
;; Press C-j to evaluate the last expression.
;; You can also press C-u C-j to evaluate the expression and pretty-print its result.
```
So C-j is not supposed to insert the result into the buffer.
After this change C-u C-j will still print/insert the result into buffer.

p4v4n: I just looked at the source of `cider-eval-print-last-sexp` and understood what the original intent was. 
Perhaps the initial-message should worded differently. But IMO it makes more sense to have C-j not insert results into the scratch buffer.

bbatsov: I just copied the behavior of Emacs's scratch buffer and I don't think it's prudent to make changes to this behavior as a lot of people are used to it.

p4v4n: Sure. I have reverted this change now.
I didn't notice any change of behaviour between C-j and C-u C-j when simple expressions and thought this was a bug.

PR #3478, opened 2023-09-22
Fix `cider-insert-commands-map` variable initialization

mvproton:
Let form must return sparse map instead of result of define-key function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Fix the `cider-xref-fn-deps` buttons to direct to the right file.
 - Fix the `cider-find-keyword` overall reliability and correctness, particularly for ClojureScript.
 - Make TRAMP functionality work when using non-standard ports.
+- Fix the `cider-insert-commands-map` variable initialization.
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


vemv:
Looks reasonable. May you provide the context as for why is this needed?

Thanks!

mvproton:
@vemv
I found this when I tried to add bindings with my functions to "C-c C-j" prefix and got an error. After that I analyzed the usage of `define-prefix-command` in another contexts and saw that sparse map is returned every time.

vemv:
Makes sense - `defvar cider-insert-commands-map` defines a var that can be used, so it should have the right type.

The pattern can found e.g. [here](https://github.com/doomemacs/doomemacs/blob/844a82c4a0cacbb5a1aa558c88675ba1a9ee80a3/modules/term/eshell/autoload/backport.el#L25-L40)

Please add a changelog entry in this precise line https://github.com/clojure-emacs/cider/blob/c4425fe9e129d98bf3347574d39b529718cabb97/CHANGELOG.md?plain=1#L41 and we're done 🙌

vemv:
Thank you.

mvproton:
Thank for your help with the contribution process.
Description of the fix added.

Issue #3279, opened 2022-12-05
Cannot run tests defined with custom macro


lmartelli:
## Expected behavior
Run tests, provided `cider-test-defining-forms` is customized for a custom test macro

## Actual behavior

`No assertions (or no tests) were run.Did you forget to use ‘is’ in your tests?`

## Steps to reproduce the problem

```clojure
(ns test
  (:require
   [clojure.test :refer [deftest is are]]))


(defmacro def-my-test [f expected]
  `(deftest ~(symbol (str f "-test"))
     (is (= ~expected (~f)))))

(defn meaning-of-life [] 42)

(def-my-test meaning-of-life 43)
```
Include `def-my-test` in `cider-test-defining-forms`.

Run test for `(def-my-test meaning-of-life 42)` with `C-c C-

Macros sems to be OK, since I can run the test by hand : 

```
test> (meaning-of-life-test)

FAIL in (meaning-of-life-test) (test.clj:12)
expected: (clojure.core/= 43 (meaning-of-life))
  actual: (not (clojure.core/= 43 42))
nil
```

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; CIDER 1.1.1 (Plovdiv), nREPL 0.8.3
;; Clojure 1.10.0, Java 17.0.5
```

### Lein / Clojure CLI version

Leiningen 2.9.1

### Emacs version

27.1

### Operating system

Ubuntu 22.04.1 LTS
Linux tp 5.15.0-56-generic #62-Ubuntu SMP

### JDK distribution
openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment (build 17.0.5+8-Ubuntu-2ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.5+8-Ubuntu-2ubuntu122.04, mixed mode, sharing)

bbatsov:
> ;; CIDER 1.1.1 (Plovdiv), nREPL 0.8.3

I assume you're not actually running CIDER 1.1.1, right? :-)

lmartelli:
> > ;; CIDER 1.1.1 (Plovdiv), nREPL 0.8.3
> 
> I assume you're not actually running CIDER 1.1.1, right? :-)

I am afraid I am, actually :-\ I had not realized I am using such an old version. Will try the latest ...

lmartelli:
I have the same behavior with 1.5.0

bbatsov:
[...]

PR #3480, opened 2023-09-24
Add support for running test from src namespace

### LINKED ISSUE #2958: Support `cider-test-run-test` from the `src` NS
Presently, this can only be called from the `test` NS.

It would be nice if you could call `cider-test-run-test` from the `src` NS, instead of having to run the whole `cider-test-run-ns-tests`.

This should be possible if your `deftest` function is named consistently with the source function. So with point on source function `foo`, invoking `cider-test-run-test` would look for a corresponding `test` named `foo` or `foo-test` (considering that [the style guide now now encourages it](https://github.com/bbatsov/clojure-style-guide/pull/132/files)).

The only alternative I see now is using `-with-filters` but that's not too convenient.

tvirolai:
Fixes #2958 

This should add the functionality requested in the issue, which I picked up on as the task seemed reasonably simple to implement. There did not seem to be any tests for this function as of yet - I was unsure about how to approach testing this so I skipped them for now.

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -853,16 +853,22 @@ is searched."
           (cider-test-update-last-test ns var)
           (cider-test-execute ns (list var)))
       ;; we're in a `clojure-mode' buffer
-      (or (when-let* ((ns  (cider-get-ns-name))
-                      (def (clojure-find-def)) ; it's a list of the form (deftest something)
-                      (deftype (car def))
-                      (var (cadr def)))
-            (if (and ns (member deftype cider-test-defining-forms))
-                (progn
-                  (cider-test-update-last-test ns (list var))
-                  (cider-test-execute ns (list var)))
-              (message "No test at point")))
-          (message "No test at point")))))
+      (when-let* ((ns  (cider-get-ns-name))
+                  (def (clojure-find-def)) ; it's a list of the form (deftest something)
+                  (deftype (car def))
+                  (var (cadr def)))
+        (cond
+         ;; the form under the point is a test
+         ((and ns (member deftype cider-test-defining-forms)) (progn
+                                                                (cider-test-update-last-test ns (list var))
+                                                                (cider-test-execute ns (list var))))
+         ;; the form under the point is a function, for which a logically named test exists
+         ;; in the corresponding test namespace
+         ((and ns (string= "defn" deftype)) (when-let ((test-var (concat var "-test"))
+                                                       (ns (funcall cider-test-infer-test-ns
+                                                                    (cider-current-ns t))))
+                                              (cider-test-execute ns (list test-var))))
+         (t (message "No test at point")))))))
 
 (defun cider-test-rerun-test ()
   "Re-run the test that was previously ran."
--- a/cider-test.el
+++ b/cider-test.el
@@ -842,7 +842,7 @@ See `cider-test-rerun-test'."
   "Run the test at point.
 The test ns/var exist as text properties on report items and on highlighted
 failed/erred test definitions.  When not found, a test definition at point
-is searched."
+or in a corresponding test namespace is searched."
   (interactive)
   (let ((ns  (get-text-property (point) 'ns))
         (var (get-text-property (point) 'var)))
--- a/doc/modules/ROOT/pages/usage/cider_mode.adoc
+++ b/doc/modules/ROOT/pages/usage/cider_mode.adoc
@@ -223,7 +223,7 @@ kbd:[C-c C-d C-e]
 | `cider-test-run-test`
 | kbd:[C-c C-t t] +
 kbd:[C-c C-t C-t]
-| Run test at point.
+| Run test at point. If the form under the point is a function, try to search and run a corresponding test.
 
 | `cider-test-rerun-test`
 | kbd:[C-c C-t a] +


vemv (commented):
Hi @tvirolai!

Thanks much for the proposal.

I've marked it as draft because I strongly believe CIDER should discover tests in a more dynamic and runtime-powered way - it's the essence of CIDER for the vast majority of our features.

Being very fair, the current state of things was already far from ideal, as shown in another issue: https://github.com/clojure-emacs/cider/issues/3279

These are the things I don't quite like:

* We "infer the test ns" (`cider-test-infer-test-ns`) instead of inspecting the environment
* We have a fixed `cider-test-defining-forms` list
* We tentatively want (in this PR and also its corresponding issue) to infer a test by merely adding a -test suffix.

I believe we can do better for all of those.

Let me explain how I'd like cider-test-run-test to work:

* It should look at the second symbol of the top level form (e.g. the var name: `foo-test` in `(deftest foo-test`, `bar` in `(defn bar`)
  * Do not check if the first symbol is `defn`, `deftest`, whatever - it shouldn't matter
* Query whether the var exists and it has `:test` metadata
  * Always true for `deftest`, all custom `deftest`-like macros that follow clojure.test customs, and the few vanilla defns in `src/` that may choose to add that metadata 'manually'
* If not, check all the vars that refer to the current var, using xref-find-references
  * This started working recently: https://github.com/clojure-emacs/orchard/pull/177
  * The idea is, this will return all the deftest-like vars that target the var at point
  * We filter them by `:test` metadata, since non-tests will also be returned
* (Especially) if nothing was returned by the previous attempt, try looking up a var named `<var>-test`
  * Nuance: look it up in all the namespaces defined in the project directories
  * In other words, we don't use `cider-test-infer-test-ns`

I'd say, all this logic can be nicely implemented as a new cider-nrepl op, e.g. `"find-test"`.

Let me know what you think (cc/ @bbatsov). I'd be willing to implement this early in the week  - these are good days to add such features.

Cheers - V

bbatsov:
I'm not arguing with @vemv's suggestions - something like `find-test` makes sense, although it'd be nice to keep some fallback logic around for the CIDER users that are using it without `cider-nrepl`. 

I'm not sure if we can fully avoid ns inference though, as it's quite possible you'll have potentially matching vars in several different namespaces, so you'll need some way to decide which one to chose. 

For the PR in question - the changes suggested there seem reasonable given the current implementation. One small concern I have about it is the message you get if you invoke this command for some function without tests, as it's a bit misleading. Probably the message should be different, depending on what you're looking at.

vemv (commented):
The concern of a fallback (for when the would-be cider-nrepl op is absent) is fair.

I suggested an eval-based refinement, which will work with forms other than `defn`, and deftest-like forms that may not be included under `cider-test-defining-forms`.

cider-test.el
```diff
@@ -853,16 +853,22 @@ is searched."
...
+         ;; the form under the point is a test
+         ((and ns (member deftype cider-test-defining-forms)) (progn
+                                                                (cider-test-update-last-test ns (list var))
+                                                                (cider-test-execute ns (list var))))
+         ;; the form under the point is a function, for which a logically named test exists
+         ;; in the corresponding test namespace
+         ((and ns (string= "defn" deftype)) (when-let ((test-var (concat var "-test"))
+                                                       (ns (funcall cider-test-infer-test-ns
+                                                                    (cider-current-ns t))))
+                                              (cider-test-execute ns (list test-var))))
```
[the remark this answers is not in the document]
vemv: We can also avoid the eval with:

```lisp
(cider-resolve--get-in ns "interns" var "test")
```

Please perform first the `cider-resolve--get-in` and then, as a fallback cider-sync-tooling-eval.

The reason is that the `cider-resolve--get-in` is powered by our "track-state" middleware, which may be absent.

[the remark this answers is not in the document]
tvirolai: To clarify: do you mean that snippet as an additional test to the branch quoted, so that we determine that the `test-var` is actually a test? Or should there more refactoring?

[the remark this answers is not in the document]
tvirolai: Ok, I'll refactor this.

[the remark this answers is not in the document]
vemv: Thanks for the willingness 🙌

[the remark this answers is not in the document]
tvirolai: Is this what you had in mind?

cider-test.el
```diff
@@ -838,11 +838,21 @@ See `cider-test-rerun-test'."
   (setq cider-test-last-test-ns ns
         cider-test-last-test-var var))
 
+(defun is-a-test-p (ns var)
+  "Determine if the VAR in NS is a test."
+  (or (cider-resolve--get-in ns "interns" var "test")
```
vemv, line 843:
```suggestion
  (or (cider-resolve--get-in ns "interns" var "test") ;; Returns nil if cider-nrepl is unavailable (or for cljs: https://github.com/clojure-emacs/cider-nrepl/blob/24707/src/cider/nrepl/middleware/track_state.clj#L149)
```

cider-test.el
```diff
@@ -853,16 +863,24 @@ is searched."
...
+                  (deftype (car def))
+                  (var (cadr def)))
+        (cond
+         ;; the form under the point is a test
+         ((and ns (member deftype cider-test-defining-forms)) (progn
+                                                                (cider-test-update-last-test ns (list var))
+                                                                (cider-test-execute ns (list var))))
+         ;; the form under the point is a function, for which a logically named test exists
+         ;; in the corresponding test namespace
+         ((and ns (not (is-a-test-p ns var))) (when-let ((test-var (concat var "-test"))
```
vemv, line 877: We can now also try running the thing at point if `(is-a-test-p ns var)`

doc/modules/ROOT/pages/usage/cider_mode.adoc
```diff
@@ -223,7 +223,7 @@ kbd:[C-c C-d C-e]
 | `cider-test-run-test`
 | kbd:[C-c C-t t] +
 kbd:[C-c C-t C-t]
-| Run test at point.
+| Run test at point. If the form under the point is a function, try to search and run a corresponding test.
```
vemv, line 226: Please add under Changelog, features added:

```
[#2958](https://github.com/clojure-emacs/cider/issues/2958), [#3279](https://github.com/clojure-emacs/cider/issues/3279): `cider-test-run-test` support arbitrary deftest-like forms, defns with :test metadata, and search for a `-test` counterpart for a given defn (following `cider-test-defining-forms`).
```

vemv (commented):
Some suggestions

If possible, please confirm that it works as intended.

Thanks!

tvirolai:
Can confirm that the changes are working as intended - tests found and run for functions that have them and the error is shown if a test does not exist.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - [#3354](https://github.com/clojure-emacs/cider/issues/3354): Add new customization variable `cider-reuse-dead-repls` to control how dead REPL buffers are reused on new connections.
 - [#3364](https://github.com/clojure-emacs/cider/pull/3364): Update enrich-classpath, adding Clojure CLI compatibility, and reworking its integration into CIDER.
   * It will be progressively refined and documented, please consider this alpha software.
+- [#2958](https://github.com/clojure-emacs/cider/issues/2958), [#3279](https://github.com/clojure-emacs/cider/issues/3279): `cider-test-run-test` support arbitrary deftest-like forms, defns with :test metadata, and search for a `-test` counterpart for a given defn (following `cider-test-defining-forms`).
```
vemv, line 17:
```suggestion
- [#2958](https://github.com/clojure-emacs/cider/issues/2958), [#3279](https://github.com/clojure-emacs/cider/issues/3279): `cider-test-run-test` support arbitrary deftest-like forms, defns with :test metadata, and search for a `-test` counterpart for a given defn (following `cider-test-infer-test-ns` logic).
  - This also makes obsolete the `cider-test-defining-forms` customization variable.
```

cider-test.el
```diff
@@ -853,16 +863,24 @@ is searched."
...
-                  (cider-test-execute ns (list var)))
-              (message "No test at point")))
-          (message "No test at point")))))
+      (when-let* ((ns  (cider-get-ns-name))
+                  (def (clojure-find-def)) ; it's a list of the form (deftest something)
+                  (deftype (car def))
+                  (var (cadr def)))
+        (cond
+         ;; the form under the point is a test
+         ((and ns (is-a-test-p ns var)) (progn
```
vemv, line 872: Please reuse the result of `(is-a-test-p ns var)` - the underlying `eval` has some cost

cider-test.el
```diff
@@ -853,16 +863,24 @@ is searched."
           (cider-test-update-last-test ns var)
           (cider-test-execute ns (list var)))
       ;; we're in a `clojure-mode' buffer
-      (or (when-let* ((ns  (cider-get-ns-name))
-                      (def (clojure-find-def)) ; it's a list of the form (deftest something)
-                      (deftype (car def))
-                      (var (cadr def)))
-            (if (and ns (member deftype cider-test-defining-forms))
```
vemv, line 860: At this point we can remove `cider-test-defining-forms` from the codebase and user manual (lives under `doc/`).

I believe it's a non-breaking change - if the user keeps customizing it, nothing will happen)

Please add where it lived:

```lisp
(make-obsolete 'cider-test-defining-forms nil "1.8.0")
```

vemv (commented):
Almost there.

Again, thanks for going through the feedback. I'm pretty interested in having as many contributors as possible familiar with some of our internals, so that future contributions will be smooth 👍

cider-test.el
```diff
@@ -838,11 +838,21 @@ See `cider-test-rerun-test'."
   (setq cider-test-last-test-ns ns
         cider-test-last-test-var var))
 
+(defun is-a-test-p (ns var)
```
vemv, line 841: This should be named `cider--test-var-p`. The important part is the prefix, for avoiding clashes.

cider-test.el
```diff
@@ -853,16 +865,25 @@ is searched."
...
-              (message "No test at point")))
-          (message "No test at point")))))
+      (when-let* ((ns  (cider-get-ns-name))
+                  (def (clojure-find-def)) ; it's a list of the form (deftest something)
+                  (deftype (car def))
+                  (var (cadr def)))
+        (let ((is-a-test (cider--test-var-p ns var)))
+          (cond
+           ;; the form under the point is a test
+           ((and ns is-a-test) (progn
```
vemv, line 875: Note that we're within a when-let*, so `ns` is true - no need to check it again

cider-test.el
```diff
@@ -68,6 +68,8 @@ Add to this list to have CIDER recognize additional test defining macros."
   :type '(repeat string)
   :package-version '(cider . "0.15.0"))
 
+(make-obsolete 'cider-test-defining-forms nil "1.8.0")
```
vemv, line 71: Remember to remove the remaining references to `cider-test-defining-forms` in doc and code

cider-test.el
tvirolai: Upon closer inspection, I noticed that the default branch is never reached. I refactored the `cond` to `if` again.

cider-test.el
```diff
@@ -853,16 +865,23 @@ is searched."
...
+            (progn
+              (cider-test-update-last-test ns (list var))
+              (cider-test-execute ns (list var))))
+        ;; the form under the point is a function, for which a logically named test exists
+        ;; in the corresponding test namespace
+        (let ((test-var (concat var "-test"))
+              (ns (funcall cider-test-infer-test-ns
+                           (cider-current-ns t))))
+          (if (cider--test-var-p ns test-var)
+              (cider-test-execute ns (list test-var))
```
vemv, line 883: Note that `cider-test-update-last-test` is missing here

In face of that, I'd suggest refactoring this such that `cider-test-update-last-test` and `cider-test-execute` called just once, at most.

tvirolai: Would this do? I refactored it to be a lot easier on the eyes, I think. Based on my testing, it should work as before.

vemv:
`cider--test-var` is called twice again.

However, I don't want to torture you with repeated feedback - I know how tedious that can be.

I'll squash this with some tweaks crediting `tvirolai@users.noreply.github.com`.

Thanks yet again!

tvirolai:
Fair enough. I'll be happy to give a go at addressing some other issues in the future as well, as I'm interested in getting into elisp development more.

bbatsov:
@vemv I think the right email is tuomo.virolainen@siili.com

vemv:
Both integrate well into github (normally I choose the `noreply.github.com` ones since it prevents bots from scraping data from Git)

PR #3481, opened 2023-09-25
Refine `cider-test-run-test`

### LINKED ISSUE #2958: Support `cider-test-run-test` from the `src` NS

### LINKED ISSUE #3279: Cannot run tests defined with custom macro

vemv:
> Fixes #2958
> Fixes #3279

Rehash of https://github.com/clojure-emacs/cider/pull/3480 with all my feedback directly applied.

Under this approach, cider-test-run-test will run the first valid test found by:

* tests marked by text properties
* deftests, and deftest-like forms (`cider-test-defining-forms` is no longer necessary)
* :test metadata that may be present in vanilla defns
* deftests, as found by `cider-test-infer-test-ns`
  * We search for a var named `foo-test`, falling back to `foo` (e.g. `deftest foo` - less idiomatic, but quite frequent)

All this lookup is very fast when cider-nrepl is present, since we use `cider-resolve--get-in`. It falls back to eval if we're on bare nrepl.

In a future, we can further search tests by leveraging `xref` functionality - but it's not as important now.

Special thanks to @tvirolai!

### DIFF
--- a/cider-test.el
+++ b/cider-test.el
@@ -875,7 +875,7 @@ or in a corresponding test namespace is searched."
          (found-ns (car found))
          (found-var (cadr found)))
     (if (not found-var)
-        (message "No test found for the function at point")
+        (message "No test found at point")
       (cider-test-update-last-test found-ns (list found-var))
       (cider-test-execute found-ns (list found-var)))))
 
--- a/cider-test.el
+++ b/cider-test.el
@@ -60,13 +60,7 @@
   :type 'boolean
   :package-version '(cider . "0.9.0"))
 
-(defcustom cider-test-defining-forms '("deftest" "defspec")
-  "Forms that define individual tests.
-CIDER considers the \"top-level\" form around point to define a test if
-the form starts with one of these forms.
-Add to this list to have CIDER recognize additional test defining macros."
-  :type '(repeat string)
-  :package-version '(cider . "0.15.0"))
+(make-obsolete 'cider-test-defining-forms nil "1.8.0")
 
 (defvar cider-test-last-summary nil
   "The summary of the last run test.")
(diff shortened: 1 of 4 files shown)


cider-test.el
```diff
@@ -838,31 +832,52 @@ See `cider-test-rerun-test'."
...
+                            ;; deftest foo-test:
+                            (or (when (cider--test-var-p derived-ns derived-var)
+                                  (list derived-ns derived-var))
+                                ;; deftest foo (less usual, but quite frequent):
+                                (when (cider--test-var-p derived-ns v)
+                                  (list derived-ns v))))))))
+         (found-ns (car found))
+         (found-var (cadr found)))
+    (if (not found-var)
+        (message "No test found at point")
```
vemv, line 878: Maybe it could be a `(user-error` instead?

cider-test.el
```diff
@@ -838,31 +832,52 @@ See `cider-test-rerun-test'."
...
-                  (cider-test-update-last-test ns (list var))
-                  (cider-test-execute ns (list var)))
-              (message "No test at point")))
-          (message "No test at point")))))
+  (let* ((ns-from-text-property (get-text-property (point) 'ns))
+         (var-from-text-property (when ns-from-text-property
+                                   ;; we're in a `cider-test-report-mode' buffer
+                                   ;; or on a highlighted failed/erred test definition
+                                   (get-text-property (point) 'var)))
+         (found (or (when (and var-from-text-property
```
bbatsov, line 859: Probably this can be extracted to a private function.

cider-test.el
```diff
@@ -838,31 +832,52 @@ See `cider-test-rerun-test'."
   (setq cider-test-last-test-ns ns
         cider-test-last-test-var var))
 
+(defun cider--test-var-p (ns var)
+  "Determines if the VAR in NS is a test."
+  (if (cider-nrepl-op-supported-p "cider/get-state")
+      (cider-resolve--get-in ns "interns" var "test")
+    (equal "true"
+           (nrepl-dict-get (cider-sync-tooling-eval
+                            (format "(clojure.core/-> %s var clojure.core/meta (clojure.core/contains? :test))"
```
bbatsov, line 841: This will do for now, but it'd be nice if we made something more portable at some point. I was also thinking it'd be handy to have a command that just dumps a var metadata in an overlay or a buffer.

vemv: > This will do for now, but it'd be nice if we made something more portable at some point. 

Speaking of, cljs functionality for our test-related functionality doesn't seem as hard as it might have been in the past.

Inspecting the cljs env, and `eval`ing cljs code seem pretty vanilla things to do nowadays. We could look into it soon enough.

vemv:
Thanks for the review!

PR #3483, opened 2023-09-27
Introduce `cider-inspector-preferred-var-names`

vemv:
I found that the `cider-inspector-def-current-val` prompt makes me think excessively.

90% of the times I want to use the same var name. So I propose CIDER suggests the name `X`, customizable.

(capital `X` to prevent clashes with locally-scoped `x` variables - very common)

The user can always customize it to something else, or choose something else on the fly.

(On-the-fly choices will take precedence in subsequent usages)

After implementing this, I found an interesting usage pattern: maybe I want to repeatedly re-define X and Y - each slot for a different kind of thing. It's really handy to have a `completing-read` for those, so that I don't forget/mistype those two names (which may be longer names if the case calls for it).

Cheers - V

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -274,7 +274,7 @@ MAX-SIZE is the new value."
   (when-let ((value (cider-sync-request:inspect-set-max-coll-size max-size)))
     (cider-inspector--render-value value)))
 
-(defcustom cider-inspector-preferred-var-names '("X")
+(defcustom cider-inspector-preferred-var-names nil
   "The preferred var names to be suggested by `cider-inspector-def-current-val'.
 
 If you choose a different one while completing interactively,
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -274,13 +274,33 @@ MAX-SIZE is the new value."
   (when-let ((value (cider-sync-request:inspect-set-max-coll-size max-size)))
     (cider-inspector--render-value value)))
 
+(defcustom cider-inspector-preferred-var-names '("X")
+  "The preferred var names to be suggested by `cider-inspector-def-current-val'.
+
+If you choose a different one while completing interactively,
+it will be included (in the first position) the next time
+you use `cider-inspector-def-current-val'."
+  :type '(repeat string)
+  :group 'cider
+  :package-version '(cider . "1.8.0"))
+
+(defun cider-inspector--read-var-name-from-user (ns)
+  "Reads a var name from the user, to be defined within NS.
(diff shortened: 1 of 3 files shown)


bbatsov:
I'll have to think about this a bit, as I'm not sure it's a problem for most people.

cider-inspector.el
```diff
@@ -274,13 +274,33 @@ MAX-SIZE is the new value."
...
+you use `cider-inspector-def-current-val'."
+  :type '(repeat string)
+  :group 'cider
+  :package-version '(cider . "1.8.0"))
+
+(defun cider-inspector--read-var-name-from-user (ns)
+  "Reads a var name from the user, to be defined within NS.
+Grows `cider-inspector-preferred-var-names' if the user chose a new name,
+making that new name take precedence for subsequent usages."
+  (let ((v (completing-read (format "Name of the var to be defined in ns %s: " ns)
```
bbatsov, line 291: One small problem with completing read is that with some backends you'll need to do extra work to enter something that's on the list of suggestions. That's why I'm wondering if this will actually save anything in practice, given that typically such vars are one letter long.

vemv: Maybe IDO has me spoiled, but it's hard to imagine for me how a popular `completing-read` backend may make it harder to fill something that's pre-completed.

Does a particular backend come to mind?

Anyway, as noted in the PR it's not only about the completion, but about remembering the names. If I'm using 2 or a handful of vars in my current session, getting the names right is good help.

(In face of that, even if the backend shows `x` and it's easier to type `x` manually, we still see an improvement because we graphically remind users of what the var name is)

bbatsov: > Maybe IDO has me spoiled, but it's hard to imagine for me how a popular completing-read backend may make it harder to fill something that's pre-completed.

I meant this makes it harder to select something that's not pre-filled (as some backends limit what else you can input). I haven't used IDO in a while, but I recall I needed to press something extra if I needed to select something other than the provided completions. Or perhaps this was only if you wanted to override the current match.

> Anyway, as noted in the PR it's not only about the completion, but about remembering the names. If I'm using 2 or a handful of vars in my current session, getting the names right is good help.

That's a fair use case indeed.

bbatsov:
I've also been thinking about the default `X` and I wonder if it's not better to just leave this empty by default, as I'm not sure if that's strong convention to begin with. Personally, I always use vars with longer-names even when I debug stuff.

vemv:
It sounds like a nice middle ground - thanks! It would zap a few concerns, indeed.

Issue #3487, opened 2023-09-28
Error notifications missing


samhedin:
A while ago cider/emacs stopped giving me as useful error notifications as it did before. I tried to debug my init.el but couldn't find anything (I guess this is on the Emacs side and not cider).
For example, evaluating
`(defn foo [] undefined-thing)`
creates a red line under the expression and jumps the marker to the start. However, there is no error information. I can see the error info in the repl buffer or by calling `cider-jump-to-compilation-error` which shows it in the modeline:
```
Syntax error compiling at (dev/user.clj:37:1).
Unable to resolve symbol: undefined-thing in this context
```
but it used to show up as a nice popup right in the buffer. I miss the nice popup.

## Steps to reproduce the problem
Unfortunately, I don't have much to say here.
We have discussed the issue on Slack and I'm apparently not the only one with the issue. However, none of us are sure what we did to make it stop working. I don't even remember when it started, maybe a few weeks ago. 

### CIDER version information
```
;; CIDER 1.8.0-snapshot (package: 20230927.1218), nREPL 1.0.0
;; Clojure 1.11.1, Java 17.0.8
```

### Lein / Clojure CLI version
clj deps 

### Emacs version
29.1

### Operating system
Fedora 38

### JDK distribution

```
openjdk version "17.0.8" 2023-07-18
OpenJDK Runtime Environment (Red_Hat-17.0.8.0.7-1.fc38) (build 17.0.8+7)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.8.0.7-1.fc38) (build 17.0.8+7, mixed mode, sharing)
```

I realize that this report isn't very helpful for debugging, I'm doing my best to figure out more details.

rrudakov:
I also noticed this problem after switching to MELPA version of CIDER.

vemv:
Hi!

I suspect I know what you're talking about, but please kindly confirm:

* If I type `(/ 2 0)` in the .clj buffer and hit `M-x eval-sexp-at-point`, the `*cider-error*` buffer with a stacktrace pops up
  * is `*cider-error*`what you call "error notifications"? Does it have the same kind of contents that you find missing?
* If I type `(defn foo [] undefined-thing)` in the .clj buffer and hit `M-x eval-sexp-at-point`, `*cider-error*` does not pop up
  * Note however that the REPL buffer will print: `Syntax error compiling at (src/haystack/parser/aviso.cljc:13:1). Unable to resolve symbol: undefined-thing in this context`

[...]

PR #3492, opened 2023-09-29
Restore `cider--display-interactive-eval-result` overlays for compilation errors

vemv:
Fixes https://github.com/clojure-emacs/cider/issues/3487

* `cider-interactive-eval-handler`: Check if the given exception's phase is ignored, and if so, display the overlay.
  * The general idea is, we either display the overlay or `*cider-error*` (not both, not none)
* Reuse the same computation ("is the phase ignored?") and if truthy, we avoid jumping to the compilation line (because for compilation errors, it's always the current one, while nrepl while tell us 0:0, which we cannot detect as spurious, client-side)

I've tested this out locally for all cases, and it works nicely.

No changelog is needed.

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -775,8 +775,8 @@ REPL buffer.  This is controlled via
             (cider--make-fringe-overlay (point)))
         (scan-error nil)))))
 
-(defun cider--error-from-ignored-phase-p (buffer)
-  "Returns whether the latest exception associated to BUFFER should be ignored."
+(defun cider--error-phase-of-latest-exception (buffer)
+  "Returns the :phase of the latest exception associated to BUFFER, if any."
   (when cider-clojure-compilation-error-phases
     (when-let ((conn (with-current-buffer buffer
                        (cider-current-repl))))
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -774,6 +775,21 @@ REPL buffer.  This is controlled via
             (cider--make-fringe-overlay (point)))
         (scan-error nil)))))
 
+(defun cider--error-from-ignored-phase-p (buffer)
+  "Returns whether the latest exception associated to BUFFER should be ignored."
+  (when cider-clojure-compilation-error-phases
+    (when-let ((conn (with-current-buffer buffer
+                       (cider-current-repl))))
+      (when (cider-nrepl-op-supported-p "analyze-last-stacktrace" conn)
+        (when-let* ((result (nrepl-send-sync-request (thread-last (map-merge 'list
```
vemv, line 784: There may be up to 2 `analyze-last-stacktrace` requests during the same interaction (one async in `defun cider-default-err-op-handler`, one sync in here).

About 1mo ago I had tried doing this sort of thing in just 1 async call, but was overly complex.

For avoiding any unnecessary costs, we can introduce some caching, cider-nrepl side. The cache would have a size of 1 (since one normally is only interested in analyzing the _last_ exception - older exceptions are uninteresting / unlikely to be re-queried)

vemv: I checked this out and while our stacktrace analysis isn't blazing-fast, runtimes in the 20-40ms range don't seem worth caching.

cider-eval.el
```diff
@@ -774,6 +775,19 @@ REPL buffer.  This is controlled via
             (cider--make-fringe-overlay (point)))
         (scan-error nil)))))
 
+(defun cider--error-phase-of-latest-exception (buffer)
```
bbatsov, line 778: Probably `last` is a better term than `latest` for this. Or `most-recent`.

bbatsov:
The changes look good to me.

PR #3493, opened 2023-09-30
Interactive evaluation: show a shorter overlay when rendering compilation errors

vemv:
e.g., the `Syntax error compiling clojure.core/let at (foo/bar.clj:10:1)` prefix is now removed.

Samples:

image

Screen Shot 2023-09-30 at 15 44 01

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -822,9 +822,15 @@ when `cider-auto-inspect-after-eval' is non-nil."
                                             (and cider-show-error-buffer
                                                  (member phase cider-clojure-compilation-error-phases)))
                                        ;; Display errors as temporary overlays
-                                       (let ((cider-result-use-clojure-font-lock nil))
+                                       (let ((cider-result-use-clojure-font-lock nil)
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -822,9 +822,15 @@ when `cider-auto-inspect-after-eval' is non-nil."
                                             (and cider-show-error-buffer
                                                  (member phase cider-clojure-compilation-error-phases)))
                                        ;; Display errors as temporary overlays
-                                       (let ((cider-result-use-clojure-font-lock nil))
+                                       (let ((cider-result-use-clojure-font-lock nil)
+                                             (trimmed-err (thread-last err
```
bbatsov, line 826: Might be good to add a comment explaining what this removes exactly, as it wasn't clear to me without reading the issue description. Potentially this can also be a small function with a unit test.

vemv: `cider-clojure-compilation-regexp` happens to have nice tests, however I'll add a docstring to it

PR #3496, opened 2023-10-05
Add cider-eval-dwim

samhedin:
If no region is active, eval the toplevel form with `cider-eval-defun-at-point`.
If a region is active, run `cider-eval-region`.
Tiny function, but I find it useful. If there are any other situations/commands to add I'm all ears.

-----------------

- [ ] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md). `checkdoc` stuck: "Documentation style:       Checking..."
- [x] You've added tests (if possible) to cover your change(s). No tests. Are they possible?
- [ ] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint). There are warnings but they are unrelated to my changes.
- [x] You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality).


vemv:
Thanks for the PR!

You can merge master in, in order for the CI warning to disappear.

doc/modules/ROOT/pages/usage/cider_mode.adoc
```diff
@@ -86,6 +86,11 @@ kbd:[C-c C-e]
 kbd:[C-c C-c]
 | Evaluate the top level form under point and display the result in the echo area.
 
+| `cider-eval-dwim`
+| kbd:[C-c C-d]
```
bbatsov, line 90: That's the documentation keymap, which we can't mess with. I'd suggest using the eval keymap `(C-c C-v someting)`.

cider-mode.el
```diff
@@ -495,6 +496,7 @@ higher precedence."
     (define-key map (kbd "M-TAB") #'complete-symbol)
     (define-key map (kbd "C-M-x")   #'cider-eval-defun-at-point)
     (define-key map (kbd "C-c C-c") #'cider-eval-defun-at-point)
+    (define-key map (kbd "C-c C-d") #'cider-eval-dwim)
```
bbatsov, line 499: Please use a keybinding in the eval keymap, instead of this one.

cider-eval.el
```diff
@@ -1361,6 +1361,14 @@ buffer, else display in a popup buffer."
                    'help-echo "Breakpoint. Reevaluate this form to remove it."
                    :type 'cider-fragile))))
 
+(defun cider-eval-dwim (&optional debug-it)
```
bbatsov, line 1364: I'm wondering if a more specific name like `cider-eval-region-or-defun` won't be better, as `dwim` might be different for different people.

vemv: Perhaps `dwim` can leave the door open to additional behaviors in the future.

e.g. it can be smart about evaluating the outer _or_ inner sexpr, depending on the context.

For instance, for cursor | in `(comment | (foo`, most certainly I want to eval `(foo` and not `(comment`.

(Do we have something like that already? IDK myself)

...but I agree that we shouldn't go too crazy - the more conservative we are, the most people will be satisfied by our definition of dwim.

cider-mode.el
samhedin: Is `C-c C-v C-d` appropriate?

cider-eval.el
samhedin: Yes, my idea was that this could support more behavior. 
The important thing to me is that the decision for what function to run isn't nested.
This scales:
``` 
(cond 
  ((region-active? eval-region)
  (inside-comment-form eval-defn-in-comment)
  ...))
```        
This doesn't:
```
(if (region-active?)
  (if (inside-comment-form)
    (....))
  (...))
```
Of course, if you think it's better to leave it with `region-or-defun`, it'll stay that way, and I'll update the name accordingly.

I agree that the behavior @vemv suggested would be cool.

cider-mode.el
```diff
@@ -495,6 +496,7 @@ higher precedence."
     (define-key map (kbd "M-TAB") #'complete-symbol)
     (define-key map (kbd "C-M-x")   #'cider-eval-defun-at-point)
     (define-key map (kbd "C-c C-c") #'cider-eval-defun-at-point)
+    (define-key map (kbd "C-c C-v C-d") #'cider-eval-dwim)
```
bbatsov, line 499: That's not the right keymap for this. There's a dedicated `cider-eval-commands-map`.

cider-mode.el
bbatsov: Yeah, that's fine, just keep in mind that all keybindings in the eval keymap should have both a variant with and without `Control` - `C-c C-v d` and `C-c C-v C-d`. (some people don't like letting go of Control in complex keybindings)

cider-eval.el
bbatsov: Fair enough - if we plan to make this command more flexible it can stay with the `dwim` name.

cider-mode.el
samhedin: `C-c C-v C-d` was already taken by `eval-defun` so I changed `d` to `s`.

doc/modules/ROOT/pages/usage/cider_mode.adoc
```diff
@@ -86,6 +86,10 @@ kbd:[C-c C-e]
 kbd:[C-c C-c]
 | Evaluate the top level form under point and display the result in the echo area.
 
+| `cider-eval-dwim`
+| kbd:[C-c C-v s]
```
bbatsov, line 90: See the entry below - there should be two keybindings listed here.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## master (unreleased)
 
 ### New features
-
+- [#3352](https://github.com/clojure-emacs/cider/pull/3496) Add `cider-eval-dwim`, If no region is active, eval the toplevel form with `cider-eval-defun-at-point`. If a region is active, run `cider-eval-region`
```
bbatsov, line 6: I think you accidentally deleted the preceding blank line.

cider-eval.el
samhedin: @vemv, the behavior you specified works if `clojure-toplevel-inside-comment-form` is set to `t`.
We could make this the default behavior by setting it in a `let`, but it would change how `eval-defun-at-point` works.
https://docs.cider.mx/cider/usage/code_evaluation.html#evaluating-code-inside-comments

If you think it makes sense as default I'm all for it.

vemv: Binding `(clojure-toplevel-inside-comment-form t)` in a fixed manner, only within the scope of `cider-eval-dwim` would sound good to me.

cider-eval.el
```diff
@@ -1365,9 +1365,10 @@ buffer, else display in a popup buffer."
   "If no region is active, call `cider-eval-defun-at-point' with DEBUG-IT.
 If a region is active, run `cider-eval-region'"
   (interactive "P")
-  (if (use-region-p)
-      (cider-eval-region (region-beginning) (region-end))
-    (cider-eval-defun-at-point debug-it)))
+  (let ((clojure-toplevel-inside-comment-form t))
```
bbatsov, line 1368: I'm guessing you should also mention this in the docstring.

vemv:
Thanks much for your contribution! I squashed it with some docstring and changelog tweaks, preserving attribution.

PR #3500, opened 2023-10-05
GHA: avoid duplicated jobs

vemv:
Explained inline. 

It's a recipe I've used in the past.


vemv:
https://github.com/clojure-emacs/cider/actions shows that this worked as intended - merging

PR #3501, opened 2023-10-06
Allow the `Member in class:` prompt to be easily skipped under ido-mode

vemv:
## Context

When using completion for Java interop, one normally browses the options by repeatedly pressing `<down>` or `<up>`.

However, if one completion is ambiguous (as for its Java class) we will perform a `completing-read`. Sample:

image

This `completing-read` breaks the user flow. Say I want to press `<down>, <down>` (go two candidates down). Then after the first candidate, the `completing-read` will interrupt the experience.

One can only exit this pressing RET (to choose an arbitrary one) or C-g. Both require excessive awareness of what's going on and how to remediate it.

Also, ambiguity happens _quite often_, so requiring users to press `<down>, RET` once for each ambigious candidate is a pretty irritating experience.

## Proposed solution

Temporarily redefine the up and down keys to mean 'abort'. This way, completion can proceed.

Only ido-mode is affected. Other modes can be tackled in a future.

I've tried this and it works as intended.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@
 - Always match friendly sessions for `cider-ancillary-buffers` (like `*cider-error*`, `*cider-result*`, etc).
 - `cider-test`: only show diffs for collections.
 - `cider-inspector-def-current-val` now can suggest a var name (default none), which can be customized via `cider-inspector-preferred-var-names`.
+- The "Member in class: " prompt can now be skipped in ido-mode by pressing `<up>` or `<down>`.
 - [#3375](https://github.com/clojure-emacs/cider/pull/3375): `cider-test`: don't render a newline between expected and actual, most times.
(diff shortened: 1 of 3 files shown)


vemv:
Reflecting a bit, while this works, perhaps it's excessively risky to modify things just for a single combination (Company + IDO).

Maybe we can provide the primitives, and document in https://docs.cider.mx/cider/usage/code_completion.html#auto-completion how to wire them up (or write your own, if needed?)

bbatsov:
I agree - it's best to leave this to the users to do it if they want to. It's generally not a good idea to optimize for any particular tool, and a combination of tools is even worse IMO.

vemv:
Refactored

bbatsov (approved):

vemv:
Thanks!

PR #3504, opened 2023-10-08
Add sensible category overrides for `cider`

vemv:
This possibility was enabled by https://github.com/clojure-emacs/cider/pull/3226

It can be seen in use in https://github.com/SystemCrafters/crafted-emacs/pull/241 or https://clojurians.slack.com/archives/C0617A8PQ/p1696684010392119?thread_ts=1696618691.247989&cid=C0617A8PQ|

It provides a better default experience for those using `orderless` or other newer styles.

(Ideally we'd just do whatever is needed to implement `orderless` etc, but we have plenty of stuff on the table. Anyone can pick up https://github.com/clojure-emacs/cider/issues/3019#issuecomment-1029553163)

Cheers - V

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -249,6 +249,11 @@ in the buffer."
                cider-company-unfiltered-candidates
                "CIDER backend-driven completion style."))
 
+;; Currently CIDER completions only work for `basic`, and not `initials`, `partial-completion`, `orderless`, etc.
+;; So we ensure that those other styles aren't used with CIDER, otherwise one would see bad or no completions at all.
+;; This `add-to-list` call can be removed once we implement the other completion styles.
+(add-to-list 'completion-category-overrides '(cider (styles basic flex)))
+
 (defun cider-company-enable-fuzzy-completion ()
   "Enable backend-driven fuzzy completion in the current buffer."
   (setq-local completion-styles '(cider)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
 
 ### Changes
 
+- Ensure that `cider` completion isn't used with completion styles that are currently unsupported (`flex` `initials` `partial-completion`, `orderless`, etc).
+  - This restores completions for users that favor those styles - otherwise the would see bad or no completions.
 - Improve support for multiple forms in the same line by replacing `beginning-of-defun` fn.
 - [#3390](https://github.com/clojure-emacs/cider/issues/3390): Enhance `cider-connect` to show all nREPLs available ports, instead of only Leiningen ones.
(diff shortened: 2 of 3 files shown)


cider-completion.el
```diff
@@ -249,6 +249,11 @@ in the buffer."
                cider-company-unfiltered-candidates
                "CIDER backend-driven completion style."))
 
+;; Currently CIDER completions only work for `basic`, and not `flex` `initials` `partial-completion`, `orderless`, etc.
+;; So we ensure that those other styles aren't used with CIDER, otherwise one would see bad or no completions at all.
+;; This `add-to-list` call can be removed once we implement the other completion styles.
+(add-to-list 'completion-category-overrides '(cider (styles basic)))
```
bbatsov, line 255: Something like this should not be at the top-level IMO, but rather at the mode init. (and should be cleaned when CIDER is disabled). It's not nice for packages to pollute the Emacs global config at load time.

bbatsov:
> (Ideally we'd just do whatever is needed to implement orderless etc, but we have plenty of stuff on the table. Anyone can pick up https://github.com/clojure-emacs/cider/issues/3019#issuecomment-1029553163)

Yeah, I never got to investigating what's actually needed for those. I'm OK with those changes provided they don't break the fuzzy completion (which Emacs calls "flex") in `company-mode`.

bbatsov:
We should also document this somewhere (what's supported and what's not - seems we don't support much today).

cider-completion.el
vemv: I understand the concern, however we already perform `add-to-list 'completion-styles-alist '(cider` above these LOCs.

If we define `cider` globally, it seems only fair to also declare globally the extent of usefulness of `cider`.

The impact of the customization is limited to our own completion style, so the alternative would not add much IMO and would be harder to keep track of (vs. having everything in contiguous LOCs)

vemv:
`flex` had to be declared for the Company fuzzy matching to work (I tested it out).

I added doc as well.

cider-completion.el
bbatsov: Hmm, it seem to me that the preceding expression should have been in the mode setup as well. It's not a big deal in general - obvious the code will work either way. It's more about mode hygiene and not making permanent modifications to global settings.

vemv: I'd be concerned about having too many moving pieces to achieve an otherwise harmless thing.

Since we only affect `cider` aspects here, all modifications are 'namespaced' and aren't too different from e.g. `defun`s (which also techically alter the global environment).

bbatsov: As I said - it's just a matter of modes keeping tidy after themselves. Obviously this completion backend is mostly useless without cider-mode being active. But there's not much point in discussing this - it will work as it is and it's not a big deal.

Issue #3495, opened 2023-10-03
Use the `cider--display-interactive-eval-result` for more cases of compilation errors


vemv:
An overlay is displayed through `cider--display-interactive-eval-result` when a compilation error happens, but only when doing interactive evaluation.

(Note that that is the intended design, and fully makes sense: we use something named cider--display-**interactive-eval**-result)

However, from a user point of view, it is desirable to show this overlay under more use cases, which may not go under the `cider-interactive-eval-handler` (the part responsible for calling `cider--display-interactive-eval-result`)

* ~cider-load-buffer~ (done)
* cider-refresh

For both cases, besides from the display of an overlay, the jumping to the right line, as introduced in https://github.com/clojure-emacs/cider/pull/3492 should also be offered. It might be more tricky for cider-refresh, since the cause of the error may reside in another file (possibly without a corresponding open buffer).

magnars:
I would be willing to give this a shot, but am not familiar enough with the inner workings of CIDER to know where to begin. If I could get a few pointers, where to start looking, what sorts of things to look out for, I could see if I can figure something out.

vemv:
cider-eval.el features a series of 'handler' defuns, please grep for: `defun.*cider.*handler` in that file.

Then, it's a matter of locating the relevant handler for cider-load-buffer, and modify it such that `cider--display-interactive-eval-result` is invoked, when adequate.

As noted in the OP we also want to reliably jump to the relevant line. That is accomplished with the `cider-handle-compilation-errors` function as can be seen in https://github.com/clojure-emacs/cider/pull/3492/files

Probably we want to only jump if the line:col is not 0:0 (some compilation errors are reported like that).

Hope it helps!

magnars:
Thanks for the info, @vemv! It turns out `cider-load-buffer` is handled by `cider-load-file-handler` which it turns out already has this feature via `cider-handle-compilation-errors`. Its docstring says:

> "Highlight and jump to compilation error extracted from MESSAGE, honor NO-JUMP.
> EVAL-BUFFER is the buffer that was current during user's interactive
> evaluation command.  Honor `cider-auto-jump-to-error'."

The question then becomes why it doesn't seem to work. Debugging reveals that it is unable to find the location from the error. It looks like the regex being used does not match the syntax error message. The pattern matches this:

> Syntax error compiling at

but my error message is 

> Syntax error reading source at

[...]

PR #3506, opened 2023-10-09
Add missing case for Clojure 1.10 syntax errors

magnars:
Loading a buffer sometimes fails silently, since `cider-handle-compilation-errors` currently does not properly parse all syntax errors.

The pattern matches:

> Syntax error compiling at
> Syntax error macroexpanding at

but needs to also match:

> Syntax error reading source at

This PR fixes this by adding the missing string to the regex.

There is a relevant issue, #3495, but that is larger in scope. Since evaluating code and getting no feedback about syntax errors is a bad experience, I propose that this be fixed without waiting for the additional quality of life improvements in that issue.

I will start work on all the checkboxes below if this seems like a fair change.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/test/cider-eval-tests.el
+++ b/test/cider-eval-tests.el
@@ -55,5 +55,7 @@
             :to-equal '("src/haystack/analyzer.clj" 18 1 cider-error-highlight-face "Syntax error compiling clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n"))
     (expect (cider-extract-error-info cider-compilation-regexp "Syntax error macroexpanding clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n")
             :to-equal '("src/haystack/analyzer.clj" 18 1 cider-error-highlight-face "Syntax error macroexpanding clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n"))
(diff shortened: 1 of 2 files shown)


vemv:
There are no more cases:

https://github.com/clojure/clojure/blob/6975553804b0f8da9e196e6fb97838ea4e153564/src/jvm/clojure/lang/Compiler.java#L6862-L6870

so LGTM!

Simply make sure to expand `test/cider-eval-tests.el`.

test/cider-eval-tests.el
```diff
@@ -55,5 +55,7 @@
...
     (expect (cider-extract-error-info cider-compilation-regexp "Syntax error macroexpanding clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n")
             :to-equal '("src/haystack/analyzer.clj" 18 1 cider-error-highlight-face "Syntax error macroexpanding clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n"))
+    (expect (cider-extract-error-info cider-compilation-regexp "Syntax error reading source clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n")
+            :to-equal '("src/haystack/analyzer.clj" 18 1 cider-error-highlight-face "Syntax error reading source clojure.core/let at (src/haystack/analyzer.clj:18:1).\n[1] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings\n"))
```
vemv, line 59: The exact message I get for e.g. `::::e` is: `"Syntax error reading source at (/Users/vemv/haystack/src/haystack/parser.cljc:13:0)."`

Please let's reflect such a message here - the added test appears to be synthetic.

magnars: Agreed, that's an improvement for sure. Done.

vemv:
Thanks!

magnars:
Thank you!

Issue #3507, opened 2023-10-09
`cider-jack-in` fails when using asdf (a runtime version manager) and `cider-enrich-classpath` is set to `t`


tucho:
## Expected behavior

When:

1) we set the local clojure version (using asdf);
2) we do not set any global clojure version (again, using asdf);
3) we call `cider-jack-in` on a clojure project;

the REPL should be started.

## Actual behavior

The REPL is not started and the we get the following messages:

```
[nREPL] Starting server via bash /home/marcel/.config/emacs/elpa/cider-20231006.659/clojure.sh /home/marcel/.asdf/shims/clojure -Sdeps \{\:deps\ \{nrepl/nrepl\ \{\:mvn/version\ \"1.0.0\"\}\ cider/cider-nrepl\ \{\:mvn/version\ \"0.39.0\"\}\ refactor-nrepl/refactor-nrepl\ \{\:mvn/version\ \"3.9.0\"\}\}\ \:aliases\ \{\:cider/nrepl\ \{\:main-opts\ \[\"-m\"\ \"nrepl.cmdline\"\ \"--middleware\"\ \"\[refactor-nrepl.middleware/wrap-refactor\,cider.nrepl/cider-middleware\]\"\]\}\}\} -M:cider/nrepl
error in process sentinel: nrepl-server-sentinel: Could not start nREPL server:  ("exited abnormally with code 126")
error in process sentinel: Could not start nREPL server:  ("exited abnormally with code 126")
```

## Steps to reproduce the problem

1) Install and configure asdf (see <https://asdf-vm.com/>).
2) Create a folder for a project.
3) Create an empty `deps.edn` file into this folder.
4) Set the local clojure version to any version available. This will create a `.tool-versions` file into the project folder.
5) Make sure that there is no global clojure version set, i.e., that there is no `.tool-versions` file into your home folder or if there is one, that there is not a clojure versions declared in it.
6) Try to launch a REPL (`cider-jack-in`) for this project.

### CIDER version information

```
;; CIDER 1.8.0-snapshot (package: 20231009.1715), nREPL 1.0.0
;; Clojure 1.11.1, Java 17.0.8.1
```

### Lein / Clojure CLI version

Clojure CLI version 1.11.1.1413

### Emacs version

GNU Emacs 29.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.17.8)

### Operating system

Arch Linux

### JDK distribution

[...]

PR #3508, opened 2023-10-09
Copy .tool-version to the enrich-classpath temporary folder

tucho:
Solve the bug described in the issue #3507.


vemv:
I merged this separately preserving attribution, mainly because it was good timing to also introduce Shellcheck.

Thanks for the contribution!

We recently added a series of enrich-classpath shine: https://clojurians.slack.com/archives/C0617A8PQ/p1696618492372739 - any feedback would be most welcome.

Cheers - V

PR #3509, opened 2023-10-10
Introduce `cider-enable-flex-completion`

vemv:
Replaces `cider-company-enable-fuzzy-completion`, which is now deprecated.

This way, the `flex` completion style becomes opt-in ([7e68df](https://github.com/clojure-emacs/cider/commit/7e68df2bf930f9ea577ffa76172a0b00ff391fe3) had inadvertently enabled it unconditionally).

I have QAd this and it works as intended:

* `iteatn<TAB>` will complete nothing by default
* after `M-x cider-enable-flex-completion`, `iteatn<TAB>` will complete `iteration`.

I also added comments so that we better understand things in a future.

Cheers - V

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -36,6 +36,7 @@
 (require 'cider-docstring)
 (require 'cider-eldoc)
 (require 'nrepl-dict)
+(require 'seq)
 
 (defcustom cider-annotate-completion-candidates t
   "When true, annotate completion candidates with some extra information."
@@ -193,7 +194,7 @@ performed by `cider-annotate-completion-function'."
               ;;
               ;; This api is better described in the section
               ;; '21.6.7 Programmed Completion' of the elisp manual.
-              (cond ((eq action 'metadata) `(metadata (category . cider)))
+              (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
                     ((eq (car-safe action) 'boundaries) nil)
                     (t (with-current-buffer (current-buffer)
                          (complete-with-action action
@@ -243,6 +244,11 @@ in the buffer."
   "Return CIDER completion candidates for STRING as is, unfiltered."
   (cider-complete string))
 
+;; defines a completion style named `cider' (which ideally would have been named `cider-fuzzy').
+;; note that there's already a completion category named `cider' (grep for `(metadata (category . cider))` in this file),
+;; which can be confusing given the identical name.
+;; The `cider' completion style should be removed because the `flex' style is essentially equivalent.
+;; (`flex' was introduced 3 years later, to be fair)
 (add-to-list 'completion-styles-alist
              '(cider
                cider-company-unfiltered-candidates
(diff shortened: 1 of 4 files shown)


cider-completion.el
```diff
@@ -252,11 +258,33 @@ in the buffer."
...
 (defun cider-company-enable-fuzzy-completion ()
-  "Enable backend-driven fuzzy completion in the current buffer."
+  "Enable backend-driven fuzzy completion in the current buffer.
+
+DEPRECATED: please use `cider-company-enable-fuzzy-completion-globally' instead."
   (setq-local completion-styles '(cider)))
 
+(make-obsolete 'cider-company-enable-fuzzy-completion 'cider-company-enable-fuzzy-completion-globally "1.8.0")
+
+(defun cider-company-enable-fuzzy-completion-globally ()
```
vemv, line 272: The old `cider-company-enable-fuzzy-completion` worked on a per-buffer fashion, which isn't necessarily the best UX.

`cider-company-enable-fuzzy-completion-globally` can afford to work globally instead, because `completion-category-overrides` works on our own `cider` category, which allows us to operate "globally _within a namespace_" i.e. safely.

bbatsov:
As mentioned on Slack my main concern is whether this should be used via a hook, given that it seems to be modifying some global settings and functions triggered by hooks usually are buffer-local.

I'm guessing this can just be some function that users can trigger manually in their setups.  Perhaps something like `cider-company-enable-flex-completion` would be a better name, as I think the term "globally" is confusing here. Probably there should also be a note that the old setup ins the only option for people on Emacs 26.

cider-completion.el
```diff
@@ -252,11 +258,33 @@ in the buffer."
...
+  "Enables `flex' (fuzzy) completion for CIDER in all buffers."
+  (interactive)
+  (when (< emacs-major-version 27)
+    (user-error "`cider-company-enable-fuzzy-completion-globally' requires Emacs 27 or later"))
+  (let ((found-styles (when-let ((cider (assq 'cider completion-category-overrides)))
```
bbatsov, line 277: This you base this code on something from another mode? I'm a bit surprised the that setting those `completion-category-overrides` has so many steps, but I'm also not very familiar with the modern completion engine in Emacs.

cider-completion.el
```diff
@@ -243,6 +244,11 @@ in the buffer."
   "Return CIDER completion candidates for STRING as is, unfiltered."
   (cider-complete string))
 
+;; defines a completion style named `cider' (which ideally would have been named `cider-fuzzy').
+;; note that there's already a completion category named `cider' (grep for `(metadata (category . cider))` in this file),
+;; which can be confusing given the identical name.
+;; The `cider' completion style should be removed because the `flex' style is essentially equivalent.
+;; (`flex' was introduced 3 years later, to be fair)
```
bbatsov, line 251: It's best to use specific years (e.g. 2020 or even better - Emacs 27) here, as comments become outdated at some point.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -140,17 +141,21 @@ CIDER-specific "fuzzy completion" by adding:
...
+This adds the `flex` completion style, as introduced in Emacs 27.
+
 Now, `company-mode` will accept certain fuzziness when matching
 candidates against the prefix. For example, typing `mi` will show you
 `map-indexed` as one of the possible completion candidates and `cji`
 will complete to `clojure.java.io`. Different completion examples are
 shown
 https://github.com/alexander-yakushev/compliment/wiki/Examples[here].
 
+NOTE: `cider-company-enable-fuzzy-completion` (now depreecated) should be used for Emacs < 27.
```
bbatsov, line 157: deprecated

bbatsov: Ah, you already had that note that I suggested elsewhere. :-)

bbatsov:
Perhaps it would be a bad idea to check what other modes that support `flex` are doing in terms of init code.

cider-completion.el
vemv: This is simply for updating a nested structure like `((foo (styles bar) (cycle t)) (cider (styles basic flex)))` in a way that preserves other entries.

Sadly we don't have something as handy as `update-in`, it seems!

vemv (commented):
> Perhaps it would be a bad idea to check what other modes that support flex are doing in terms of init code.

The best thing we could do is to implement completion comprehensively such that all styles will work OOTB. It just happens to not be so easy.

That way we could remove the `completion-category-overrides` customizations, I'd consider it a temporary measure for having control of what really intend to be functional as of today.

In a future, there would be no `cider-company-enable-fuzzy-completion-globally` or any similar function - users would just add `flex` in a standard fashion.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -140,17 +141,21 @@ CIDER-specific "fuzzy completion" by adding:
 
 [source,lisp]
 ----
-(add-hook 'cider-repl-mode-hook #'cider-company-enable-fuzzy-completion)
-(add-hook 'cider-mode-hook #'cider-company-enable-fuzzy-completion)
+(add-hook 'cider-repl-mode-hook #'cider-company-enable-fuzzy-completion-globally)
```
bbatsov, line 144: I don't think this needs to be a hook, given what the function does - it should just be something the users add to their CIDER config.

cider-completion.el
bbatsov: I see.

cider-completion.el
bbatsov: Yeah, I get this, although for the end users it'd be more or less the same (as the hook would be triggered in any `cider-mode` buffer), so I'm big concerned the name might be confusing. That was part of the reason I suggested the use of `flex` instead of `fuzzy`. I'm also thinking that is not company specific, so we should probably drop `company` from the name.

vemv:
Ready again!

cider-completion.el
```diff
@@ -252,11 +259,35 @@ in the buffer."
...
+
+DEPRECATED: please use `cider-enable-flex-completion' instead."
   (setq-local completion-styles '(cider)))
 
+(make-obsolete 'cider-company-enable-fuzzy-completion 'cider-enable-flex-completion "1.8.0")
+
+(defun cider-enable-flex-completion ()
+  "Enables `flex' (fuzzy) completion for CIDER in all buffers.
+
+Only affects the `cider' comlpetion category.`"
```
bbatsov, line 276: completion

bbatsov:
Other than the small typo I mentioned - looks good to go.

vemv:
Thanks!

PR #3513, opened 2023-10-12
Use `cider-nrepl` 0.39.1

vemv:
https://github.com/clojure-emacs/cider-nrepl/blob/v0.39.1/CHANGELOG.md#0391-2023-10-12 
https://clojars.org/cider/cider-nrepl/versions/0.39.1


vemv:
Tried locally

PR #3515, opened 2023-10-12
Update up_and_running.adoc

behrica:
mention 2 other problematic cases apart from "Docker" (multi-hop ssh and across hosts mounted files)

### DIFF
--- a/doc/modules/ROOT/pages/basics/up_and_running.adoc
+++ b/doc/modules/ROOT/pages/basics/up_and_running.adoc
@@ -351,8 +351,8 @@ will reuse existing SSH connection's parameters (like port and username) for est
 The same will happen if you try to `cider-connect-\*` to a host that matches the one you're currently
 connected to.
 
-NOTE: For Docker containers, running `cider-jack-in-\*` over TRAMP may technically work but it may give mixed results.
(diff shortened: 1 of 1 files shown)


doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -351,8 +351,8 @@ will reuse existing SSH connection's parameters (like port and username) for est
 The same will happen if you try to `cider-connect-\*` to a host that matches the one you're currently
 connected to.
 
-NOTE: For Docker containers, running `cider-jack-in-\*` over TRAMP may technically work but it may give mixed results.
-Please check out the following section for the recommended approaches.
+NOTE: For Docker containers,multihop ssh connections or across hosts mounted files running `cider-jack-in-\*` over TRAMP may technically work but it may give mixed results.
```
vemv, line 354: What does `across hosts mounted files` mean?

behrica: when I have files from one machine "mounted" on an other.
This means that the "mapping" of a "path" back to the machine might fail.
This happens between host<->docker often, but could hapen as well host<->host

Issue #3522, opened 2023-10-15
Finer-grained opt-in to overlays


daveliepmann:
I'm thankful for CIDER 1.8.0's improved use of Clojure 1.10+ error reporting features. One way I'd like to leverage this change is to get short/non-stacktrace error messages as an overlay, _without_ changing the behavior of 1. stacktrace/_\*cider-error\*_-opening errors or 2. normal evaluation. 

That is, my dream scenario is:

1. successful/uneventful form evaluation, e.g. `(range 100)` --> no overlay, see result in minibuffer
2. early-phase error (reading, macroexpansion, compilation), e.g. `:::5` or `(def 5)` --> error overlay
3. late-phase error, e.g. `(throw (ex-info "boom" {:a "a"}))` --> open _\*cider-error\*_ with stacktrace, no overlay

Because `cider-use-overlays` is coarse-grained, it seems 1 and 2 are currently incompatible. Would it be possible to get more fine-grained config here?

Cheers and thanks

(relevant [clojurians thread](https://clojurians.slack.com/archives/C0617A8PQ/p1697390114500909))

vemv:
Thanks!

It seems feasible. `defcustom cider-use-overlays` currently accepts `'both`, `t`, or `nil`.

We could add a `errors-only` option, which would be like `t`. Errors show in the repl in all cases, so we don't need `both-errors-only`, `t-errors-only`, etc.

daveliepmann:
Cool!

> Errors show in the repl in all cases, so we don't need `both-errors-only`, `t-errors-only`, etc.

This doesn't matter to me because I don't mind if situation 2 (early-phase errors) show in both minibuffer and the overlay, but I can't speak to the preferences of others. As long as 1, 2, and 3 are separately configurable I'm satisfied.

vemv:
We have https://github.com/clojure-emacs/cider/pull/3524 in transit, feel free to chime in

PR #3524, opened 2023-10-15
Introduce `errors-only` value for `cider-use-overlays`

vemv:
Fixes https://github.com/clojure-emacs/cider/issues/3522

It works as intended.

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -865,7 +865,8 @@ when `cider-auto-inspect-after-eval' is non-nil."
                                          (cider--display-interactive-eval-result
                                           trimmed-err
                                           end
-                                          'cider-error-overlay-face)))
+                                          'cider-error-overlay-face
+                                          :is-error)))
 
                                      (cider-handle-compilation-errors err
                                                                       eval-buffer
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -840,7 +840,7 @@ when `cider-auto-inspect-after-eval' is non-nil."
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  (lambda (_buffer value)
                                    (setq res (concat res value))
-                                   (cider--display-interactive-eval-result res end))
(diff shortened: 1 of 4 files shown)


cider-overlays.el
```diff
@@ -290,9 +292,10 @@ overlay."
 
 
 ;;; Displaying eval result
-(defun cider--display-interactive-eval-result (value &optional point overlay-face)
+(defun cider--display-interactive-eval-result (value &optional point overlay-face error-p)
```
vemv, line 295: (We probably should make `error-p` mandatory so that consumers don't forget specifying this)

bbatsov: That's a private API, so it's fine to change it. Perhaps it could be `value-type` instead, so it'd give us more flexibility in the future. I guess it can also stay as optional right after `value`, so at least the ordering of the args is more natural.

cider-eval.el
```diff
@@ -865,7 +865,8 @@ when `cider-auto-inspect-after-eval' is non-nil."
                                          (cider--display-interactive-eval-result
                                           trimmed-err
                                           end
-                                          'cider-error-overlay-face)))
+                                          'cider-error-overlay-face
+                                          :is-error)))
```
bbatsov, line 869: In Elisp it's more common to use symbols instead of keywords, so I'd go with `'is-error`.

doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -189,7 +189,16 @@ variable.
 (setq cider-overlays-use-font-lock t)
 ----
 
-You can disable overlays entirely (and display results in the echo-area at the
+You can restrict overlays to only be shown for errors
```
bbatsov, line 192: We might also explain why this may make sense. Without some extra context I doubt many people, other than the person who asked for this, will end up using it.

PR #3530, opened 2023-10-16
Preserve the font size as one navigates through the CIDER inspector

### LINKED ISSUE #3527: Inspector: keep the font size
I noticed that while using the inspector, if I `M-x text-scale-increase`, that font size won't be kept as one navigates.

vemv:
Fixes #3527

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -389,6 +389,9 @@ MAX-COLL-SIZE if non nil."
 (defun cider-inspector--render-value (value)
   "Render VALUE."
   (let ((font-size (when-let ((b (get-buffer cider-inspector-buffer)))
+                     ;; The font size is lost between inspector 'screens',
+                     ;; because on each re-rendering, we wipe everything, including the mode.
+                     ;; Enabling cider-inspector-mode is the specific step that loses the font size.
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -388,8 +388,10 @@ MAX-COLL-SIZE if non nil."
 ;; Render Inspector from Structured Values
 (defun cider-inspector--render-value (value)
   "Render VALUE."
-  (cider-make-popup-buffer cider-inspector-buffer 'cider-inspector-mode 'ancillary)
-  (cider-inspector-render cider-inspector-buffer value)
+  (let ((font-size (when-let ((b (get-buffer cider-inspector-buffer)))
```
bbatsov, line 391: I'd add a note here explaining why we need to do this.

bbatsov:
I'm guessing we might have a similar problem in other buffers with drill-down functionality.

vemv:
> I'm guessing we might have a similar problem in other buffers with drill-down functionality.

I could use a refresher as for which are those!

bbatsov:
I could use one myself. 😆 I just have some vague memory there was something else besides this. (maybe the ns browser?) Anyways, not a big deal.

PR #3531, opened 2023-10-16
Introduce `cider-shorten-error-overlays` customization option

### LINKED ISSUE #3525: Make the error overlay trimming optional
## Context

https://github.com/clojure-emacs/cider/discussions/3338#discussioncomment-7286764

## Task 

Extract https://github.com/clojure-emacs/cider/blob/b83c997ec2473254f97e8754c217ce2d485b9d33/cider-eval.el#L860-L864 to a defun, and make its application optional.

vemv:
Fixes #3525

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -821,13 +821,15 @@ REPL buffer.  This is controlled via
                                                      conn)))
           (nrepl-dict-get result "phase"))))))
 
-(defcustom cider-shorten-error-overlays t
-  "If t, overlays that reflect error messages will be more concise.
-Specifically, the prefix matched by `cider-clojure-compilation-regexp',
-and the suffix matched by `cider-module-info-regexp' will be removed."
+(defcustom cider-inline-error-message-function #'cider--shorten-error-message
+  "A function that will shorten a given error message,
+as shown in overlays / the minibuffer (per `cider-use-overlays').
+
+The function takes a single arg.  You may want to use `identity',
+for leaving the message as-is."
   :type 'boolean
   :group 'cider
-  :package-version '(cider . "0.19.0"))
+  :package-version '(cider . "1.19.0"))
 
 (defun cider--shorten-error-message (err)
   "Removes from ERR the prefix matched by `cider-clojure-compilation-regexp',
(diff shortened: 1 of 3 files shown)


bbatsov:
Seems to me we don't really need a defcustom for this. Given that the offending line would already be underlined this should simply be the only behavior. There's no need to make everything customizable. :-)

bbatsov:
Hmm, it seems to me we're solving the wrong problem here - if the highlighting is not precise/useful probably we should improve it. Obviously adding this extra defcustom is not a big deal for us, but it just seems weird to me to have such a high granularity of customizations, that probably almost no one will ever touch.

vemv:
My take would be that different Clojure developers need/want different degrees of detail, for errors.

Personally, I feel that I have seen certain errors so many times, that just a short string will suffice to inform me.

But not everyone operates like that.

bbatsov:
> My take would be that different Clojure developers need/want different degrees of detail, for errors.

This part I get, but I believe that level of granularity is something like a line in an overlay vs a stacktrace buffer. Now we're discussing whether we need a defcustom for a message. Every such branch/defcustom adds some maintenance complexity, that's why it's good to consider carefully their value and their alternatives.

cider-eval.el
```diff
@@ -821,6 +821,24 @@ REPL buffer.  This is controlled via
                                                      conn)))
           (nrepl-dict-get result "phase"))))))
 
+(defcustom cider-shorten-error-overlays t
+  "If t, overlays that reflect error messages will be more concise.
+Specifically, the prefix matched by `cider-clojure-compilation-regexp',
+and the suffix matched by `cider-module-info-regexp' will be removed."
+  :type 'boolean
+  :group 'cider
+  :package-version '(cider . "0.19.0"))
```
bbatsov, line 830: Btw, this should be 1.9, not 0.19. :-)

vemv:
Maybe a middle ground solution would be to make `cider--shorten-error-message` public such that users can advice it to act like the identity function.

cider-eval.el
```diff
@@ -821,6 +821,24 @@ REPL buffer.  This is controlled via
                                                      conn)))
           (nrepl-dict-get result "phase"))))))
 
+(defcustom cider-shorten-error-overlays t
```
bbatsov, line 824: I'd name this `cider-shorten-inline-errors` or `cider-shorten-inline-error-messages`. I think that the overlay is an implementation detail, that might change in the future, so it's better to focus on the essence (that we should an error message inline).

doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -207,6 +207,15 @@ bottom) with the `cider-use-overlays` variable:
 (setq cider-use-overlays nil)
 ----
 
+Overlays that indicate errors are by default trimmed of file/line/phase information.
+If you find it useful, you can customize:
```
bbatsov, line 211: Might be good to also put two error messages side by side here, so it clearer to people how the messages look in both cases.

PR #3532, opened 2023-10-16
Fix breaks with running current set of unit tests on GH

ikappaki:
Hi, 

as discussed in https://github.com/clojure-emacs/cider/pull/3461, there are a couple of unit tests that fail if ran on GH MS-Windows, which will of benefit to the `clojure-ts-mode` PR

The purpose of this PR is to fix them. We start by having a run of the tests on GH, observe their failures and fix them in a follow up commit

Thanks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,10 +86,6 @@ jobs:
     - run: npm install shadow-cljs@2.20.13 -g
     - run: npm install nbb@1.1.152 -g
 
-    - name: Test the project
-      run: |
-        eldev -p -dtTC test || eldev -p -dtTC test
-
     - name: Test integration
       run: |
         # The tests occasionally fail on macos&win in what is seems to
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,10 @@ jobs:
     - run: npm install shadow-cljs@2.20.13 -g
     - run: npm install nbb@1.1.152 -g
 
+    - name: Test the project
+      run: |
+        eldev -p -dtTC test || eldev -p -dtTC test
+
     - name: Test integration
       run: |
         # The tests occasionally fail on macos&win in what is seems to
--- a/test/cider-client-tests.el
+++ b/test/cider-client-tests.el
@@ -136,7 +136,7 @@
   (before-each
     (spy-on 'cider-connected-p :and-return-value t)
     (spy-on 'cider-classpath-entries :and-return-value
-            '("/a" "/b" "/c" "/c/inner" "/base/clj" "/base/clj-dev"))
+            '("/cider--a" "/cider--b" "/cider--c" "/cider--c/inner" "/cider--base/clj" "/cider--base/clj-dev"))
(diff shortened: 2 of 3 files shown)


ikappaki:
ok, so windows-latest, 27.2 run failed with the following two errors as expected

cider-expected-ns returns the namespace matching the given string path

  buttercup-fail("%s" "Expected `(cider-expected-ns \"/a/foo/bar/baz_utils.clj\")' to be `equal' to `\"foo.bar.baz-utils\"', but instead it was `\"a.foo.bar.baz-utils\"' which does not match because: (arrays-of-different-length 19 17 \"a.foo.bar.baz-utils\" \"foo.bar.baz-utils\" first-mismatch-at 0).")
  signal(buttercup-failed "Expected `(cider-expected-ns \"/a/foo/bar/baz_utils.clj\")' to be `equal' to `\"foo.bar.baz-utils\"', but instead it was `\"a.foo.bar.baz-utils\"' which does not match because: (arrays-of-different-length 19 17 \"a.foo.bar.baz-utils\" \"foo.bar.baz-utils\" first-mismatch-at 0).")
FAILED: Expected `(cider-expected-ns "/a/foo/bar/baz_utils.clj")' to be `equal' to `"foo.bar.baz-utils"', but instead it was `"a.foo.bar.baz-utils"' which does not match because: (arrays-of-different-length 19 17 "a.foo.bar.baz-utils" "foo.bar.baz-utils" first-mismatch-at 0).

cider-provide-file can handle multibyte characters

```
  buttercup-fail("%s" "Expected `(cider-provide-file filename)' to be `equal' to `\"8J+Nuw==\"', but instead it was `\"\"' which does not match because: (arrays-of-different-length 0 8 \"\" \"8J+Nuw==\" first-mismatch-at 0).")
  signal(buttercup-failed "Expected `(cider-provide-file filename)' to be `equal' to `\"8J+Nuw==\"', but instead it was `\"\"' which does not match because: (arrays-of-different-length 0 8 \"\" \"8J+Nuw==\" first-mismatch-at 0).")
FAILED: Expected `(cider-provide-file filename)' to be `equal' to `"8J+Nuw=="', but instead it was `""' which does not match because: (arrays-of-different-length 0 8 "" "8J+Nuw==" first-mismatch-at 0).

Ran 423 out of 424 specs, 2 failed, in 26.59s.
[00:29.344]  Found test `.feature' files: none (0)
[00:29.344]  Finished erroneously on Mon Oct 16 20:42:59 2023
[00:29.339]  2 Buttercup tests failed
Error: Process completed with exit code 1.
```

fixes to follow

ikappaki:
Everything is now passing, fixes are working, next and final step is to remove the job from GHA.

image

vemv:
Awesome - I could follow along. Thanks for operating so clearly.

ikappaki:
> Awesome - I could follow along. Thanks for operating so clearly.

Thanks. I think we are done here, the test job has been removed from the GHA. The above demonstrated the two fixes work if we are to run the tests in GH.

Issue #3541, opened 2023-10-18
cider-jack-in throws `Wrong type argument: stringp, nil` when invoked from TRAMP ssh-buffer


adrech:
## Expected behavior
In a tramp buffer connected to my rasperry pi via ssh, `cider-jack-in` creates a remote repl and connects to it.

## Actual behaviour
`cider-jack-in` creates a remote repl and throws on connect:
"error in process filter: Wrong type argument: stringp, nil"

## Steps
```
find-file /ssh:****@pi.local:~/hello/deps.edn [containing only "{}"]
M-x cider-jack-in

[nREPL] Starting server via clojure -Sdeps \{\:deps\ \{nrepl/nrepl\ \{\:mvn/version\ \"1.0.0\"\}\ cider/cider-nrepl\ \{\:mvn/version\ \"0.40.0\"\}\}\ \:aliases\ \{\:cider/nrepl\ \{\:main-opts\ \[\"-m\"\ \"nrepl.cmdline\"\ \"--middleware\"\ \"\[cider.nrepl/cider-middleware\]\"\]\}\}\} -M:cider/nrepl
[nREPL] server started on 40699
...
[nREPL] Falling back to SSH tunneled connection ...
[nREPL] Establishing SSH tunneled connection to pi.local:40699 ...
error in process filter: Wrong type argument: stringp, nil
```

## Environment & Version Information
GNU Emacs 29.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.37, cairo version 1.16.0)
CIDER 1.8.3-snapshot, nREPL 1.0.0
Clojure 1.11.1, Java 17.0.8

## Likely Reason & Attempted Solutions
The problem is that `nrepl--ssh-tunnel-connect` tries to string-match on `(buffer-file-name)` - which is `nil`, because the current execution context is the nrepl buffer:
```
;; (current-buffer) in edebug
#<buffer *nrepl-server hello:localhost*<13>>
```
Is this expected / did something change since this feature was introduced (https://github.com/clojure-emacs/cider/pull/3264)?

To confirm, I changed this in `nrepl--ssh-tunnel-connect`:
```diff
(defun nrepl--ssh-tunnel-connect (host port)
  "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
  (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
- (let* ((current-buf (buffer-file-name))
[...]

PR #3544, opened 2023-10-19
Fix `cider-jack-in` failing in remote ssh buffers and adapt the tramp example project

adrech:
> Fixes https://github.com/clojure-emacs/cider/issues/3541

- [x] All code passes the linter (`eldev lint`) 

## Fix `cider-jack-in` failing in remote ssh buffers

[Updated]
cca3e3651e5e3c4a1d455dd5f6e2f3ea062f0c1c
```
Calling cider-jack-in from a tramp buffer on a ssh-remote would throw:
"error in process filter: Wrong type argument: stringp, nil"

This happens because nrepl--ssh-tunnel-connect is trying to infer the
hostname from the current filename: It string-matches
on (buffer-file-name) which returns nil in
the *cider-uninitialized-repl* buffer.

For the jack-in use-case, the local var nrepl-project-dir will
always be bound when connecting the client, so we can use
it as a fallback.

Fixes:
https://github.com/clojure-emacs/cider/issues/3541
https://github.com/clojure-emacs/cider/issues/3303 (partially)
```

## Additional Issues

Trying `cider-jack-in` with the [tramp-sample-project](https://github.com/clojure-emacs/cider/tree/master/dev/tramp-sample-project) uncovered some issues.

I followed the [README](https://github.com/clojure-emacs/cider/blob/master/dev/tramp-sample-project/README.md), was able to start a repl and connect to it as described:
```
* In one terminal tab, run `make run` to run the Docker image
* Once it's ready, from another tab, run `make ssh` and start a repl manually from there
...
```

but had no luck with `M-x cider-jack-in` from `(dired "/sshx:root@localhost#8022:/usr/src/app")`:

### `M-x cider-jack-in` in `/usr/src/app` failed at server creation: `java` not in `PATH`
```
Tramp: Opening connection nrepl-server for root@localhost using sshx...done
[nREPL] Starting server via lein update-in :dependencies conj \[nrepl/nrepl\ \"1.0.0\"\] -- update-in :plugins conj \[cider/cider-nrepl\ \"0.40.0\"\] -- repl :headless :host localhost
error in process sentinel: let: Could not start nREPL server: /usr/local/bin/lein: line 224: type: java: not found
Leiningen couldn't find 'java' executable, which is required.
Please either set JAVA_CMD or put java (>=1.6) in your $PATH (/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin).
 ("exited abnormally with code 1")
```

But when manually connecting through `make ssh`, `java` is available (and there's some duplication going on):

```
make ssh
...
root@747a04bdfcfa:~# echo $PATH
/opt/java/openjdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin/
```

When shelling out in the remote dired buffer, the envars that are written to `.bashrc` at the bottom of the [Dockerfile](https://github.com/clojure-emacs/cider/blob/master/dev/tramp-sample-project/Dockerfile) (e.g. `LEIN_ROOT`) seem to be picked up correctly, but `PATH` is different:
```
(dired "/sshx:root@localhost#8022:/usr/src/app")
(async-shell-command "echo $LEIN_ROOT")
;; => 1
(async-shell-command "echo $PATH")
;; => /bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
```

Where are the extra `PATH` elements coming from when using `make ssh`? It's just
```
ssh:
	ssh-keygen -R "[localhost]:8022"
	echo "Password is: cider"
	ssh root@localhost -p 8022

```
#### Wild guess
I stumbled over a comment in `start-file-process-shell-command` (used by `nrepl-start-server-process`):
 ```
 ;; On remote hosts, the local `shell-file-name' might be useless.
 ```
Maybe some `sh` vs `bash` thing going on here? But why are the `.bashrc` envars working then...?
Giving  up for now, because my Docker experience is limited.

#### Workaround
Adding `$JAVA_CMD` - as suggested by `lein` - made it work and the remote repl would spin up,
see: 442bdc353783a851c6bb7ad20a101465c9989979

### The remote filename was not detected properly by `nrepl--ssh-tunnel-connect`

With `"/sshx:root@localhost#8022:/usr/src/app"`, the `tramp-file-regex`
- did not match "sshx"
- and its host capture group ("`localhost`") did not match the `host` arg (`"localhost#8022"`)

**DANGER**: These are corrected in e3c933df722316db37e9e14b598e0a18cba4233d, but I am unsure if including the `#...` part in the host capture group is the right thing to do here and not breaking anything?

#### Cosmetics and Documentation

5e41923e76cebd280c5536db60bfd91aefb052d7
cc2ae8d978b77090a83808d4b920896f5e4fe5ff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Bugs fixed
 
-- [#3544](https://github.com/clojure-emacs/cider/pull/3544): Fix cider-jack-in failing with ssh-remotes #3541 #3303
+- [#3541](https://github.com/clojure-emacs/cider/issues/3541): Fix `cider-jack-in` failing with SSH remotes.
 - [#3559](https://github.com/clojure-emacs/cider/issues/3559): Don't apply [dynamic syntax highlighting](https://docs.cider.mx/cider/config/syntax_highlighting.html) over buffers belonging to unrelated Sesman sessions.
 
 ## 1.9.0 (2023-10-24)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
+- [#3544](https://github.com/clojure-emacs/cider/pull/3544): Fix cider-jack-in failing with ssh-remotes #3541 #3303
 - [#3559](https://github.com/clojure-emacs/cider/issues/3559): Don't apply [dynamic syntax highlighting](https://docs.cider.mx/cider/config/syntax_highlighting.html) over buffers belonging to unrelated Sesman sessions.
 
 ## 1.9.0 (2023-10-24)
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -271,6 +271,12 @@ Discards it if it can be determined that the port is not active."
                           (shell-command-to-string (format "lsof -i:%s" port-number))))
           port-string)))))
 
+(defun nrepl--ssh-file-name-matches-host-p (file-name host)
+  "Return t, if FILE-NAME is a tramp-file-name on HOST via ssh."
+  (when (tramp-tramp-file-p file-name)
+    (with-parsed-tramp-file-name file-name v
+      (and (member v-method '("ssh" "sshx" "sshfs"))
+           (member host (list v-host (concat v-host "#" v-port)))))))
 
 ;;; Bencode
 
@@ -587,14 +593,12 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
 (defun nrepl--ssh-tunnel-connect (host port)
   "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
   (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
-  (let* ((current-buf (buffer-file-name))
-         (tramp-file-regexp "/ssh:\\(.+@\\)?\\(.+?\\)\\(:\\|#\\).+")
+  (let* ((file-name (or (buffer-file-name) nrepl-project-dir))
          (remote-dir (cond
                       ;; If current buffer is a TRAMP buffer and its host is
                       ;; the same as HOST, reuse its connection parameters for
                       ;; SSH tunnel.
-                      ((and (string-match tramp-file-regexp current-buf)
-                            (string= host (match-string 2 current-buf))) current-buf)
+                      ((nrepl--ssh-file-name-matches-host-p file-name host) file-name)
                       ;; Otherwise, if HOST was provided, use it for connection.
                       (host (format "/ssh:%s:" host))
                       ;; Use default directory as fallback.
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -275,7 +275,7 @@ Discards it if it can be determined that the port is not active."
   "Return t, if FILE-NAME is a tramp-file-name on HOST via ssh."
   (when (tramp-tramp-file-p file-name)
     (with-parsed-tramp-file-name file-name v
-      (and (member v-method '("ssh" "sshx" "sshfs"))
+      (and (member v-method '("ssh" "sshx"))
            (member host (list v-host (concat v-host "#" v-port)))))))
 
 ;;; Bencode
--- a/test/nrepl-client-tests.el
+++ b/test/nrepl-client-tests.el
@@ -158,6 +158,30 @@
         (expect (match-string 1 msg)
                 :to-equal "nrepl.sock"))))
 
+(describe "nrepl--ssh-file-name-matches-host-p"
+  (it "works in the most basic case"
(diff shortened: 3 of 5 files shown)


nrepl-client.el
```diff
@@ -584,17 +585,19 @@ If NO-ERROR is non-nil, show messages instead of throwing an error."
...
-  "Connect to a remote machine identified by HOST and PORT through SSH tunnel."
+(defun nrepl--ssh-tunnel-connect (host port &optional jack-in-buffer)
+  "Connect to a remote machine identified by HOST and PORT through SSH tunnel.
+If JACK-IN-BUFFER is given and a tramp buffer, reuse its connection parameters.
+ Defaults to the current buffer."
   (message "[nREPL] Establishing SSH tunneled connection to %s:%s ..." host port)
-  (let* ((current-buf (buffer-file-name))
-         (tramp-file-regexp "/ssh:\\(.+@\\)?\\(.+?\\)\\(:\\|#\\).+")
+  (let* ((file-name (buffer-file-name (or jack-in-buffer (current-buffer))))
+         (tramp-file-regexp "/sshx?:\\(.+@\\)?\\(.+?\\):.+")
```
vemv, line 594: `#` refers to the port - one can use a non-standard one.

Is anything necessary at the moment for supporting that use case?

vemv (commented):
Looking great - thanks for the detailed PR!

nrepl-client.el
adrech: I assume not - this change makes the ssh port an (optional) part of the hostname instead of throwing it away. But I've only tested this against the example project:

On `nrepl--ssh-tunnel-connect` invocation,
HOST is `"localhost#8022"`and PORT `35371`.
So PORT is the nrepl-port and `#port` will always be part of HOST in that context.

vemv: Awesome, thanks!

If you have it at hand, please extract `(and (string-match tramp-file-regexp file-name) (string= host (match-string 2 file-name)))` to a defun with a nice docstring and unit test (that has at least two cases: no port, custom port)

vemv:
Alright! The code looks reasonable as-is, but I'll be curious about the improvement.

adrech:
> Alright! The code looks reasonable as-is, but I'll be curious about the improvement.

When coming from `cider-jack-in`, we can fall back to `nrepl-project-dir` which should always be set at that point.
This fixes the original issue while (hopefully) leaving other use-cases untouched, where the execution context is different.

Updated the branch.


Outside the scope of this PR:
- when using `cider-connect-clj` the user gets prompted for a hostname that might not be used, because the inference from the tramp-file-name happens afterwards. Confusing?
- comparing `host` with the filename might not work for babashka, because it gets the host differently (from the server log https://github.com/clojure-emacs/cider/blob/95a19f09999d57d0c78d8a01795a455b6501e042/nrepl-client.el#L1186C34-L1186C34)

test/nrepl-client-tests.el
```diff
@@ -158,6 +158,30 @@
...
+            :to-be-truthy)
+    (expect (nrepl--ssh-file-name-matches-host-p
+             "/ssh:tester@host#8022:~/test/" "host#7777")
+            :to-be nil))
+  (it "works with tramps other ssh methods"
+    (expect (nrepl--ssh-file-name-matches-host-p
+             "/sshx:tester@host:~/test/" "host")
+            :to-be-truthy)
+    (expect (nrepl--ssh-file-name-matches-host-p
+             "/sshfs:tester@host:~/test/" "host")
```
vemv, line 179: Feel free to skip this test on Emacs 27.

IDK why it only fails on 27 - maybe it had a transient issue?

adrech: I quickly checked with Emacs 27 and it's a real issue. Not sure whats going on, but I just cut "sshfs" functionality for now. That wasn't supported before anyways.

vemv:
LGTM - thanks much!

Can you confirm that the code works as-is for connecting to your Raspberry?

adrech:
Confirmed. Also checked it with the docker build in the tramp example project.

vemv:
Perfect! Please merge master in, and reflect the fix in the changelog

adrech:
Done

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### Bugs fixed
 
+- [#3544](https://github.com/clojure-emacs/cider/pull/3544): Fix cider-jack-in failing with ssh-remotes #3541 #3303
```
vemv, line 17:
```suggestion
- [#3541](https://github.com/clojure-emacs/cider/issues/3541): Fix `cider-jack-in` failing with SSH remotes.
```

vemv:
🍻

Expect a new snapshot within a couple hours

Issue #3546, opened 2023-10-20
Inspector: (feature request) Syntax highlighting for the java methods


anonimitoraf:
First of all, I want to say that the CIDER inspector is awesome! Can't live without it.

**Is your feature request related to a problem? Please describe.**

I find it hard to grok the methods list when they are long, have a lot of arguments and/or there are a lot of them.

For example:

<img width="600" alt="image" src="https://github.com/clojure-emacs/cider/assets/15933322/4db35046-2773-4e94-bef8-3f37b151ad87">

**Describe the solution you'd like**

It would be nice to have Java syntax highlighting for the method names, arguments, etc.

**Describe alternatives you've considered**

I currently have a very (very) hacky solution that adds highlighting via overlays

```elisp
(defvar ++ov-cider-inspector-keywords)
(defvar ++ov-cider-inspector-method-names)
(defvar ++ov-cider-inspector-method-param-lists)
(defun ++ov-cider-inspector ()
  (ov-clear)
  (let ((case-fold-search nil))
    (setq ++ov-cider-inspector-keywords (ov-regexp "^public\\|private\\|protected\\|final\\|native\\|static$")
     ++ov-cider-inspector-method-names (ov-regexp "\\([a-zA-Z0-9.$]\\)+(")
      ++ov-cider-inspector-method-param-lists (ov-regexp "(\\([][<>a-zA-Z0-9.$,]*\\))?")))
  (ov-set ++ov-cider-inspector-keywords 'face `(:foreground ,(doom-color 'magenta)))
  (ov-set ++ov-cider-inspector-method-names 'face `(:foreground ,(doom-color 'red)))
  (ov-set ++ov-cider-inspector-method-param-lists 'face `(:foreground "white")))
```

but even this gives me something very basic:

<img width="600" alt="image" src="https://github.com/clojure-emacs/cider/assets/15933322/02b03562-2fbd-4f1b-a336-20a6723c8205">

vemv:
This is the snippet I use for similar use cases:

```lisp
(defun vemv/to-java-string (s)
  "Returns a propertized string with Java syntax coloring"
  (with-temp-buffer
    (insert s)
    (java-mode)
    (font-lock-ensure)
    (buffer-string)))
```

Feel free to give it a try.

Besides, I noticed in your screenshot another point of improvement. Created https://github.com/clojure-emacs/orchard/issues/198

anonimitoraf:
[...]

PR #3547, opened 2023-10-21
Inspector: render Java items using `java-mode` syntax coloring

vemv:
Fixes https://github.com/clojure-emacs/cider/issues/3546

Demo:

image

Cheers - V

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -427,23 +427,54 @@ MAX-COLL-SIZE if non nil."
         (error (insert "\nInspector error for: " str))))
     (goto-char (point-min))))
 
+(defvar cider-inspector-looking-at-java-p nil)
+
 (defun cider-inspector-render* (elements)
   "Render ELEMENTS."
+  (setq cider-inspector-looking-at-java-p nil)
   (dolist (el elements)
     (cider-inspector-render-el* el)))
 
+(defun cider--to-java-string (s)
+  "Given `S`, returns a propertized string with Java syntax coloring."
+  (with-temp-buffer
+    (insert s)
+    (java-mode)
+    (font-lock-ensure)
+    (buffer-string)))
+
+(defconst cider--inspector-java-headers
+  '("--- Interfaces:" "--- Constructors:" "--- Fields:" "--- Methods:" "--- Imports:"))
+
 (defun cider-inspector-render-el* (el)
   "Render EL."
-  (cond ((symbolp el) (insert (symbol-name el)))
-        ((stringp el) (insert (propertize el 'font-lock-face 'font-lock-keyword-face)))
-        ((and (consp el) (eq (car el) :newline))
-         (insert "\n"))
-        ((and (consp el) (eq (car el) :value))
-         (cider-inspector-render-value (cadr el) (cl-caddr el)))
-        (t (message "Unrecognized inspector object: %s" el))))
+  (let ((header-p (or (member el cider--inspector-java-headers)
+                      (and (stringp el)
+                           (string-prefix-p "--- " el)))))
+    ;; Headers reset the Java syntax coloring:
+    (when header-p
+      (setq cider-inspector-looking-at-java-p nil))
+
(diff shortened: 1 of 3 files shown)


cider-inspector.el
```diff
@@ -427,23 +427,54 @@ MAX-COLL-SIZE if non nil."
...
+                           (string-prefix-p "--- " el)))))
+    ;; Headers reset the Java syntax coloring:
+    (when header-p
+      (setq cider-inspector-looking-at-java-p nil))
+
+    (cond ((symbolp el) (insert (symbol-name el)))
+          ((stringp el) (insert (if cider-inspector-looking-at-java-p
+                                    (cider--to-java-string el)
+                                  (propertize el 'font-lock-face (if header-p
+                                                                     'font-lock-comment-face
```
vemv, line 462: Headers now are rendered as comments - else it's all too homogeneous.

rrudakov:
Starting from emacs-29, build-in `java-ts-mode` is also available. Maybe this could be customizable, to let users pick the desired mode?

vemv:
Good eye!

It seems most desirable to simply pick java-ts-mode if available, improving performance I guess (or in any case having color consistency, in case there's a discrepancy between both modes. Assuming that the user uses java-ts-mode regularly).

Would you be able to try this diff locally and confirm that it works well with java-ts-mode?

rrudakov:
> Good eye!
> 
> It seems most desirable to simply pick java-ts-mode if available, improving performance I guess (or in any case having color consistency, in case there's a discrepancy between both modes. Assuming that the user uses java-ts-mode regularly).
> 
> Would you be able to try this diff locally and confirm that it works well with java-ts-mode?

Sure, I can try it out tomorrow. The only concern I have is that emacs can be built without the tree-sitter support or the grammar can be missing, I guess `java-ts-mode` won't work properly in this case.

vemv:
Is that condition detectable?

rrudakov:
@vemv, yes, it's possible:

```emacs-lisp
(require 'treesit)
(treesit-available-p)                   ;check if Emacs is built with tree-sitter library
(treesit-language-available-p 'java)    ;Make sure Emacs can find the language grammar you want to use
```

bbatsov:
Given the simple nature of this highlighting in the inspector, it seems to me that using `java-mode` unconditionally will be perfectly fine. 

I'd suggest a defcustom to enable/disable this, and perhaps a second one about the preferred java-mode to use, if someone feels strongly that using `java-ts-mode` will result in any meaningful improvements.

cider-inspector.el
```diff
@@ -427,23 +427,54 @@ MAX-COLL-SIZE if non nil."
...
 
+(defvar cider-inspector-looking-at-java-p nil)
+
 (defun cider-inspector-render* (elements)
   "Render ELEMENTS."
+  (setq cider-inspector-looking-at-java-p nil)
   (dolist (el elements)
     (cider-inspector-render-el* el)))
 
+(defun cider--to-java-string (s)
```
bbatsov, line 438: I think we had a similar function in `cider-util.el` already that you can reuse (it takes the mode to use as its param).

vemv:
@rrudakov although your suggestion was a fair call, it seems best addressed later, and also backed by a specific problem statement (not that I don't like the idea, but we'd like to see some performance or presentational difference first).

My main reasoning is that we now have 5 `cider-font-lock-as 'java-mode` calls, so it would be a broader refactoring than initially thought.

rrudakov:
vemv ***@***.***> writes:

> @rrudakov although your suggestion was a fair call, it seems best 
> addressed later, and also backed by a specific problem statement 
> (not that I don't like the idea, but we'd like to see some 
> performance or presentational difference first).
>
> My main reasoning is that we now have 5 `cider-font-lock-as 
> 'java-mode` calls, so it would be a broader refactoring than 
> initially thought.

Sure, sounds fair. Thanks for considering this anyway.
--
Best regards, Roman

Issue #3548, opened 2023-10-23
add new  function `cider-inspector-tap-current-val `


behrica:
completes the work started on:

https://github.com/clojure-emacs/orchard/pull/167
https://github.com/clojure-emacs/cider-nrepl/issues/770

PR will follow

behrica:
fixed by PR #3549

Issue #3311, opened 2023-01-24
Allow to tap> current value from inside cider inspector


behrica:
**Is your feature request related to a problem? Please describe.**
I would like to visualize graphical parts of a complex data structure using the current value
of the `cider-inspector`,  so combining the simple drilling into the data with a way to render
on key-press  the data in a rich, graphical way

**Describe the solution you'd like**
As "visualizing of Clojure data"  has a multitude of different solutions, I would like 
to see an generic solution, which allows to easily `tap>` the current value in cider-inspector.
Then the data renderer could be plugged in as a tap handler
So maybe adding a function such as `cider-inspector-tap-current-value`

**Describe alternatives you've considered**
Using `cider-def-current-value` can be used but requires "2 steps",  namely define a var and then tap the var.
I could as well write ad-hoc code which drills down and taps, but using the cider-inspector for drill down is more comfortable.

**Additional context**
There was an inn initial discussion and some proof of concept code on Slack:
https://app.slack.com/client/T03RZGPFR/C0617A8PQ/thread/C0617A8PQ-1674486881.210099

To fully solve the problem, we would need something else.
The "data-rendered" available for Clojure  (example Clerk) cannot know  "how to render data".
They expect, that a human user provide them with some "help", how to render. In Clerk this is called "viewer selection".

So ideally we would have a interactive function in Cider which "request" this information from the user and sens it together with the value to "tap>" , as part of the `cider-inspector-tap-current-value`
This information could be probably a simple "string", a kind of "viewer id", which options could come from a list, which the user of cider can configure in a variable.

Without this feature, Clojure code would need to "guess" the viewer from the data, which is very likeley not working well.

The code linked above contains as well a proof of concept of such function:
`clerk-tap-last-sexp-with-viewer`

This does the tapping of the "last sexp", but the same could be applied for `cider-inspector-tap-current-value`

behrica:
Just to add to the "viewer selection":
A "static list" is not the "best" solution, even-though being a very good start.

At least in Clerk , viewers can be added and removed at runtime,
so ideally the lisp function should be calling back into Clojure, to ask for the "current list of viewer ids".
To have this generic, the user should be able to configure which is the Clojure call back function to be called for retrieving the "list of viewers".

I have experimented with the viewer selection (with fixed list), and it is very comfortable.
[...]

PR #3549, opened 2023-10-23
add new inspector function to tap current value

behrica:
> Fixes https://github.com/clojure-emacs/cider/issues/3548

add new interactive function  `cider-inspector-tap-current-val` to inspector.

It sends the current inspector value to the tap>.
It uses a new middleware op,  `inspect-tap-current-value` which has been added recently to cider-nrepl.

-----------------

- [ x] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
- [ x] You've updated the [user manual](../blob/master/doc) (if adding/changing user-visible functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Changes
 
 - [#3546](https://github.com/clojure-emacs/cider/issues/3546): Inspector: render Java items using `java-mode` syntax coloring.
+- [#3548](https://github.com/clojure-emacs/cider/issues/3548)
+Inspector: add function to tap> the current inspector value
 
 ### Bugs fixed
 
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -308,7 +308,7 @@ current-namespace."
     (message "%s#'%s/%s = %s" cider-eval-result-prefix ns var-name value)))
 
 (defun cider-inspector-tap-current-val ()
-  "Sends current value to tap>"
+  "Sends current value to tap>."
   (interactive)
   (setq cider-inspector--current-repl (cider-current-repl))
   (when-let* ((value (cider-sync-request:inspect-tap-current-val)))
@@ -378,7 +378,7 @@ MAX-SIZE is the new value."
                 (nrepl-dict-get "value")))
 
 (defun cider-sync-request:inspect-tap-current-val ()
-  "Sends current inspector value to tap>"
+  "Sends current inspector value to tap>."
   (thread-first `("op" "inspect-tap-current-value")
                 (cider-nrepl-send-sync-request cider-inspector--current-repl)
                 (nrepl-dict-get "value")))
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -307,6 +307,14 @@ current-namespace."
     (cider-inspector--render-value value)
     (message "%s#'%s/%s = %s" cider-eval-result-prefix ns var-name value)))
 
+(defun cider-inspector-tap-current-val ()
+  "Sends current value to tap>"
+  (interactive)
+  (setq cider-inspector--current-repl (cider-current-repl))
+  (when-let* ((value (cider-sync-request:inspect-tap-current-val)))
+    (cider-inspector--render-value value)
+    (message "%s# tapped %s" cider-eval-result-prefix value)))
+
 ;; nREPL interactions
 (defun cider-sync-request:inspect-pop ()
   "Move one level up in the inspector stack."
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Changes
 
 - [#3546](https://github.com/clojure-emacs/cider/issues/3546): Inspector: render Java items using `java-mode` syntax coloring.
+- [#3548](https://github.com/clojure-emacs/cider/issues/3548)
+Inspector: add function to tap> the current inspector value
```
vemv, line 9:
```suggestion
- [#3548](https://github.com/clojure-emacs/cider/issues/3548) Inspector: add function to tap> the current inspector value
```

doc/modules/ROOT/pages/debugging/inspector.adoc
```diff
@@ -60,6 +60,10 @@ You'll have access to additional keybindings in the inspector buffer
 
 | kbd:[d]
 | Defines a var in the REPL namespace with current inspector value. If you tend to always choose the same name(s), you may want to set the `cider-inspector-preferred-var-names` customization option.
+
+| kbd:[t]
```
vemv, line 64: (I don't see this binding in the PR diff yet)

vemv (commented):
Looking good

doc/modules/ROOT/pages/debugging/inspector.adoc
behrica: just added it

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changes
 
 - [#3546](https://github.com/clojure-emacs/cider/issues/3546): Inspector: render Java items using `java-mode` syntax coloring.
+- [#3548](https://github.com/clojure-emacs/cider/issues/3548) Inspector: add function to tap> the current inspector value
```
vemv, line 8:
```suggestion
- [#3548](https://github.com/clojure-emacs/cider/issues/3548): Inspector: add function to tap> the current inspector value.
```

cider-inspector.el
```diff
@@ -307,6 +308,14 @@ current-namespace."
     (cider-inspector--render-value value)
     (message "%s#'%s/%s = %s" cider-eval-result-prefix ns var-name value)))
 
+(defun cider-inspector-tap-current-val ()
+  "Sends current value to tap>."
+  (interactive)
+  (setq cider-inspector--current-repl (cider-current-repl))
+  (when-let* ((value (cider-sync-request:inspect-tap-current-val)))
+    (cider-inspector--render-value value)
+    (message "%s# tapped %s" cider-eval-result-prefix value)))
```
vemv, line 317: I reckon that `value` could be huge, leaving the echo area difficult to read.

I'd suggest instead: `(message "%s# tapped the value at %s" cider-eval-result-prefix path)`

(where path looks like `:foo :bar (nth 20)`)

bbatsov:
Might be good to expand on this a bit in the debugging techniques/inspector part of the docs.

cider-inspector.el
behrica: I commited a change which does that.

behrica:
I added in docu a TIP to tap values from inspector in order to render them graphically using other tools.
This is at least my main use case.

behrica:
I would have one more question on this PR.
My "initial use case" requires as well that the "tapped values"  needed to be "transformed" by Clojure code before they get tapped.

This gets complicated by the fact, that the user would need to give "feedback" when this happens.

The example is "having a map deep in a value which represents 'vega-lite'  data"
Clerk has a special vega-lite render, which can be "selected" by "wrapping" the value with some `:meta` before tapping it.
The point is that the user needed to be "asked" before tapping "what type of data" its is (= vega lite). Tihs cannot be inferred from the data as such.
Maybe this becomes more clear by look at an adhoc function I did for this purpose:
https://github.com/behrica/clerk-mode/blob/bff0aa274eae0923b6b0f3b64d32ab1fa3333c7f/clerk-mode.el#L80

We had as well a discussion on this here: https://github.com/clojure-emacs/cider/issues/3311

Do you think the current code  of this PR allows this easely, as an extension by using "advice" for example ?

vemv:
I got these changed into https://github.com/clojure-emacs/cider/commit/966cae463c65d0392d1148eb252912f31c03c881 preserving attribution.

Thanks much!

> Do you think the current code of this PR allows this easely, as an extension by using "advice" for example ?

Probably not, but feel free to give it a try.

A clean solution would be to make the cider-nrepl accept a `transformer` function (fully-qualified symbol) - something like that?

Users could bind a specific key in the inspector map to perform `cider-inspector-tap-current-val` passing that extra argument.

Lastly, I dropped the path code, it turned out to excessively complicate things. It would be easier if cider-nrepl simply returned a (well-formed!) :path as part of every reponse that included a `:value`.

PR #3550, opened 2023-10-24
Add option to skip ssh hosts lookup on cider-connect

nivekuil:
This slows down connect and causes serious annoyances if you have an `authinfo.gpg` and the pinentry path breaks for some reason (it often does)

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 - [#3546](https://github.com/clojure-emacs/cider/issues/3546): Inspector: render Java items using `java-mode` syntax coloring.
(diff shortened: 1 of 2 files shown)


vemv:
Mmm, I'd rather have a report for those annoyances, and have them eventually fixed.

Otherwise we try to avoid an abundance of defcustoms.

A nice Emacs-y way to have a workaround in the meantime is to `advide-add` the defun so that it returns nil.

Thanks - V

vemv:
I'll close the PR for that reason for the time being, but as mentioned an issue would be most welcome - perhaps a fix would not be that hard.

Issue #3519, opened 2023-10-13
new command  `cider-copy-jack-in`


behrica:
**Is your feature request related to a problem? Please describe.**
Using CIDER on a remote machine would be eased in complex remote scenarious by:
- open a shell in Emacs and "go ito the final remote server"
- generate teh "jack-in-command' and "copy" it
- paste it into remote shell

**Describe the solution you'd like**
A Cider build in commnad such as:
```clojure
(defun cider-copy-jack-in (params)
  "docstring"
  (interactive "P")
  (let ((params (thread-first params
                  (cider--update-project-dir)
                  (cider--update-jack-in-cmd))))
    (kill-new (plist-get params :jack-in-cmd))))
```
but full-featured, in case tghis command does not a consider all cases

**Describe alternatives you've considered**
`cider-connect` and picking a remote host. This does only work for 2simple" ssh connection, but not for "complex" connections (such as multihop, or remote+docker)

**Additional context**
There was a discussion here:
https://clojurians.slack.com/archives/C0617A8PQ/p1688529209320339

dpsutton:
some design considerations:
should allow for the cljs form, you might also need the cljs startup forms (shadow/select :app , etc)

PR #3555, opened 2023-10-26
introduce cider-start-server command

behrica:
Split  cider-jack-in in two and therefore obtain  interactive function cider-start-server,
which starts the nrpel server without trying to connect afterwards

see here:
https://clojurians.slack.com/archives/C0617A8PQ/p1698268379712609?thread_ts=1698268132.937489&cid=C0617A8PQ

-----------------

- [ x] All tests are passing (`eldev test`)
- [ ]x All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes

Thanks!

manual][1] extremely useful.*


cider.el
```diff
@@ -1338,11 +1350,23 @@ double prefix prompt for all these parameters."
...
+    (cider--do-start-server params
+                        (lambda (server-buffer)
+                          (cider-connect-sibling-clj params server-buffer)))))
+
+
+(defun cider-start-server (params)
+   "Start an nREPL server for the current project.
+PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
+With the prefix argument, allow editing of the start server in command; with a
+double prefix prompt for all these parameters."
```
vemv, line 1362:
```suggestion
   "Start an nREPL server for the current project, but don't connect to it.
PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
With the prefix argument, allow editing of the start server in command; with a
double prefix prompt for all these parameters.

See also: `cider-jack-in-clj'."
```

cider.el
```diff
@@ -1326,6 +1326,18 @@ nil."
     map)
   "CIDER jack-in and connect keymap.")
 
+
+(defun cider--do-start-server (params on-port-callback)
+  "Starts nrepl server and passes the call-back to it.
+PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
+ON-PORT-CALLBACK is a function of one argument (server buffer)
+which is called by the process filter once the port of the connection has
+been determined.  Can be nil."
```
vemv, line 1335:
```suggestion
been determined.  Can be nil.

See also: `cider-start-server'."
```

vemv (commented):
Looking good!

Can you confirm that this change does accomplish your end goal?

behrica:
> Looking good!
> 
> Can you confirm that this change does accomplish your end goal?

To a large extend.
I can start now the nREPL server without bothering with the dependencies.

And then an easy cider connect using host and port I know anyhow (as I fix the port in order to have tunneling working)

Something similar could be reached as well by:
https://github.com/clojure-emacs/cider/issues/3519

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 ## 1.9.0 (2023-10-24)
 
 ### New features
-
+- new command `cider-start-server` which does the same as `cider-jack-in` but without trying to connect to the started nREPL server
```
vemv, line 8: Please place this under `## master (unreleased)`, and as `### Changes`

cider.el
```diff
@@ -1326,6 +1326,20 @@ nil."
     map)
   "CIDER jack-in and connect keymap.")
 
+
+(defun cider--do-start-server (params on-port-callback)
+  "Starts nrepl server and passes the call-back to it.
```
vemv, line 1331:
```suggestion
  "Starts an nrepl server and passes the callback to it.
```

cider.el
```diff
@@ -1338,11 +1352,25 @@ double prefix prompt for all these parameters."
...
+With the prefix argument, allow editing of the start server in command; with a
+double prefix prompt for all these parameters.
+
+See also: `cider-jack-in-clj'."
+  (interactive "P")
+  (let ((params (thread-first
+                  params
+                  (cider--update-project-dir)
+                  (cider--check-existing-session)
+                  (cider--update-jack-in-cmd))))
```
vemv, line 1372: This thread computation is duplicated, please extract it to a private function

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -267,6 +267,18 @@ Here is an example Nbb Jack-In command, providing a custom `:jack-in-cmd`.
...
+in which CIDER cannot reliably detect to which server:port to connect and
+would therefore fail.
+This assumes that the user executes a `cider-connect` command manually afterwards
+and specifies host:port .
+
+For this there is the `cider-start-server` command, which optionally takes the same
+parameters then `cider-jack-in`.
+
 
 == Connect to a Running nREPL Server
```
vemv, line 283:
```suggestion
==== Starting nREPL server without trying to connect to it ====

In some situations, it might be useful to only start a nREPL server process, without
connecting to it. This can support complex setups
for which CIDER cannot reliably detect to which server/port to connect, and
would therefore fail.
This assumes that the user will execute a `cider-connect` command manually afterwards,
specifying host/port.

For this scenario, the `cider-start-server` command is offered, which optionally takes the same
parameters than `cider-jack-in`.

Sample usage:

[source,lisp]
----
M-x cider-start-server
----

== Connect to a Running nREPL Server
```

cider.el
```diff
@@ -1326,6 +1326,20 @@ nil."
...
 
+
+(defun cider--do-start-server (params on-port-callback)
+  "Starts nrepl server and passes the call-back to it.
+PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
+ON-PORT-CALLBACK is a function of one argument (server buffer)
+which is called by the process filter once the port of the connection has
+been determined.  Can be nil.
+
+See also: `cider-start-server'."
```
vemv, line 1337: Ah sorry, I meant the this `See also` bit to go in cider-jack-in-clj's docstring

cider.el
```diff
@@ -1326,23 +1326,51 @@ nil."
...
+                  params
+                  (cider--update-project-dir)
+                  (cider--check-existing-session)
+                  (cider--update-jack-in-cmd))))
+    (cider--do-start-server params
+                            (lambda (server-buffer)
+                              (cider-connect-sibling-clj params server-buffer)))))
+
+
+(defun cider-start-server (params)
```
bbatsov, line 1360: Probably we should name this `cider-start-nrepl-server` (or at least `cider-start-repl-server`, so it's clearer what it is for. And there's also the chance CIDER will support some other type of server down the road.

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 # Changelog
 
 ## master (unreleased)
+- new command `cider-start-server` which does the same as `cider-jack-in` but without trying to connect to the started nREPL server
```
bbatsov, line 4: You should add this under `### New features` and the changelog entry should have the save format as other entries in it.

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
...
+
 
 ## 1.9.0 (2023-10-24)
 
 ### New features
-
```
bbatsov, line 8: I assume you deleted this blank line by mistake.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -267,6 +267,24 @@ Here is an example Nbb Jack-In command, providing a custom `:jack-in-cmd`.
...
+would therefore fail.
+This assumes that the user will execute a `cider-connect` command manually afterwards,
+specifying host/port.
+
+For this scenario, the `cider-start-server` command is offered, which optionally takes the same
+parameters than `cider-jack-in`.
+
+Sample usage:
+
+[source,lisp]
```
bbatsov, line 284: That's not really Lisp code, so it's better to use `kbd` macro instead.

cider.el
```diff
@@ -1326,23 +1326,51 @@ nil."
     map)
   "CIDER jack-in and connect keymap.")
 
+
+(defun cider--do-start-server (params on-port-callback)
```
bbatsov, line 1330: I don't quite get the name `do-start-server`, so I'd suggest to clarify it a bit (I understand why this function exists, I just don't think the name is optimal).

CHANGELOG.md
behrica: Not sure, what you meant with same format, but I change it to be under new features

doc/modules/ROOT/pages/basics/up_and_running.adoc
behrica: I removed it even. I think it's enough to mention the command name.

cider.el
behrica: renamed it to `cider--start-nrepl-server-with-callback`

cider.el
```diff
@@ -1326,24 +1326,53 @@ nil."
     map)
   "CIDER jack-in and connect keymap.")
 
+
+(defun cider--start-nrepl-server-with-callback (params on-port-callback)
```
bbatsov, line 1330: I'm guessing it'd be better if you made the callback optional, given it is optional. Then this is also not needed in the function's name.

cider.el
```diff
@@ -1326,24 +1326,53 @@ nil."
...
+
 ;;;###autoload
 (defun cider-jack-in-clj (params)
   "Start an nREPL server for the current project and connect to it.
 PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
 With the prefix argument, allow editing of the jack in command; with a
-double prefix prompt for all these parameters."
+double prefix prompt for all these parameters.
+
+See also: `cider-start-server'."
```
bbatsov, line 1348: I'm guessing you'll want to mention this in the other jack-in commands if you've mentioned it here. Or in none of them, which will also be consistent IMO.

bbatsov:
I think you can also add some keybinding for this in the jack-in keymap and mention in the relevant command menu.

cider.el
vemv: Please address this duplication:

```
 (cider--update-project-dir)
(cider--check-existing-session)
(cider--update-jack-in-cmd)
```

behrica:
If / when #3563 get decided on and this is merged, we could slightly change the docu again.
Because a user has the basically three options when working with remote hosts, in decreasing order of comfort
and increasing guarantee of success

1. Rely on `cider-jack-in` and its configurations to heuristically do the correct thing for start and connect
2. `cider-nrepl-start-server` followed by `cider-connect`
3. Manual start of nREPL , ev. manual port tunneling, and `cider-connect`

cider.el
```diff
@@ -1318,6 +1318,7 @@ nil."
     (define-key map (kbd "C-c C-s") #'cider-connect-cljs)
     (define-key map (kbd "C-c C-m") #'cider-connect-clj&cljs)
     (define-key map (kbd "s j") #'cider-connect-sibling-clj)
+    (define-key map (kbd "s n") #'cider-start-nrepl-server)
```
bbatsov, line 1321: You should also add the variants with Control - see the keybindings below. Basically we have 3 options for each keybinding for the preferences of different users.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -267,6 +267,18 @@ Here is an example Nbb Jack-In command, providing a custom `:jack-in-cmd`.
...
+==== Starting nREPL server without trying to connect to it ====
+
+In some situations, it might be useful to only start a nREPL server process, without
+connecting to it. This can support complex setups
+for which CIDER cannot reliably detect to which server/port to connect, and
+would therefore fail.
+This assumes that the user will execute a `cider-connect` command manually afterwards,
+specifying host/port.
+
+For this scenario, the `cider-start-nrepl-server` command is offered, which optionally
```
bbatsov, line 279: You might want to mention the keybindings for the command here.

cider.el
```diff
@@ -1326,24 +1329,50 @@ nil."
...
        (cider-connect-sibling-clj params server-buffer)))))
 
+
+(defun cider-start-nrepl-server (params)
+  "Start an nREPL server for the current project, but don't connect to it.
+PARAMS is a plist optionally containing :project-dir and :jack-in-cmd.
+With the prefix argument, allow editing of the start server in command; with a
+double prefix prompt for all these parameters."
+  (interactive "P")
+  (let ((params (cider--update-params params)))
```
bbatsov, line 1373: The `let` here is redundant, because the `params` are used just once. Not a big deal, though.

cider.el
```diff
@@ -1313,11 +1314,13 @@ nil."
     (define-key map (kbd "c m") #'cider-connect-clj&cljs)
     (define-key map (kbd "C-c j") #'cider-connect-clj)
     (define-key map (kbd "C-c s") #'cider-connect-cljs)
+    (define-key map (kbd "C-c n") #'cider-start-nrepl-server)
     (define-key map (kbd "C-c m") #'cider-connect-clj&cljs)
     (define-key map (kbd "C-c C-j") #'cider-connect-clj)
     (define-key map (kbd "C-c C-s") #'cider-connect-cljs)
     (define-key map (kbd "C-c C-m") #'cider-connect-clj&cljs)
     (define-key map (kbd "s j") #'cider-connect-sibling-clj)
+    (define-key map (kbd "s n") #'cider-start-nrepl-server)
```
bbatsov, line 1323: I meant adding `C-s n` and `C-s C-n` (if you want to stick with this prefix). Frankly, I'm not sure whether the jack-in or the connect keymap is the best, but I think that the siblings is probably not. 

I'm leaning towards jack-in, so I'd go with `C-j n`, `C-j C-n` and `j n`.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -267,6 +267,18 @@ Here is an example Nbb Jack-In command, providing a custom `:jack-in-cmd`.
...
+==== Starting nREPL server without trying to connect to it ====
+
+In some situations, it might be useful to only start a nREPL server process, without
+connecting to it. This can support complex setups
+for which CIDER cannot reliably detect to which server/port to connect, and
+would therefore fail.
+This assumes that the user will execute a `cider-connect` command manually afterwards,
+specifying host/port.
+
+For this scenario, the `cider-start-nrepl-server` kbd:[C-c C-x s n]  command is offered, which optionally
```
bbatsov, line 279: This should say `j n`.

cider.el
```diff
@@ -1318,6 +1320,7 @@ nil."
     (define-key map (kbd "C-c C-s") #'cider-connect-cljs)
     (define-key map (kbd "C-c C-m") #'cider-connect-clj&cljs)
     (define-key map (kbd "s j") #'cider-connect-sibling-clj)
+    (define-key map (kbd "j n") #'cider-start-nrepl-server)
```
bbatsov, line 1323: This line should be earlier in the keymap (next to the other `j something` keybindings).

cider.el
```diff
@@ -1326,24 +1329,49 @@ nil."
...
+been determined.  Can be nil."
+  (nrepl-start-server-process
+   (plist-get params :project-dir)
+   (plist-get params :jack-in-cmd)
+   on-port-callback))
+
+
+(defun cider--update-params (params)
+  "Completes the passed in PARAMS from user input.
+Updates :project-dir, confirmation for existing seession and :jack-in-cmd."
```
bbatsov, line 1346: seesion -> session

bbatsov:
Thanks!

PR #3556, opened 2023-10-28
Expand `cider-font-lock-dynamically` documentation

vemv:
cljs.core protocols started getting getting syntax highlighting starting from 1.8.x

e.g. the colored IAtom:

img_2081

That surprised a user, and it wasn't immediately obvious to me why that behavior was happening.

This PR expands the doc to make it more obvious to everyone.

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -621,7 +621,9 @@ that should be font-locked:
    `var': Any non-local var gets the `font-lock-variable-name-face'.
    `deprecated' (default): Any deprecated var gets the `cider-deprecated-face'
    face.
-   `core' (default): Any symbol from clojure.core (face depends on type).
+   `core' (default): Any symbol from clojure.core/cljs.core.  The selected face will depend on type.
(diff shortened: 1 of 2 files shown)


cider-mode.el
```diff
@@ -621,7 +621,9 @@ that should be font-locked:
    `var': Any non-local var gets the `font-lock-variable-name-face'.
    `deprecated' (default): Any deprecated var gets the `cider-deprecated-face'
    face.
-   `core' (default): Any symbol from clojure.core (face depends on type).
+   `core' (default): Any symbol from clojure.core/cljs.core.  The selected face will depend on type.
+   Note that while rendering `core', all types of vars (`macro', `function', `var', `deprecated')
+   will be honored, regardless of the user'ss customization value.
```
bbatsov, line 626: ss -> s

vemv, line 626:
```suggestion
   will be honored, regardless of the user's customization value.
```

Issue #3553, opened 2023-10-25
Error not conveyed when using `cider-pprint-eval-last-sexp`


danielneal:
## Expected behavior

When an expression is evaluated using `C-c C-p` `(cider-pprint-eval-last-sexp)`, the user should be alerted that there is an error, either via an overlay or an error buffer. 

## Actual behavior

The `*cider-result*` buffer is shown, but it's empty. The error is printed to the REPL. 

## Steps to reproduce the problem

Evaluate an expression that has an error e.g. `(let [x])` using `C-c C-p`

### CIDER version information

CIDER 1.8.4-snapshot (package: 1.8.3), nREPL 1.0.0
Clojure 1.11.1, Java 11.0.13

### Lein / Clojure CLI version

Clojure CLI version 1.11.1.1413

### Emacs version

29.1

### Operating system

Mac OS Ventura 13.5.2

### JDK distribution

openjdk version "11.0.13" 2021-10-19

vemv:
Done!

Expect a new snapshot (and stable release, probably) within a few hours.

Cheers - V

danielneal:
Awesome, thanks @vemv , that's going to greatly improve my daily experience :)

PR #3558, opened 2023-10-28
`cider-pprint-eval-last-sexp`: use error overlays to indicate failure

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3553

`cider-pprint-eval-last-sexp`, `cider-pprint-eval-last-sexp-to-repl`: use error overlays to indicate failure

This also avoids showing an empty `*cider-result*` buffer.

I've tried this locally over the entire matrix of possibilities:

* command: `cider-pprint-eval-last-sexp`, `cider-eval-last-sexp-to-repl`, `cider-pprint-eval-last-sexp-to-repl`
* sexpr
  * `42` (success)
  * `(/ 2 0)` (runtime error)
  * `(let 1)` (error with a :phase)
* `cider-show-error-buffer`
  * `nil` (`*cider-error*` never shown)
  * `'always` (`*cider-error*` shown _iff_ there's an error with a :phase)

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -776,7 +776,7 @@ The handler simply inserts the result value in BUFFER."
                                               (end (or (car-safe (cdr-safe bounds)) bounds))
                                               (end (when end
                                                      (copy-marker end))))
-                                         (cider--display-error-as-overlay phase err end))))
+                                         (cider--maybe-display-error-as-overlay phase err end))))
 
                                    (cider-handle-compilation-errors err eval-buffer))
                                  ;; done handler:
@@ -863,7 +863,7 @@ and the suffix matched by `cider-module-info-regexp'."
                                          "")
                (string-trim)))
 
-(defun cider--display-error-as-overlay (phase err end)
+(defun cider--maybe-display-error-as-overlay (phase err end)
   "Possibly display ERR as an overlay honoring END,
 depending on the PHASE."
   (when (or
@@ -908,7 +908,7 @@ when `cider-auto-inspect-after-eval' is non-nil."
 
                                    (let ((phase (cider--error-phase-of-last-exception buffer)))
 
-                                     (cider--display-error-as-overlay phase err end)
+                                     (cider--maybe-display-error-as-overlay phase err end)
 
                                      (cider-handle-compilation-errors err
                                                                       eval-buffer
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -746,24 +746,46 @@ evaluation command.  Honor `cider-auto-jump-to-error'."
...
+SOURCE-BUFFER the original buffer,
+and ON-SUCCESS-CALLBACK an optional callback.
+
 The handler simply inserts the result value in BUFFER."
   (let ((eval-buffer (current-buffer))
-        (res ""))
+        (res "")
+        (failed nil))
     (nrepl-make-response-handler (or buffer eval-buffer)
+                                 ;; value handler:
```
bbatsov, line 760: At some point we should just make this take named params (e.g. with `cl-defun`, so it's easier to understand complex handlers.

cider-eval.el
```diff
@@ -841,6 +863,24 @@ and the suffix matched by `cider-module-info-regexp'."
                                          "")
                (string-trim)))
 
+(defun cider--display-error-as-overlay (phase err end)
```
bbatsov, line 866: Given the name of the function it's kind of weird this may or may not render an overlay. Perhaps there should be separate `display-error` and `display-error-as-overlay` functions (one invoking the other internally) or only `display-error` (which is essentially the current function).

cider-eval.el
```diff
@@ -1329,23 +1382,31 @@ If INSERT-BEFORE is non-nil, insert before the form, otherwise afterwards."
...
   (interactive "P")
-  (cider-interactive-eval nil
-                          (cider-insert-eval-handler (cider-current-repl))
-                          (cider-last-sexp 'bounds)
-                          (cider--nrepl-print-request-map fill-column))
-  (when prefix
-    (cider-switch-to-repl-buffer)))
+  (let ((bounds (cider-last-sexp 'bounds)))
+    (cider-interactive-eval nil
+                            (cider-insert-eval-handler (cider-current-repl)
```
bbatsov, line 1402: Probably you can extract the duplicated handler code (I see the same code twice in this source file) in a private function.

cider-eval.el
vemv: TIL!

Yes, would be a good guideline to follow in a future.

cider-eval.el
vemv: I'd be wary about two functions that have to be invoked in concert.

I've renamed it to `cider--maybe-display-error-as-overlay` which surely cuts it.

PR #3568, opened 2023-11-03
`cider-popup-buffer-display`: honor `special-display-buffer-names` if customized for a given CIDER buffer name

vemv:
Before this PR, if I `(add-to-list 'special-display-buffer-names '("*cider-inspect*" my/display-fn))`, the buffer would be rendered twice - once through one's custom function, and once again through CIDER.

I verified it works as intended.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+- `cider-popup-buffer-display`: honor `special-display-buffer-names` if customized for a given CIDER buffer name (e.g. `*cider-inspect*`), avoiding the double-rendering of the given buffer.
+
 ## 1.10.0 (2023-10-31)
 
 ### New features
--- a/cider-popup.el
+++ b/cider-popup.el
@@ -46,29 +46,53 @@ If major MODE is non-nil, enable it for the popup buffer.
 If ANCILLARY is non-nil, the buffer is added to `cider-ancillary-buffers'
 and automatically removed when killed."
   (thread-first (cider-make-popup-buffer name mode ancillary)
+                (buffer-name)
                 (cider-popup-buffer-display select)))
 
-(defun cider-popup-buffer-display (buffer &optional select)
-  "Display BUFFER.
-If SELECT is non-nil, select the BUFFER."
-  (let ((window (get-buffer-window buffer 'visible)))
-    (when window
-      (with-current-buffer buffer
-        (set-window-point window (point))))
-    ;; If the buffer we are popping up is already displayed in the selected
-    ;; window, the below `inhibit-same-window' logic will cause it to be
-    ;; displayed twice - so we early out in this case. Note that we must check
-    ;; `selected-window', as async request handlers are executed in the context
-    ;; of the current connection buffer (i.e. `current-buffer' is dynamically
-    ;; bound to that).
-    (unless (eq window (selected-window))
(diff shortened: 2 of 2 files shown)


bbatsov:
I've never heard of this Emacs functionality, but now that we support it we should probably mention it somewhere in the docs. Even I'm not sure why would the end users want to play with this right now, so some examples would be welcome.

vemv:
I considered it, but most of all I didn't find an adequate page in the manual.

How about: for each possible target (Inspector, etc), mention in their specific page:

> You can control how the popup is rendered using the `special-display-buffer-names` variable blah blah....

vemv:
I'll merge this so as not to pile up PRs (few incoming), but LMK about the best page(s) to doc this

bbatsov:
> How about: for each possible target (Inspector, etc), mention in their specific page:

That sounds reasonable. It can also mentioned only once around the popup-related settings.

rrudakov:
This update introduced a bug: when I evaluate code using `C-c C-p` the result should be rendered in the `*cider-result*` buffer, now this buffer is empty.

Before:

image

After:

image

rrudakov:
Also according to the docs, variable `special-display-buffer-names` is obsolete since emacs version 24.3:

```
special-display-buffer-names is a variable defined in ‘window.el’.

Its value is nil

This variable is obsolete since 24.3; use ‘display-buffer-alist’
instead.
```

vemv:
Thanks!

I think I know the cause - there were two kinds of code paths for raising popups.

Checking

vemv:
Anyway, what I find odd is that you perceive any difference given that your `special-display-buffer-names` value is nil.

The code branch that does not handle `special-display-buffer-names` is essentially untouched - can you spot a bug?

rrudakov:
The only difference I can see is that function `cider-popup-buffer-display` returns string now (`buffer-name`) and before it was returning a buffer. I guess this returned value was used to insert the result to the new buffer.

vemv:
Good catch - that was it!

Pushing to master now

vemv:
I'll also see about the deprecated var. The old one has worked for me for some 5 years but yeah, we should primarily support the modern alternative

That can wait though

rrudakov:
I've been using `display-buffer-alist` for a while to control windows behavior, wasn't even aware of `special-display-buffer-names` until this moment :)

vemv:
I'll have to check it out. Had you successfully used it with cider buffers?

rrudakov:
Yes, sure. 

Here's the relevant part of my config if you're curious (pretty big, but it's easy to spot CIDER-related buffers): https://git.sr.ht/~rrudakov/dotfiles/tree/feature/dotfiles-ng/item/templates/emacs.d/emacs.org#L630-859

Also:
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Displaying-Buffers.html - official documentation
- https://www.masteringemacs.org/article/demystifying-emacs-window-manager - good article on the topic

bbatsov:
> This variable is obsolete since 24.3; use ‘display-buffer-alist’

CIDER only supports Emacs 26+, so it's pointless to support the legacy variable IMO. I didn't know about it, but I certainly know about `display-buffer-alist`. :-)

vemv:
I'll revert it today after migrating my personal stuff `special-display-buffer-names` -> `display-buffer-alist` and verifying it keeps working as usual w/ cider buffers.

PR #3570, opened 2023-11-04
Inspector - misc

### LINKED ISSUE #3565: Streamline navigating from `*cider-error*` to the inspector
## Context

`*cider-error*` shows ex-data when displaying a given exception. Said ex-data can be intrincate, and therefore jumping to the Inspector is a natural thing to want.

Currently, if I press `RET`, nothing will happen (other than `newline: Buffer is read-only: #<buffer *cider-error*>`)

## Proposal

When hitting `RET`, the given exception should be inspected.

## Notes

`*cider-error*` can have N `cause`s. `RET` ing each cause should inspect a different object.

It probably would be idea to store in a var (e.g. `cider.nrepl.middleware.ex/xxx`) the current exception being shown in `*cider-error*`, given that it doesn't necessarily corresponds to `*e`.

vemv:
* Refine `cider-find-keyword` docstring
* Inspector: display Java class/method/field info when available
* `*cider-error*`: open a given exception in the Inspector by clicking it 
  * Fixes #3565

Cheers - V

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -236,6 +236,17 @@ See `cider-sync-request:inspect-push' and `cider-inspector--render-value'"
       (cider-inspector--render-value result 'v2)
       (cider-inspector-next-inspectable-object 1))))
 
+(defun cider-inspector-inspect-last-exception (index)
+  "Inspects the exception in the cause stack identified by INDEX."
+  (interactive)
+  (cl-assert (numberp index))
+  (setq cider-inspector--current-repl (cider-current-repl))
+  (let ((result (cider-sync-request:inspect-last-exception index 'v2)))
+    (when (nrepl-dict-get result "value")
+      (push (point) cider-inspector-location-stack)
+      (cider-inspector--render-value result 'v2)
+      (cider-inspector-next-inspectable-object 1))))
+
 (defun cider-inspector-previous-sibling ()
   "Inspect the previous sibling value within a sequential parent.
 See `cider-sync-request:inspect-previous-sibling' and `cider-inspector--render-value'"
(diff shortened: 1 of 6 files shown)


cider-inspector.el
```diff
@@ -208,93 +208,116 @@ current buffer's namespace."
...
-                      cider-inspector-max-atom-length
-                      cider-inspector-max-coll-size)))
-    (cider-inspector--render-value value)))
+  (let ((result (cider-sync-request:inspect-expr
+                 expr ns
+                 cider-inspector-page-size
+                 cider-inspector-max-atom-length
+                 cider-inspector-max-coll-size
+                 'v2)))
+    (when (nrepl-dict-get result "value")
```
vemv, line 217: I had to do this `v2` refactoring which keeps backward compat.

All of the inspector usage here was coupled to the `"value"` attr, which was hindering access to other attrs (the `"doc-fragments"` attr in particular, which needs separate tratment given that it's processed through `cider--render-docstring`)

doc/modules/ROOT/pages/usage/dealing_with_errors.adoc
```diff
@@ -295,3 +295,21 @@ Cider recognizes stacktraces printed in the following formats:
...
+This allows you to better understand intrincate `ex-data`.
+
+This clicking is defined and customizable in `cider-stacktrace-exception-map`, which has the following defaults:
+
+=== Keybindings
+
+|===
+| Action | Description
+
+| kbd:[click] or kbd:[p]
```
bbatsov, line 313: Why not "i" for "inspect"?

vemv: cider-debug has bound:

    ["Inspect current value" (cider-debug-mode-send-reply ":inspect") :keys "p"]

, so it would seem consistent.

However we could use both p and i - seems harmless given that this keymap is constrained to a single given string.

bbatsov: Yeah, that'd be good. I'm guessing the debugger couldn't use `i` as it was already taken by `inject`. Lots of functionality got weird keybindings if it was added later in the lifecycle of the project and the natural keybindings were already taken by something else.

cider-stacktrace.el
```diff
@@ -799,19 +799,52 @@ the NAME.  The whole group is prefixed by string INDENT."
...
+
+                                     inspect-index
+                                     ,inspect-index
+
+                                     keymap
+                                     ,cider-stacktrace-exception-map)
             (insert (format "%d. " num)
                     (propertize note 'font-lock-face 'font-lock-comment-face) " "
-                    (propertize class 'font-lock-face class-face)
+                    (propertize class 'font-lock-face class-face 'mouse-face 'highlight)
```
vemv, line 847: @bbatsov - added `'mouse-face 'highlight` following Slack feedback

doc/modules/ROOT/pages/usage/dealing_with_errors.adoc
vemv: Added `i`!

vemv:
I tried the refactored one by one and they keep working as usual - merging

PR #3571, opened 2023-11-05
Possibly display error overlays on `cider-load-buffer`

vemv:
> Part of https://github.com/clojure-emacs/cider/issues/3495

I verified locally that it works nicely.

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -951,7 +951,16 @@ Optional argument DONE-HANDLER lambda will be run once load is complete."
                                    (cider-emit-interactive-eval-output value))
                                  (lambda (_buffer err)
                                    (cider-emit-interactive-eval-err-output err)
-                                   (cider-handle-compilation-errors err eval-buffer))
+                                   ;; 1.- Jump to the error line:
+                                   (cider-handle-compilation-errors err eval-buffer)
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -951,7 +951,16 @@ Optional argument DONE-HANDLER lambda will be run once load is complete."
...
                                  (lambda (_buffer err)
                                    (cider-emit-interactive-eval-err-output err)
-                                   (cider-handle-compilation-errors err eval-buffer))
+                                   ;; 1.- Jump to the error line:
+                                   (cider-handle-compilation-errors err eval-buffer)
+                                   (with-current-buffer eval-buffer
+                                     (let* ((phase (cider--error-phase-of-last-exception buffer))
+                                            ;; 2.- Calculate the overlay position, which is the point (per the previous jump),
+                                            ;;     and then end-of-line (for ensuring the overlay will be rendered properly):
+                                            (end (save-excursion
```
bbatsov, line 960: Won't this be problematic on long lines? I think it's probably best to just have the overlay at the point's position after the jump to the error line.

vemv: Done!

PR #3574, opened 2023-11-07
Add a command to delete history item at point

Zzull:
This PR adds a command to delete an history item at point.

~The first commit is important because it makes `cider-repl-input-history` global. It was already sort of global before because the history was shared across REPLs and saved to a unique file but now it's downright a `defvar` instead of a `defvar-local`. I went in that direction because it felt like it was the simplest way of dealing with acknowledgment of deletion of an item across multiple REPLs.~

The first commit now stores histories on a per-project basis

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+- [#3574](https://github.com/clojure-emacs/cider/issues/3574): New value `per-project` for `cider-repl-history-file` to save the history on a per-project basis.
+
 ### Bugs fixed
 
 - [#3763](https://github.com/clojure-emacs/cider/issues/3763): Fix `cider-docview-render` completion popup error when symbol being completed does not have a docstring.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### New features
+- CIDER [History](https://docs.cider.mx/cider/repl/history.html): Add a command to delete history item at point.
+
 ### Changes
 
 - [#3574](https://github.com/clojure-emacs/cider/issues/3574): New value `per-project` for `cider-repl-history-file` to save the history on a per-project basis.
--- a/cider-repl-history.el
+++ b/cider-repl-history.el
@@ -575,6 +575,16 @@ text from the *cider-repl-history* buffer."
   (with-current-buffer cider-repl-history-repl-buffer
     (undo)))
 
+(defun cider-repl-history-delete-entry-at-point ()
+  "Delete history entry (at point)."
+  (interactive)
+  (let* ((orig (point))
+         (str (cider-repl-history-current-string orig)))
+    (with-current-buffer cider-repl-history-repl-buffer
+      (delete str cider-repl-input-history))
+    (cider-repl-history-update)
+    (goto-char orig)))
+
 (defun cider-repl-history-setup (repl-win repl-buf history-buf &optional regexp)
   "Setup.
 REPL-WIN and REPL-BUF are where to insert commands;
@@ -693,6 +703,7 @@ HISTORY-BUF is the history, and optional arg REGEXP is a filter."
     (define-key map (kbd "g")   #'cider-repl-history-update)
     (define-key map (kbd "q")   #'cider-repl-history-quit)
     (define-key map (kbd "U")   #'cider-repl-history-undo-other-window)
+    (define-key map (kbd "D")   #'cider-repl-history-delete-entry-at-point)
     (define-key map (kbd "?")   #'describe-mode)
     (define-key map (kbd "h")   #'describe-mode)
     map))
--- a/cider-repl-history.el
+++ b/cider-repl-history.el
@@ -220,7 +220,6 @@ call `cider-repl-history' again.")
 (defvar cider-repl-history-previous-overlay nil
   "Previous overlay within *cider-repl-history* buffer.")
 
-
 (defun cider-repl-history-get-history ()
   "Function to retrieve history from the REPL buffer."
   (if cider-repl-history-repl-buffer
--- a/cider-repl-history.el
+++ b/cider-repl-history.el
@@ -636,16 +636,17 @@ HISTORY-BUF is the history, and optional arg REGEXP is a filter."
                         #'cider-repl-history-update-highlighted-entry
                         nil t))
             (message
-             (let ((entry (if (= 1 (length cider-command-history))
-                              "entry"
-                            "entries")))
+             (let* ((history-length (length cider-command-history))
+                    (entry (if (= 1 history-length)
+                               "entry"
+                             "entries")))
                (concat
                 (if (and (not regexp)
                          cider-repl-history-display-duplicates)
                     (format "%s %s in the command history."
-                            (length cider-command-history) entry)
+                            history-length entry)
                   (format "%s (of %s) %s in the command history shown."
-                          (length items) (length cider-command-history) entry))
+                          (length items) history-length entry))
                 (substitute-command-keys
                  (concat "  Type \\[cider-repl-history-quit] to quit.  "
                          "\\[describe-mode] for help.")))))
--- a/cider-repl.el
+++ b/cider-repl.el
@@ -194,9 +194,6 @@ CIDER 1.7."
 This property value must be unique to avoid having adjacent inputs be
 joined together.")
 
-(defvar-local cider-repl-input-history '()
-  "History list of strings read from the REPL buffer.")
-
 (defvar-local cider-repl-input-history-items-added 0
   "Variable counting the items added in the current session.")
 
@@ -1468,6 +1465,9 @@ WIN, BUFFER and POS are the window, buffer and point under mouse position."
 (defvar cider-repl-history-pattern nil
   "The regexp most recently used for finding input history.")
 
+(defvar cider-repl-input-history '()
+  "History list of strings read from the REPL buffer.")
+
 (defun cider-repl--add-to-input-history (string)
   "Add STRING to the input history.
 Empty strings and duplicates are ignored."
@@ -1593,9 +1593,11 @@ If USE-CURRENT-INPUT is non-nil, use the current input."
   :safe #'integerp)
 
 (defcustom cider-repl-history-file nil
-  "File to save the persistent REPL history to."
-  :type 'string
-  :safe #'stringp)
+  "File to save the persistent REPL history to.
+If this is set to a path the history will be global to all projects.  If this is
+set to `per-project', the history will be stored in a file (.cider-history) at
+the root of each project."
+  :type '(choice string symbol))
 
 (defun cider-repl--history-read-filename ()
   "Ask the user which file to use, defaulting `cider-repl-history-file'."
@@ -1619,11 +1621,27 @@ defined filenames can be used to read special history files.
 
 The value of `cider-repl-input-history' is set by this function."
   (interactive (list (cider-repl--history-read-filename)))
-  (let ((f (or filename cider-repl-history-file)))
-    ;; TODO: probably need to set cider-repl-input-history-position as well.
-    ;; in a fresh connection the newest item in the list is currently
-    ;; not available.  After sending one input, everything seems to work.
-    (setq cider-repl-input-history (cider-repl--history-read f))))
+  (cond
+   (filename (setq cider-repl-history-file filename))
+   ((equal 'per-project cider-repl-history-file)
+    (make-local-variable 'cider-repl-input-history)
+    (when-let ((dir (clojure-project-dir)))
+      (setq-local
+       cider-repl-history-file (expand-file-name ".cider-history" dir)))))
+  (when cider-repl-history-file
+    (condition-case nil
+        ;; TODO: probably need to set cider-repl-input-history-position as
+        ;; well. In a fresh connection the newest item in the list is
+        ;; currently not available.  After sending one input, everything
+        ;; seems to work.
+        (setq
+         cider-repl-input-history
+         (cider-repl--history-read cider-repl-history-file))
+      (error
+       (message
+        "Malformed cider-repl-history-file: %s" cider-repl-history-file)))
+    (add-hook 'kill-buffer-hook #'cider-repl-history-just-save t t)
+    (add-hook 'kill-emacs-hook #'cider-repl-history-save-all)))
 
 (defun cider-repl--history-write (filename)
   "Write history to FILENAME.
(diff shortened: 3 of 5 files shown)


Zzull:
On a side note, it looks like `eldev linter` doesn't agree with the CI on how to indent
```
cider-repl.el:354:0 (indent)     !          (contents)
cider-repl.el:355:0 (indent)     !          (insert-before-markers
cider-repl.el:356:0 (indent)     !           (propertize
cider-repl.el:357:0 (indent)     !            (if (string-blank-p contents) ";;\n" (concat ";; " contents "\n"))
cider-repl.el:358:0 (indent)     !            'font-lock-face 'font-lock-comment-face))))
Found 5 warnings in file `cider-repl.el'
```

vemv:
Thanks for the PR!

> The first commit is important because it makes cider-repl-input-history global. 

Although it may be true, I sense that having per-project repl histories would be the best new direction - WDYT?

vemv:
> On a side note, it looks like eldev linter doesn't agree with the CI on how to indent

Probably it's a matter of using our Makefile instead

Zzull:
> > On a side note, it looks like eldev linter doesn't agree with the CI on how to indent
> 
> Probably it's a matter of using our Makefile instead

`make lint` and `eldev lint` yielded the same result (that `cider-repl.el` has a block poorly indented)

Zzull:
> Thanks for the PR!
> 
> > The first commit is important because it makes cider-repl-input-history global. It was already sort of global before because the history was shared across REPLs
> 
> Although it may be true, I sense that having per-project repl histories would be the best new direction - WDYT?

I can totally try to split the history on a per-project basis

vemv:
> make lint and eldev lint yielded the same result (that cider-repl.el has a block poorly indented)

Maybe you're on a different Emacs version - we use 28 for linting. The Makefile has a comment indicating how to override the local Emacs.

> I can totally try to split the history on a per-project basis

Thanks!

I assume it's never been like that / it's something we want? @bbatsov 

To me it makes sense because a given project's sexprs are often irrelevant / invalid (not compile-able) in another project.

Zzull:
> Maybe you're on a different Emacs version - we use 28 for linting. The Makefile has a comment indicating how to override the local Emacs.

Absolutely. Thank you

> To me it makes sense because a given project's sexprs are often irrelevant / invalid (not compile-able) in another project.

It's done in the first commit (which is a bit hard to read…)

bbatsov:
> I assume it's never been like that / it's something we want? @bbatsov

I'd be fine by this, although I'm not sure how it can be made reliably.(e.g. what would you use as project id for remote connections?) I'll take a look at the proposed implementation.

cider-repl.el
```diff
@@ -1612,18 +1605,27 @@ It does not yet set the input history."
...
-  (interactive (list (cider-repl--history-read-filename)))
-  (let ((f (or filename cider-repl-history-file)))
-    ;; TODO: probably need to set cider-repl-input-history-position as well.
-    ;; in a fresh connection the newest item in the list is currently
-    ;; not available.  After sending one input, everything seems to work.
-    (setq cider-repl-input-history (cider-repl--history-read f))))
+The value of `cider-repl-input-history' and `cider-repl-history-file' are set by
+this function."
+  (when-let* ((dir (clojure-project-dir))
+              (file (concat dir "/.cider-history")))
```
bbatsov, line 1610: I'd better to use `expand-file-name` instead of `concat`. Also - I'd move the file name to a constant.

cider-repl.el
```diff
@@ -1592,16 +1595,6 @@ If USE-CURRENT-INPUT is non-nil, use the current input."
   :type 'integer
   :safe #'integerp)
 
-(defcustom cider-repl-history-file nil
```
bbatsov, line 1595: You can't just delete this - you'll have to add a deprecation warning for it, so people would know it's no longer being used. If we want to avoid breaking changes we'd probably want to keep the old behavior if this variable is set.

cider-repl.el
```diff
@@ -194,6 +194,9 @@ CIDER 1.7."
 This property value must be unique to avoid having adjacent inputs be
 joined together.")
 
+(defvar-local cider-repl-history-file nil
```
bbatsov, line 197: Related to my other remark - you should probably use a different name for the buffer-local var, to avoid breaking changes.

bbatsov:
I quickly went over the code and it looks reasonable overall, but it also introduces a breaking change that I think would be prudent to ignore. Probably it'd be best to keep the old behavior if someone set an file name explicitly and switch to the logging per project if they haven't.

cider-repl.el
```diff
@@ -1612,28 +1601,35 @@ It does not yet set the input history."
...
-  (interactive (list (cider-repl--history-read-filename)))
-  (let ((f (or filename cider-repl-history-file)))
-    ;; TODO: probably need to set cider-repl-input-history-position as well.
-    ;; in a fresh connection the newest item in the list is currently
-    ;; not available.  After sending one input, everything seems to work.
-    (setq cider-repl-input-history (cider-repl--history-read f))))
+The value of `cider-repl-input-history' and `cider-repl-history-file' are set by
+this function."
+  (when-let* ((dir (clojure-project-dir))
+              (file (concat dir "/.cider-history")))
```
vemv, line 1610: Probably we should have one file per platform: clj, cljs, bb

cider-repl.el
```diff
@@ -1612,28 +1601,35 @@ It does not yet set the input history."
...
-The value of `cider-repl-input-history' is set by this function."
-  (interactive (list (cider-repl--history-read-filename)))
-  (let ((f (or filename cider-repl-history-file)))
-    ;; TODO: probably need to set cider-repl-input-history-position as well.
-    ;; in a fresh connection the newest item in the list is currently
-    ;; not available.  After sending one input, everything seems to work.
-    (setq cider-repl-input-history (cider-repl--history-read f))))
+The value of `cider-repl-input-history' and `cider-repl-history-file' are set by
+this function."
+  (when-let* ((dir (clojure-project-dir))
```
vemv, line 1609: It may be better to pick the first, non-nil, non-tramp value of `(list nrepl-project-dir clojure-project-dir default-directory)`

This way we don't write files to remote TRAMP targets, which may surprise most people - or bring unforeseen issues.

Zzull:
Hello,

Almost all comments have been addressed.

I still have two things I would like your advice on:

* I don't have a remote machine accessible with clojure on it to make sure that `cider-repl--find-dir-for-history` behaves the way it should. Do you think `tramp-tramp-file-p` is enough?
* I couldn't find a way to get the type of the REPL to suffix the name of the history file with. It seems that it is too early to call `(cider-runtime)` (returns `'generic`) nor get the value of `cider-repl-type` (nil) when `cider-repl-history-load` is called. And anyways, after initialization, `(cider-runtime) returns `'clojure` for a cljs project and  `cider-repl-type` has the value `clj` for a babashka project. Do you have other leads?

vemv:
Thanks much!

It's looking great.

>  Do you think tramp-tramp-file-p is enough?

Most definitely.

> Do you have other leads?

If you don't mind I'll pick up this branch, fix some nits and then also add the 'platform' refinement. Then I'll squash everything, keeping you as the author. It's how I operate normally, to avoid long feedback loops.

We can keep the PR open in the meantime.

cider-repl-history.el
```diff
@@ -576,6 +577,16 @@ text from the *cider-repl-history* buffer."
   (with-current-buffer cider-repl-history-repl-buffer
     (undo)))
 
+(defun cider-repl-history-delete ()
```
bbatsov, line 580: delete-item-at-point

bbatsov: (or entry)

doc/modules/ROOT/pages/repl/configuration.adoc
```diff
@@ -340,12 +340,15 @@ reset automatically by the `track-state` middleware.
...
 
 [source,lisp]
 ----
 (setq cider-repl-history-file "path/to/file")
 ----
 
-Note that CIDER writes the history to the file when you kill the REPL
-buffer, which includes invoking `cider-quit`, or when you quit Emacs.
+Note that CIDER stores the history in a file named by
+`cider-repl-local-history-name` located at the root of your project if
```
bbatsov, line 351: Probably this should be named `cider-repl-project-history-file` or something along those lines, as I think `local` history might be slightly misleading in this case. (and yeah - I realize there might not always be a project associated with a REPL)

bbatsov:
I'll also add that currently CIDER doesn't have persistent history by default and I think it's a good idea to keep this behavior. Frankly, I really care about history between sessions and I guess I'm not the only one, given there has been little demand to change the default behavior over the years.

vemv:
> Frankly, I really care about history between sessions

I didn't get this - what behavior should remain unchanged?

bbatsov:
I meant to write "don't care". I think this should be some opt-in, not a default. (just as storing the global history to file is)

vemv:
From my side it's something I've always missed (really), but it was more in the 'nice-to-have' area. There was always something more important 😄

Maybe we can keep it under a defcustom indeed, enabling it later after some months for gaining confidence in the solution.

bbatsov:
Yeah, that's fair. I'm just always a bit wary of serializing data to files (and reading it back) as often people run into issues with broken entries due to various reason. (probably less so in this case, given the simple nature of the history format, but I've had my fair share of issues with seemingly trivial serialization)

bbatsov:
So, what are we doing about this PR?

bbatsov:
@vemv Any updates on this front?

vemv:
Perhaps this weekend, sorry for the delay.

bbatsov:
@vemv Friendly reminder about this one. :-) I want us to start finalizing the scope of the next release.

vemv:
Sure, I'm finally closing a sprint which didn't leave a lot of OSS time available.

Zzull:
Is there any way I can help?

vemv:
I can't allocate time currently so you can go ahead in any way that works / passes review from @bbatsov

Zzull:
> I'll also add that currently CIDER doesn't have persistent history by default and I think it's a good idea to keep this behavior. Frankly, I really don't care much about history between sessions and I guess I'm not the only one, given there has been little demand to change the default behavior over the years.

I changed the implementation so that CIDER stays without a persistent history by default.
The documentation of `cider-repl-history-file` went from:
```
"File to save the persistent REPL history to.
If this is set, the history will be global to all projects.  Otherwise, the
history is local per project and stored in a file (named by
`cider-repl-local-history-name') at its root."
```
to
```
"File to save the persistent REPL history to.
If this is set to a path the history will be global to all projects.  If this is
set to `per-project', the history will be stored in a file (named by
`cider-repl-project-history-name') at the root of each project."
```

I still haven't found a way to get the REPL type though. I described it earlier:
> Hello,
> … 
>     * I couldn't find a way to get the type of the REPL to suffix the name of the history file with. It seems that it is too early to call `(cider-runtime)` (returns `'generic`) nor get the value of `cider-repl-type` (nil) when `cider-repl-history-load` is called. And anyways, after initialization, `(cider-runtime)` returns `'clojure` for a cljs project and  `cider-repl-type` has the value `'clj` for a babashka project. Do you have other leads?

CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
 
 ## master (unreleased)
 
+### New features
```
bbatsov, line 5: You'll need to rebase, as we cut a new CIDER release recently.

cider-repl.el
```diff
@@ -1612,28 +1614,52 @@ It does not yet set the input history."
           (read (current-buffer))))
     '()))
 
+(defun cider-repl--find-dir-for-history ()
+  "Find the first suitable directory to store the project's history."
+  (seq-find
+   (lambda (dir) (and (not (null dir)) (not (tramp-tramp-file-p dir))))
```
bbatsov, line 1620: Is there an issue with storing the history on a remote filesystem?

bbatsov:
>  * I couldn't find a way to get the type of the REPL to suffix the name of the history file with. It seems that it is too early to call (cider-runtime) (returns 'generic) nor get the value of cider-repl-type (nil) when cider-repl-history-load is called. And anyways, after initialization, (cider-runtime) returns 'clojure for a cljs project and cider-repl-type has the value 'clj for a babashka project. Do you have other leads?

There's not an easy way around this, so I guess we can skip it for the first version. I think the per-project history will be quite the improvement, even without being so granular.

doc/modules/ROOT/pages/repl/configuration.adoc
```diff
@@ -340,12 +340,15 @@ reset automatically by the `track-state` middleware.
...
+* To share all the REPL histories in a single file:
 
 [source,lisp]
 ----
 (setq cider-repl-history-file "path/to/file")
 ----
 
-Note that CIDER writes the history to the file when you kill the REPL
-buffer, which includes invoking `cider-quit`, or when you quit Emacs.
+Note that CIDER stores the history in a file named by
```
bbatsov, line 350: I'd probably expand this to show a full configuration snippet, so it's clearer how people are supposed to set this up. Perhaps you can just have a bullet point for the project-specific configuration so it stands out a bit more?

CHANGELOG.md
Zzull: Done

cider-repl.el
Zzull: I actually don't know and I don't have access to a remote machine on which I can start CIDER on to try. I followed @vemv advice:

> It may be better to pick the first, non-nil, non-tramp value of (list nrepl-project-dir clojure-project-dir default-directory)

> This way we don't write files to remote TRAMP targets, which may surprise most people - or bring unforeseen issues.

Zzull:
> > * I couldn't find a way to get the type of the REPL to suffix the name of the history file with. It seems that it is too early to call (cider-runtime) (returns 'generic) nor get the value of cider-repl-type (nil) when cider-repl-history-load is called. And anyways, after initialization, (cider-runtime) returns 'clojure for a cljs project and cider-repl-type has the value 'clj for a babashka project. Do you have other leads?
> 
> There's not an easy way around this, so I guess we can skip it for the first version. I think the per-project history will be quite the improvement, even without being so granular.

I removed the commit introducing that change

doc/modules/ROOT/pages/repl/configuration.adoc
Zzull: Done

cider-repl.el
bbatsov: Fair enough.

bbatsov:
@Zzull Can you also rebase on `master` to resolve the Changelog conflict?

bbatsov (approved):

CHANGELOG.md
bbatsov: Sorry about the slow responses here - I had a couple of busy weeks, but I'll use the holidays to catch up with oss work and finally merge your PR.

Zzull: No worries

Zzull:
> @Zzull Can you also rebase on `master` to resolve the Changelog conflict?

Done

bbatsov:
And we are finally done here! 🎉 🚀 

Happy New Year!!!

Zzull:
Happy New Year!

PR #3579, opened 2023-11-12
Inspector: retain `truncate-lines` values across screens

vemv:
Retains the user's t/nil preference across Inspector screens.

This is particularly important now that we render Java doc, which can be quite lengthy.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -537,14 +537,18 @@ instead of just its \"value\" entry."
                       ;; because on each re-rendering, we wipe everything, including the mode.
                       ;; Enabling cider-inspector-mode is the specific step that loses the font size.
                       (buffer-local-value variable b)))
-         (truncate-lines-variable 'truncate-lines)
          (truncate-lines-defined (when-let* ((b (get-buffer cider-inspector-buffer)))
-                                   (local-variable-p truncate-lines-variable b)))
+                                   (local-variable-p 'truncate-lines b)))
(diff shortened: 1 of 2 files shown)


cider-inspector.el
```diff
@@ -557,12 +563,14 @@ instead of just its \"value\" entry."
       (when cider-inspector-page-location-stack
         (goto-char (pop cider-inspector-page-location-stack))))))
 
-(defun cider-inspector-render (buffer str &optional font-size fragments block-tags)
+(defun cider-inspector-render (buffer str &optional font-size truncate-lines-defined truncate-lines-p fragments block-tags)
```
bbatsov, line 566: The signature looks pretty bad with so many optional params, so it's better to switch to named params at this point or just fold all the rendering params in some alist or plist.

cider-inspector.el
```diff
@@ -536,9 +536,15 @@ instead of just its \"value\" entry."
                       ;; The font size is lost between inspector 'screens',
                       ;; because on each re-rendering, we wipe everything, including the mode.
                       ;; Enabling cider-inspector-mode is the specific step that loses the font size.
-                      (buffer-local-value variable b))))
+                      (buffer-local-value variable b)))
+         (truncate-lines-variable 'truncate-lines)
```
bbatsov, line 540: I'd use the symbol directly instead of introducing this variable - it just makes the subsequent code more confusing IMO.

vemv:
Never too late to use `cl-defun` for the first time

Thanks!

PR #3567, opened 2023-11-12
Remote `cider-jack-in` breaks with `cider-enrich-classpath`: `No such file or directory`

caadr:
This is an attempt to fix enrich-classpath erroring out, when `default-directory` is remote, by creating a copy of the `lein.sh|clojure.sh` script on the remote before jacking in.

The [core of it is happening here](https://github.com/clojure-emacs/cider/blob/4a1a33eb398fe75d45676f6b9dea2b36daeb4481/cider.el#L440).

**_Conversation from the original Issue below._**

-----------------

- [x] You've added (very few) tests to cover your change(s)

-----------------

## Expected behavior

With point in a project-file visited via ssh/tramp and `cider-enrich-classpath` = `t`,
`cider-jack-in-clj` creates a remote repl and connects to it.

## Actual behavior

`nrepl-server-sentinel` throws:
```
error("Could not start nREPL server: %s (%S)" "bash: /home/**/cider/clojure.sh: No such file or directory\n" "exited abnormally with code 127")
```

## Reason

Cider is trying to use the local path (from `cider--get-enrich-classpath-clojure-cli-script`) on the remote, where it does not exist.

## Possible Workaround

Copy the script to the remote before starting the server.


adrech:
Ah, first time I tried creating a PR from magit/forge and it makes a mess :o

> * If it's at hand, prefer `mktemp -d -t cider.XXXXXX` to `/tmp` This
should essentially do the same thing in most cases:
```
(make-nearby-temp-file "cider." :dir) ;; =>
/ssh:root@localhost#8022:/tmp/cider.OrkB00 
``` 
But for this draft, instead of creating a directory I copy directly to
`<remote-tempdir>/.cider__<script-name>__XXXXXX`
because cleanup is simpler. 
The way it is right now, the file could end up in the project directory (if the remote tmpdir is inaccessible) and I didn't want the tempfiles to pile up there. There are some holes in the way cleanup is implemented right now - when something goes wrong before the script is executed it will stay - but it was fun playing around :)

> * If so, ~setq-local~ bind `cider-enrich-classpath nil` and proceed
This is happening now in (`cider--update-jack-in-cmd`) [https://github.com/adrech/cider/blob/ae20f4e72b0e2206955d82ac328bf8660a81a429/cider.el#L1686].
by (checking if the command contains one of the scripts names) [https://github.com/adrech/cider/blob/ae20f4e72b0e2206955d82ac328bf8660a81a429/cider.el#L418].
To keep that simple but reliable, I prefixed the files with "enriched_[lein|clojure].sh".

In general, I tried to avoid introducing new `defvar-locals`, but I'm not sure it's worth it.

So far this worked and failed as expected with `emacs- Q` on my pi and in clojure / lein docker containers (all via ssh though).

Let me know if I'm moving in the right direction and I'll fill up the docstrings, add tests.

vemv:
Happy to see these iterations 🙌

Please don't rename `lein.sh → enrich_lein.sh` as it makes some UIs more verbose, and it also would require a PR against https://github.com/melpa/melpa which is extra friction.

Cheers - V

adrech:
Aye, of course, reverted the filenames.

adrech:
There's a potential problem (with handling remotes in general):

All of this assumes, that tramp and the remote PATH are set up to work together. It will fail [early in the process](https://github.com/clojure-emacs/cider/blob/8183c2b1300c036bf28da5c8c9d1bceae3f5383f/cider.el#L2143)
otherwise.

This would happen e.g. when one connects via ssh + relies on a PATH setup in the remotes `.bashrc` or `.profile` + uses default tramp. More specifically, it they don't have this line in their tramp config, which makes tramp grab the env from a login shell:
```
(add-to-list 'tramp-remote-path 'tramp-own-remote-path)
```

For more background, see: `(info "(tramp) Remote progams")` and the ramblings at 
https://github.com/clojure-emacs/cider/blob/92c9694161fa7638b5e9d05bb797ffefde613237/dev/tramp-sample-project/Dockerfile.template#L4

A couple of ways to approach this:
- rebinding `tramp-remote-path` somewhere high up in the callchain (only if it is untouched)
- informing the user that they might have this problem and how to fix it, when resolving a remote command fails
- reverting to the old behaviour of `cider--resolve-command`, that would just no-op on remotes and let the user run into an error message like `java not found...` when the server tries to start, which would give them an idea that something is wrong with their PATH setup.

bbatsov:
Alternatively now you can try using https://metaredux.com/posts/2025/01/17/a-simpler-way-to-deal-with-java-sources-in-cider.html

caadr:
Perfect! The new approach via `cider-download-java-sources` works seamlessly on my remotes. Thanks for the great work!
Closing this.

PR #3586, opened 2023-11-15
Inspector: introduce `o`, `:` keybindings

vemv:
* Inspector: introduce `:` keybinding which performs `cider-inspect-expr`

Now the inspector can also double as a REPL 😄

* Inspector: introduce `o` keybinding which performs `cider-open-url-at-point`

It's relatively common to inspect objects that contain a URL in it.

`u` now can open those in the browser.

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -89,11 +89,13 @@ by clicking or navigating to them by other means."
           "\\|[+-.0-9]+")            ; nor numbers. Note: BigInts, ratios etc. are interesting
   "Regexp of uninteresting and skippable values.")
 
-(defun cider-open-url-at-point ()
-  "Opens the url at point if found, without prompting."
+(defun cider-inspector-open-thing-at-point ()
+  "Opens the thing at point if found, without prompting."
   (interactive)
-  (when-let ((url (thing-at-point 'url)))
-    (browse-url url)))
+  (if-let ((url (thing-at-point 'url)))
+      (browse-url url)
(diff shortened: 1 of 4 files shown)


vemv:
@bbatsov I slightly broadened the scope of this PR.

doc/modules/ROOT/pages/debugging/inspector.adoc
```diff
@@ -84,6 +84,14 @@ You'll have access to additional keybindings in the inspector buffer
 | `cider-inspector-next-sibling`
 | Navigates to the next sibling, within a sequential collection.
 
+| kbd:[u]
+| `cider-open-url-at-point`
+| Opens the url at point if found.
+
+| kbd:[r]
```
bbatsov, line 91: Perhaps you can use `:` for that, as it'd be closer to the cider-mode keybinding for reading an expression from the minibuffer.

cider-inspector.el
```diff
@@ -89,13 +89,20 @@ by clicking or navigating to them by other means."
           "\\|[+-.0-9]+")            ; nor numbers. Note: BigInts, ratios etc. are interesting
   "Regexp of uninteresting and skippable values.")
 
+(defun cider-open-url-at-point ()
```
bbatsov, line 92: Commands in the inspector should be starting with `cider-inspector-`.

bbatsov: On a more general note - probably it'd be better to have a single "open" command that just tries a few options. (e.g. file, url, etc).

PR #3589, opened 2023-11-24
Introduce `cider-inspector-tap-at-point`

vemv:
- Introduce `cider-inspector-tap-at-point`
- Use enrich-classpath [1.18.6](https://github.com/clojure-emacs/enrich-classpath/compare/v1.18.4...v1.18.6)
- Use cider-nrepl [0.44.0](https://github.com/clojure-emacs/cider-nrepl/blob/44da162f51765464192ba04102398c5982f01638/CHANGELOG.md#0440-2023-11-24)
- Use clojure-mode [5.18.1](https://github.com/clojure-emacs/clojure-mode/blob/v5.18.1/CHANGELOG.md#5181-2023-11-24)

### DIFF
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -116,6 +116,7 @@ by clicking or navigating to them by other means."
     (define-key map "c" #'cider-inspector-set-max-coll-size)
     (define-key map "d" #'cider-inspector-def-current-val)
     (define-key map "t" #'cider-inspector-tap-current-val)
+    (define-key map "1" #'cider-inspector-tap-at-point)
     (define-key map [tab] #'cider-inspector-next-inspectable-object)
     (define-key map "\C-i" #'cider-inspector-next-inspectable-object)
     (define-key map "n" #'cider-inspector-next-inspectable-object)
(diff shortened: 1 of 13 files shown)


cider-inspector.el
```diff
@@ -386,6 +387,24 @@ current-namespace."
...
     (user-error "No CIDER session found")))
 
+(defun cider-inspector-tap-at-point ()
+  "Sends the current Inspector current sub-value (per POINT) to `tap>'."
+  (interactive)
+  ;; NOTE: we don't set `cider-inspector--current-repl', because we mean to tap the current value of an existing Inspector,
+  ;; so whatever repl was used for it, should be used here.
+  (if cider-inspector--current-repl
+      (seq-let (property value) (cider-inspector-property-at-point)
+        (cl-case property
```
bbatsov, line 397: It's generally better to use `pcase` instead of `cl-case`.

vemv: We had a similar occurence in this file. Will fix both

cider-inspector.el
```diff
@@ -386,6 +387,24 @@ current-namespace."
...
+          (cider-value-idx
+           (let* ((idx value)
+                  (response (cider-sync-request:inspect-tap-indexed idx)))
+             (nrepl-dbind-response response (value err)
+               (if value
+                   (message "Successfully tapped the Inspector item at point")
```
bbatsov, line 403: Do we even need a message if things are going as expected?

vemv: Yep, because the `tap>` is a side-effect, nothing else is visible

One could as well have hit an unbound key, so the confirmation is useful.

Issue #3588, opened 2023-11-20
configure cider to use PowerShell 7.x  and not Windows PowerShell 5.1 on windows 


shishini:
Cider fails to jack in a repl , and this initially seem to be because its trying to start the repl using 

`c:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`
instead of 
`C:\Program Files\PowerShell\7\pwsh.exe` 

I installed Clojure CLI in a path that is only accessible for PowerShell 7.x not for Windows PowerShell 5.1

So how can i configure cider (or emacs) to use pwsh (which is the command for Powershell 7.x) not powershell  (which is the command for the older Windows Powershell 5.1)

---

I can confirm that adding the clojure cli module to a Windows Powershell 5.1 modules path, fixes my initial issue
so i just want the ability to configure either cider or emacs to use pwsh not powershell

vemv:
https://github.com/clojure-emacs/cider/blob/5261886ad7c0a8828d022812c7c1c6b503ddd6ba/cider.el#L139 is customizable.

Let us know if you have further questions!

Cheers - V

shishini:
I am really not that experience
Can you give more detail on how to use this feature to fix the problem
what do i need to add in my .emacs for example ?

vemv:
Yes, please add something like `(setq cider-clojure-cli-command "pwsh")` in your Emacs .init file. After a restart, that should be it

shishini:
thanks i did just that but now i am getting this error

```
error in process sentinel: nrepl-server-sentinel: Could not start nREPL server: Error while parsing option "--config-data {:deps {nrepl/nrepl {:mvn/version \"\"1.0.0\"\"} cider/cider-nrepl {:mvn/version \"\"0.43.3\"\"}} :aliases {:cider/nrepl {:main-opts [\"\"-m\"\" \"\"nrepl.cmdline\"\" \"\"--middleware\"\" \"\"[cider.nrepl/cider-middleware]\"\"]}}}": java.lang.NumberFormatException: Invalid number: 1.0.0

 ("finished")
[...]

PR #3590, opened 2023-11-28
Fix pwsh compatibility bug

shishini:
The command line needs a different formatting for pwsh compared to powershell

More details can be found in this bug reports #3588 
In summary, initializing a repl works when installing clojure-cli to Powershell 5.1 (Windows Powershell) but was failing when clojure-cli was installed in pwsh (when  (setq cider-clojure-cli-command "pwsh") )

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -800,7 +800,7 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   (string-equal command "powershell"))
 
 (defun cider--shell-quote-argument (argument &optional command)
-    "Quotes ARGUMENT like `shell-quote-argument', suitable for use with COMMAND.
+  "Quotes ARGUMENT like `shell-quote-argument', suitable for use with COMMAND.
 
 Uses `shell-quote-argument' to quote the ARGUMENT, unless COMMAND is given
 and refers to PowerShell, in which case it uses (some limited) PowerShell
--- a/cider.el
+++ b/cider.el
@@ -791,21 +791,19 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   "Removes the duplicates in DEPS."
   (cl-delete-duplicates deps :test 'equal))
 
+(defun cider--jack-in-cmd-pwsh-p (command)
+  "Returns whether COMMAND is pwsh."
+  (string-equal command "pwsh"))
+
 (defun cider--jack-in-cmd-powershell-p (command)
-  "Returns whether COMMAND is PowerShell."
-  (or (string-equal command "powershell")
-      (string-equal command "pwsh")))
+  "Returns whether COMMAND is Powershell."
+  (string-equal command "powershell"))
 
 (defun cider--shell-quote-argument (argument &optional command)
-  "Quotes ARGUMENT like `shell-quote-argument', suitable for use with COMMAND.
-
-Uses `shell-quote-argument' to quote the ARGUMENT, unless COMMAND is given
-and refers to PowerShell, in which case it uses (some limited) PowerShell
-rules to quote it."
-  (if (cider--jack-in-cmd-powershell-p command)
-      ;; please add more PowerShell quoting rules as necessary.
-      (format "'%s'" (replace-regexp-in-string "\"" "\"\"" argument))
-    (shell-quote-argument argument)))
+  "Patch to skip the quoting on windows"
+  (cond ((cider--jack-in-cmd-pwsh-p command) (format "'%s'" argument))
+	((cider--jack-in-cmd-powershell-p command) (format "'%s'" (replace-regexp-in-string "\"" "\"\"" argument)))
+	(t (shell-quote-argument argument))))
 
 (defun cider--powershell-encode-command (cmd-params)
   "Base64 encode the powershell command and jack-in CMD-PARAMS for clojure-cli."
@@ -1625,7 +1623,7 @@ Params is a plist with the following keys (non-exhaustive)
 (defun cider--format-cmd (command-resolved command cmd-params)
   "Format COMMAND-RESOLVED or COMMAND followed by CMD-PARAMS."
   (format "%s %s" command-resolved
-          (if (cider--jack-in-cmd-powershell-p command)
+          (if (or (cider--jack-in-cmd-pwsh-p command) (cider--jack-in-cmd-powershell-p command))
               (cider--powershell-encode-command cmd-params)
             cmd-params)))
 
--- a/cider.el
+++ b/cider.el
@@ -800,10 +800,14 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
   (string-equal command "powershell"))
 
 (defun cider--shell-quote-argument (argument &optional command)
-  "Patch to skip the quoting on windows"
+    "Quotes ARGUMENT like `shell-quote-argument', suitable for use with COMMAND.
+
+Uses `shell-quote-argument' to quote the ARGUMENT, unless COMMAND is given
+and refers to PowerShell, in which case it uses (some limited) PowerShell
+rules to quote it."
   (cond ((cider--jack-in-cmd-pwsh-p command) (format "'%s'" argument))
-	((cider--jack-in-cmd-powershell-p command) (format "'%s'" (replace-regexp-in-string "\"" "\"\"" argument)))
-	(t (shell-quote-argument argument))))
+        ((cider--jack-in-cmd-powershell-p command) (format "'%s'" (replace-regexp-in-string "\"" "\"\"" argument)))
+        (t (shell-quote-argument argument))))
 
 (defun cider--powershell-encode-command (cmd-params)
   "Base64 encode the powershell command and jack-in CMD-PARAMS for clojure-cli."


vemv:
Looks good to me - great work!

If you have it easy, please make the failing CI pass. Othewise we can cherry-pick + amend your commit.

@ikappaki review also welcome.

Thanks - V

shishini:
Not really sure how to 
Do I add a new commit to my branch, will it show up automatically, ?

vemv:
Yes, commit and push

cider.el
```diff
@@ -791,21 +791,23 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
...
   "Quotes ARGUMENT like `shell-quote-argument', suitable for use with COMMAND.
 
 Uses `shell-quote-argument' to quote the ARGUMENT, unless COMMAND is given
 and refers to PowerShell, in which case it uses (some limited) PowerShell
 rules to quote it."
-  (if (cider--jack-in-cmd-powershell-p command)
-      ;; please add more PowerShell quoting rules as necessary.
-      (format "'%s'" (replace-regexp-in-string "\"" "\"\"" argument))
-    (shell-quote-argument argument)))
+  (cond ((cider--jack-in-cmd-pwsh-p command) (format "'%s'" argument))
```
bbatsov, line 808: Might be good to add some comment explaining the need for the different escaping rules.

bbatsov:
Don't forget to add a changelog entry about this bugfix.

cider.el
shishini: Its important to clarify that this for me, is workaround for a bug that is either in the clojure cli tool or in powershell
clojure cli should behave the same on both shells, since as per powerhsell docs the quoting rules are the same 

I created a question on ask clojure , suggesting that this is a bug in the clojure command line tool, since i cannot replicate this behavior with any other command 

Are you ok with documenting this as a workaround, or do your prefer another solution, i think the clava plugin in windows skips the cli tool and use the java command directly , so they dont seem to have this issue 

they seem to run this command

```
; Starting Jack-in Terminal: pushd c:\dev\lang\clojure\project\ukg & java -jar ".calva\deps.clj.jar" -Sdeps "{:deps {nrepl/nrepl {:mvn/version,""1.0.0""},cider/cider-nrepl {:mvn/version,""0.28.5""}}}" -M -m nrepl.cmdline --middleware "[cider.nrepl/cider-middleware]" & popd
```

bbatsov: Documenting this as a workaround is fine - e.g. I just learned what this is about and I imagine others reading the codebase will be in the same boat as me. :-)

bbatsov: (I thought `powershell` and `pwsh` were exactly the same thing)

ikappaki:
Hi all

I wanted to add some analysis on why we're seeing a different behavior with `pwsh`. It appears the double quoting rules have changed with [pwsh 7.3 as per this stackoverflow post](https://stackoverflow.com/a/74440425), using double double quotes is not required with that version or newer. 

Here I'm using the GNU coreutils echo command to demonstrate the change in behaviour:

powershell.exe treats `""` as `"` while pwsh.exe treats `""` as `""` and thus the error in pwsh case.

``` powershell
# powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.22621.2428

PS> C:\local\msys64\usr\bin\echo.exe -Sdeps '{:deps {nrepl/nrepl {:mvn/version ""1.0.0""}}}'
-Sdeps {:deps {nrepl/nrepl {:mvn/version "1.0.0"}}}
```

``` powershell
# pwsh
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.0

PS > C:\local\msys64\usr\bin\echo.exe -Sdeps '{:deps {nrepl/nrepl {:mvn/version ""1.0.0""}}}'
-Sdeps {:deps {nrepl/nrepl {:mvn/version ""1.0.0""}}}
```

@shishini's hack is good. 

There is also a one liner hack that forces pwsh.exe to the old behavior with `$PSNativeCommandArgumentPassing = 'Legacy'` as mentioned in the stackoverflow post. So it is possible to get away with the following in `cider.el`, i.e. prepending that at the front of the clojure command execution at https://github.com/clojure-emacs/cider/blob/0134a0bc09c8cd332b100cb1c37d375204253ab6/cider.el#L813

``` elisp
(defun cider--powershell-encode-command (cmd-params)
  "Base64 encode the powershell command and jack-in CMD-PARAMS for clojure-cli."
  (let* ((quoted-params cmd-params)
         (command (format "$PSNativeCommandArgumentPassing = 'Legacy'; clojure %s" quoted-params))
         (utf-16le-command (encode-coding-string command 'utf-16le)))
    (format "-encodedCommand %s" (base64-encode-string utf-16le-command t))))
```

I've tested this to work with both powershell.exe and pwsh.exe. Perhaps this is a slightly better solution because we won't have to maintain two branches of logic for each powershell executable?

Irrespective of the solution, i've stumbled upon a problem with `cider-nrepl` and specifically one of its dependencies, [orchard](https://github.com/clojure-emacs/orchard), when I try to jack in with only pwsh in the PATH, it complains that it can't find `powershell`

```
;;  Startup: "d:/removeme/pwsh/pwsh.exe" -encodedCommand JABQAFMATgBhAHQAaQB2AGUAQwBvAG0AbQBhAG4AZABBAHIAZwB1AG0AZQBuAHQAUABhAHMAcwBpAG4AZwAgAD0AIAAnAEwAZQBnAGEAYwB5ACcAOwAgAGMAbABvAGoAdQByAGUAIAAtAFMAZABlAHAAcwAgACcAewA6AGQAZQBwAHMAIAB7AG4AcgBlAHAAbAAvAG4AcgBlAHAAbAAgAHsAOgBtAHYAbgAvAHYAZQByAHMAaQBvAG4AIAAiACIAMQAuADAALgAwACIAIgB9ACAAYwBpAGQAZQByAC8AYwBpAGQAZQByAC0AbgByAGUAcABsACAAewA6AG0AdgBuAC8AdgBlAHIAcwBpAG8AbgAgACIAIgAwAC4ANAAwAC4AMAAiACIAfQB9ACAAOgBhAGwAaQBhAHMAZQBzACAAewA6AGMAaQBkAGUAcgAvAG4AcgBlAHAAbAAgAHsAOgBtAGEAaQBuAC0AbwBwAHQAcwAgAFsAIgAiAC0AbQAiACIAIAAiACIAbgByAGUAcABsAC4AYwBtAGQAbABpAG4AZQAiACIAIAAiACIALQAtAG0AaQBkAGQAbABlAHcAYQByAGUAIgAiACAAIgAiAFsAYwBpAGQAZQByAC4AbgByAGUAcABsAC8AYwBpAGQAZQByAC0AbQBpAGQAZABsAGUAdwBhAHIAZQBdACIAIgBdAH0AfQB9ACcAIAAtAE0AOgBjAGkAZABlAHIALwBuAHIAZQBwAGwA
ERROR: Unhandled REPL handler exception processing message {:op init-debugger, :nrepl.middleware.print/stream? 1, :nrepl.middleware.print/print cider.nrepl.pprint/pprint, :nrepl.middleware.print/quota 1048576, :nrepl.middleware.print/buffer-size 4096, :nrepl.middleware.print/options {:right-margin 70}, :session a731b0b0-85e4-4c4f-9710-465d0e9671dc, :id 6}
Syntax error macroexpanding at (clojuredocs.clj:19:35).
	at clojure.lang.Compiler$InvokeExpr.eval(Compiler.java:3719)
...
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.io.IOException: Cannot run program "powershell.exe": CreateProcess error=2, The system cannot find the file specified
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1143)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1073)
	at orchard.util.os$run_commands.invokeStatic(os.clj:28)
	at orchard.util.os$run_commands.invoke(os.clj:24)
	at orchard.util.os$get_win_dirs.invokeStatic(os.clj:56)
	at orchard.util.os$get_win_dirs.invoke(os.clj:35)
	at orchard.util.os$cache_dir.invokeStatic(os.clj:73)
	at orchard.util.os$cache_dir.invoke(os.clj:58)
	at clojure.lang.AFn.applyToHelper(AFn.java:152)
	at clojure.lang.AFn.applyTo(AFn.java:144)
	at clojure.lang.Compiler$InvokeExpr.eval(Compiler.java:3714)
	... 158 more

...
```

This is due to orchard having hardcoded `powershell` as the executable of choice for invoking a command (I haven't look in detail why it does that)

https://github.com/clojure-emacs/orchard/blob/8169deff8abfdf68bccf3ceb7b1e953cc0a8918b/src/orchard/util/os.clj#L56

Thus, if we only have a `pwsh` executable in the PATH and no `powershell` executable, cider is going to fail at the start of `cider-jack-in` due to orchard.  @shishini are you able to jack in with pwsh without getting the above issue with your patch?

Thanks

vemv:
Thanks for another amazing analysis, @ikappaki 👏

Sounds like we should improve Orchard.

[This](https://github.com/clojure-emacs/orchard/blob/8169deff8abfdf68bccf3ceb7b1e953cc0a8918b/src/orchard/util/os.clj#L37-L53) code looks overly complicated to me...

Perhaps there's some pure-JVM approach we could use instead?

Else we could try/catch and fallback to some other choice.

But I wouldn't be keen on adding dynamically choosing pwsh.exe, powershell.exe, which might mean maintaining _two_ scripts?

shishini:
Hi @ikappaki 

Well I could only temporarily hide powershell.exe from my path by modifying the $env:path variable
I tried renaming powerhsell.exe but windows wont let me ( i didnt try to delete it, but i think i probably cant) 

Anyway, when I did i got this error, in my message buffer 


```
[nREPL] Starting server via "c:/Program Files/PowerShell/7/pwsh.exe" -encodedCommand YwBsAG8AagB1AHIAZQAgAC0AUwBkAGUAcABzACAAJwB7ADoAZABlAHAAcwAgAHsAbgByAGUAcABsAC8AbgByAGUAcABsACAAewA6AG0AdgBuAC8AdgBlAHIAcwBpAG8AbgAgACIAMQAuADAALgAwACIAfQAgAGMAaQBkAGUAcgAvAGMAaQBkAGUAcgAtAG4AcgBlAHAAbAAgAHsAOgBtAHYAbgAvAHYAZQByAHMAaQBvAG4AIAAiADAALgA0ADQALgAwACIAfQB9ACAAOgBhAGwAaQBhAHMAZQBzACAAewA6AGMAaQBkAGUAcgAvAG4AcgBlAHAAbAAgAHsAOgBtAGEAaQBuAC0AbwBwAHQAcwAgAFsAIgAtAG0AIgAgACIAbgByAGUAcABsAC4AYwBtAGQAbABpAG4AZQAiACAAIgAtAC0AbQBpAGQAZABsAGUAdwBhAHIAZQAiACAAIgBbAGMAaQBkAGUAcgAuAG4AcgBlAHAAbAAvAGMAaQBkAGUAcgAtAG0AaQBkAGQAbABlAHcAYQByAGUAXQAiAF0AfQB9AH0AJwAgAC0ATQA6AGMAaQBkAGUAcgAvAG4AcgBlAHAAbAA=
error in process sentinel: nrepl-server-sentinel: Could not start nREPL server: Execution error (IOException) at java.lang.ProcessImpl/create (ProcessImpl.java:-2).

CreateProcess error=2, The system cannot find the file specified



Full report at:

C:\Users\XXXXXX~1\AppData\Local\Temp\clojure-7741569257158880857.edn

 ("finished")
error in process sentinel: Could not start nREPL server: Execution error (IOException) at java.lang.ProcessImpl/create (ProcessImpl.java:-2).

CreateProcess error=2, The system cannot find the file specified



Full report at:

C:\Users\XXXXXX~1\AppData\Local\Temp\clojure-7741569257158880857.edn

 ("finished")
 
```
 
 And this line was one of the line in clojure-7741569257158880857.edn
```
     :message
    "Cannot run program \"powershell.exe\": CreateProcess error=2, The system cannot find the file specified"
 
```

So in short my patch doesnt fix this other issue, i think the orchard module should use `cider-clojure-cli-command`

ikappaki:
> Hi @ikappaki
> 
> Well I could only temporarily hide powershell.exe from my path by modifying the $env:path variable I tried renaming powerhsell.exe but windows wont let me ( i didnt try to delete it, but i think i probably cant)
> 
> Anyway, when I did i got this error, in my message buffer
> ...
> ```
>     "Cannot run program \"powershell.exe\": CreateProcess error=2, The system cannot find the file specified"
>  
> ```
> 
> So in short my patch doesnt fix this other issue, i think the orchard module should use `cider-clojure-cli-command`

It seems as if `powershell.exe` is included in every recent version of Windows as an integral part of the system? Therefore, it is highly improbable that Orchard's code will cause any issues, unless in deliberately engineered setups where the user has removed the powershell.exe from the PATH.

Out of curiosity, why are you opting for `pwsh.exe` when it looks like `powershell.exe` is already accessible in the PATH?

Thanks

shishini:
@ikappaki in retrospect, i probably should just have used powershell

but my luck kinda, i installed clojure cli in pwsh, in a folder only access by pwsh 
because its the new shell and its what i use, and from that point i came across those issues

again in retrospect, i should have thought of this as bash vs sh situation, where they can coexist and some scripts might only be compatible wish sh and not bash 

but i just kinda thought pwsh should work its what i use

ikappaki:
> @ikappaki in retrospect, i probably should just have used powershell
> 
> but my luck kinda, i installed clojure cli in pwsh, (in a folder only accessed by pwsh , only in pwsh module path) because its the new shell and its what i use, and from that point i came across those issues
> 
> again in retrospect, i should have thought of this as bash vs sh situation, where they can coexist and some scripts might only be compatible wish sh and not bash
> 
> but i just kinda thought pwsh should work its what i use

This is a very good reason why we should support pwsh.

It would also be nice if we can update the documentation at the below link with the above info?

 https://github.com/clojure-emacs/cider/blob/0134a0bc09c8cd332b100cb1c37d375204253ab6/doc/modules/ROOT/pages/basics/up_and_running.adoc?plain=1#L211-L215 to 

To summarise
1. There are real use cases to support the pwsh executable in CIDER.
2. This patch fixes it. There is also a one-liner alternative as mentioned above using `$PSNativeCommandArgumentPassing = 'Legacy'`
3. It will also be nice if we could update the documentation with regards to pwsh.
4. There is an issue with orchard if we only have pwsh.exe in the PATH but not powershell.exe. We think this is a non issue, since `powershell.exe` is always in the path and can't think of a reason why someone might manually remove it.

thanks

ikappaki:
> > This patch fixes it. There is also a one-liner alternative as mentioned above using $PSNativeCommandArgumentPassing = 'Legacy'
> 
> I think I would prefer this approach - the code would have no branches.
> 
> Do you agree @ikappaki ?

Hi @vemv, I'd prefer the one line edit, but I don't want to deter @shishini in their first check in :)

Issue #3591, opened 2023-11-30
Logback logs not showing in `*cider-log*` buffer


tatut:
## Expected behavior
Expected to see logback logs in `*cider-log*` buffer after adding an appender.
 
## Actual behavior

`*cider-log*` buffer is empty after doing logging calls.

## Steps to reproduce the problem
- Start REPL with `cider-jack-in-clj`
- Invoke `cider-log` and add appender with `a a` and `a` keys
- Evaluate `(clojure.tools.logging/info "hello")` in user namespace
- See "hello" log message in `*cider-log*  buffer

## Environment & Version information

- Logback dependency version `ch.qos.logback/logback-classic {:mvn/version "1.4.7"}`

### CIDER version information
CIDER 1.13.0-snapshot (package: 1.13.0-snapshot), nREPL 1.0.0
Clojure 1.11.1, Java 17.0.8

CIDER commit sha: 0134a0bc09c8cd332b100cb1c37d375204253ab6

### Lein / Clojure CLI version
Clojure CLI version 1.11.1.1347 

### Emacs version
GNU Emacs 29.1 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.60 Version 12.6.6 (Build 21G646)) of 2023-07-30

### Operating system
macOS 14.1.1 (23B81)

### JDK distribution
openjdk version "17.0.8" 2023-07-18

vemv:
> To view log events and stream them to your client, type es (Search log events) followed by s. This will open the *cider-log* buffer showing any log events captured thus far. It will also add a log consumer to this buffer, which receives newly-arriving log events.

Did you follow these steps?

https://docs.cider.mx/cider/debugging/logging.html

Cheers - V

tatut:
Enabling message logging with `nrepl-toggle-message-logging`, I can see the evaluation call and its responses in the `*nrepl-messages ...*` buffer, but nothing related to logs.

vemv:
I meant, you should ensure there's a consumer besides from an appender, it seems.

tatut:
> > To view log events and stream them to your client, type es (Search log events) followed by s. This will open the _cider-log_ buffer showing any log events captured thus far. It will also add a log consumer to this buffer, which receives newly-arriving log events.
> 
> Did you follow these steps?
>

Ok, it's a user error on my part 😌 

After doing the `e s` & `s`, the logs start appearing.

vemv:
No issue. Let us know how it goes!

[...]

PR #3592, opened 2023-12-02
Introduce `cider-log-dwim`

vemv:
Following https://github.com/clojure-emacs/cider/issues/3591, I was checking how to use CIDER Log in a more streamlined fashion.

My thinking is, in day-to-day usage, I should be able to run a single command that doesn't make me think about anything at all, and have the `*cider-log*` rendered.

I did get an appreciation for transient-mode, but ideally one would only need it for fine-grained manipulations.

So this PR:

* Introduces `defvar cider-log-preferred-framework nil`
  * Necessary for that "don't make me think" experience 😄
* Warns when Logview isn't installed
  *  TODO: also warn when no consumer is associated, if not using cider-log-dwim
* Introduces `cider-log-dwim`, meant as an idempotent command for showing the logs.

Leaving it as draft for now - UX tweaks welcome. There may be also room for improvents as for how I used the internals?

Cheers - V

### DIFF
--- a/cider-log.el
+++ b/cider-log.el
@@ -85,12 +85,17 @@
   :safe #'integerp
   :type 'integer)
 
-(defcustom cider-log-use-logview (fboundp 'logview-mode)
-  "Whether to use `logview-mode' or not."
+(defcustom cider-log-use-logview t
+  "Whether to use `logview-mode'."
   :group 'cider
   :package-version '(cider . "1.8.0")
   :safe #'booleanp
   :type 'boolean)
+
+(defun cider-log-use-logview ()
+  "Whether to use `logview-mode'."
+  (and (fboundp 'logview-mode)
+       cider-log-use-logview))
 
 (defvar logview-mode-map)
 (declare-function logview--guess-submode "logview" () t)
@@ -580,7 +585,7 @@ The KEYS are used to lookup the values and are joined by SEPARATOR."
                 (seq-doseq (buffer buffers)
                   (with-current-buffer buffer
                     (cider-log--insert-events buffer (list cider/log-event))
-                    (when (and cider-log-use-logview (not (logview-initialized-p)))
+                    (when (and (cider-log-use-logview) (not (logview-initialized-p)))
                       (let ((framework cider-log-framework)
                             (appender cider-log-appender)
                             (consumer cider-log-consumer))
@@ -738,11 +743,23 @@ The KEYS are used to lookup the values and are joined by SEPARATOR."
                (cider-log-consumer-display-name consumer)
                (cider-log-appender-display-name appender)))))
 
+(defvar cider-log-preferred-framework nil
+  "The name of the preferred framework.
+
+This is primarily intended to be set in .dir-locals.el.
+
+Example value: \"Logback\"")
+
 (defun cider-log-select-framework ()
   "Select the log framework."
   (let ((frameworks (cider-sync-request:log-frameworks)))
     (cond ((= 1 (length frameworks))
            (car frameworks))
+
+          ((and cider-log-preferred-framework
+                (cider-log-framework-by-name frameworks cider-log-preferred-framework))
+           (cider-log-framework-by-name frameworks cider-log-preferred-framework))
+
           (t (let ((name (cider-log--read-framework-name)))
                (cider-log-framework-by-name frameworks name))))))
 
@@ -812,7 +829,7 @@ The KEYS are used to lookup the values and are joined by SEPARATOR."
 
 (defvar cider-log-mode-map
   (let ((map (make-sparse-keymap))
-        (parent (if cider-log-use-logview logview-mode-map special-mode-map)))
+        (parent (if (cider-log-use-logview) logview-mode-map special-mode-map)))
     (set-keymap-parent map parent)
     (define-key map (kbd "C-c M-l a") #'cider-log-appender)
     (define-key map (kbd "C-c M-l c") #'cider-log-consumer)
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm not a big fan of `dwim` commands, as often people understand those different and they are usually something that could be resolved at a different level (e.g. by tweaking the base functionality). Perhaps it'd be nice to start by outlining what exactly are the problems this is aiming to resolve? E.g. I think the preferred-log-framework can easily be integrated into the existing command to start the mode.

vemv:
>  Perhaps it'd be nice to start by outlining what exactly are the problems this is aiming to resolve?

In my view, the default workflow represents an excess of steps for getting to see the logs.

...I like the book title `Don't Make Me Think` as a general UX guideline 😄

In specific terms:

* I don't want to think about the concepts of producer and consumer every time
* I won't want to see/use a `transient` UI for the "80%" use case
* I want to immediately see a buffer with the logs, practically every time (even if I might perform some other log-related action afterwards)

So, the proposed `cider-log-dwim` accomplishes that.

I understand the concern about dwim naming. However, one can observe:

* "show me the logs" is a pretty universal intent
  * people are used to `tail -f` / `less +F` and see logs immediately
* We can write down the 'contract' for what dwim means here, in the docstring.

Ultimately, we could find an alternative name, but e.g. `cider-log-setup-idempotently-and-show` might not exactly help.

`cider-log-show` maybe? (the "setup idempotently" part would be implied)

bbatsov:
My point was that we can probably enhance `cider-log` instead of introducing a different command or make it behave differently with a prefix arg (e.g. it can try to infer some of the things that are explicitly set today). Fewer commands usually result in better UX.

vemv:
`cider-log` currently isn't a vanilla function, but a `transient-define-prefix` thing. So one couldn't prevent the `transient` UI from popping up, even if it wasn't necessary for the "80% case".

Maybe we can rename `transient-define-prefix cider-log` to `transient-define-prefix cider-log-transient`, and move some of its logic to a new `cider-log` function which would do what `cider-log-dwim` does as of this PR.

But that would be a breaking change.

bbatsov:
Okay, let's go with the `cider-log-dwim` naming then and document clearly the differences between both commands.

vemv:
@r0man - last chance if you'd like to add any observation!

r0man:
Hi,

I just tried `cider-log-dwim`, but I'm not sure if we actually need
another command for this. I believe we mostly have what it tries to
do. And maybe we can enhance/rename the existing commands.

To setup CIDER log mode, you don't necessarily have to go through the
`cider-log` command. Here is what you could do instead:

1. Run `cider-jack-in` in the cider-nrepl repository.

2. Run `cider-log-event`. This will ask you for the log framework,
will add the log appender, to capture events, and switch to the
`*cider-log*` buffer.

3. Open the `cider.nrepl.middleware.log-tests` namespace and run the
tests (which do log some events).

4. Take a look at the `*cider-log*` buffer and you will see the events.

Those are the minimum steps needed to do. Step 1 and 4 are always
needed, so I think we can optimize 2 and 3.

I have a couple of suggestions:

- If you have `cider-log-framework-name` customized, you won't get
  asked for the framework in step 2. So maybe we should better
  document this. If you have exactly 1 framework then I believe you
  won't get asked at all. But this is a rare case, since on a JDK you
  always have the JUL framework (which nobody uses).

- Maybe the `cider-log` transient is too much for a new user. We could
  rename `cider-log` to `cider-log-menu`, and make `cider-log-event`
  the main entry point by renaming it to `cider-log`.

- I could imagine another issue for confusion is, that you have to
  attach a log appender before you actually see events appearing in
  the buffer. We could automatically add the appender if
  `cider-log-framework-name` is customized after jack in. But I would
  add this behind a feature flag. I think it really depends when you
  want logs to be captured. For example if you have a project that
  uses generative testing a lot, and some code paths are logging, you
  don't want this enabled by default.

- Maybe add a quick-start section to the documentation that shows the
  steps needed as a numbered list, by showing how to set it up with
  the JUL framework.

Wdyt?

bbatsov:
Great suggestions by @r0man - seems we really need to expands the docs a bit. Also:

> Maybe the cider-log transient is too much for a new user. We could
rename cider-log to cider-log-menu, and make cider-log-event
the main entry point by renaming it to cider-log.

I'd be fine with this suggestion, although perhaps we can come up with a better name than `cider-log-menu`.

vemv:
Thanks a lot for the input, both!

I'm not sure about the breaking changes part. Perhaps documenting the two main approaches (programmatic and transient-driven) would suffice.

We could leave renaming for later if it turned out to be a frequent pain point?

I'll close this PR and will be checking out `cider-log-event` - might propose doc tweaks afterwards.

vemv:
@r0man - I cider-log-event and while it works (honoring the defcustom and setting things up, as you mentioned), it's also transient-based.

It also doesn't show the cider-log buffer instantly (step 4) - personally I strongly feel it should show without user intervention.

I think I'll give another stab at this, keeping the feedback in mind.

r0man:
Hi @vemv, maybe we can change `cider-log-event` in a way that it only shows the transient if the `*cider-log-buffer*` is already open (to refine options, or adjust search options) and otherwise open it directly? 
Please feel free to experiment, whatever you think is most efficient. I have nothing against that.

vemv:
Got it! Feel free to check out https://github.com/clojure-emacs/cider/pull/3602

> maybe we can change cider-log-event in a way that it only shows the transient if the *cider-log-buffer* is already open (to refine options, or adjust search options) and otherwise open it directly?
Please feel free to experiment, whatever you think is most efficient. I have nothing against that.

I don't know, it seems to me that it would be desirable to have predictable UXs: one (`cider-log-show`) that never uses transient-mode, and one (`cider-log-event`) that always does.

That would seem clearer to users and maintainers alike.

PR #3593, opened 2023-12-02
pwsh 7.3 quoting rules compatibility

### LINKED ISSUE #3588: configure cider to use PowerShell 7.x  and not Windows PowerShell 5.1 on windows

ikappaki:
Hi,

could you please review patch to maintain compatibility with pwsh 7.3. It fixes #3588 based on #3590 by @shishini.

It uses the `PSNativeCommandArgumentPassing = 'Legacy'` flag as explained at https://stackoverflow.com/a/59036879.

A integration test was added for the same,  powershell quoting tests updated to accommodate the above var,  docs updated.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+- [#3588](https://github.com/clojure-emacs/cider/issues/3588): Compatibility with pwsh 7.3 quoting rules.
+
 ## 1.12.0 (2023-11-24)
 
 ### Changes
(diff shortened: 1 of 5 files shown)


vemv:
Thanks much @ikappaki @shishini!

A MELPA snapshot will be available soon. The docs are being deployed as well

PR #3600, opened 2023-12-22
Add scittle to runtime versions

benjamin-asdf:

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
   - Fixes JS completions for nested objects.
   - Bundles other fixes and reliability improvements.
 - Fix scittle jack in when using `cider-jack-in-clj` ()
+- [#3600](https://github.com/clojure-emacs/cider/pull/3600) Fix scittle jack in when using `cider-jack-in-clj`
 
 ## 1.12.0 (2023-11-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
   - Fixes JS completions for nested objects.
   - Bundles other fixes and reliability improvements.
 - Fix scittle jack in when using `cider-jack-in-clj` ()
-- [#3600](https://github.com/clojure-emacs/cider/pull/3600) Fix scittle jack in when using `cider-jack-in-clj`
+- [#3600](https://github.com/clojure-emacs/cider/pull/3600): Fix scittle jack in when using `cider-jack-in-clj.
 
 ## 1.12.0 (2023-11-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
   - Introduces Timbre compatibility for [CIDER Log Mode](https://docs.cider.mx/cider/debugging/logging.html).
   - Fixes JS completions for nested objects.
   - Bundles other fixes and reliability improvements.
+- Fix scittle jack in when using `cider-jack-in-clj` ()
 
 ## 1.12.0 (2023-11-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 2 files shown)


benjamin-asdf:
Scittle nrepl was saying a version but we don't take it into account for runtime classification.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - [#3588](https://github.com/clojure-emacs/cider/issues/3588): Compatibility with pwsh 7.3 quoting rules.
 - Bump the injected `enrich-classpath` to [1.19.0](https://github.com/clojure-emacs/enrich-classpath/compare/v1.18.6...v1.19.0).
 - Bump the `parseedn` required version to 1.2.1.
+- [#3600](https://github.com/clojure-emacs/cider/pull/3600) Fix scittle jack in when using `cider-jack-in-clj`
```
bbatsov, line 10: You're missing a `:` and a `.` here.

cider-connection.el
```diff
@@ -456,12 +457,19 @@ But helps us know if this is a nbb repl, or not."
     (when nrepl-versions
       (nrepl-dict-get nrepl-versions "nbb-nrepl"))))
 
+(defun cider--scittle-nrepl-version ()
+  "Likewise for scittle. See `cider--nbb-nrepl-version'."
```
bbatsov, line 461: I'd rather copy the docstring from `nbb`, so this can be read standalone.

vemv (commented):
I'd appreciate understanding, does scittle have a nrepl server? Mind to share some links?

If so ideally it would bave the same capabilities as a bb-nrepl server, namely `ns-list` and `classpath` ops, for correct Sesman functioning.

vemv:
You could also expand the user manual to show that this is possible, its use case, sample setup etc.

...It would also help us verify this works as intended - I've never used scittle but would gladly QA this branch

bbatsov:
@benjamin-asdf ping :-)

bob-just-bob:
https://github.com/babashka/scittle/tree/main/doc/nrepl

benjamin-asdf:
Hi, the write up is here. 
https://github.com/babashka/scittle/tree/main/doc/nrepl

vemv:
> But we removed this mainly because of the startup performance of doing a cljs eval or something?

Please check the git logs as I don't immediately recall what that thing was named, or why it was removed.

> The problem right now is that if somebody like Jack Rusher writes a new clojurescript repl, it won't work directly.

Not sure of how things came to that point (if they really are). Per:

```lisp
(defun cider-runtime ()
  "Return the runtime of the nREPl server."
  (cond
   ((cider--clojure-version) 'clojure)
   ((cider--babashka-version) 'babashka)
   ((cider--nbb-nrepl-version) 'nbb)
   ((cider--scittle-nrepl-version) 'scittle)
   (t 'generic)))
```

...the only cljs repls that need a special identification happen to come from the same author. There's a nice variety of cljs repls that don't need that.

bbatsov:
@benjamin-asdf Ping :-)

benjamin-asdf:
@bbatsov Jo I fixed the remaining lint error (I hope)

cider-connection.el
```diff
@@ -457,12 +458,22 @@ But helps us know if this is a nbb repl, or not."
     (when nrepl-versions
       (nrepl-dict-get nrepl-versions "nbb-nrepl"))))
 
+(defun cider--scittle-nrepl-version ()
+  "Retrieve the underlying connection's scittle version."
+  (with-current-buffer
```
bbatsov, line 463: I'd use here more compact indentation, as the code reads better this way - e.g. like in the function above for `nbb`.

bbatsov:
I think you'll need to rebase on top of `master`, as the changelog seems outdated on this branch. I also left a few small remarks.

benjamin-asdf:
@bbatsov Cool, I fixed the remaining issues

bbatsov:
Thanks!

PR #3602, opened 2024-01-01
Introduce `cider-log-show` function

vemv:
Rehash of https://github.com/clojure-emacs/cider/pull/3592, with misc feedback applied.

My take is summarized in the expanded documentation - there are three approaches, two of which are higher-level and therefore most recommended as an entrypoint to most users.

My thinking is that of the high-level commands, one (`cider-log-show`) would never use transient-mode, and the other one (`cider-log-event`) always would. This way we have predictable UXs for users.

Commits:

* Introduce `cider-log-show` function
* Prefer logview-mode unconditionally, and warn when it's absent 
  * This way, users will know that logview-mode exists.
  * There's still fallback, and optionality.

Happy new year,

-V


cider-log.el
```diff
@@ -1154,12 +1178,17 @@ the CIDER Inspector and the CIDER stacktrace mode.
       (let ((inhibit-read-only t))
         (erase-buffer)))))
 
+(defun cider-log-switch-to-buffer* (buffer)
```
r0man, line 1181: Should we make this private, e.g. use 2 dashes `cider-log--switch-to-buffer` ?

doc/modules/ROOT/pages/debugging/logging.adoc
```diff
@@ -103,6 +145,15 @@ https://stackoverflow.com/a/17842174/12711900[difficulties] with
 configuration changes made at runtime, which are wiped out by the
 Log4j2 reconfiguration mechanism.
 
+Timbre support is WIP as well.
+
+TIP: If your logging framework of choice is not currently supported by CIDER Log Mode,
```
r0man, line 150: Nice!

r0man (commented):
LGTM! Thanks for improving this @vemv !

cider-log.el
bbatsov: Yeah, probably. In general the external API for `cider-log` should be quite limited.

bbatsov:
The changes look good to me overall.

doc/modules/ROOT/pages/debugging/logging.adoc
```diff
@@ -20,34 +20,76 @@ logging related actions.
...
+available, and its use can be customized via `cider-log-use-logview`.
 
 == Usage
 
-To use CIDER Log Mode, type kbd:[C-c M-l l] or kbd:[M-x cider-log] in
-any buffer that has a CIDER https://github.com/vspinu/sesman[Sesman]
-session attached to it. The first time you run the command, it will
+To use CIDER Log Mode, there are a few ways to get started:
+
+* Traditionally, `M-x cider-log` has been considered the main entrypoint, although it can be slightly low-level for casual usage.
```
bbatsov, line 42: That's a brand new feature, so I don't think we can talk about "traditional" usage. :-) Let's just focus on the use-cases for each and not on whether something is newer/older.

bbatsov:
My only bigger feedback is that even after reading the docs and knowing what the functions do, I find the differences between `cider-log`, `cider-log-event` and `cider-log-show` unclear. 

You have to understand that most users will have 0 idea what `transient-mode` is, so you should not focus too much on implementation details, but rather on use-cases for the 3 commands.

doc/modules/ROOT/pages/debugging/logging.adoc
r0man: I was also thinking this, it's still marked as experimental in the docs.

bbatsov: As it should be. I don't want to mark it as stable until we've solidified the interface and I want us to have some leeway for breaking changes to get to the optimal interface.

vemv:
Thanks folks, all feedback applied.

Now there are two documented ways to get started, and one function left as advanced usage in another section.

All ideas certainly welcome as for how to further improve workflows.

Will be deploying docs in a few minutes.

vemv:
https://docs.cider.mx/cider/debugging/logging.html

r0man:
Nice! Thanks @vemv!

Issue #3606, opened 2024-01-16
Lots of compilation warnings since update to Emacs 29


DerGuteMoritz:
## Expected behavior

Loading CIDER should work without warnings.

## Actual behavior

Since upgrading from Emacs 28 to 29, I get a lot of compilation warnings when loading CIDER. They pop up in a `*Warnings*` buffer like this:

![2024-01-16_11-07-18_cider](https://github.com/clojure-emacs/cider/assets/19733/c664e949-cb71-426a-8c49-adc47e6f3130)

[Here's the full list](https://gist.github.com/DerGuteMoritz/6f29653c66e1ced286151937daee21f4).

## Steps to reproduce the problem

Load CIDER in Emacs 29.1 with native compilation enabled.

### CIDER version information

```
;; CIDER 1.13.0 (Santiago), nREPL 1.0.0
;; Clojure 1.11.1, Java 17.0.7
```

### Lein / Clojure CLI version

Clojure CLI version 1.11.1.1413

### Emacs version

GNU Emacs 29.1 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.18.0, Xaw3d scroll bars)

### Operating system

NixOS 24.05 (Uakari)

### JDK distribution

OpenJDK

vemv:
Thanks much for the heads up. As far as CircleCI goes, we run `test-lint` only in Emacs 28. It seems about time that we run it for 29 instead.

(probably it's not a good idea to lint for our entire matrix - I think I've seen issues that if solved for one version, then create an issue for a different version, and so on)

@ikappaki may enjoy fixing this, but that's only an invitation :)

bbatsov:
> (probably it's not a good idea to lint for our entire matrix - I think I've seen issues that if solved for one version, then create an issue for a different version, and so on)

I agreed. Generally it's best to lint only on the latest supported version, at typically the lint warning are accretive anyways.

purcell:
Somewhat relatedly: I might suggest adding Emacs HEAD to the matrix too, but allowing failures for that version. Then you kinda get a heads-up about potential future issues. I noticed that `sesman` doesn't compile in current Emacs HEAD, for example, so Cider is uninstallable there. It could be an Emacs bug, but who knows?

bbatsov:
Hmm, I thought we had done it, but I guess this changed amidst all the CI tweaks. It's always a good idea.

purcell:
[...]

PR #3607, opened 2024-01-17
Avoid `cider--error-phase-of-last-exception` recursive loop

### LINKED ISSUE #3605: `cider--error-phase-of-last-exception` can call itself when the cider-nrepl middleware is faulty
One thing I've noticed with recent reports is that whenever the cider-nrepl `analyze-stacktrace` op was faulty, `cider--error-phase-of-last-exception` would enter into a recursion loop.

Worth investigating

vemv:
Fixes #3605

Cheers - V

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -765,7 +765,7 @@ The handler simply inserts the result value in BUFFER."
   (let ((eval-buffer (current-buffer))
         (res "")
         (failed nil)
-        (error-phase-requested nil))
+        (error-phase-requested nil)) ;; avoid requesting the phase more than once - can happen if there are errors during the phase nrepl sync request.
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  ;; value handler:
                                  (lambda (_buffer value)
@@ -914,7 +914,7 @@ when `cider-auto-inspect-after-eval' is non-nil."
          (end (when end (copy-marker end)))
          (fringed nil)
          (res "")
-         (error-phase-requested nil))
+         (error-phase-requested nil)) ;; avoid requesting the phase more than once - can happen if there are errors during the phase nrepl sync request.
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  ;; value handler:
                                  (lambda (_buffer value)
@@ -963,7 +963,7 @@ when `cider-auto-inspect-after-eval' is non-nil."
 Optional argument DONE-HANDLER lambda will be run once load is complete."
   (let ((eval-buffer (current-buffer))
         (res "")
-        (error-phase-requested nil))
+        (error-phase-requested nil)) ;; avoid requesting the phase more than once - can happen if there are errors during the phase nrepl sync request.
     (nrepl-make-response-handler (or buffer eval-buffer)
                                  (lambda (buffer value)
                                    (cider--display-interactive-eval-result value 'value)
@@ -1093,7 +1093,7 @@ This is used by pretty-printing commands."
   ;; NOTE: cider-eval-register behavior is not implemented here for performance reasons.
   ;; See https://github.com/clojure-emacs/cider/pull/3162
   (let ((chosen-buffer (or buffer (current-buffer)))
-        (error-phase-requested nil))
+        (error-phase-requested nil)) ;; avoid requesting the phase more than once - can happen if there are errors during the phase nrepl sync request.
     (nrepl-make-response-handler
      chosen-buffer
      ;; value handler:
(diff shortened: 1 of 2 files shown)


cider-eval.el
```diff
@@ -905,17 +913,24 @@ when `cider-auto-inspect-after-eval' is non-nil."
...
                                    (cider--display-interactive-eval-result res 'value end))
+                                 ;; stdout handler:
                                  (lambda (_buffer out)
                                    (cider-emit-interactive-eval-output out))
+                                 ;; stderr handler:
                                  (lambda (buffer err)
                                    (cider-emit-interactive-eval-err-output err)
 
-                                   (let ((phase (cider--error-phase-of-last-exception buffer)))
+                                   (let ((phase (if error-phase-requested
```
vemv, line 930: This is the main fix

I could only immediately reproduce it for `cider-interactive-eval-handler`. But I used the same pattern in the other handlers as it's easy enough, and harmless.

cider-eval.el
```diff
@@ -845,12 +849,16 @@ REPL buffer.  This is controlled via
...
-                                                     conn)))
-          (nrepl-dict-get result "phase"))))))
+        (let ((nrepl-err-handler (lambda (&rest _))) ;; ignore any errors during this request to avoid any recursion
+              (nrepl-sync-request-timeout 4)) ;; ensure that this feature cannot possibly create an overly laggy UX
+          (when-let* ((result (nrepl-send-sync-request (thread-last (map-merge 'list
+                                                                               '(("op" "analyze-last-stacktrace"))
+                                                                               (cider--nrepl-print-request-map fill-column))
+                                                                    (seq-mapcat #'identity))
+                                                       conn
+                                                       'abort-on-input ;; favor responsiveness over this feature, in case something went wrong.
```
vemv, line 859: When I reproduced the issue, 'abort-on-input made it less terrible, so it seems wise to leave this here.

cider-eval.el
```diff
@@ -781,7 +782,10 @@ The handler simply inserts the result value in BUFFER."
                                    (when (and source-buffer
                                               (listp bounds)) ;; if it's a list, it represents bounds, otherwise it's a string (code) and we can't display the overlay
                                      (with-current-buffer source-buffer
-                                       (let* ((phase (cider--error-phase-of-last-exception buffer))
+                                       (let* ((phase (if error-phase-requested
```
bbatsov, line 785: It seems this `if` is duplicated a few times, so it might be good to make it some utility function instead.

cider-eval.el
```diff
@@ -845,12 +849,16 @@ REPL buffer.  This is controlled via
...
                        (cider-current-repl))))
       (when (cider-nrepl-op-supported-p "analyze-last-stacktrace" conn)
-        (when-let* ((result (nrepl-send-sync-request (thread-last (map-merge 'list
-                                                                             '(("op" "analyze-last-stacktrace"))
-                                                                             (cider--nrepl-print-request-map fill-column))
-                                                                  (seq-mapcat #'identity))
-                                                     conn)))
-          (nrepl-dict-get result "phase"))))))
+        (let ((nrepl-err-handler (lambda (&rest _))) ;; ignore any errors during this request to avoid any recursion
+              (nrepl-sync-request-timeout 4)) ;; ensure that this feature cannot possibly create an overly laggy UX
```
bbatsov, line 853: Why 4? That's a oddly specific number. :-)

bbatsov:
Apart from my small remarks the changes look reasonable to me.

cider-eval.el
bbatsov: You might want to add some comment explaining the need for the toggling between `t/nil` of this flag, as it's definitely not going to be obvious to most people.

cider-eval.el
vemv: Not sure if an abstraction could capture the necessary pattern

At most I can imagine a macro that ensured that a local variable `error-phase-requested` exists.

cider-eval.el
vemv: 4 - ample enough to give it time to be slow (especially if there was some sort of cache to be hit), tight enough to not negatively impact the UX

I'd be fine with 3 or 5 as well 😄

vemv:
Thanks for the review!

I was hesitant about an abstraction (even more so with https://github.com/clojure-emacs/cider/issues/3606), but if you felt it was needed we could keep it in the tracker.

bbatsov:
No, it's not a big deal - I'm just always wary of repeated code, as usually we get to a situation where someone forgets to make some change at every location. Anyways, I doubt this code is going to see much change going forward.

Issue #3587, opened 2023-11-16
Timbre logs can show in the minibuffer


vemv:
https://clojurians.slack.com/archives/C099W16KZ/p1700123927174669

Partial workaround: `(setq inhibit-message t)`

rrudakov:
Besides printing errors to the minibuffer, I'm also getting an overlay when I run tests.

<img width="1261" alt="cider" src="https://github.com/clojure-emacs/cider/assets/6093590/627a384b-4bbc-44ad-b896-d42084adbce9">

vemv:
Are you on Timbre as well?

rrudakov:
Yes.

vemv:
(Mostly a note to self)

The cause might be that Clojure compilation errors and Timbre errors are both captured under a regex.

So maybe stuff that was exclusively meant for Clojure compilation errors is also being triggered for Timbre.

vemv:
One thing I've noticed at the nrepl-messages level is that replies with Timbre err output correspond to overly old requests:

```
(-->
  id         "22"
  op         "test"
  session    "8286fdd4-64f1-4787-b6a7-22fd1af87ed0"
  time-stamp "2023-11-16 20:22:15.516106000"
  fail-fast  "true"
  load?      "true"
  ns         #("timbre" 0 6 (fontified t cider-locals nil cider-block-dynamic-font-lock t face font-lock-type-face))
  tests      ("foo")
)
(<--
  id         "15"
  session    "8286fdd4-64f1-4787-b6a7-22fd1af87ed0"
  time-stamp "2023-11-16 20:22:15.883850000"
  err        "2023-11-16T19:22:15.517Z vemvs-MacBook-Pro.local ERROR [timbre:8] - 234
"
)
```

Note how the request has id "22" but the error reply is for "15" - quite a few messages behind.

It's as if the original `eval` for 15 got "stuck" or was "greedy" in correctly releasing err.

We haven't changed stderr handling at nrepl level. We have https://github.com/clojure-emacs/cider/issues/3206 though.

I don't think this issue is releated to the overlay improvements since there's also the minifbuffer issue.

I'll see what happens if I downgrade Timbre.

rrudakov:
I've created a simple project with the following test:
```clojure
(ns com.example.app-test
  (:require
   [clojure.test :refer [deftest is]]
   [taoensso.timbre :as timbre]))

(deftest foo-test
  (try
    (throw (ex-info "Some error" {}))
[...]

PR #3608, opened 2024-01-21
Avoid overlays and `message`s on stderr that is unrelated to exception handling

vemv:
Fixes https://github.com/clojure-emacs/cider/issues/3587


vemv:
@rrudakov kindly LMK if this would fix it

(from my experiments/analysis it would, but I never got to reproduce the Timbre error)

It would solve the overlays and messages alike, since the `message` is produced by the overlay function

Cheers - V

vemv:
Mmm, I'll have to refine this approach, because it essentially relies on `analyze-last-stacktrace`.

There's no guarantee that the 'last stacktrace' is related to the 'most recent eval' - it could store an unrelated exception from prior evals.

This could be quite easily fixed by changing `analyze-last-stacktrace` -> `(analyze-stacktrace err)`

(So back to WIP - sorry for the noise @rrudakov)

PR #3609, opened 2024-01-21
Don't void `nrepl-err-handler`

vemv:
This was an extra measure included as part of https://github.com/clojure-emacs/cider/pull/3607

It resulted in `*cider-error*` not being shown, in certain scenarios.

That particular measure wasn't critical in avoiding the issue tackled in that PR.


vemv:
Not much to review as this simply undoes a recent addition - merging

PR #3615, opened 2024-01-31
Add a visual progress for when evaluating a directory recursively

J0sueTM:
Recently, since I'm working with a big clj project, I periodically use `cider-eval-all-files`, and it works fine. The only problem is that I always underestimate the time it takes to compile all the files in a directory, and endup evaluating another directory while the first one is still in progress.

That's what I try to solve with this PR. By adding a visual progress, it becomes easier to know if its OK to continue evaluating other modules of my code that depends on the thing that's currently being done.

BTW, I ran `eldev lint`, and some errors showed up, but none of them had to do with my changes.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1738,14 +1738,15 @@ passing arguments."
       (widen)
       (substring-no-properties (buffer-string)))))
 
-(defun cider-load-buffer (&optional buffer callback undef-all)
+(defun cider-load-buffer (&optional buffer callback undef-all progress)
   "Load (eval) BUFFER's file in nREPL.
 If no buffer is provided the command acts on the current buffer.  If the
 buffer is for a cljc file, and both a Clojure and ClojureScript REPL exists
 for the project, it is evaluated in both REPLs.
 Optional argument CALLBACK will override the default ‘cider-load-file-handler’.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
-all ns aliases and var mappings from the namespace before reloading it."
+all ns aliases and var mappings from the namespace before reloading it.
+PROGRESS is a formatted progress status thats shown on the loading message."
   (interactive (list (current-buffer) nil (equal current-prefix-arg '(4))))
   (setq buffer (or buffer (current-buffer)))
   ;; When cider-load-buffer or cider-load-file are called in programs the
@@ -1780,22 +1781,18 @@ all ns aliases and var mappings from the namespace before reloading it."
                                        (file-name-nondirectory filename)
                                        repl
                                        callback)))
-          (message "Loading %s..." filename))))))
+          (message (concat "Loading " progress "%s...") filename))))))
 
-(defun cider-load-file (filename &optional undef-all pos total)
+(defun cider-load-file (filename &optional undef-all progress)
   "Load (eval) the Clojure file FILENAME in nREPL.
 If the file is a cljc file, and both a Clojure and ClojureScript REPL
 exists for the project, it is evaluated in both REPLs.  The heavy lifting
 is done by `cider-load-buffer'.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
 all ns aliases and var mappings from the namespace before reloading it.
-POS and TOTAL, when not nil, are formatted as progress helpers in the
-interactive buffer while loading."
+PROGRESS is a formatted progress status thats shown on the loading message."
   (interactive (list
-                (read-file-name (concat "Load file"
-                                        (when (and pos total)
-                                          (format " [%d/%d]" pos total))
-                                        ": ")
+                (read-file-name "Load file:"
                                 nil nil nil
                                 (when (buffer-file-name)
                                   (file-name-nondirectory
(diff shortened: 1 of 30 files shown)


lanjoni (approved):
LGTM

cider-eval.el
```diff
@@ -1805,8 +1811,13 @@ Useful when the running nREPL on remote host.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
 all ns aliases and var mappings from the namespaces being reloaded"
   (interactive "DLoad files beneath directory: \nP")
-  (mapcar (lambda (file) (cider-load-file file undef-all))
-          (directory-files-recursively directory "\\.clj[cs]?$")))
+  (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
+         (total (length files)))
+    (mapcar (lambda (file)
+              (cider-load-file file undef-all
+                               (+ (cl-position file files :test #'equal) 1)
```
vemv, line 1818: It would seem to me that you could use a simple counter that would be incremented on each iteration

vemv (commented):
Seems like a reasonable addition - kudos!

Please simply satisfy the small feedback and ensure there's a green build.

Cheers - V

cider-eval.el
```diff
@@ -1782,15 +1782,21 @@ all ns aliases and var mappings from the namespace before reloading it."
                                        callback)))
           (message "Loading %s..." filename))))))
 
-(defun cider-load-file (filename &optional undef-all)
+(defun cider-load-file (filename &optional undef-all pos total)
```
bbatsov, line 1785: I've got some concern as to whether this should be in the public API of the command, given that it's needed only by the counter. Perhaps we can introduce a helper function to hide this?

bbatsov:
@J0sueTM ping :-)

J0sueTM:
Thanks for pinging @bbatsov completely forgot about it! My time hasn't been the most generous recently :sweat_smile: . I'll be coming with changes in the coming days.

cider-eval.el
```diff
@@ -1831,33 +1832,51 @@ all ns aliases and var mappings from the namespace before reloading it."
...
-          (directory-files-recursively directory "\\.clj[cs]?$")))
+  (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
+         (indexed-files (index-files files)))
+    (mapcar (lambda (indexed-file)
+              (let ((filename (car indexed-file))
+                    (progress (car (cdr indexed-file))))
+                (cider-load-file filename undef-all progress)))
+            indexed-files)))
+
+;; (cider-load-all-files "/home/jtm/dev/moclojer/moclojer-app/src/components/" t)
```
J0sueTM, line 1879:
```suggestion
```

forgot to rm a comment :sweat_smile:

cider-eval.el
```diff
@@ -1831,33 +1832,50 @@ all ns aliases and var mappings from the namespace before reloading it."
...
 Useful when the running nREPL on remote host.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
-all ns aliases and var mappings from the namespaces being reloaded"
+all ns aliases and var mappings from the namespaces being reloaded."
   (interactive "DLoad files beneath directory: \nP")
-  (mapcar (lambda (file) (cider-load-file file undef-all))
-          (directory-files-recursively directory "\\.clj[cs]?$")))
+  (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
+         (indexed-files (index-files files)))
+    (mapcar (lambda (indexed-file)
```
vemv, line 1873: While `number-sequence` is a nice function, I'd prefer it you used local mutation as it would seem easier to follow here.

An example:

```lisp
(let* ((files '("a" "b" "c"))
       (i 0))
        (mapcar (lambda (file)
                  (message (format "%s %s" file i))
                  (setq i (inc i)))
                files))
```

So, you can use the existing `mapcar` to mutate a local counter, so we have one mapcar instead of two (plus the helper function).

vemv (commented):
Thanks for picking up this!

Left a last round, please ensure a green build + no conflicts as well

cider-eval.el
J0sueTM: Thanks for pointing out. My elisp knowledge is really limited and that helped a lot. Thanks!

cider-eval.el
```diff
@@ -1831,33 +1832,41 @@ all ns aliases and var mappings from the namespace before reloading it."
...
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
-all ns aliases and var mappings from the namespaces being reloaded"
+all ns aliases and var mappings from the namespaces being reloaded."
   (interactive "DLoad files beneath directory: \nP")
-  (mapcar (lambda (file) (cider-load-file file undef-all))
-          (directory-files-recursively directory "\\.clj[cs]?$")))
+  (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
+         (file-count (length files))
+         (i 0))
+    (mapcar (lambda (file)
```
bbatsov, line 1865: It's better to use `mapc` instead of `mapcar` if all you care about are the side-effects.

cider-eval.el
```diff
@@ -1831,33 +1832,41 @@ all ns aliases and var mappings from the namespace before reloading it."
                                        (file-name-nondirectory filename)
                                        repl
                                        callback)))
-          (message "Loading %s..." filename))))))
+          (message (concat "Loading " progress "%s...") filename))))))
```
bbatsov, line 1835: You should not use `concat` in `message`, as you can just use format specifiers like `%s`, `%d`, etc. It's similar to Clojure's `format`.

cider-eval.el
bbatsov: Looking again at the code - probably using `dolist` would be even better than `mapc`, as it reflects best what we want to do anyways.

J0sueTM:
Do you guys think this change should be brought to `cider-load-all-project-ns` as well?

bbatsov:
Btw, I still don't like that we're now passing some text around. Might be better to make the message optional or move the progress reporting to the function that deals with many namespaces or to rename the `progress` param to something like `msg`, so it's clearer what this is.

Also - for a lot of files some status bar might be a better option (e.g. https://www.gnu.org/software/emacs/manual/html_node/elisp/Progress.html or spinner.el that we're using elsewhere in the code)

bbatsov:
@J0sueTM ping :-)

J0sueTM:
hey @bbatsov. I'm probably not able to finish this PR in the way you guys want. Nothing wrong with that, of course, I just don't know much about ELisp or have enough time in order to accomodate your quality requests :sweat_smile:.

I'm sorry for taking your time. Could you please either close the PR or backlog it so, if on interest of anyone, implement this better later?

Thanks for the great lib. I use it daily and it work flawlessly!

vemv:
It's no issue. Normally in these cases someone from us can pick up the branch and do whatever's needed, preserving attribution to you.

(Will do)

bbatsov:
@katomuso That's a fairly small and simple task, that might be interesting to you if you're looking for more tickets to tackle.

katomuso:
> @katomuso That's a fairly small and simple task, that might be interesting to you if you're looking for more tickets to tackle.

I will look into it sometime this month. Do I need to create a separate PR and link to this one, or is there a better way to do this?

bbatsov:
@katomuso You can just do a separate PR - as noted in my comments, I think it's much better (and simpler) to just use a real progress indicator (either a built-in one or coming from spinner.el, which is a CIDER dep) instead of indicating progress via a message printed in the minibuffer.

katomuso:
> @katomuso You can just do a separate PR - as noted in my comments, I think it's much better (and simpler) to just use a real progress indicator (either a built-in one or coming from spinner.el, which is a CIDER dep) instead of indicating progress via a message printed in the minibuffer.

As far as I understand, `spinner.el` can't show exact progress (number of evaluated files / number of total files). I will look into the ways in which this is possible, as it is not very reassuring to wait a minute or two without knowing how much longer it will take.

bbatsov:
That's the classic built-in approach - https://www.gnu.org/software/emacs/manual/html_node/elisp/Progress.html I mentioned spinner just because some indication that something is happening will still be an improvement over no indication. :-)

PR #3616, opened 2024-02-01
Avoid overlays and `message`s on stderr that is unrelated to exception handling

### LINKED ISSUE #3587: Timbre logs can show in the minibuffer

vemv:
* Introduce `cider-clojure-runtime-error-regexp`
  * Similar to our compilation errors regex - we didn't need runtime-oriented ones till now
  * [Ref](https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/clj/clojure/main.clj#L326-L340)
* Avoid overlays and `message`s on stderr that is unrelated to exception handling
  * Fixes #3587

After https://github.com/clojure-emacs/cider/pull/3608 and other attempts I'm finally confident about this solution - it's simple, robust and aligned with other similar solutions.

@rrudakov QAing would be most welcome.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [#3605](https://github.com/clojure-emacs/cider/issues/3605): avoid `cider--error-phase-of-last-exception` recursive loop.  
 - [#3613](https://github.com/clojure-emacs/cider/issues/3613): adapt `cider-completion-context.el` to upstream changes in Compliment.  
+- [#3587](https://github.com/clojure-emacs/cider/issues/3587): avoid overlays and `message`s on stderr that is unrelated to exception handling.
 
 ## 1.13.0 (2024-01-14)
 
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -615,6 +615,8 @@ It delegates the actual error content to the eval or op handler."
                                   (optional ":" (group-n 4 (one-or-more digit)))
                                   " - "))
 
+;; Please keep this in sync with `cider-clojure-compilation-error-regexp',
+;; which is a subset of these regexes.
 (defconst cider-clojure-compilation-regexp
   (eval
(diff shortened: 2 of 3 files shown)


rrudakov:
I've done a quick test with the minimal project. The problem is not reproducible. I'll also try to test it with real project for some time.

rrudakov:
> I've done a quick test with the minimal project. The problem is not reproducible. I'll also try to test it with real project for some time.

I still can see it on a real project:

image

But those are reflection warnings. I don't see timbre error logs in the overlay anymore. Not sure if it's an expected behaviour though.

I have the following in my `user.clj`:
```clojure
(alter-var-root #'*warn-on-reflection* (constantly true))
```

nrepl messages:

```
(<--
  id         "4"
  session    "bb23f216-4dc0-4420-b48c-1df6be8e6b08"
  time-stamp "2024-02-01 16:46:40.763034000"
  err        "Performance warning, com/adgoji/mollie/client.clj:86:22 - ca..."
)
(<--
  id         "11"
  session    "bb23f216-4dc0-4420-b48c-1df6be8e6b08"
  time-stamp "2024-02-01 16:46:44.966841000"
  err        "Reflection warning, net/cgrand/xml.clj:85:4 - call to method..."
)
```

rrudakov:
I'd say it's good enough for me, I only see those messages once after starting the REPL.

vemv:
That was super useful. Thankfully we have a regex for reflection warnings, so it's an easy fix.

Commit incoming

vemv:
Ready again

rrudakov:
I've just checked, looks great now, reflection/performance warnings are also gone.

PR #3617, opened 2024-02-01
Address most Emacs 29 warnings

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3606

Addresses indentation and other warnings.

Sadly we cannot immediately upgrade the CI linting, because:

* Emacs 29 deprecates lax-plist-* functions
* The alternative is based on an arity that isn't present in Emacs < 29.

Similarly, xref-go-back deprecates xref-pop-marker-stack, while the former only exists on Emacs 29, so there's no possible fix, AFAIK.

However, a great deal of warnings is now fixed, which should be less noisy to users, and pave the path future refinements.

Cheers - V

### DIFF
--- a/cider-browse-ns.el
+++ b/cider-browse-ns.el
@@ -266,12 +266,11 @@ list of items."
 (defun cider-browse-ns--render-items (items)
   "Render the nrepl-dict ITEMS to the browse-ns buffer."
   (let* ((max-length (cider-browse-ns--column-width items)))
-    (cl-labels
-        ((keys-from-pred
-          (pred items)
-          (nrepl-dict-keys (nrepl-dict-filter (lambda (_ var-meta)
-                                                (funcall pred var-meta))
-                                              items))))
+    (cl-labels ((keys-from-pred
+                 (pred items)
+                 (nrepl-dict-keys (nrepl-dict-filter (lambda (_ var-meta)
+                                                       (funcall pred var-meta))
+                                                     items))))
       (cond
        ((eql cider-browse-ns-group-by 'type)
         (let* ((func-keys (keys-from-pred #'cider-browse-ns--meta-function-p items))
@@ -453,7 +452,7 @@ var-meta map."
 
 (defun cider-browse-ns--thing-at-point ()
   "Get the thing at point.
-Return a list of the type ('ns or 'var) and the value."
+Return a list of the type (\='ns or \='var) and the value."
   (let ((ns-p (get-text-property (point) 'ns))
         (line (car (split-string (string-trim (thing-at-point 'line)) " "))))
     (if (or ns-p (string-match "\\." line))
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -64,8 +64,8 @@
 
 (defcustom cider-show-error-buffer t
   "Control the popup behavior of cider stacktraces.
-The following values are possible t or 'always, 'except-in-repl,
-'only-in-repl.  Any other value, including nil, will cause the stacktrace
+The following values are possible t or \='always, \='except-in-repl,
+\='only-in-repl.  Any other value, including nil, will cause the stacktrace
 not to be automatically shown.
 
 Irrespective of the value of this variable, the `cider-error-buffer' is
@@ -88,7 +88,7 @@ in order to void its effect."
 (defcustom cider-auto-jump-to-error t
   "Control the cursor jump behavior in compilation error buffer.
 When non-nil automatically jump to error location during interactive
-compilation.  When set to 'errors-only, don't jump to warnings.
+compilation.  When set to \='errors-only, don\='t jump to warnings.
 When set to nil, don't jump at all."
   :type '(choice (const :tag "always" t)
                  (const errors-only)
@@ -120,7 +120,7 @@ Only applies when the *cider-inspect* buffer is currently visible."
 (defcustom cider-save-file-on-load 'prompt
   "Controls whether to prompt to save the file when loading a buffer.
 If nil, files are not saved.
-If 'prompt, the user is prompted to save the file if it's been modified.
+If \='prompt, the user is prompted to save the file if it\='s been modified.
 If t, save the file without confirmation."
   :type '(choice (const prompt :tag "Prompt to save the file if it's been modified")
                  (const nil :tag "Don't save the file")
@@ -405,13 +405,12 @@ Returns the position at which PROPERTY was found, or nil if not found."
 _ARG and _RESET are ignored, as there is only ever one compilation error.
 They exist for compatibility with `next-error'."
   (interactive)
-  (cl-labels ((goto-next-note-boundary
-               ()
-               (let ((p (or (cider-find-property 'cider-note-p)
-                            (cider-find-property 'cider-note-p t))))
-                 (when p
-                   (goto-char p)
-                   (message "%s" (get-char-property p 'cider-note))))))
+  (cl-labels ((goto-next-note-boundary ()
+                                       (let ((p (or (cider-find-property 'cider-note-p)
+                                                    (cider-find-property 'cider-note-p t))))
+                                         (when p
+                                           (goto-char p)
+                                           (message "%s" (get-char-property p 'cider-note))))))
     ;; if we're already on a compilation error, first jump to the end of
     ;; it, so that we find the next error.
     (when (get-char-property (point) 'cider-note-p)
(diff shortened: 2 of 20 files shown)


cider-mode.el
```diff
@@ -594,10 +594,8 @@ re-visited."
     (if-let* ((meta (cider-resolve-var ns symbol-name))
               (indent (or (nrepl-dict-get meta "style/indent")
                           (nrepl-dict-get meta "indent"))))
-        (let ((format (format ":indent metadata on ‘%s’ is unreadable! \nERROR: %%s"
-                              symbol-name)))
-          (with-demoted-errors format
-            (cider--deep-vector-to-list (read indent))))
+        (with-demoted-errors "Some :indent metadata is unreadable! \nERROR: %s"
```
vemv, line 597: Sadly the byte compiler kept complaining about the original shape of code, despite the many variations I tried.

I think it's ultimately because this linter assumes a string literal as an argument to `with-demoted-errors`.

So we have to give up on the informative string.

cider-mode.el
```diff
@@ -776,41 +774,40 @@ with the given LIMIT."
...
-                       (let ((is-macro (nrepl-dict-get meta "macro"))
-                             (is-function (or (nrepl-dict-get meta "fn")
-                                              (nrepl-dict-get meta "arglists"))))
-                         (cond ((and do-macro is-macro)
-                                (push sym macros))
-                               ((and do-function is-function)
-                                (push sym functions))
-                               ((and do-var (not is-function) (not is-macro))
-                                (push sym vars)))))))))
+    (cl-labels ((handle-plist (plist)
```
bbatsov, line 777: The indentation here looks very wrong to me. If `cl-labels` is so broken in terms of indentation now I guess it'd be better to use a private function instead.

cider-util.el
```diff
@@ -438,7 +438,7 @@ plugin or dependency with:
 (defvar cider-version)
 
 (defconst cider-manual-url "https://docs.cider.mx/cider/%s"
-  "The URL to CIDER's manual.")
+  "The URL to CIDER\='s manual.")
```
bbatsov, line 441: Hmm, that's not really a symbol.

cider-util.el
```diff
@@ -451,7 +451,7 @@ plugin or dependency with:
     (concat (cider-version-sans-patch) "/")))
 
 (defun cider-manual-url ()
-  "The CIDER manual's url."
+  "The CIDER manual\='s url."
```
bbatsov, line 454: Ditto.

cider-util.el
```diff
@@ -474,7 +474,7 @@ to."
     (buffer-string)))
 
 (defconst cider-refcard-url "https://github.com/clojure-emacs/cider/raw/%s/refcard/cider-refcard.pdf"
-  "The URL to CIDER's refcard.")
+  "The URL to CIDER\='s refcard.")
```
bbatsov, line 477: Ditto.

bbatsov:
> Similarly, xref-go-back deprecates xref-pop-marker-stack, while the former only exists on Emacs 29, so there's no possible fix, AFAIK.

Normally in such cases I copied stuff from the newer Emacs into `cider-compat.el`. (not sure if it currently exists) Backporting newer code is the only reasonable solution in such cases IMO. 

I noticed that many of the changes are related to `thread-first/last` and `cl-labels`? Did they change their indentation settings? Seems like a pretty weird upstream change if so.

cider-mode.el
vemv: I agree that this one looks odd, I couldn't produce any other accepted indentation.

> it'd be better to use a private function instead.

👍

cider-util.el
vemv: One has to use it for all `'`s no matter what (which yes, looks fairly ugly)

https://emacs.stackexchange.com/posts/73048/revisions

vemv (commented):
> I noticed that many of the changes are related to thread-first/last and cl-labels? Did they change their indentation settings? 

As far as I could try with Eldev (i.e. no personal settings involved), yes.

> Normally in such cases I copied stuff from the newer Emacs into cider-compat.el. (not sure if it currently exists) Backporting newer code is the only reasonable solution in such cases IMO.

Yes, it was removed in c60598fa4df6cdd3331c29b8e319cc23de1b7cc6

I could give it a shot. However we'd redefine ` lax-plist-*` functions to receive an extra arity. That would seem a bit hacky and perhaps we'd be better off with the warnings.

We could upgrade the CI linter to emacs 29, but running a search/replace beforehand in the CI job for such cases? So that it doesn't complain for the cases that aren't reasonably fixable.

cider-mode.el
bbatsov: I'm always worried that if people edit the code on different versions of Emacs they'd get different indentations, which always results in a more painful contribution process. (e.g. I'm still using Emacs 28, as I didn't have a big incentive to move to Emacs 29)

vemv: In general, for thread-* and cl-labels I rearranged things in such a way that Emacs 29 and <29 will agree, which is why the PR's build remain green despite the indentation changes and no CI changes.

(for thread-*, it also happens to look more clojurey)

bbatsov:
I just saw this in the Emacs 29 changelog:

```
Indentation of 'cl-flet' and 'cl-labels' has changed.
These forms now indent like this:

    (cl-flet ((bla (x)
                (* x x)))
      (bla 42))

This change also affects 'cl-macrolet', 'cl-flet*' and
'cl-symbol-macrolet'.
```

cider-browse-ns.el
```diff
@@ -266,12 +266,11 @@ list of items."
...
   "Render the nrepl-dict ITEMS to the browse-ns buffer."
   (let* ((max-length (cider-browse-ns--column-width items)))
-    (cl-labels
-        ((keys-from-pred
-          (pred items)
-          (nrepl-dict-keys (nrepl-dict-filter (lambda (_ var-meta)
-                                                (funcall pred var-meta))
-                                              items))))
+    (cl-labels ((keys-from-pred
+                 (pred items)
```
bbatsov, line 270: I think we should pull the params on the same line as the function name.

cider-eval.el
```diff
@@ -405,13 +405,12 @@ Returns the position at which PROPERTY was found, or nil if not found."
...
 They exist for compatibility with `next-error'."
   (interactive)
-  (cl-labels ((goto-next-note-boundary
-               ()
-               (let ((p (or (cider-find-property 'cider-note-p)
-                            (cider-find-property 'cider-note-p t))))
-                 (when p
-                   (goto-char p)
-                   (message "%s" (get-char-property p 'cider-note))))))
+  (cl-labels ((goto-next-note-boundary ()
```
bbatsov, line 408: Weirdly this doesn't look like the indentation in the Emacs 29 changelog.

p4v4n:
> * Emacs 29 deprecates lax-plist-* functions
> * The alternative is based on an arity that isn't present in Emacs < 29.

IIUC it's possible to just replace `lax-plist-get/put` with `plist-get/put`.
`plist-get/put` got an optional 3rd argument in emacs29+ but with 2 arguments they behave the same as `lax-plist-*`

cider-eval.el
p4v4n: I am using 29.2 and can confirm that the indentation is not correct here.

vemv:
> Edit: Not exactly the same but similar. lax-plist-get uses equal for comparison and plist-get uses eq.

It may have been that prevented me from using a direct fix - I recall trying something and see the builds fail for E<29

---

I'll re-take this PR in March

metayan:
Just commenting so that this PR doesn't get closed, hoping that it will get merged eventually.
Feel free to scold me if this is the wrong thing to do...
Or inform me if there is a better way. ;)

bbatsov:
The PR has been stuck for a while, so I have some doubts it will get to the finish line. That being said - eventually we'll address those warnings. It's a fairly simple thing to do, and it wasn't prioritized mostly because the warnings are harmless.

metayan:
Understood, and yes, the warnings are harmless - I just know from experience that lots of harmless warnings might have a tendency to hide important ones.

jcs090218:
I'm seeing a lot of warnings with this package. 🤔 Hopefully, the maintainer can take them a bit more seriously, since they could lead to potential issues. 😅

bbatsov:
@jcs090218 I pan to look into this soon.

bbatsov:
I've pushed a few changes that mostly address the `lax-plist-*` deprecations and a few other small problems. Now most of the warnings remaining are about escaping `'` in docstrings with `=\`, which I find pretty weird but I'll do at some point.

alexander-yakushev:
@bbatsov You've hidden most of the warnings behind custom wrapper functions, right? I think this can be closed now; it would require a complete do-over anyway.

PR #3623, opened 2024-02-25
introduced variable `cider-clojure-cli-global-aliases`

behrica:
This PR introduces a variable `cider-clojure-cli-global-aliases`
Discussed here: https://clojurians.slack.com/archives/C0617A8PQ/p1708857780583699

It allows a user to have a "global" system wide set of aliases, which get added by cider to each repl start.
It allows to combine "project specific aliases" and" global aliases"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## master (unreleased)
 
-- new configuration variable `cider-clojure-cli-global-aliases`
+
 
 ### New features
-
+- [#3632](https://github.com/clojure-emacs/cider/pull/3623): new configuration variable `cider-clojure-cli-global-aliases`
 - [#3366](https://github.com/clojure-emacs/cider/pull/3366): Support display of error overlays with #dbg! and #break! reader macros.
 - [#3622](https://github.com/clojure-emacs/cider/pull/3461): Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode).
   - The `clojure-mode` dependency is still required for CIDER to function.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- new configuration variable `cider-clojure-cli-global-aliases`
+
 ### New features
 
 - [#3622](https://github.com/clojure-emacs/cider/pull/3461): Basic support for using CIDER from [clojure-ts-mode](https://github.com/clojure-emacs/clojure-ts-mode)
--- a/cider.el
+++ b/cider.el
@@ -176,14 +176,15 @@ then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
 
 (defcustom cider-clojure-cli-global-aliases
   nil
-  "A list of global aliases to include when using the clojure cli.
-It's value will be prepended to the value of `cider-clojure-cli-aliases`
+  "Global aliases to include when jacking in with the clojure CLI.
+This value should be a string of the form \":foo:bar\", and
+will be prepended to the value of `cider-clojure-cli-aliases'.
 Alias names should be of the form \":foo:bar\".
 Leading \"-A\" \"-M\" \"-T\" or \"-X\" are stripped from aliases
 then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
   :type 'string
   :safe #'stringp
-  :package-version '(cider . "1.1"))
+  :package-version '(cider . "1.14"))
 
 
 (defcustom cider-shadow-cljs-command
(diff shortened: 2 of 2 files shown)


vemv:
Looks like a reasonable start, let's see what @bbatsov thinks as well

(note as future steps: there's test coverage for this area - easy to extend)

bbatsov:
I get the intent, but I'm curious why is the current variable considered to be project-specific, given that nothing about it implies it. Might be better to actually add something name `cider-clojure-cli-project-aliases` instead that's meant to be used only via `.dir-locals.el`. 

If we go with your approach the documentation for the existing defcustom will need to be updated. I also think this will need some coverage in the user manual.

vemv:
> Might be better to actually add something name cider-clojure-cli-project-aliases instead that's meant to be used only via .dir-locals.el.

This would work as well (it's basically the same endgame as proposed in this PR - two variables).

However over the last few years we've taught people to set `cider-clojure-cli-aliases` in .dir-locals.el (and/or that's what people do anyway), so changing the intended semantics for `cider-clojure-cli-aliases` would seem overly disruptive.

bbatsov:
Yeah, that's fair. As I never used `clojure-cli` much, it's hard for me to assess whether people have more need for global aliases or project aliases (e.g. how like it is that those differ at all, as with many build system I have more or less the same setup across many projects)

behrica:
I think that most people have need for 2 things:
- global alias for some functions they want to have always available, in any project
- project specific alias, like `test` , `dev` in slide variations of name

-> ideally you want be able to "add" them together
-> the globals, should usually be first (they do not define :main entry points, the add functions to ns 'user for example)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- new configuration variable `cider-clojure-cli-global-aliases`
```
bbatsov, line 5: This should be under new features and mention the PR that introduces it.

cider.el
```diff
@@ -173,6 +173,19 @@ then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
...
+(defcustom cider-clojure-cli-global-aliases
+  nil
+  "A list of global aliases to include when using the clojure cli.
+It's value will be prepended to the value of `cider-clojure-cli-aliases`
+Alias names should be of the form \":foo:bar\".
+Leading \"-A\" \"-M\" \"-T\" or \"-X\" are stripped from aliases
+then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.1"))
```
bbatsov, line 186: This should say 1.14

cider.el
```diff
@@ -850,14 +863,19 @@ your aliases contain any mains, the cider/nrepl one will be the one used."
...
             deps-quoted
-            (if cider-clojure-cli-aliases
-                ;; remove exec-opts flags -A -M -T or -X from cider-clojure-cli-aliases
-                ;; concatenated with :cider/nrepl to ensure :cider/nrepl comes last
-                (let ((aliases (format "%s" (replace-regexp-in-string "^-\\(A\\|M\\|T\\|X\\)" "" cider-clojure-cli-aliases))))
-                  (if (string-prefix-p ":" aliases)
-                      aliases
-                    (concat ":" aliases)))
-              "")
+            (let ((final-cider-clojure-cli-aliases
```
bbatsov, line 866: I'd probably name this just `clojure-cli-aliases` or `combined-aliases`.

bbatsov: Btw, we can also move the who part about aliases to a helper function, as it's fairly big now and it'd also be easier to test as a separate unit.

vemv:
Hi @behrica, feel free to polish this PR these days, would be good timing in preparation for the next major CIDER release.

cider.el
```diff
@@ -173,6 +173,19 @@ then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
   :safe #'stringp
   :package-version '(cider . "1.1"))
 
+
+(defcustom cider-clojure-cli-global-aliases
+  nil
+  "A list of global aliases to include when using the clojure cli.
```
yuhan0, line 179: This wording (also the existing docstring for cider-clojure-cli-aliases) seems potentially confusing - one might mistake them for an actual `'(":list" ":of" ":strings")`. I'd suggest something like:
> Global aliases to include when jacking in with the clojure CLI.
> This value should be a string of the form \":foo:bar\", and
> will be prepended to the value of `cider-clojure-cli-aliases'.

cider.el
behrica: I created a helper function extracting the code accordingly

cider.el
behrica: I updated comment as proposed.

Regarding the "type" I was not sure about it, so left it as is

PR #3624, opened 2024-03-02
Support cider.clj-reload/reload ops

filipesilva:
See https://github.com/clojure-emacs/cider-nrepl/pull/850 for more.

Notes:
- this ns only had one trivial test, I didn't add more
- listing shows existing linting errors, I made sure I wasn't adding more

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -118,6 +118,14 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-ns-refresh-tool 'tools.namespace
+  "Which tool to use for ns refresh.
+Current options: tools.namespace and clj-reload."
+  :group 'cider
+  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
+                 (const :tag "clj-reload https://github.com/tonsky/clj-reload" clj-reload))
+  :package-version '(cider . "1.13.1"))
+
 (defun cider-ns--present-error (error)
   "Render the `ERROR' stacktrace,
 and jump to the adequate file/line location,
@@ -156,7 +164,7 @@ presenting the error message as an overlay."
 
 (defun cider-ns-refresh--handle-response (response log-buffer)
   "Refresh LOG-BUFFER with RESPONSE."
-  (nrepl-dbind-response response (out err reloading status error error-ns after before)
+  (nrepl-dbind-response response (out err reloading progress status error error-ns after before)
     (cl-flet* ((log (message &optional face)
                     (cider-emit-into-popup-buffer log-buffer message face t))
 
@@ -183,6 +191,9 @@ presenting the error message as an overlay."
 
        (reloading
         (log-echo (format "Reloading %s\n" reloading) 'font-lock-string-face))
+
+       (progress
+        (log-echo progress 'font-lock-string-face))
 
        ((member "reloading" (nrepl-dict-keys response))
         (log-echo "Nothing to reload\n" 'font-lock-string-face))
@@ -222,6 +233,19 @@ Its behavior is controlled by `cider-ns-save-files-on-refresh' and
               (seq-some (lambda (dir)
                           (file-in-directory-p buffer-file-name dir))
                         dirs)))))))
+
+(defun cider-ns-refresh--refresh-op (op-name)
+  "Return the refresh operation to use.
+Based on OP-NAME and the value of cider-ns-refresh-tool defcustom."
+  (list "op"
+        (cond
+         ((eq cider-ns-refresh-tool 'tools.namespace)
+          op-name)
+
+         ((eq cider-ns-refresh-tool 'clj-reload)
+          (cond ((string= op-name "refresh")       "cider.clj-reload/reload")
+                ((string= op-name "refresh-all")   "cider.clj-reload/reload-all")
+                ((string= op-name "refresh-clear") "cider.clj-reload/reload-clear"))))))
 
 ;;;###autoload
 (defun cider-ns-reload (&optional prompt)
@@ -275,6 +299,7 @@ refresh functions (defined in `cider-ns-refresh-before-fn' and
   (interactive "p")
   (cider-ensure-connected)
   (cider-ensure-op-supported "refresh")
+  (cider-ensure-op-supported "cider.clj-reload/reload")
   (cider-ns-refresh--save-modified-buffers)
   (let ((clear? (member mode '(clear 16)))
         (refresh-all? (member mode '(refresh-all 4)))
@@ -292,11 +317,11 @@ refresh functions (defined in `cider-ns-refresh-before-fn' and
                                           nil
                                           t))
           (when clear?
-            (cider-nrepl-send-sync-request '("op" "refresh-clear") conn))
+            (cider-nrepl-send-sync-request (cider-ns-refresh--refresh-op "refresh-clear") conn))
           (cider-nrepl-send-request
            (thread-last
              (map-merge 'list
-                        `(("op" ,(if refresh-all? "refresh-all" "refresh")))
+                        `(,(cider-ns-refresh--refresh-op (if refresh-all? "refresh-all" "refresh")))
                         (cider--nrepl-print-request-map fill-column)
                         (when (and (not inhibit-refresh-fns) cider-ns-refresh-before-fn)
                           `(("before" ,cider-ns-refresh-before-fn)))
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -118,13 +118,13 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
-(defcustom cider-ns-refresh-tool 'tools.namespace
+(defcustom cider-ns-code-reload-tool 'tools.namespace
   "Which tool to use for ns refresh.
 Current options: tools.namespace and clj-reload."
   :group 'cider
-  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
+  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools.namespace)
                  (const :tag "clj-reload https://github.com/tonsky/clj-reload" clj-reload))
-  :package-version '(cider . "1.13.1"))
+  :package-version '(cider . "1.14.0"))
 
 (defun cider-ns--present-error (error)
   "Render the `ERROR' stacktrace,
@@ -234,18 +234,18 @@ Its behavior is controlled by `cider-ns-save-files-on-refresh' and
                           (file-in-directory-p buffer-file-name dir))
                         dirs)))))))
 
-(defun cider-ns-refresh--refresh-op (op-name)
-  "Return the refresh operation to use.
-Based on OP-NAME and the value of cider-ns-refresh-tool defcustom."
+(defun cider-ns--reload-op (op-name)
+  "Return the reload operation to use.
+Based on OP-NAME and the value of cider-ns-code-reload-tool defcustom."
   (list "op"
-        (cond
-         ((eq cider-ns-refresh-tool 'tools.namespace)
-          op-name)
+        (if (eq cider-ns-code-reload-tool 'tools.namespace)
+            (cond ((string= op-name "reload")       "refresh")
+                  ((string= op-name "reload-all")   "refresh-all")
+                  ((string= op-name "reload-clear") "refresh-clear"))
 
-         ((eq cider-ns-refresh-tool 'clj-reload)
-          (cond ((string= op-name "refresh")       "cider.clj-reload/reload")
-                ((string= op-name "refresh-all")   "cider.clj-reload/reload-all")
-                ((string= op-name "refresh-clear") "cider.clj-reload/reload-clear"))))))
+          (cond ((string= op-name "reload")       "cider.clj-reload/reload")
+                ((string= op-name "reload-all")   "cider.clj-reload/reload-all")
+                ((string= op-name "reload-clear") "cider.clj-reload/reload-clear")))))
 
 ;;;###autoload
 (defun cider-ns-reload (&optional prompt)
@@ -302,7 +302,7 @@ refresh functions (defined in `cider-ns-refresh-before-fn' and
   (cider-ensure-op-supported "cider.clj-reload/reload")
   (cider-ns-refresh--save-modified-buffers)
   (let ((clear? (member mode '(clear 16)))
-        (refresh-all? (member mode '(refresh-all 4)))
+        (all? (member mode '(refresh-all 4)))
         (inhibit-refresh-fns (member mode '(inhibit-fns -1))))
     (cider-map-repls :clj
       (lambda (conn)
@@ -317,11 +317,11 @@ refresh functions (defined in `cider-ns-refresh-before-fn' and
                                           nil
                                           t))
           (when clear?
-            (cider-nrepl-send-sync-request (cider-ns-refresh--refresh-op "refresh-clear") conn))
+            (cider-nrepl-send-sync-request (cider-ns--reload-op "reload-clear") conn))
           (cider-nrepl-send-request
            (thread-last
              (map-merge 'list
-                        `(,(cider-ns-refresh--refresh-op (if refresh-all? "refresh-all" "refresh")))
+                        `(,(cider-ns--reload-op (if all? "reload-all" "reload")))
                         (cider--nrepl-print-request-map fill-column)
                         (when (and (not inhibit-refresh-fns) cider-ns-refresh-before-fn)
                           `(("before" ,cider-ns-refresh-before-fn)))
--- a/doc/modules/ROOT/pages/usage/misc_features.adoc
+++ b/doc/modules/ROOT/pages/usage/misc_features.adoc
@@ -130,6 +130,13 @@ and `cider-ns-reload-all` commands can be used instead. These commands
 invoke Clojure's `+(require ... :reload)+` and `+(require
 ... :reload-all)+` commands at the REPL.
 
+You can also use https://github.com/tonsky/clj-reload[clj-reload] instead:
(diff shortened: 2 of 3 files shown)


filipesilva:
@vemv just to get the ball rolling, is this roughly what you had in mind?

cider-ns.el
```diff
@@ -118,6 +118,12 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-ns-use-clj-reload nil
```
vemv, line 121: Initially I thought of having one defcustom per op. Admittedly this is more convenient for users, but if a third option appeared, it would get messy.

A middle ground:

```lisp
(defcustom cider-blah 'tools-namespace
  "Docstring"
  :group 'cider
  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
                 (const :tag "clj-reload https://github.com/tonsky/clj-reload" clj-reload))
  :package-version '(cider . "0.x.x"))
```

This way, a third option could be eventually added cleanly.

vemv (commented):
Looks reasonable to me - thanks!

cider-ns.el
filipesilva: sgtm, added

filipesilva:
Btw have you seen this eldev error? It happens on build and test:
```
[I] filipesilva@Filipes-MacBook-Pro ~/s/cider (clj-reload) [1]> eldev test
Native compiler error: (lambda (arg0 &rest arg1) (let ((f #'message)) (apply f arg0 arg1))), "Compiling /Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln...
-macosx_version_min has been renamed to -macos_version_min
ld: library 'emutls_w' not found
libgccjit.so: error: error invoking gcc driver
Internal native compiler error: \"failed to compile\", \"/Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln\", \"error invoking gcc driver\"

Error: native-ice (\"failed to compile\" \"/Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln\" \"error invoking gcc driver\")
  mapbacktrace(#f(compiled-function (evald func args flags) #<bytecode -0xc883f0734512f81>))
  debug-early-backtrace()
  debug-early(error (native-ice \"failed to compile\" \"/Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln\" \"error invoking gcc driver\"))
  comp--compile-ctxt-to-file(\"/Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln\")
  comp-compile-ctxt-to-file(\"/Users/filipesilva/.emacs.d/eln-cache/29.1-272591eb/subr--trampoline-6d657373616765_message_0.eln\")
  comp-final1()
  load-with-code-conversion(\"/private/var/folders/hx/dl3d0pqn6615z78g83gc26300000gn/T/emacs-int-comp-subr--trampoline-6d657373616765_message_0-re9NpE.el\" \"/private/var/folders/hx/dl3d0pqn6615z78g83gc26300000gn/T/emacs-int-comp-subr--trampoline-6d657373616765_message_0-re9NpE.el\" nil t)
  command-line-1((\"-l\" \"/var/folders/hx/dl3d0pqn6615z78g83gc26300000gn/T/emacs-int-comp-subr--trampoline-6d657373616765_message_0-re9NpE.el\"))
  command-line()
  normal-top-level()
"
```

filipesilva:
If anyone else comes across that error, reinstalling emacs from homebrew seemed to fix it.

filipesilva:
@vemv was able to get a local dev env working, debugged the feature, and I believe this PR is ready to be reviewed. I left support for the clear operation, but will continue discussion of it in https://github.com/clojure-emacs/cider-nrepl/pull/850 instead.

cider-ns.el
```diff
@@ -118,9 +118,17 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-ns-refresh-tool 'tools.namespace
+  "Which tool to use for ns refresh.
+Current options: tools.namespace and clj-reload."
+  :group 'cider
+  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
+                 (const :tag "clj-reload https://github.com/tonsky/clj-reload" clj-reload))
+  :package-version '(cider . "1.13.1"))
```
filipesilva, line 127: I'm not sure what version should be here. Please advise.

cider-ns.el
```diff
@@ -148,6 +156,9 @@ namespace-qualified function of zero arity."
        (reloading
         (log-echo (format "Reloading %s\n" reloading) 'font-lock-string-face))
 
+       (progress
+        (log-echo progress 'font-lock-string-face))
```
filipesilva, line 160: The `progress` key is a passthrough of the logging `cli-reload` outputs. It doesn't give you a list of files being reloaded per se, just logs as they are individually unloaded/reloaded. Examples:

```
cider-ns-refresh: Nothing to reload
cider-ns-refresh: Reloading successful
```

```
cider-ns-refresh: Unloading scratch.scratch
cider-ns-refresh: Loading scratch.scratch
cider-ns-refresh: Reloading successful
```

cider-ns.el
vemv: 1.14.0 i.e. the next major version

cider-ns.el
vemv: This seems nice. However for large projects it might get excessively noisy?

I'd suggest a defcustom, default "no logging" (since one isn't debugging stuff by default)

vemv (commented):
LGTM!

Please let's wait till the `clear` op is there.

That also gives you the opportunity to QA the feature for a few days - I tend to suggest that as it can easily uncover the need for fixes/refinements.

Cheers - V

cider-ns.el
filipesilva: I can see how this can get very spammy if you're looking at the `*messages*` buffer, and agree it should be good to prevent that spam.

I worry preventing it might be premature though... it is adding another customisation, and I imagine this feature might not have a lot of users for a while. Perhaps it could prove unnecessary.

I defer to your opinion though. I don't look at `*messages*` a lot. Maybe spam there is a big issue.

filipesilva: WRT the default of no logging, there's some tension there. Reloading, especially on larger projects, can take a while. Seeing the progress is helpful, and `tools.namespace` not showing the progress is actually to its detriment.

So larger projects, which are the ones where the progress report would be more spammy, are also the ones that might care more about the current progress.

cider-ns.el
```diff
@@ -186,6 +197,19 @@ Its behavior is controlled by `cider-ns-save-files-on-refresh' and
                           (file-in-directory-p buffer-file-name dir))
                         dirs)))))))
 
+(defun cider-ns-refresh--refresh-op (op-name)
```
bbatsov, line 200: I'd name this just `cider-ns--refresh-op`. Although I'm not fully sold on the current approach, as I find it weird to use one set of ops as baseline instead of something more generic. Also - the "refresh" name comes from a function with this name in `c.t.n.`.

I get that the current approach will likely result in the least amount of code changes, but I still want to think about this a bit more.

cider-ns.el
```diff
@@ -186,6 +197,19 @@ Its behavior is controlled by `cider-ns-save-files-on-refresh' and
...
+  "Return the refresh operation to use.
+Based on OP-NAME and the value of cider-ns-refresh-tool defcustom."
+  (list "op"
+        (cond
```
bbatsov, line 204: I think the first `cond` should just be an `if/else`.

cider-ns.el
```diff
@@ -118,9 +118,17 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-ns-refresh-tool 'tools.namespace
+  "Which tool to use for ns refresh.
+Current options: tools.namespace and clj-reload."
+  :group 'cider
+  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
```
bbatsov, line 125: You've made a typo here - `tools-namespace`.

doc/modules/ROOT/pages/usage/misc_features.adoc
```diff
@@ -130,6 +130,13 @@ and `cider-ns-reload-all` commands can be used instead. These commands
 invoke Clojure's `+(require ... :reload)+` and `+(require
 ... :reload-all)+` commands at the REPL.
 
+You can also use https://github.com/tonsky/clj-reload[clj-reload] instead:
```
bbatsov, line 133: Hmm, won't there be any configuration differences that should be noted here? The way it's written one would expect that changing the "refresh" tool will be mostly transparent action and I'm not sure that's really the case.

cider-ns.el
```diff
@@ -118,9 +118,17 @@ namespace-qualified function of zero arity."
   :group 'cider
   :package-version '(cider . "0.10.0"))
 
+(defcustom cider-ns-refresh-tool 'tools.namespace
```
bbatsov, line 121: I'm not sure I like "refresh-tool" as the refresh terminology is coming from a function in `c.t.n.`. I'm pretty sure most people think of this as "code-reload-tool" or something along those lines.

cider-ns.el
vemv: CIDER generally uses `spinner` for progress indication needs.

(it's an item that popped up in a recent PR)

It could be added at some point. Please do not feel pressured to do so, but it's what I'd expect before the "spammy" default :)

doc/modules/ROOT/pages/usage/misc_features.adoc
vemv: Yes, please indicate that one should generally `(reload/init {:dirs dirs})` and that cider won't do that for users

(I'm aware that that became optional, but that was a very recent change and _easily_ a suboptimal one, based on t.n experience)

vemv:
@filipesilva : following https://github.com/clojure-emacs/cider/pull/3625 you may rebase master in to get cider-nrepl with your changes in 🎉

cider-ns.el
bbatsov: I spinner would be quite easy to add indeed. I agree that the extra logging should probably be optional (and maybe displayed in a standalone buffer one can quickly discard). Not a big deal at this stage, though.

bbatsov: Btw, there's already `cider-ns-refresh-show-log-buffer`, which I guess it makes sure to reuse if possible.

bbatsov: And because we're already using `log-echo` it seems we don't have to worry much about spamming Messages here:

```
(log-echo (message &optional face)
                         (log message face)
                         (unless cider-ns-refresh-show-log-buffer
                           (let ((message-truncate-lines t))
                             (message "cider-ns-refresh: %s" message))))
```

cider-ns.el
bbatsov: I now like `cider-ns--reload-op` even more, as I guess we'll move away from the "refresh" naming at some point. So maybe you can name the `clj-reload` op names (sans the namespace) the "default" here and calculate the "refresh" ops in terms of them. Something like `(cider-ns--reload-op 'reload)`.

vemv:
...and following https://github.com/clojure-emacs/cider/pull/3627 you should should see jumping to the culprit file/line on errors.

(No backend changes were needed at all)

Please let us know if it works here as well.

doc/modules/ROOT/pages/usage/misc_features.adoc
filipesilva: I'm happy to describe how users should add the init first but I don't see why that would come up as a difference. I don't see any direct mention of the initialisation procedure for tools.namespace either, although it seems to be implied, but also not necessary.

filipesilva: Another difference is that `cider-ns-refresh-before-fn` and `cider-ns-refresh-after-fn` are not used anymore, but cli-reload supports [unload hooks](https://github.com/tonsky/clj-reload?tab=readme-ov-file#usage-unload-hooks) on individual namespaces.

cider-ns.el
filipesilva: I think I effected the changes as you indicated, but please let me know if understood it wrong.

cider-ns.el
filipesilva: I'm sorry, I'm not sure what the end result is here. Is it that I should add a spinner, or hat I shouldn't do anything since `cider-ns-refresh-show-log-buffer` already exists?

filipesilva:
@vemv @bbatsov I've pushed changes that address your review items. The review items that were trivial to incorporate I marked as resolved to better keep track of things.

cider-ns.el
```diff
@@ -118,13 +118,13 @@ namespace-qualified function of zero arity."
...
-(defcustom cider-ns-refresh-tool 'tools.namespace
+(defcustom cider-ns-code-reload-tool 'tools.namespace
   "Which tool to use for ns refresh.
 Current options: tools.namespace and clj-reload."
   :group 'cider
-  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools-namespace)
+  :type '(choice (const :tag "tools.namespace https://github.com/clojure/tools.namespace" tools.namespace)
                  (const :tag "clj-reload https://github.com/tonsky/clj-reload" clj-reload))
-  :package-version '(cider . "1.13.1"))
+  :package-version '(cider . "1.40.0"))
```
bbatsov, line 127: 1.14.0 :-)

cider-ns.el
bbatsov: It seems to me you don't need to do anything, as those logs will end in the dedicated buffer (when enabled). 

We can discuss the need for adding a spinner down the road.

doc/modules/ROOT/pages/usage/misc_features.adoc
vemv: > I don't see any direct mention of the initialisation procedure for tools.namespace either,

I added https://github.com/clojure-emacs/cider/blob/23540fdfcdac02b2d18e51a170cd07b32d20e80d/cider-ns.el#L258-L260 this week

> Another difference is that cider-ns-refresh-before-fn and cider-ns-refresh-after-fn are not used anymore, but cli-reload supports [unload hooks](https://github.com/tonsky/clj-reload?tab=readme-ov-file#usage-unload-hooks) on individual namespaces.

We should pay attention to this. `reset`ting a system is a fundamental step when working when any commercial app.

I'd suggest that it's discussed + tackled in a follow-up PR?

vemv:
@filipesilva : a very common .dir-locals.el is some variation of:

```lisp
((nil . ((cider-ns-refresh-before-fn . "integrant.repl/suspend")
         (cider-ns-refresh-after-fn  . "integrant.repl/resume")
,,,
```

So that reloaded changes can actually affect an app.

I'd be OK with merging this asap but please confirm if you'd intend to further refine the solution

cider-ns.el
bbatsov: Yep, it did exactly what I had in mind.

cider-ns.el
filipesilva: Whoops, fixed!

filipesilva:
@vemv I'm not really thinking of adding that extra functionality, no. That looks like it should be part of base functionality for `clj-reload`, not something that needs CIDER to work. It might already work just with unload hooks, but I'm not sure. @tonsky what are your thoughts on this?

filipesilva:
Meanwhile I found a bug where reload-all isn't working because I passed the wrong option in cider-nrepl. I'm working on fixing it and making the tests more robust.

filipesilva:
https://github.com/clojure-emacs/cider-nrepl/pull/854 is the fix for the reload-all op.

tonsky:
> @vemv I'm not really thinking of adding that extra functionality, no. That looks like it should be part of base functionality for `clj-reload`, not something that needs CIDER to work. It might already work just with unload hooks, but I'm not sure. @tonsky what are your thoughts on this?

I think this is just calling a function immediately before `clj-reload.core/reload` and immediately after. What kind of support are you thinking? Can clj-reload really do something here?

vemv:
Thanks!

While t.n supports an `:after` fn, in cider-nrepl, (somewhat surprisingly) we invoke the functions ourselves.

So yeah no upstream changes are required.

vemv:
I'm squash-merging this separately as we speak.

Thanks much for your work @filipesilva !

vemv:
It's in `master` now

https://github.com/clojure-emacs/cider/commit/c4fa1a84a3b3d03ef5f61cc5d33ff4e91b9a1dce

A melpa snapshot will be automatically created later today.

We'll cut a stable release soon enough.

Cheers - V

filipesilva:
Awesome, thank you so much!

filipesilva:
Just following up on the "reloaded changes can actually affect an app" case.

I can verify I can add this fn in a given ns:
```
(defn before-ns-unload []
  (println "stopping processes")

(defn after-ns-reload []
  (println "restarting processes")
```

And it is running before/after the reload, both when I call reload manually and when I call it via cider. I just used it to restart a server in an app of mine I've moved to clj-reload. So clj-reload can handle this use case via its hooks.

vemv:
Thanks!

Anyway, cider-ns-refresh-before-fn / cider-ns-refresh-after-fn are now supported, so it should work as well.

cichli:
> While t.n supports an `:after` fn, in cider-nrepl, (somewhat surprisingly) we invoke the functions ourselves.

The middleware interleaves responses such as `{:status :ok}` / `{:status :invoking-after}` / `{:status :invoked-after}` with the execution of the refresh and invoking `:after` – if we used the t.n option, we wouldn't be able to do that.

PR #3629, opened 2024-03-07
`cider-ns-refresh`: summarize errors as an overlay

### LINKED ISSUE #3628: cider-ns-refresh: summarize errors as an overlay
Most times, code reloading errors come from a 'silly' mistake that doesn't deserve much stacktrace inspection.

So, in addition to the displayed stacktrace, we could present the gist of the problem as a red overlay, similar to what we've done in recent times.

While we're there, we could add defcustoms controlling both widgets (overlay, stacktrace)

vemv:
> Fixes #3628

Sample after a `refresh`:

image

For the moment I think it will be best not to offer defcustoms for hiding the overlay or the stacktrace. At some point we could hide the stacktrace depending on the nature of the exception, but that needs some work.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Changes
 
 - [#3626](https://github.com/clojure-emacs/cider/issues/3626): `cider-ns-refresh`: jump to the relevant file/line on errors.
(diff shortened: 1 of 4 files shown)


bbatsov (approved):

bbatsov:
Looks good to me!

PR #3632, opened 2024-03-10
`cider-ns` - refinements

vemv:
- Strengthen `cider-ns--present-error`
- `cider-ns--present-error`: prefer `end-of-sexp`
  -  One has to use structural movements in order to present the overlay without errors.
  - By using this finer-grained function, it's more likely that the overlay will be rendered at the right place (and not too far away at the bottom).
- `cider-ns-refresh`: accept `clear-and-inhibit` mode
  - It often makes sense not to run the before/after functions around the `clear`ing.

Cheers - V

### DIFF
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -137,8 +137,9 @@ presenting the error message as an overlay."
                                              ;; jars are unlikely sources of user errors, so we favor the next `cause-dict':
                                              (not (string-prefix-p "jar:" file-url))
                                              line)
-                                    (setq buf (cider--find-buffer-for-file file-url))
-                                    (list buf (cons line column)))))
+                                    (when-let ((found (cider--find-buffer-for-file file-url)))
+                                      (setq buf found)
+                                      (list buf (cons line column))))))
                               error)))
     (when jump-args
       (apply #'cider-jump-to jump-args)
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -153,7 +153,7 @@ presenting the error message as an overlay."
             (cider--display-interactive-eval-result trimmed-err
                                                     'error
                                                     (save-excursion
-                                                      (end-of-defun)
+                                                      (end-of-sexp)
                                                       (point))
                                                     'cider-error-overlay-face)))))
     (cider--render-stacktrace-causes error)
--- a/cider-ns.el
+++ b/cider-ns.el
@@ -286,24 +286,27 @@ Uses the configured 'refresh dirs' \(defaults to the classpath dirs).
 With a single prefix argument, or if MODE is `refresh-all', reload all
 namespaces on the classpath dirs unconditionally.
 
-With a double prefix argument, or if MODE is `clear', clear the state of
-the namespace tracker before reloading.  This is useful for recovering from
+With a double prefix argument, or if MODE is `clear' (or `clear-and-inhibit'),
+clear the state of the namespace tracker before reloading.
+
+This is useful for recovering from
 some classes of error (for example, those caused by circular dependencies)
 that a normal reload would not otherwise recover from.  The trade-off of
 clearing is that stale code from any deleted files may not be completely
 unloaded.
 
-With a negative prefix argument, or if MODE is `inhibit-fns', prevent any
-refresh functions (defined in `cider-ns-refresh-before-fn' and
+With a negative prefix argument,
+or if MODE is `inhibit-fns' (or `clear-and-inhibit'),
+ prevent any refresh functions (defined in `cider-ns-refresh-before-fn' and
 `cider-ns-refresh-after-fn') from being invoked."
   (interactive "p")
   (cider-ensure-connected)
   (cider-ensure-op-supported "refresh")
   (cider-ensure-op-supported "cider.clj-reload/reload")
   (cider-ns-refresh--save-modified-buffers)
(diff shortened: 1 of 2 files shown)


cider-ns.el
```diff
@@ -153,7 +154,7 @@ presenting the error message as an overlay."
             (cider--display-interactive-eval-result trimmed-err
                                                     'error
                                                     (save-excursion
-                                                      (end-of-defun)
+                                                      (end-of-sexp)
```
vemv, line 157: Hmm, the linter suggests `thing-at-point--end-of-sexp` due to deprecations, but that one looks internal.

Anything better that comes to mind? @bbatsov

cider-ns.el
```diff
@@ -286,24 +286,27 @@ Uses the configured 'refresh dirs' \(defaults to the classpath dirs).
 With a single prefix argument, or if MODE is `refresh-all', reload all
 namespaces on the classpath dirs unconditionally.
 
-With a double prefix argument, or if MODE is `clear', clear the state of
-the namespace tracker before reloading.  This is useful for recovering from
+With a double prefix argument, or if MODE is `clear' (or `clear-and-inhibit'),
```
bbatsov, line 290: Probably we'll need to document this somewhere in the manual as well, otherwise few people are going to find it.

cider-ns.el
vemv: > clojure-forward-logical-sexp

Will try!

> I'm not sure what you're trying to achieve here, though, as there might be quite a big difference between the boundary of a sexp (which might be just the symbol at point) and a top-level form.

Yes, end of top-level seems the safest option. But it can easily mean that no overlay will be visible.

tbh I don't understand why displaying an overlay can't simply be done at the end of the current line, no matter what. [cider--make-result-overlay ](https://github.com/clojure-emacs/cider/blob/c4fa1a84a3b3d03ef5f61cc5d33ff4e91b9a1dce/cider-overlays.el#L183) is pretty complex.

With a bit of luck `clojure-forward-logical-sexp` will do the trick (which would make sense given that `cider--make-result-overlay` uses `clojure-backward-logical-sexp`)

cider-ns.el
vemv: Sure. While we're here, should the new mode be invokable through some prefix incantation?

tbh I don't use prefixes at all so I wouldn't know how to add a new variation here.

cider-ns.el
bbatsov: > tbh I don't understand why displaying an overlay can't simply be done at the end of the current line, no matter what. [cider--make-result-overlay ](https://github.com/clojure-emacs/cider/blob/c4fa1a84a3b3d03ef5f61cc5d33ff4e91b9a1dce/cider-overlays.el#L183)is pretty complex.

Mostly because normally we want to have the overlay attached to some expression, as that makes it most clear where some result originates from. Obviously we can create a different variant of this function if needed.

cider-ns.el
bbatsov: That would be nice indeed. I think multiple prefixes had to be checked as some number value in the code to differentiate them. There must be some examples in the code. I do think, however, we should start moving towards `transient` for prefix arguments as it's UI is so much better, and it's bundled with Emacs 28+.

vemv: I expanded the user manual

>  I do think, however, we should start moving towards transient for prefix arguments as it's UI is so much better

In face of this maybe we can refrain from doubling down on prefixes (at least here).

Transient rocks and would love to see a few features / examples of how it would look for CIDER

vemv (commented):
Ready again with a `cider--semantic-end-of-line` which seems to do the right thing

codespell.txt
```diff
@@ -1,4 +1,5 @@
 edn
+fo
```
bbatsov, line 2: Why so? That's often a misspelled `of`.

bbatsov:
I'll take a closer look tomorrow.

cider-util.el
```diff
@@ -826,6 +826,28 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
+(defun cider--semantic-end-of-line ()
+  "Returns POINT at the closest EOL that we can move to semantically,
```
bbatsov, line 830: I'm not sure I understand this. What's an EOL to which we can move semantically? I checked the tests as well, but all assertions are in a single tests, so it's pretty hard to figure out what you're testing for exactly.

doc/modules/ROOT/pages/usage/misc_features.adoc
```diff
@@ -62,17 +62,24 @@ Typing kbd:[C-c M-n r] or kbd:[C-c M-n M-r] will invoke
 `cider-ns-refresh` and reload all modified Clojure files on the
 classpath.
 
-Adding a prefix argument, kbd:[C-u C-c M-n r], will reload all
+Adding a prefix argument, kbd:[C-u C-c M-n r], it will reload all
```
bbatsov, line 65: I think "Adding" is not the right terminology here, which affects the rest of the sentence (which reads a bit awkwardly to me). I'd write something like "If you invoke cider-ns-refresh with a prefix argument it will" or something along those lines.

bbatsov:
Ping :-)

PR #3634, opened 2024-03-14
`cider-eval`: Introduce new `cider-auto-inspect-after-eval-select-window` defcustom

vemv:
My overall intent is to have the `*cider-inspect*` permanently on a background frame. cider-eval should have that frame's cider-inspect buffer re-rendered, _without switching to it_, given that I personally find that only a fraction of evals deserve inspection.

This wasn't possible before, as CIDER would detect that my frame wasn't visible, so it would render the buffer in a random window of the visible frame.

Commits:

- `cider-popup-buffer-display`: Do nothing if the intended buffer is visible in another frame and `select` is nil
  - This seems the right thing to do, for a wide variety of use cases.
- `cider-eval`: Introduce new `cider-auto-inspect-after-eval-select-window` defcustom
  - It allows one to have the `*cider-inspect*` buffer rendered without switching to it.
- cider-overlays: Make `Result truncated` message more accurate
  -  If there's already a `*cider-inspect*` buffer visible, we shouldn't suggest the user to inspect the value again.

Cheers - V


vemv:
I reworked this PR to not introduce any defcustoms - ready now!

bbatsov:
Thanks for the update, I'll review it tomorrow.

cider-util.el
```diff
@@ -826,6 +826,24 @@ KIND can be the symbols `ns', `var', `emph', `fn', or a face name."
              (t x)))
           menu-list))
 
+;; Defined here to avoid circular dependencies
+(defconst cider-inspector-buffer "*cider-inspect*")
```
bbatsov, line 830: Hmm, I'm not sure I like this, because the way I see it the inspector shouldn't be depending on `cider-eval` and now we're kind of sweeping this odd dependency under the rug.

bbatsov:
I checked the `cider-inspector.el` code and it seems the dependency on `cider-eval.el` is there just to access a single variable from there. Perhaps a better fix would be a simple `declare`.

bbatsov:
@alexander-yakushev Any thoughts on this PR?

PR #3648, opened 2024-04-23
Use enrich-classpath 1.19.2

vemv:
https://github.com/clojure-emacs/enrich-classpath/compare/v1.19.0...v1.19.2


vemv:
GHA is red for older emacsen https://github.com/purcell/nix-emacs-ci/issues/267 and Circle has a different hiccup.

Issue #3653, opened 2024-05-05
`cider-enable-flex-completion` is not a sufficient replacement for `cider-company-enable-fuzzy-completion`


alexander-yakushev:
After I enabled `cider-enable-flex-completion` instead of `cider-company-enable-fuzzy-completion`, I noticed that my completion was acting up. Certain prefixes that used to complete correctly now failed to complete, and weird suggestions were popping up. The weird suggestions are understandable because of `flex` aggressive fuzzy policy, but that's not that big of an issue. The issue is the following completion scenarios supported by compliment that don't work in `flex`:

1. Namespace completion by first letters of namespaces:
  - `(cider-complete "cji")` => `(#("clojure.java.io" 0 1 (ns nil type "namespace")))`
  - `Tab` with `flex` => no completions.
2. FQ classname completions by first word:
- `(cider-complete "InputSt")` => `(#("java.io.InputStream" 0 1 (ns nil type "class")) #("java.io.InputStreamReader" 0 1 (ns nil type "class")) ...`
 - `Tab` with `flex` => no completions.
3. Fuzzy completion behind namespace prefix:
  - `(cider-complete "str/sl")` => `(#("str/split-lines" 0 1 (ns "clojure.string" type "function")))`
  - `Tab` with `flex` => no completions.

There are a couple of others but they are not as important as these.

Additionally, Emacs achieves this flex behavior by sending two requests to the backend: one with the normal prefix and one with empty prefix (`""`). Since we have no upper bound on the number of candidates we return, this may add some extra overhead/delay for the completion because CIDER has to deserialize all that output.

vemv:
Thanks for the issue!

We should thread carefully here because we shouldn't necessarily be smarter than (or reinvent) Emacs' mechanisms.

Most of all, it's a concern of simplicity - Emacs has an abundance of styles, and there's also an abundance of Emacs packages building up on those, so creating our own standards can possibly make things even messier for everyone involved.

I don't have a specific suggestion/stance except: we can try first hard at adhering at Emacs standards and squeezing performance out of them.

If we document reasoning/facts reflecting that that's impossible, we could keep exploring solutions.

vemv:
Quick q - would it make sense to pass the completion style to Compliment?

In my mind it would make sense - then Compliment could infer what's wanted and return fewer / more relevant results.

alexander-yakushev:
[...]

PR #3659, opened 2024-05-13
Grab the completion prefix correctly

toniz4:
This PR makes the CIDER completions behave correctly. Reworking the method to grab the prefix in `cider-complete-at-point`. Making it compatible with "fuzzy" completion styles.

Tested with the `flex` and `orderless` completion styles.

What should we do with `cider-company-enable-fuzzy-completion`? With
this we can remove the CIDER completion style. Should
`cider-company-enable-fuzzy-completion` just mirror the behavior of `cider-enable-flex-completion`?

This also makes CIDER use the default completion style, instead of setting `completion-category-overrides` to make CIDER use the basic completion style.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -299,6 +299,8 @@ Only affects the `cider' completion category.`"
     (setq completion-category-overrides (seq-remove (lambda (x)
                                                       (equal 'cider (car x)))
                                                     completion-category-overrides))
+    (unless found-styles
+      (setq found-styles '(styles)))
     (unless (member 'flex found-styles)
       (setq found-styles (append found-styles '(flex))))
     (add-to-list 'completion-category-overrides (apply #'list 'cider found-styles (when found-cycle
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -179,24 +179,46 @@ performed by `cider-annotate-completion-function'."
            (ns (cider-completion--get-candidate-ns symbol)))
       (funcall cider-annotate-completion-function type ns))))
 
-(defvar cider--completion-cache nil
-  "Cache used by `cider--complete-with-cache'.
-this is a cons cell of (BOUNDS . COMPLETIONS).")
-
-(defun cider--complete-with-cache (bounds)
-  "Return completions to the symbol at `BOUNDS' with caching.
-If the completion of `bounds' is cached, return the cached completions,
-otherwise, call `cider-complete', set the cache, and return the completions."
-  (let* ((prefix (or (buffer-substring-no-properties (car bounds) (cdr bounds)) ""))
-         (completions nil))
-    (when (and (consp cider--completion-cache)
-               (string= prefix (car cider--completion-cache)))
-      (setq completions (cdr cider--completion-cache)))
-    (unless completions
-      (let ((resp (cider-complete prefix)))
-        (setq cider--completion-cache `(,prefix . ,resp)
-              completions resp)))
-    completions))
+(defun cider-complete-at-point ()
+  "Complete the symbol at point."
+  (when-let* ((bounds (or (bounds-of-thing-at-point 'symbol)
+                          (cons (point) (point))))
+              (bounds-string (buffer-substring (car bounds) (cdr bounds))))
+    (when (and (cider-connected-p)
+               (not (or (cider-in-string-p) (cider-in-comment-p))))
+      (let* (last-bounds-string
+             last-result
+             (complete
+              (lambda ()
+                ;; Not using the prefix extracted within the (prefix pred action) lambda.
+                ;; In certain completion styles, the prefix might be an empty string,
+                ;; which is unreliable. A more dependable method is to use the string
+                ;; defined by the bounds of the symbol at point.
+                ;;
+                ;; Caching just within the function is sufficient. Keeping it local ensures
+                ;; that it will not extend across diferent CIDER sessions.
+                (unless (string= bounds-string last-bounds-string)
+                  (setq last-bounds-string bounds-string)
+                  (setq last-result (cider-complete bounds-string)))
+                last-result)))
+        (list (car bounds) (cdr bounds)
+              (lambda (prefix pred action)
+                ;; When the 'action is 'metadata, this lambda returns metadata about this
+                ;; capf, when action is (boundaries . suffix), it returns nil. With every
+                ;; other value of 'action (t, nil, or lambda), 'action is forwarded to
+                ;; (complete-with-action), together with (cider-complete), prefix and pred.
+                ;; And that function performs the completion based on those arguments.
+                ;;
+                ;; This api is better described in the section
+                ;; '21.6.7 Programmed Completion' of the elisp manual.
+                (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
+                      ((eq (car-safe action) 'boundaries) nil) ; boundaries
+                      (t (complete-with-action action (funcall complete) prefix pred))))
+              :annotation-function #'cider-annotate-symbol
+              :company-kind #'cider-company-symbol-kind
+              :company-doc-buffer #'cider-create-compact-doc-buffer
+              :company-location #'cider-company-location
+              :company-docsig #'cider-company-docsig)))))
 
 (defun cider-completion-flush-caches ()
   "Force Compliment to refill its caches.
@@ -205,43 +227,6 @@ and methods from dependencies that were loaded dynamically after the REPL
 has started."
   (interactive)
   (cider-sync-request:complete-flush-caches))
-
-(defun cider--clear-completion-cache (_ _)
-  "Clears the completion cache."
-  (cider-completion-flush-caches)
-  (setq cider--completion-cache nil))
-
-(defun cider-complete-at-point ()
-  "Complete the symbol at point."
-  (when-let* ((bounds (or (bounds-of-thing-at-point 'symbol)
-                          (cons (point) (point))))
-              (bounds-string (buffer-substring (car bounds) (cdr bounds))))
-    (when (and (cider-connected-p)
-               (not (or (cider-in-string-p) (cider-in-comment-p))))
-      (list (car bounds) (cdr bounds)
-            (lambda (prefix pred action)
-              ;; When the 'action is 'metadata, this lambda returns metadata about this
-              ;; capf, when action is (boundaries . suffix), it returns nil. With every
-              ;; other value of 'action (t, nil, or lambda), 'action is forwarded to
-              ;; (complete-with-action), together with (cider-complete), prefix and pred.
-              ;; And that function performs the completion based on those arguments.
-              ;;
-              ;; This api is better described in the section
-              ;; '21.6.7 Programmed Completion' of the elisp manual.
-              (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
-                    ((eq (car-safe action) 'boundaries) nil) ; boundaries
-                    ((eq action 'lambda) ; test-completion
-                     (test-completion prefix (cider--complete-with-cache bounds)))
-                    ((null action)  ; try-completion
-                     (try-completion prefix (cider--complete-with-cache bounds)))
-                    ((eq action t)  ; all-completions
-                     (all-completions "" (cider--complete-with-cache bounds)))))
-            :annotation-function #'cider-annotate-symbol
-            :company-kind #'cider-company-symbol-kind
-            :company-doc-buffer #'cider-create-compact-doc-buffer
-            :company-location #'cider-company-location
-            :company-docsig #'cider-company-docsig
-            :exit-function #'cider--clear-completion-cache))))
 
 (defun cider-company-location (var)
   "Open VAR's definition in a buffer.
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -219,7 +219,7 @@ has started."
     (when (and (cider-connected-p)
                (not (or (cider-in-string-p) (cider-in-comment-p))))
       (list (car bounds) (cdr bounds)
-            (lambda (pattern pred action)
+            (lambda (prefix pred action)
               ;; When the 'action is 'metadata, this lambda returns metadata about this
               ;; capf, when action is (boundaries . suffix), it returns nil. With every
               ;; other value of 'action (t, nil, or lambda), 'action is forwarded to
@@ -231,9 +231,9 @@ has started."
               (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
                     ((eq (car-safe action) 'boundaries) nil) ; boundaries
                     ((eq action 'lambda) ; test-completion
-                     (test-completion pattern (cider--complete-with-cache bounds)))
+                     (test-completion prefix (cider--complete-with-cache bounds)))
                     ((null action)  ; try-completion
-                     (try-completion pattern (cider--complete-with-cache bounds)))
+                     (try-completion prefix (cider--complete-with-cache bounds)))
                     ((eq action t)  ; all-completions
                      (all-completions "" (cider--complete-with-cache bounds)))))
             :annotation-function #'cider-annotate-symbol
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -213,7 +213,8 @@ performed by `cider-annotate-completion-function'."
                 ;; '21.6.7 Programmed Completion' of the elisp manual.
                 (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
                       ((eq (car-safe action) 'boundaries) nil)
-                      (t (complete-with-action action (funcall complete) prefix pred))))
+                      (t (with-current-buffer (current-buffer)
+                           (complete-with-action action (funcall complete) prefix pred)))))
               :annotation-function #'cider-annotate-symbol
               :company-kind #'cider-company-symbol-kind
               :company-doc-buffer #'cider-create-compact-doc-buffer
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -187,12 +187,16 @@ this is a cons cell of (BOUNDS . COMPLETIONS).")
   "Return completions to the symbol at `BOUNDS' with caching.
 If the completion of `bounds' is cached, return the cached completions,
 otherwise, call `cider-complete', set the cache, and return the completions."
-  (cdr-safe
-   (if (and (consp cider--completion-cache)
-            (eq bounds (car cider--completion-cache)))
-       cider--completion-cache
-     (setq cider--completion-cache
-           `(,bounds . ,(cider-complete (buffer-substring (car bounds) (cdr bounds))))))))
+  (let* ((prefix (or (buffer-substring-no-properties (car bounds) (cdr bounds)) ""))
+         (completions nil))
+    (when (and (consp cider--completion-cache)
+               (equal prefix (car cider--completion-cache)))
+      (setq completions (cdr cider--completion-cache)))
+    (when (null completions)
+      (let ((resp (cider-complete prefix)))
+        (setq cider--completion-cache `(,prefix . ,resp)
+              completions resp)))
+    completions))
 
 (defun cider-completion-flush-caches ()
   "Force Compliment to refill its caches.
@@ -202,7 +206,7 @@ has started."
   (interactive)
   (cider-sync-request:complete-flush-caches))
 
-(defun cider--clear-completion-cache (&optional _ _)
+(defun cider--clear-completion-cache (_ _)
   "Clears the completion cache."
   (cider-completion-flush-caches)
   (setq cider--completion-cache nil))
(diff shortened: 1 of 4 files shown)


vemv:
Hi @toniz4 !

Thanks much for your efforts.

Caching / redundant trip avoidance was recently tackled in https://github.com/clojure-emacs/cider/pull/3655/files . Had you noticed?

From my side I don't quite have the time to go through it however I'd constructively suggest:

* Please try to sync up with the most relevant folks for the discussions, in this case from https://github.com/clojure-emacs/cider/issues/3006 it can be appreciated that @alexander-yakushev made a series of analyses (and at least one subsequent PR)
  * most of all I'm trying to avoid having competing ideas - not the most efficient outcome!
* When presenting a PR, just as valuable as the code changes, is an analysis of what was wrong and how you've addressed those problems.
  * without that we'd have to take that it works at face value - not always reassuring, especially for one of our most important features

Once that's all clear, probably we'd appreciate unit tests for this area - it's a dense one and very few of us have the capability to maintain it in detail over the years.

None of this is to discourage you - I'm simply trying to set up us all for success.

Cheers - V

toniz4:
> Hi @toniz4 !
> 
> Thanks much for your efforts.
> 
> Caching / redundant trip avoidance was recently tackled in https://github.com/clojure-emacs/cider/pull/3655/files . Had you noticed?
> 
> From my side I don't quite have the time to go through it however I'd constructively suggest:
> 
>     * Please try to sync up with the most relevant folks for the discussions, in this case from [Cider completion style is invalid #3006](https://github.com/clojure-emacs/cider/issues/3006) it can be appreciated that @alexander-yakushev made a series of analyses (and at least one subsequent PR)
>       
>       * most of all I'm trying to avoid having competing ideas - not the most efficient outcome!
> 
>     * When presenting a PR, just as valuable as the code changes, is an analysis of what was wrong and how you've addressed those problems.
>       
>       * without that we'd have to take that it works at face value - not always reassuring, especially for one of our most important features
> 
> 
> Once that's all clear, probably we'd appreciate unit tests for this area - it's a dense one and very few of us have the capability to maintain it in detail over the years.
> 
> None of this is to discourage you - I'm simply trying to set up us all for success.
> 
> Cheers - V

Of course, I saw the commits made by @alexander-yakushev after I started working on this, but we kinda followed the same course. 

But in this PR I'm trying to address another issue, the way we are getting the prefix are insufficient. So I tried to take the string inside `bounds`

```elisp
(defun cider--complete-with-cache (bounds)
  "Return completions to the symbol at `BOUNDS' with caching.
If the completion of `bounds' is cached, return the cached completions,
otherwise, call `cider-complete', set the cache, and return the completions."
  (cdr-safe
   (if (and (consp cider--completion-cache)
            (eq bounds (car cider--completion-cache)))
       cider--completion-cache
     (setq cider--completion-cache
           `(,bounds . ,(cider-complete (buffer-substring (car bounds) (cdr bounds))))))))
```

But after some testing, this is also insufficient. The bounds is set when the completion starts.

We can see in `robe`, that the prefix gathering is much more complex:

https://github.com/dgutov/robe/blob/6bc8a07fc483407971de0966d367a11006b3ab80/robe.el#L949-L952

cider-completion.el
```diff
@@ -202,7 +206,7 @@ has started."
   (interactive)
   (cider-sync-request:complete-flush-caches))
 
-(defun cider--clear-completion-cache (&optional _ _)
+(defun cider--clear-completion-cache (_ _)
```
bbatsov, line 209: Do we even need those unused params?

cider-completion.el
```diff
@@ -187,12 +187,16 @@ this is a cons cell of (BOUNDS . COMPLETIONS).")
...
-  (cdr-safe
-   (if (and (consp cider--completion-cache)
-            (eq bounds (car cider--completion-cache)))
-       cider--completion-cache
-     (setq cider--completion-cache
-           `(,bounds . ,(cider-complete (buffer-substring (car bounds) (cdr bounds))))))))
+  (let* ((prefix (or (buffer-substring-no-properties (car bounds) (cdr bounds)) ""))
+         (completions nil))
+    (when (and (consp cider--completion-cache)
+               (equal prefix (car cider--completion-cache)))
```
bbatsov, line 193: I generally prefer the use of `string=` for string comparison.

cider-completion.el
```diff
@@ -187,12 +187,16 @@ this is a cons cell of (BOUNDS . COMPLETIONS).")
...
+      (setq completions (cdr cider--completion-cache)))
+    (when (null completions)
```
bbatsov, line 195: You can use `unless` here.

cider-completion.el
```diff
@@ -215,7 +219,7 @@ has started."
     (when (and (cider-connected-p)
                (not (or (cider-in-string-p) (cider-in-comment-p))))
       (list (car bounds) (cdr bounds)
-            (lambda (prefix pred action)
+            (lambda (pattern pred action)
```
bbatsov, line 222: You was this renamed?

bbatsov:
> Should cider-company-enable-fuzzy-completion just mirror the behavior of cider-enable-flex-completion?

If we're confident in the solution - I think yes. I don't want to kill `cider-enable-flex-completion` right away, as it might break some user setups. But we can mark it officially as deprecated and scheduled for removal. 

I'll have to ponder a bit more on your PR, as I'm not sure I quite follow the issue with the bounds that it's trying to solve. Perhaps @alexander-yakushev will understand this better and have an easier time evaluating it.

alexander-yakushev:
> Should
cider-company-enable-fuzzy-completion just mirror the behavior of cider-enable-flex-completion?

It also depends on how important the support for Emacs <27 is, I had an impression that there was also a compatibility dimension to this.

toniz4:
In the latest commit, it passes all the "tests" @alexander-yakushev presented here https://github.com/clojure-emacs/cider/issues/3653#issue-2279654647

Completing cji -> clojure.java.io
image
Completing InputSt -> java.io.InputStream
image

Completing `str/sl` is a bit different. When I'm in `company-mode`, I can type `str/sl` as expected and it will return the completions you would expect. But corfu works differently, I have to type "str", select "str" in the completions, and after that press "/". Probably because str returns multiple candidates, so corfu doesn't make assumptions. This doesn't happens if the namespace is the sole completion candidate.

alexander-yakushev:
@toniz4 thanks for picking this up!
 
You should take a look at https://github.com/clojure-emacs/cider/pull/3655 for the context. It is crucial that the completion is as fresh as possible and correct too. We figured out there that being a closed over variable inside `cider-complete-at-point` is the best place for the cache. With your current implementation, if you understand correctly, the cache transcends even different CIDER sessions which is quite dangerous.

cider-completion.el
toniz4: Renamed to debug some issue, forgot to revert it

toniz4:
@alexander-yakushev Yeah, I think that's a better way to go. The way I implemented the functions looks cleaner, but I was not thinking about multiple sessions. But it would be really improbable to "hit" the cache even on another buffer, but it's possible.

Either way, putting this stuff in a closure is the right way

cider-completion.el
toniz4: This function is being used in the `:exit-function`, it expects a function of 2 arguments: 

https://www.gnu.org/software/emacs/manual/html_node/elisp/Completion-Variables.html#index-completion_002dextra_002dproperties

But we don't need to use those arguments.

bbatsov:
The CI is failing with:

```
In cider-complete-at-point:
             cider-completion.el:222:34: Error: Unused lexical argument `pred'
```

You can ignore the integration tests, as something's wrong with their container images since last week. We'll have to address this separately.

bbatsov:
>  We figured out there that being a closed over variable inside cider-complete-at-point is the best place for the cache.

We should also probably add a comment in the code for this, so it won't get forgotten down the road. :-)

cider-completion.el
```diff
@@ -225,8 +229,13 @@ has started."
...
               (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
-                    ((eq (car-safe action) 'boundaries) nil)
-                    (t (complete-with-action action (cider--complete-with-cache bounds) bounds-string pred))))
+                    ((eq (car-safe action) 'boundaries) nil) ; boundaries
+                    ((eq action 'lambda) ; test-completion
+                     (test-completion pattern (cider--complete-with-cache bounds)))
+                    ((null action)  ; try-completion
+                     (try-completion pattern (cider--complete-with-cache bounds)))
+                    ((eq action t)  ; all-completions
+                     (all-completions "" (cider--complete-with-cache bounds)))))
```
alexander-yakushev: Are these (`test-completion`, `try-completion`, `all-completions`) worth keeping in the latest solution?

toniz4:
Updated the function to keep basically the same caching behavior that @alexander-yakushev introduced. But it differs in the way we are getting the prefix for the completions. Kinda confusing having 2 things that serve as a prefix, but we are working with emacs. Upon reviewing other packages capf's, I observed that commonly they had a specific mechanism to retrieve the prefix, rather than directly using the prefix from the lambda.

Another thing, I removed the hard coded style override, that were setting the basic completion style for CIDER completions. Is this ok? Currently it will use the completion style configured by the user (by default is `(basic partial-completion emacs22)`)

cider-completion.el
toniz4: Nope, was doing that early to ease with debugging, using complete-with-action has the same behavior.

By the way, `(with-current-buffer (current-buffer) ...` is necessary there?

bbatsov:
> Another thing, I removed the hard coded style override, that were setting the basic completion style for CIDER completions. Is this ok? Currently it will use the completion style configured by the user (by default is (basic partial-completion emacs22))

That's fine by me (from what I got everything should work now, so those are completely redundant), but I'll leave it to @alexander-yakushev to decide how to proceed here.

alexander-yakushev:
LGTM too! Once this is merged, I'll try using enable-flex again and see if there are any disparity with a custom completion style.

toniz4:
About tests, @vemv do you have any idea how can we make unit tests for `cider-complete-at-point`?

cider-completion.el
```diff
@@ -181,16 +181,25 @@ performed by `cider-annotate-completion-function'."
...
-      (let* (last-prefix
+      (let* (last-bounds-string
              last-result
              (complete
-              (lambda (prefix)
-                (unless (string-equal last-prefix prefix)
-                  (setq last-prefix prefix)
-                  (setq last-result (cider-complete prefix)))
+              (lambda ()
+                ;; Not using the prefix extracted within the (prefix pred action) lambda.
```
vemv, line 193:
```suggestion
                ;; We are not using the prefix extracted within the (prefix pred action) lambda.
```

cider-completion.el
```diff
@@ -203,10 +212,8 @@ performed by `cider-annotate-completion-function'."
                 ;; This api is better described in the section
                 ;; '21.6.7 Programmed Completion' of the elisp manual.
                 (cond ((eq action 'metadata) `(metadata (category . cider))) ;; defines a completion category named 'cider, used later in our `completion-category-overrides` logic.
-                      ((eq (car-safe action) 'boundaries) nil)
-                      (t (with-current-buffer (current-buffer)
```
vemv, line 207: I'd feel most comfortable not removing `(with-current-buffer (current-buffer)` - although it looks redundant, it's fair to think that it might be there for a good reason.

Not fun to possibly break something and later debug it (possibly for a different user)

cider-completion.el
```diff
@@ -203,10 +212,8 @@ performed by `cider-annotate-completion-function'."
...
-                           (complete-with-action action
-                                                 (funcall complete prefix) prefix pred)))))
+                      ((eq (car-safe action) 'boundaries) nil) ; boundaries
```
vemv, line 215:
```suggestion
                      ((eq (car-safe action) 'boundaries) nil)
```

cider-completion.el
```diff
@@ -264,12 +271,6 @@ in the buffer."
                cider-company-unfiltered-candidates
                "CIDER backend-driven completion style."))
 
-;; Currently CIDER completions only work for `basic`, and not `initials`, `partial-completion`, `orderless`, etc.
-;; So we ensure that those other styles aren't used with CIDER, otherwise one would see bad or no completions at all.
-;; This `add-to-list` call can be removed once we implement the other completion styles.
-;; (When doing that, please refactor `cider-enable-flex-completion' as well)
-(add-to-list 'completion-category-overrides '(cider (styles basic)))
```
vemv, line 271: Similar code is referred to in doc/modules/ROOT/pages/usage/code_completion.adoc - please update references in the manual to completion-category-overrides, flex, fuzzy... anything like that

cider-completion.el
```diff
@@ -292,6 +293,8 @@ Only affects the `cider' completion category.`"
     (setq completion-category-overrides (seq-remove (lambda (x)
                                                       (equal 'cider (car x)))
                                                     completion-category-overrides))
+    (unless found-styles
+      (setq found-styles '(styles)))
```
vemv, line 297: Shouldn't the default include `basic`?

vemv (commented):
Looking reasonable!

wrt unit tests, I don't have ideas at the moment. Might get back at it later today.

Either way I'd appreciate much if after the last round of feedback, you can try this branch over some 24-48 hours.

That QA-ing tends to catch most issues.

cider-completion.el
toniz4: lol, didn't remember that, I probably did it to keep the same behavior of completio-table-dynamic, but it's probably unnecessary. Maybe with multiple sessions it matters? In multiple buffers the completions work as expected.

alexander-yakushev: I assume it may be important when using `completion-at-point` directly and the interaction happens in the minibuffer but the code has to change in the code buffer, obviously.

toniz4: makes sense. Later this day I will be able to resolve all the reviews. I will also try to test it using the default emacs completions too. Thanks for all your help btw! It's awesome having people reviewing stuff so quickly.

cider-completion.el
toniz4: Yup, I think it's safer setting the competition override to '(styles flex basic).

bbatsov: Fine by me.

toniz4: @vemv, do you think we should put the flex style in the first position on the list, after 'styles? Currently it is appending the style list.

vemv: If you mean that the current produced list is wrong, sure, feel free to improve it

toniz4: > If you mean that the current produced list is wrong, sure, feel free to improve it

Not necessarily wrong, it just puts the flex completion style in a lower priority.

https://www.gnu.org/software/emacs/manual/html_node/emacs/Completion-Styles.html 

Updated the function and tests in the latest commit, let me know if there's a better way of doing this.

vemv: If you don't mind a lot / there isn't a huge drawback I'd put the list in the existing order (`basic flex`) so that there are no surprises for users.

It's something easily customizable anyway.

bbatsov:
Yeah, using `add-to-list` is better - it also adds only new elements and can add new elements to the beginning or end of a list.

cider-completion.el
toniz4: done

cider-completion.el
toniz4: Working on that. Sometimes fixing the docs is harder than fixing the code

toniz4: Updated the docs, let me know if the wording is fine. Writing in English is not my forte

test/cider-completion-tests.el
```diff
@@ -36,15 +36,27 @@
     (let ((old-value completion-category-overrides))
       (unwind-protect
           (progn
-            (it "adds `flex'"
+            (it "adds `flex' and `basic' as a fallback"
               (cider-enable-flex-completion)
               (expect (member 'flex (assq 'styles (assq 'cider completion-category-overrides)))
+                      :to-be-truthy)
+              (expect (member 'basic (assq 'styles (assq 'cider completion-category-overrides)))
```
vemv, line 43: While I wouldn't remove the existing assertions, I'd be included to add an extra one like this:

```
(expect completion-category-overrides :to-eq BLAH)
```

These exact expectations remove any doubt about the real behavior.

(and yes there's some redundancy between partial and exact expectations - but I prefer to err on that side here)

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -37,18 +37,27 @@ is already properly indented.
...
-The `cider` completion category currently only supports `basic` completion styles (and not `partial-completion`, `orderless`, etc),
-and `flex`, optionally (read more below).
+The CIDER completion at point function supports most completion styles, including
+`partial-completion`, `orderless` and `flex` (read more below).
 
-CIDER declares so in this fashion:
+
+Sometimes the user may want to use a different completion style just for the CIDER
+complete at point function. That can be achieved by setting
+`completion-category-overrides`, overriting the completion style of the CIDER
```
vemv, line 50:
```suggestion
`completion-category-overrides`, overwriting the completion style of the CIDER
```

vemv (approved):
LGTM!

Some final suggestions:

* I think the PR doesn't introduce caching anymore? If so please update PR title and description
* Please add a changelog entry

Cheers - V

toniz4:
Before merging, I would like to QA it more over the weekend. This week I was off work, so I couldn't test it in the real world.

bbatsov:
@toniz4 Sure. No rush.

toniz4:
I've been testing this since Sunday, it seems to be working great! If there's no more reviews to be made, I think it's ready to merge.

vemv:
Great work, kudos!

I'll be sure to use Cider master just now to give it some extra QA.

toniz4:
> Great work, kudos!
> 
> I'll be sure to use Cider master just now to give it some extra QA.

Thanks! Feel free to ping me if any problem appears

alexander-yakushev:
Installed the latest MELPA Cider and switched to `(cider-enable-flex-completion)` from `cider-company-enable-fuzzy-completion`. Works like a charm!

bbatsov:
Excellent! Seems we can hard-deprecate the company-specific code then.

toniz4:
> Excellent! Seems we can hard-deprecate the company-specific code then.

By removing completely the completion style? And make the old function just echo the deprecation notice

vemv:
I'd pursue the path of least obstrusiveness, i.e. try to keep things functional, and echo things once at most (and not on every completion).

Assuming that removing the cider completion style is a good idea, I still would keep it around assuming that it has no real cost. At least during the course of one stable release.

Users rarely enjoy starting a work day to broken setups :)

vemv:
`the company-specific code` also includes stuff that most certainly shouldn't be removed at all.

This code doesn't have a pure-Emacs equivalent and powers Company _and_ Corfu:

```
              :company-kind #'cider-company-symbol-kind
              :company-doc-buffer #'cider-create-compact-doc-buffer
              :company-location #'cider-company-location
              :company-docsig #'cider-company-docsig
```

toniz4:
> `the company-specific code` also includes stuff that most certainly shouldn't be removed at all.
> 
> This code doesn't have a pure-Emacs equivalent and powers Company _and_ Corfu:
> 
> ```
>               :company-kind #'cider-company-symbol-kind
>               :company-doc-buffer #'cider-create-compact-doc-buffer
>               :company-location #'cider-company-location
>               :company-docsig #'cider-company-docsig
> ```

yup, this is a valid implementation and should not be removed. 

I would just remove the competition style. But I agree on waiting a couple of stable releases first.

vemv:
We're talking about the `cider` completion style, correct?

It's not immediately obvious to me why it became redundant, mind to generously state here for posterity?

Last but not least - great work! No surprises since I adopted this. Between this and @alexander-yakushev's improvements I'm pretty positive that it feels snappier (especially for the doc part).

toniz4:
By my understanding and @bbatsov's confirmation, the completion style was made to fix flex completions in company. That was necessary because the complete-at-point function was not grabbing the prefix in a way that would be compatible with "flex-ish" competition styles.

And by the analysis made by @minad in #3006, the completion style by itself does not follow the specifications of a completion style. But somehow it appears to work with company.

The flex completion style has been on emacs for some years too.

toniz4:
Another option is to fix the completion style making it compliant. But since I don't use it and testing anything involving completions in emacs is a PITA, I think I would let someone pickup that

vemv:
Thanks for the summary!

A PR that marked all the now-redundant bits with comments would seem a good start.

In that PR itself we can iterate over the preferred soft/hard deprecation mechanism.

bbatsov:
> By my understanding and @bbatsov's confirmation, the completion style was made to fix flex completions in company. That was necessary because the complete-at-point function was not grabbing the prefix in a way that would be compatible with "flex-ish" competition styles.

Yeah, that is what I was referring to. We can keep around other `company`-specific code, but I'm not sure whether the old completion style adds any value. Down the road we can move the company logic outside of `cider-completion.el` (e.g. to `cider-company.el`), so we can load it only when someone actually uses company as a small additional optimization and to clarify better what's core completion logic vs what's specific to company.

Anyways, none of this is particularly important.

toniz4:
> Down the road we can move the company logic outside of `cider-completion.el` (e.g. to `cider-company.el`), so we can load it only when someone actually uses company as a small additional optimization and to clarify better what's core completion logic vs what's specific to company.

I think it's fine having the `:company-*` stuff there, corfu still uses some of those functions and maybe other completion frameworks use it too. Basically is a non standard standard.

PR #3660, opened 2024-05-13
Fix cider-inspector-def-current-val

alexander-yakushev:
Two issues with it:

1. The message in the minibuffer after `d` is used is ugly. 
2. The namespace where the values are defined is always `user`, because `*cider-inspect*` buffer is not associated with the code buffer where it was called from.

Before:
Screenshot 2024-05-13 at 11 09 31
After:
Screenshot 2024-05-13 at 11 14 58


bbatsov:
I'm fine with the proposed change. I guess the old message didn't consider that a value could be pretty long and won't fit the minibuffer well.

alexander-yakushev:
> I guess the old message didn't consider that a value could be pretty long and won't fit the minibuffer well.

That, and also that cider-nrepl op probably used to return the defined value as `"value"` in the nREPL message, but now it returns the rendered inspector state.

vemv (approved):

Issue #3636, opened 2024-03-15
Ability to auto-inspect on repl eval


vemv:
There's `cider-auto-inspect-after-eval` which only affects interactive eval, not repl eval.

This mostly makes sense, since it would be confusingly disruptive if, right after sending a form over the repl, Emacs focused to the Inspector buffer.

However after https://github.com/clojure-emacs/cider/pull/3634 it will be possible to keep the Inspector in the background.

So I'm thinking of a new defcustom, `cider-auto-inspect-after-repl-eval`, default nil.

When set to t, after every repl eval, the Inspector buffer would be refreshed with the corresponding value. This is handy for intrincate results.

bbatsov:
You can also make the old config accept symbols to have more granular behavior - e.g. `'interactive`, `'repl`, `'all`. Something along those lines. We've done such changes quite a few times in the past.

vemv:
Nice! Would be all for it

bbatsov:
@alexander-yakushev Is this anything you'd be interested in tackling?

PR #3661, opened 2024-05-13
Limit the maximum string size to be displayed in echo area

alexander-yakushev:
Although there is still work to be done (and thoughts to be thought) in cider-nrepl or nREPL regarding how much we want to print, it makes sense to solve such an issue on CIDER side too. Currently, if `cider-use-overlays` is not `t`, the standard `cider-interactive-eval-handler` will try to display the whole printed result in the minibuffer (echo area). For some reason, pushing huge strings through the echo area is much slower than printing it to the REPL. E.g., doing `(println (range 200000))` is almost instant while evaluating `(range 200000)` in the source buffer takes ~3 seconds on my machine to display the truncated result. Emacs would obviously still truncate the echo area output but not before printing the whole response to `*Messages*`.

This PR proposes to truncate the long output ahead of time. We can also add the similar message that we currently show in overlays, something like `"Result truncated. Type M-x cider-inspect-last-result to inspect it.` if it is deemed necessary.

Example of how the truncated result would look like:
image

### DIFF
--- a/cider-overlays.el
+++ b/cider-overlays.el
@@ -304,10 +304,10 @@ Note that, while POINT can be a number, it's preferable to be a marker, as
 that will better handle some corner cases where the original buffer is not
 focused."
   (cl-assert (symbolp value-type)) ;; We assert because for avoiding confusion with the optional args.
-  (let* ((font-value (if cider-result-use-clojure-font-lock
+  (let* ((value (string-trim-right value))
+         (font-value (if cider-result-use-clojure-font-lock
                          (cider-font-lock-as-clojure value)
                        value))
-         (font-value (string-trim-right font-value))
          (used-overlay (when (and point
                                   cider-use-overlays
                                   (if (equal 'error value-type)
@@ -316,10 +316,15 @@ focused."
                          (cider--make-result-overlay font-value
                            :where point
                            :duration cider-eval-result-duration
-                           :prepend-face (or overlay-face 'cider-result-overlay-face)))))
+                           :prepend-face (or overlay-face 'cider-result-overlay-face))))
+         (msg (format "%s%s" cider-eval-result-prefix value))
+         (max-msg-length (* (floor (max-mini-window-lines)) (frame-width)))
+         (msg (if (> (string-width msg) max-msg-length)
+                  (format "%s..." (substring msg 0 (- max-msg-length 3)))
+                msg)))
     (message
      "%s"
-     (propertize (format "%s%s" cider-eval-result-prefix font-value)
+     (propertize msg
                  ;; The following hides the message from the echo-area, but
                  ;; displays it in the Messages buffer. We only hide the message
                  ;; if the user wants to AND if the overlay succeeded.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [#3626](https://github.com/clojure-emacs/cider/issues/3626): `cider-ns-refresh`: jump to the relevant file/line on errors.
 - [#3628](https://github.com/clojure-emacs/cider/issues/3628): `cider-ns-refresh`: summarize errors as an overlay.
 - [#3660](https://github.com/clojure-emacs/cider/issues/3660): Fix `cider-inspector-def-current-val` always defining in `user` namespace.
+- [#3661](https://github.com/clojure-emacs/cider/issues/3661): Truncate echo area output ahead of time.
 - Bump the injected `enrich-classpath` to [1.19.3](https://github.com/clojure-emacs/enrich-classpath/compare/v1.19.0...v1.19.3).
 - Bump the injected nREPL to [1.1.1](https://github.com/nrepl/nrepl/blob/v1.1.1/CHANGELOG.md#111-2024-02-20).
 - Bump the injected `cider-nrepl` to [0.47.0](https://github.com/clojure-emacs/cider-nrepl/blob/v0.47.0/CHANGELOG.md#0470-2024-03-10).


cider-overlays.el
```diff
@@ -316,7 +316,11 @@ focused."
                          (cider--make-result-overlay font-value
                            :where point
                            :duration cider-eval-result-duration
-                           :prepend-face (or overlay-face 'cider-result-overlay-face)))))
+                           :prepend-face (or overlay-face 'cider-result-overlay-face))))
+         (max-message-width (* 10 (window-width))))
+    (when (> (string-width font-value) max-message-width)
+      (setq font-value (concat (substring font-value 0 (- max-message-width 3))
```
vemv, line 322: Two nits:

* seems better to reuse the `let*` rather than to setq - it's less imperative
* string-trim-right seems slightly better performed after the trimming?

cider-overlays.el
```diff
@@ -316,7 +316,11 @@ focused."
                          (cider--make-result-overlay font-value
```
vemv, line 316: `cider--make-result-overlay font-value` is using the non-trimmed value

alexander-yakushev:
I guess this intersects with #3635 (only spotted it now).

While I think that PR solves a more general problem, it may make sense to restrict it here as well. Dunno, let's discuss.

vemv (commented):
Seems a good improvement - thanks!

cider-overlays.el
alexander-yakushev: That's because `cider--make-result-overlay` performs its own trimming, so I guess it makes sense to pass the full result to it. In case, in the future, it decides to do something else with the result, so we don't obstruct that here.

alexander-yakushev:
I've looked through https://github.com/clojure-emacs/cider/pull/3635 and now I think that my PR is unnecessary. While with that PR, the echo area would still receive more than it could handle (e.g., 10k characters instead of ~1500 with this PR), the difference in perceived latency is not noticeable. On the plus side, we won't be doing similar work in two places.

 If we can fast forward #3635 – that will be great.

vemv:
> While I think that PR solves a more general problem, it may make sense to restrict it here as well. Dunno, let's discuss.

I don't mind about having both in place - this can be a last-resource measure.

A drawback being that users (/ maintainers) have to understand/tweak more stuff (e.g. let's imagine that each of the solutions has a different defcustom, or in any case, a different algo).

Maybe for the sake of containing complexity (i.e. favoring existing tech, namely the "print quotas"), let's go for https://github.com/clojure-emacs/cider/pull/3635 . I can polish it later this week or if it feels like a blocker, I'd be OK with you taking over it.

vemv:
> If we can fast forward https://github.com/clojure-emacs/cider/pull/3635 – that will be great.

You beat me to it ^^

as mentioned, I can go back to it soon, else you can take over.

alexander-yakushev:
Actually, let me reopen.

I just understood that because an overflowing message in the echo area truncates the beginning:

image

then, after #3635 is applied (which truncates the end), an overflowing message will show the middle of printed value, which is going to be quite ugly.

So, I think, we'll still need this PR, and perhaps even more so after #3635.

vemv:
If a small enough quota was requested for ops which output goes to the echo area, why would there be truncating in the beginning?

alexander-yakushev:
That implies that different quota is used when the output is going to the minibuffer vs the REPL vs somewhere else. From #3635 I gathered that it isn't the case.

Besides, this goes exactly against how `cider--display-interactive-eval-result` is implemented (at least, currently). The function receives a single result, and then sends it to two destinations (if `cider-use-overlays = 'both`). I guess, the echo area limit is still larger than the overlay limit (the overlay one is now at 3*frame-width), but basing the presentation behavior on that fells kinda flaky to me.

cider-overlays.el
alexander-yakushev: Ah, you meant trimming as in string-trim.

alexander-yakushev:
Addressed the comments.

> A drawback being that users (/ maintainers) have to understand/tweak more stuff (e.g. let's imagine that each of the solutions has a different defcustom, or in any case, a different algo).

I don't think we would need a `defcustom` for this one. Here, we are lucky to get "help" from the minibuffer implementation by being kinda "broken" currently since it can't go over `(max-mini-window-lines)`. So, in my opinion, any solution that we provide is better than status quo, and hence, we aren't going to break any current UX for people to miss.

vemv:
> That implies that different quota is used when the output is going to the minibuffer vs the REPL vs somewhere else. From https://github.com/clojure-emacs/cider/pull/3635 I gathered that it isn't the case.

You're partially right, but in that PR some parts were intentionally unchanged to not tighten the quota.

So indeed the 'overlay' and 'echo area' limits would be coupled.

Honestly I believe that it would be a very minor problem / edge case, because:

* most values are small
* for big values, most people won't carefully read every character in the echo area if they nearly fill their screen
  * in the inline overlay we can pretty-print things, with font-locking
  * there's `defcustom cider-auto-inspect-after-eval` as well.

So my thinking is that the echo area can help as a quick confirmation that evaluation succeeded - its rough value.

Because it's 'transient', it only has relative value as it can easily go away with keypresses or other events.

So, tldr I don't exactly oppose this PR but it only seems useful for making very specific things like `(range 20000)` nicer. For most exprs there isn't a huge difference or a problem to begin with.

I'd prefer to keep things simple and not add further moving pieces, since they only address edge cases in the less favored of the available UIs.

alexander-yakushev:
I agree with you that the problem is not a huge one. But I'd still prefer to see the beginning of the value in the minibuffer if the value is too large. At least, it will be consistent with how we trim values everywhere else.

`(range 20000)` is an example of the value that we will only show the middle of after #3636 is merged. But for any value over `(max-mini-window-lines)`, we'll still show the tail instead of the head, and seeing the head at a quick glance, in my opinion, gives much better information about the structure of the response, even if the tail is truncated.

cider-overlays.el
vemv: No, I didn't mean string-trim, but length trim

> That's because cider--make-result-overlay performs its own trimming

Alright then 👍

vemv (approved):

alexander-yakushev:
> since they only address edge cases in the less favored of the available UIs.

To your point about the UI being less favored – it is true that the default for `cider-use-overlays` is `'both`, however:

1. Both means we still print to minibuffer by default, so it is still visible to the user, even if the same value is also shown as an overlay. So we can assume that most users see the minibuffer output.
2. While I may be in the minority of users who disable overlays and only print the result to the minibuffer, the minibuffer is still the primary communication vessel in Emacs, so I would avoid writing it off as redundant or obsolete.

So, I think, if we can improve the UX of it with small costs, then it is reasonable to do so.

alexander-yakushev:
Thanks for the review!

vemv:
Cheers.

Frankly I'd review the default of `cider-use-overlays` at some point - overlays have gotten better and better over the last year, so showing duplicated info (and that can often look worse for the echo area part) makes little sense.

cider-overlays.el
```diff
@@ -316,10 +316,15 @@ focused."
...
                            :where point
                            :duration cider-eval-result-duration
-                           :prepend-face (or overlay-face 'cider-result-overlay-face)))))
+                           :prepend-face (or overlay-face 'cider-result-overlay-face))))
+         (msg (format "%s%s" cider-eval-result-prefix value))
```
bbatsov, line 320: Seems we need to print `font-value` instead of `value`, otherwise we'd lose the font-locking (if it was done).

cider-overlays.el
```diff
@@ -304,10 +304,10 @@ Note that, while POINT can be a number, it's preferable to be a marker, as
 that will better handle some corner cases where the original buffer is not
 focused."
   (cl-assert (symbolp value-type)) ;; We assert because for avoiding confusion with the optional args.
-  (let* ((font-value (if cider-result-use-clojure-font-lock
+  (let* ((value (string-trim-right value))
+         (font-value (if cider-result-use-clojure-font-lock
```
bbatsov, line 308: Probably here `font-locked-value` would have been a better name, although depending on the config it might not be font-locked. Naming is hard...

Issue #3663, opened 2024-05-13
Change `cider-interactive-eval-override` to make it more usable


chpill:
**Is your feature request related to a problem? Please describe.**

The `cider-interactive-eval-override` variable can be used to change the way forms are evaluated by `cider-interactive-eval`, but contrary to what is said [in the docstring](https://github.com/clojure-emacs/cider/blob/8b2ef5fc731e2b07956c5d10486d866bb37c5bf3/cider-eval.el#L1219-L1220), it does not actually forward `cider-interactive-eval`'s last optional argument `additional-params`.

Consider the following example, where I want to be able to wrap every evaluation with a custom form which simply prints `:wrapped` before the target form is evaluated:

```elisp
(defun println-override (form &optional callback bounds)
  (let ((cider-interactive-eval-override nil))
    (cider-interactive-eval (concat "(do (println :wrapped) " form ")")
                            callback
                            bounds)))

(setq cider-interactive-eval-override #'println-override)
```

This *almost* works, it correctly wraps any form evaluated and the `println` happens, but the pretty printing of some commands (like `cider-pprint-eval-last-sexp`) does not work anymore, because the last argument was not forwarded.

**Describe the solution you'd like**

Adding the missing argument to [`cider-interactive-eval-override`'s call site](https://github.com/clojure-emacs/cider/blob/8b2ef5fc731e2b07956c5d10486d866bb37c5bf3/cider-eval.el#L1232) `(funcall cider-interactive-eval-override form callback bounds additional-params)` is very easy, but it would break backward compatibility. I'm not very fluent in Elisp so maybe there is a way to work around that.

With that change, the previous code can be changed to the following:

```elisp
(defun println-override (form &optional callback bounds additional-params)
  (let ((cider-interactive-eval-override nil))
    (cider-interactive-eval (concat "(do (println :wrapped) " form ")")
                            callback
                            bounds
                            additional-params)))

(setq cider-interactive-eval-override #'println-override)
```

Which seems to work perfectly.

**Additional context**

As I searched for examples of correct usage of `cider-interactive-eval-override` on Github, it seems that it is in fact very rarely used. There might be a better way to wrap evaluation than this.

vemv:
Thanks for the accurate report!

Taking a look at the commit that introduced this https://github.com/clojure-emacs/cider/commit/02c4e801ecab254778e71eeacc19938229dc541f - originally all params were forwarded.

Surely that was unawarely broken later.

[...]

PR #3666, opened 2024-05-14
Fix cider-interactive-eval-override invocation

chpill:
This fixes the shortcomings of `cider-interactive-eval-override` described in [#3663](https://github.com/clojure-emacs/cider/issues/3663) in a backward compatible way.

-----------------


vemv (approved):
Thanks much!

bbatsov:
Looking good! Thanks!

PR #3669, opened 2024-05-17
Issue/gh action macos aarch64

### LINKED ISSUE #3668: Integration tests are broken
They all fail during setup:

```
Run purcell/setup-emacs@master
Run /Users/runner/work/_actions/purcell/setup-emacs/master/install.sh
Installing Nix
Configuring build cache and installing Emacs
  copying path '/nix/store/4ab6vrcph07w6ra79bc04fy8bbcmb9r0-source' from 'https://cache.nixos.org/'...
  error: flake 'github:purcell/nix-emacs-ci' does not provide attribute 'packages.aarch64-darwin.emacs-26-3', 'legacyPackages.aarch64-darwin.emacs-26-3' or 'emacs-26-3'
         Did you mean one of emacs-29-3, emacs-28-1, emacs-28-2, emacs-29-1 or emacs-29-2?
```

@ikappaki can please check what changed with the images?

ikappaki:
Hi,

Could you please review the patch to add a separate flow in the CI for macOS on Apple Silicon? It fixes #3668 

The macos-latest runner now targets Apple Silicon, and nix-emacs-ci supports version 28.x and above on this platform. I don't see this as an issue since I've also restored the old flow on the Intel macos 13.

By the way, as I've mentioned on other occasions, my personal opinion is that it is sufficient to only run/support the latest Emacs version on non-Unix platforms. :)

https://github.com/actions/runner-images/issues/9741

https://github.com/purcell/nix-emacs-ci/commit/ff5333aa28f65844da04fe6d81a404cff75f4f33#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

Thanks
 
-----------------

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,9 +21,17 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
-        emacs_version: ['26.3', '27.2', '28.2', '29.1']
+        os: [macos-13, ubuntu-latest, windows-latest]
+        emacs_version: ['26.3', '27.2', '28.2', '29.3']
         java_version: ['11', '17']
+        include:
+          # aarch64 (macos-13 is Intel)
+          - os: macos-latest
+            emacs_version: '29.3'
+            java_version: '11'
+          - os: macos-latest
+            emacs_version: '28.2'
+            java_version: '11'
 
     steps:
     - name: Set up Emacs


vemv:
Thanks for giving this a shot!

At the moment it's failing with:

> The template is not valid. .github/workflows/test.yml (Line: 38, Col: 18): A sequence was not expected

Moreover, IIRC macos-13 will be retired from GHA.

I'd be fine with removing the combinations that won't possibly work (newer macOS + older emacsen).

However `ci/circleci: test-ubuntu-emacs-26` is also failing. I don't know the reason for those ones.

Would be good to run emacs 26 on some or other CI.

ikappaki:
Hi @vemv :)

> At the moment it's failing with:
> 
> > The template is not valid. .github/workflows/test.yml (Line: 38, Col: 18): A sequence was not expected
> 


I think it should be fine now. I forgot  the matrix/include section does not accept arrays, thus I had to repeat the entries twice, and due to economy of space I've only included java version=11, let me know if you want me to add two more entries.

> Moreover, IIRC macos-13 will be retired from GHA.

Let me know what you think we should do. I'm fine supporting it as long as it is around.

> I'd be fine with removing the combinations that won't possibly work (newer macOS + older emacsen).

👍 

> However `ci/circleci: test-ubuntu-emacs-26` is also failing. I don't know the reason for those ones.

This is due to this commit https://github.com/clojure-emacs/cider/commit/8b2ef5fc731e2b07956c5d10486d866bb37c5bf3, unrelated to this commit.

`max-mini-window-lines` was introduced in v28 I think

> Would be good to run emacs 26 on some or other CI.

https://github.com/clojure-emacs/cider/issues/3207 :)

vemv:
> let me know if you want me to add two more entries for java=17 to get it inline with the rest.

Mmm, we should have it as there's some JVM-level testing. Another PR welcome.

> 👍 , just a note, this is a limitation of nix-emacs-ci, not a hard limit. If we try really hard enough we might be able to get older versions (but can't see the merit of it).

So arm macs can run older emacsen? That would seem worth looking into. No rush atm.

(the merit being to support our very diverse user base)

> > Would be good to run emacs 26 on some or other CI.

> https://github.com/clojure-emacs/cider/issues/3207 :)

The incident seems to support the choice of using both GHA and CCI - the redundancy keeps us (more-or-less) isolated from churn.

vemv:
> This is due to this commit https://github.com/clojure-emacs/cider/commit/8b2ef5fc731e2b07956c5d10486d866bb37c5bf3 https://github.com/clojure-emacs/cider/pull/3661 , unrelated to this patch.

Well spotted, thanks!

@alexander-yakushev ,  could you please take a look at the small breakage?

PR #3671, opened 2024-05-19
add macos aarch64 java/17 integration testing

ikappaki:
Hi,

could you please review follow patch up to #3669, to add integration testing for java/17 on the apple silicon on @vemv's request.

I've also added the badge for the ci integration tests on the github workflows, which was long overdue.

Thanks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,12 @@ jobs:
           - os: macos-latest
             emacs_version: '28.2'
             java_version: '11'
+          - os: macos-latest
+            emacs_version: '29.3'
+            java_version: '17'
+          - os: macos-latest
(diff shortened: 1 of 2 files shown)


bbatsov:
Seems to me we might be overdoing the platforms (OS + JVM) on which we run the integration tests, as I don't recall much problems historically that they have caught, but I'll merge this regardless. Adding the badge is a good idea. Thanks!

vemv:
Thanks, @ikappaki !

> Seems to me we might be overdoing the platforms (OS + JVM) on which we run the integration tests, as I don't recall much problems historically that they have caught,

It's a relatively young solution (< 1 year), a good chunk of its value being that we don't unawarely break the underlying feature that introduced those tests.

Issue #3672, opened 2024-05-21
False assumption in cider-popup-buffer-display (with fix)


PhilHudson:
## Expected behavior

Invoke `cider-doc` in the REPL, pop to fully-populated buffer `*cider-doc*`

## Actual behavior

Error message: 

`eval-expression: Wrong type argument: bufferp, "*cider-doc*"`

(where `*cider-doc*` is the first non-list element of `special-display-buffer-names`)

This is because code in function `cider-popup-display-buffer` falsely assumes that each element of `special-display-buffer-names` is a list. In fact and as documented, each element can be either a string or a list (the `car` of which is a string).

## Steps to reproduce the problem

Type eg `(str)` at the REPL prompt, then `C-c C-d C-d`.

## Fix

Change line 67 of `cider-popup.el` from:

```
(equal (car entry) buffer-name))
```

to:

```
(equal (if (listp entry) (car entry) entry) buffer-name))
```

I will submit a PR.

### CIDER version information

```
;; CIDER 1.14.0-snapshot (package: 20240514.1334), nREPL 1.1.1
;; Clojure 1.11.1, Java 1.8.0_402
```

### Lein / Clojure CLI version

Leiningen 2.10.0 on Java 1.8.0_402 OpenJDK 64-Bit Server VM

### Emacs version

GNU Emacs 28.1 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.18.0, Xaw3d scroll bars) of 2024-01-11

### Operating system

Linux **** 6.7.12-1-MANJARO #1 SMP PREEMPT_DYNAMIC Thu Apr  4 04:31:03 UTC 2024 x86_64 GNU/Linux

### JDK distribution

openjdk version "1.8.0_402"
OpenJDK Runtime Environment (build 1.8.0_402-b06)
OpenJDK 64-Bit Server VM (build 25.402-b06, mixed mode)

bbatsov:
Fixed by #3673.

PR #3673, opened 2024-05-21
Fix buggy special-display-buffer-names check.

PhilHudson:

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider-popup.el
+++ b/cider-popup.el
@@ -64,6 +64,9 @@ by adding BUFFER-NAME to the `special-display-buffer-names' list."
     ;; another time through `cider-popup-buffer-display'):
     (if (and (boundp 'special-display-buffer-names)
              (seq-find (lambda (entry)
+                         ;; Fix issue #3672 Phil Hudson 2024-05-21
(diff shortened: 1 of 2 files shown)


PhilHudson:
Fixes issue #3672

cider-popup.el
```diff
@@ -64,7 +64,7 @@ by adding BUFFER-NAME to the `special-display-buffer-names' list."
     ;; another time through `cider-popup-buffer-display'):
     (if (and (boundp 'special-display-buffer-names)
              (seq-find (lambda (entry)
-                         (equal (car entry) buffer-name))
+                         (equal (if (listp entry) (car entry) entry) buffer-name))
```
bbatsov, line 67: > This is because code in function cider-popup-display-buffer falsely assumes that each element of special-display-buffer-names is a list. In fact and as documented, each element can be either a string or a list (the car of which is a string).

Can you add some variant of this as a comment here for clarity's sake?

bbatsov:
I'm fine with the proposed change. Just add some comment about it and a changelog entry as well.

PR #3676, opened 2024-05-23
Update cheatsheet to v55

### LINKED ISSUE #3674: Update cheatsheet
I want to update cheatsheet to match the latest version of the official cheatsheet (v55). Right now, it kind of looks like v43 of the official cheatsheet, but there are a lot of little differences here and there, so it is hard to compare and update it to the latest version. For example, there are unrelated functions such as `nil?`, `some?`, `identical?` in the section with number test functions, while functions such as `number?`, `rational?`, `integer?` are not there.

What do you think about making it identical to the latest official cheatsheet, so it will be easier to maintain in the future? If you want to keep the current differences, I can just do a little cleanup and add the changes from the latest version.

katomuso:
Update cheatsheet to match v55 of the official cheatsheet. Also, reorder sections as they are ordered in the official cheatsheet and move additional sections, which are not in the official cheatsheet, to the end.

Closes #3674

### DIFF
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -32,56 +32,40 @@
 (require 'seq)
 
 (defconst cider-cheatsheet-hierarchy
-  '(("Primitives"
+  '(("Documentation"
+     ("REPL"
+      (clojure.repl doc find-doc apropos dir source pst)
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

Issue #3678, opened 2024-05-24
Current way of displaying cheatsheet is inconvenient


katomuso:
Initially, cheatsheet was designed to display the entire hierarchy in a single Helm buffer (as seen in the screenshot at https://github.com/clojure-emacs/clojure-cheatsheet), allowing the user to search through the entire cheatsheet at once. However, the current implementation using `completing-read` requires us to navigate one level of the hierarchy at a time. This is much less convenient because we need to know where the item we are looking for is located, and even if we do know, it takes additional steps to do so.

Instead of having multiple completions for each level of the hierarchy, we can represent the hierarchy in a flat view, so a single line will display the full path to an item, with the path elements separated by a specific separator. For example, lines might look like `Collections > Maps > Change > clojure.core/assoc` with `>` as the separator. Using a fuzzy completion style (such as the one provided by the `orderless.el` package), we will be able to search for any element within the path, potentially displaying matches from multiple sections at the same time, which is especially convenient when displaying completions vertically (for example, with the `vertico.el` package).

This can be implemented by transforming the hierarchy into an appropriate flat format that will be displayed using `completing-read`. When the user selects a line, we split it by the separator and get the last item, which is the symbol. Then, we show the documentation for it as usual. I can do this, but I'm not sure if it's better to implement it as a separate function that can be called alternatively to `cider-cheatsheet` (something like `cider-cheatsheet-flat`) or to alternate the behavior of `cider-cheatsheet` when a prefix argument is passed.

johngit22:
Thank you for the tidbits of information 😌

John Boyd

bbatsov:
> Instead of having multiple completions for each level of the hierarchy, we can represent the hierarchy in a flat view, so a single line will display the full path to an item, with the path elements separated by a specific separator. For example, lines might look like Collections > Maps > Change > clojure.core/assoc with > as the separator. Using a fuzzy completion style (such as the one provided by the orderless.el package), we will be able to search for any element within the path, potentially displaying matches from multiple sections at the same time, which is especially convenient when displaying completions vertically (for example, with the vertico.el package).

I'd be fine with this. It's true that the current UI is a bit clunky with the separate section selection step.

katomuso:
Good, I have some preliminary work done, now I just need to think about how to best implement it from a user interaction perspective.

PR #3681, opened 2024-05-26
Display cheatsheet in a separate buffer

katomuso:
While thinking about #3678 on how we can improve the current cheatsheet interface, I came up with the idea that we need to make it possible to display cheatsheet in a separate buffer.

The current cheatsheet is a subset of the official cheatsheet, displaying only functions using completion interface. However, by displaying cheatsheet in a buffer, we can have a richer output and make it identical to the official cheatsheet in terms of the information it provides. For example, we can make headings clickable to link to documentation, and we can display literals and other strings from the official cheatsheet that are currently omitted.

At this point, it looks like this: 
cheatsheet-buffer

Following the naming pattern of `cider-apropos` and `cider-apropos-select`, I've divided this functionality into two commands: `cider-cheatsheet`, which displays cheatsheet in the `*cider-cheatsheet*` buffer, and `cider-cheatsheet-select`, which provides a completion interface (as the current `cider-cheatsheet` command does).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### New features
 
+- [#3681](https://github.com/clojure-emacs/cider/pull/3681): Add an alternative way to display cheatsheet in a buffer and make it the default.
+  - Current `cider-cheatsheet` command is renamed to `cider-cheatsheet-select`.
+  - New way to display cheatsheet in a buffer is available with `cider-cheatsheet` command.
 - [#3632](https://github.com/clojure-emacs/cider/pull/3623): Add new configuration variable `cider-clojure-cli-global-aliases`.
 - [#3366](https://github.com/clojure-emacs/cider/pull/3366): Support display of error overlays with `#dbg!` and `#break!` reader macros.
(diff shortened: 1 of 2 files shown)


bbatsov:
Can you share some screenshot showing the proposed UI?

bbatsov:
Seems we have forgotten to document this functionality completely, so it will also be nice to add some section about the `cider-cheatsheet` commands here https://docs.cider.mx/cider/usage/misc_features.html

katomuso:
> Can you share some screenshot showing the proposed UI?

I've updated the description with a screenshot. Right now, it is nothing fancy, but it is a base on which we can iterate.

katomuso:
> Seems we have forgotten to document this functionality completely, so it will also be nice to add some section about the `cider-cheatsheet` commands here https://docs.cider.mx/cider/usage/misc_features.html

Okay, I'll do it. In the near future, I plan a series of changes. Should I do this now or after that?

bbatsov:
> Okay, I'll do it. In the near future, I plan a series of changes. Should I do this now or after that?

You can do this later then. For now just add some changelog entry about the new way to display the cheatsheet.

katomuso:
I've updated the changelog. Since this is my first time doing so, I hope I did it correctly, though I was unsure about the order of the entries, as they don't appear to be ordered by PR number.

PR #3683, opened 2024-05-27
jack-in support for the Basilisp Clojure dialect in Python

### LINKED ISSUE #3682: jack-in support for Basilisp

`jack-in` support for [Basilisp ](https://github.com/basilisp-lang/basilisp).

Ability to `jack-in` to Basilisp.

Use CIDER connect to the basilisp nREPL server, but this require to manually start the server.

**Additional context**

Basilisp is a Clojure compatible dialect implemented in Python.

I'll follow up with a PR.

ikappaki:
Hi,

can you please consider merging this patch to support Basilisp jack-in in CIDER. it resolves #3682.

I have added added an integration test and updated the documentation accordingly.

Currently the integration test currently fails on MS-Windows and has been temporarily disabled. This issue is addressed by https://github.com/basilisp-lang/basilisp/issues/865 but the fix has not been released yet in PyPi. The problem is related flushing the stdout after a `println`. It only fails in CI, but works fine when invoked from a user's session.

Additionally, it requires `clojure-mode` v5.19 to support recognizing Basilisp's `.lpy` file extension

Thanks

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -12,7 +12,7 @@
 ;; Maintainer: Bozhidar Batsov <bozhidar@batsov.dev>
 ;; URL: https://www.github.com/clojure-emacs/cider
 ;; Version: 1.14.0-snapshot
-;; Package-Requires: ((emacs "26") (clojure-mode "5.18.1") (parseedn "1.2.1") (queue "0.2") (spinner "1.7") (seq "2.22") (sesman "0.3.2") (transient "0.4.1"))
+;; Package-Requires: ((emacs "26") (clojure-mode "5.19") (parseedn "1.2.1") (queue "0.2") (spinner "1.7") (seq "2.22") (sesman "0.3.2") (transient "0.4.1"))
 ;; Keywords: languages, clojure, cider
 
 ;; This program is free software: you can redistribute it and/or modify
@@ -273,6 +273,30 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.6.0"))
 
+(defcustom cider-basilisp-command
+  "basilisp"
+  "The command used to execute Basilisp.
+
+   If Basilisp is installed in a virtual environment, update this to the
+   full path of the Basilisp executable within that virtual environment."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.14.0"))
+
+(defcustom cider-basilisp-global-options
+  nil
+  "Command line options used to execute Basilisp."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.14.0"))
+
+(defcustom cider-basilisp-parameters
+  "nrepl-server"
+  "Params passed to Basilisp to start an nREPL server via `cider-jack-in'."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.14.0"))
+
 (make-obsolete-variable 'cider-lein-global-options 'cider-lein-parameters "1.8.0")
 (make-obsolete-variable 'cider-boot-global-options 'cider-boot-parameters "1.8.0")
 (make-obsolete-variable 'cider-clojure-cli-global-options 'cider-clojure-cli-parameters "1.8.0")
@@ -280,6 +304,7 @@ By default we favor the project-specific shadow-cljs over the system-wide."
 (make-obsolete-variable 'cider-gradle-global-options 'cider-gradle-parameters "1.8.0")
 (make-obsolete-variable 'cider-babashka-global-options 'cider-babashka-parameters "1.8.0")
 (make-obsolete-variable 'cider-nbb-global-options 'cider-nbb-parameters "1.8.0")
+(make-obsolete-variable 'cider-basilip-global-options 'cider-basilisp-parameters "1.8.0")
 
 (defcustom cider-jack-in-default
   (if (executable-find "clojure") 'clojure-cli 'lein)
@@ -296,7 +321,8 @@ to Leiningen."
                  (const shadow-cljs)
                  (const gradle)
                  (const babashka)
-                 (const nbb))
+                 (const nbb)
+                 (const basilisp))
   :safe #'symbolp
   :package-version '(cider . "0.9.0"))
 
@@ -316,6 +342,7 @@ command when there is no ambiguity."
                  (const gradle)
                  (const babashka)
                  (const nbb)
+                 (const basilisp)
                  (const :tag "Always ask" nil))
   :safe #'symbolp
   :package-version '(cider . "0.13.0"))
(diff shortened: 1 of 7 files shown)


cider.el
```diff
@@ -273,13 +273,38 @@ By default we favor the project-specific shadow-cljs over the system-wide."
...
+  "basilisp"
+  "The command used to execute Basilisp.
+
+   If Basilisp is installed in a virtual environment, update this to the
+   full path of the Basilisp executable within that virtual environment."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.14.0"))
+
+(defcustom cider-basilisp-global-options
```
bbatsov, line 286: I think it's best to avoid adding this, given it's deprecated out of the box. :-)

cider.el
```diff
@@ -488,6 +518,7 @@ Throws an error if PROJECT-TYPE is unknown."
     ('shadow-cljs cider-shadow-cljs-global-options)
     ('gradle      cider-gradle-global-options)
     ('nbb         cider-nbb-global-options)
+    ('basilisp    cider-basilisp-global-options)
```
bbatsov, line 521: Here you can just hardcode `nil`.

cider.el
```diff
@@ -963,6 +995,10 @@ middleware and dependencies."
            global-opts
            (unless (seq-empty-p global-opts) " ")
            params))
+    ('basilisp (concat
```
bbatsov, line 998: This can also be hardcoded.

bbatsov:
The PR looks good to me overall, sans my small remarks. 

Does Basilisp implement the nREPL protocol fully? I'm wondering if there are some caveats that need to be documented if some ops are not implemented.

cider.el
ikappaki: Removed.

ikappaki:
I've just realized that jacking in on MS-Windows won't work even in a user session (I left a `flush` in while testing), contrary to what I mentioned earlier. I think it's better to wait until the new Basilisp release with the fix is out fix before merging ...

> Currently the integration test currently fails on MS-Windows and has been temporarily disabled. This issue is addressed by https://github.com/basilisp-lang/basilisp/issues/865 but the fix has not been released yet in PyPi. The problem is related flushing the stdout after a `println`. It only fails in CI, but works fine when invoked from a user's session.

bbatsov:
> The Basilisp nREPL server currently implements the followings ops: eval, describe, info, eldoc, clone, close, load-file and complete, which cover the fundamental ops in my opinion. If you believe there are any caveats that need to be documented, please let me know.

I'm guessing you mean `completions` and `lookup`, not `complete` and `info`, right? Not a big difference for CIDER, but the former are nREPL ops and the latter are `cider-nrepl` ops, which served as the basis for the nREPL ops. See https://metaredux.com/posts/2020/06/15/nrepl-0-8-evolving-the-protocol.html for details.

ikappaki:
Hi @bbatsov,

 I've checked in an update referencing the latest basilisp version with the recently released fix (thanks @chrisrink10). All tests now pass.

> > The Basilisp nREPL server currently implements the followings ops: eval, describe, info, eldoc, clone, close, load-file and complete, which cover the fundamental ops in my opinion. If you believe there are any caveats that need to be documented, please let me know.
> 
> I'm guessing you mean `completions` and `lookup`, not `complete` and `info`, right? Not a big difference for CIDER, but the former are nREPL ops and the latter are `cider-nrepl` ops, which served as the basis for the nREPL ops. See https://metaredux.com/posts/2020/06/15/nrepl-0-8-evolving-the-protocol.html for details.

I failed to mention earlier that the nREPL server is a port of the nbb nREPL server. I've kept the same ops, minus the `macroexpand` (which I couldn't find a use for at the time) and `classpath` which I plan to follow up at some point.

I haven't looked at the standard supported ops, but I'm reluctant of remove or renaming anything at this stage in fear I might break Calva support. I'll clean it up in the next basilisp nREPL update.

By the way, the `complete` keyword does appear to be used by CIDER (not sure if you meant it is not). 

thanks

```elisp
(-->
  id         "3"
  op         "describe"
  session    "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
)
(<--
  id         "3"
  session    "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
  ops        (dict
               clone     (dict)
               close     (dict)
               complete  (dict)
               describe  (dict)
               eldoc     (dict)
               eval      (dict)
               info      (dict)
               load-file (dict))
  status     ("done")
  versions   (dict
               basilisp (dict
                          incremental    "1"
                          major          "0"
                          minor          "."
                          version-string "0...1...0.b.2")
               python   (dict
                          incremental    "1"
                          major          "3"
                          minor          "12"
                          version-string "3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]"))
)
(-->
  id                        "10"
  op                        "complete"
  session                   "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
  context                   "nil"
  enhanced-cljs-completion? "t"
  ns                        "user"
  prefix                    #("printl" 0 1 (fontified t) 1 2 (fontified t) 2 3 (fontified t) 3 4 (fontified t) 4 5 (fontified t) 5 6 (font-lock-multiline t fontified t))
)
(<--
  id          "10"
  session     "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
  completions ((dict "candidate" "println" "ns" "basilisp.core" "type" "function")
 (dict "candidate" "println-str" "ns" "basilisp.core" "type" "function"))
  status      ("done")
)
(-->
  id         "11"
  op         "eldoc"
  session    "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
  context    "nil"
  ns         "user"
  sym        "println"
)
(<--
  id         "11"
  session    "b8a83d6d-84ee-4ab3-a1c0-92cd548634cf"
  docstring  "Print the arguments to the stream bound to :lpy:var:`*out*` ..."
  eldoc      (nil
 ("x")
 ("x" "&" "args"))

  name       "println"
  ns         "basilisp.core"
  status     ("done")
  status     ("done")
  type       "function"
)

```

ikappaki:
Hi @bbatsov 

> CIDER users both it's own ops (if present) and the standard nREPL ops if they are not. In general I think it's best for most nREPL implementations to stick to the official ops for completions and lookup, as this means they would work with more clients. 

Is this the official list of operations supported by the nREPL protocol that servers should adhere to? https://nrepl.org/nrepl/ops.html

> I've meant for a while to rename the CIDER ops to something like `cider/complete` and `cider/info`, so it's easier to figure out what's part of the protocol and what's a CIDER extension, but I never got to doing so.

That would be very helpful indeed.

PR #3686, opened 2024-05-29
Add an alternative way to display `cider-cheatsheet-select`

katomuso:
Instead of having the multi-step selection process, which is the default `cider-cheatsheet-select` behavior, we represent each candidate as a full path to a var when `cider-cheatsheet-select` is called with a prefix argument. This can be handy with fuzzy completion style and vertical candidates display, as discussed in #3678.

Here's how the default multi-step selection process looks when you need to select every single section before choosing a var:
cider-cheatsheet-select-default

In contrast, when `cider-cheatsheet-select` is called with a prefix argument, each candidate is a full path to a var:
cider-cheatsheet-select-flat

### DIFF
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -32,6 +32,7 @@
 (require 'cl-lib)
 (require 'map)
 (require 'seq)
+(require 'subr-x)
 
 (defconst cider-cheatsheet-hierarchy
   '(("Documentation"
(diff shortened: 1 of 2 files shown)


cider-cheatsheet.el
```diff
@@ -549,18 +549,35 @@ This list is supposed to have the following format:
...
-  (interactive)
-  (let ((hierarchy cider-cheatsheet-hierarchy))
-    (while (stringp (caar hierarchy))
-      (let* ((sections (mapcar #'car hierarchy))
-             (section (completing-read "Select section: " sections)))
-        (setq hierarchy (map-elt hierarchy section))))
-    (let* ((vars (seq-mapcat #'cider-cheatsheet--expand-vars hierarchy))
-           (var (completing-read "Select var: " vars)))
-      (cider-doc-lookup var))))
+(defun cider-cheatsheet-select (&optional flat)
```
bbatsov, line 562: Perhaps this can be a defcustom instead and you can use the prefix arg to invert whatever the value of the defcustom is.

bbatsov:
Nice improvement - i definitely like it.

cider-cheatsheet.el
katomuso: Good idea! I've seen such a pattern a few times, but haven't thought about it this time. I planned to think about customizable variables later, so I will keep this in mind.

katomuso:
I wonder why the lint check failed because `eldev lint` shows no warnings...

katomuso:
> I wonder why the lint check failed because `eldev lint` shows no warnings...

Oh, I found it. It has to do with byte compilation and the `string-split` function. I don't need to fix this kind of warnings, right?

bbatsov:
You just need to require `subr-x` and the error will go away.

katomuso:
> You just need to require `subr-x` and the error will go away.

Hmm, it seems that the error is still present, as `string-split` is located in `subr.el`, while other string functions indeed are located in `subr-x.el`.

bbatsov:
Ah, my bad. Then it's very weird, as `subr.el` is auto-required always.

bbatsov:
Hmm, I see that the function is actually named `split-string`, so it seems the warning has discovered an actual error. :-)

cider-cheatsheet.el
```diff
@@ -32,6 +32,7 @@
 (require 'cl-lib)
 (require 'map)
 (require 'seq)
+(require 'subr-x)
```
bbatsov, line 35: In light of my last comment you can remove this.

cider-cheatsheet.el
```diff
@@ -549,18 +550,35 @@ This list is supposed to have the following format:
...
+(defun cider-cheatsheet-select (&optional flat)
+  "Navigate cheatsheet sections and show documentation for selected var.
+
+With a prefix argument FLAT, represent each candidate as a full path to var."
+  (interactive "P")
+  (if flat
+      (let* ((hierarchy (cider-cheatsheet--flatten-hierarchy cider-cheatsheet-hierarchy))
+             (paths (mapcar (lambda (sections) (string-join sections " > ")) hierarchy))
+             (path (completing-read "Select path: " paths))
+             (var (car (last (string-split path " > ")))))
```
bbatsov, line 572: It's actually `split-string`. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Strings.html

No idea how this ever worked for you, if the name of the function is wrong.

cider-cheatsheet.el
katomuso: As I actually use `string-join` from `subr-x`, I think it will be okay to be explicit about it by requiring it.

cider-cheatsheet.el
katomuso: Well, this is an alias. I guess that may be the reason it can't find it during byte compilation.

bbatsov: It also seems this alias was only added in Emacs 29.

katomuso: Yep, changing the name fixed it. It still seems weird, but I'll be more cautious with aliases from now on :)

katomuso: > It also seems this alias was only added in Emacs 29.

I see, so that's the reason. Thanks for helping me figure it out.

katomuso:
So now that all the checks are passing, can I add a new entry to the changelog so we can merge the change? I'll add customizable variables in the future as I am still not sure about some of the details.

bbatsov:
Sure.

bbatsov:
Instead of having the multi-step selection process, which is the default cider-cheatsheet-select behavior, we represent each candidate as a full path to a var when cider-cheatsheet-select is called with a prefix argument. This can be handy with fuzzy completion style and vertical candidates display.

bbatsov:
Thanks!

Issue #3687, opened 2024-05-29
Inline error overlays not displaying


yuhan0:
The changes introduced in #3616 to filter out stderr messages cause actual eval errors to be swallowed up when `cider-show-error-buffer` is set to nil.

## Expected behavior
Error messages should always be displayed to the user.

## Actual behavior
In certain situations, there is no indication is given of the eval command having successfully completed/ failed, which may be misinterpreted as ongoing evaluation or a REPL crash.

## Steps to reproduce the problem
Elisp:
```lisp
(setq cider-show-error-buffer nil)
```

### JVM clojure

`cider-eval` the following and observe the lack of an error overlay (the only feedback being an error output in the possibly hidden REPL buffer)

```clj
(slurp "does-not-exist.txt")
```
Stderr output:
```
Execution error (FileNotFoundException) at java.io.FileInputStream/open0 (FileInputStream.java:-2).
does-not-exist.txt (No such file or directory)
```

### Babashka
Jack into a bb repl and trigger any exception, observe that all error overlays are suppressed.
```clj
(ns repro)
(inc nil)
```
Stderr output: 
```bash
java.lang.NullPointerException repro /Users/yuhan/.local/scripts/repro.bb:2:1
```

## Additional context
The issue can be traced to this conditional 
https://github.com/clojure-emacs/cider/blob/a74dff972c2f941cb562ac05251ebbf649eed4df/cider-eval.el#L925
which only displays the error overlay if the stderr output matches an expected clojure-1.10 regexp, notice that the exception thrown from `slurp`  doesn't match due to the 'line number' being negative:

```
Execution error (FileNotFoundException) at java.io.FileInputStream/open0 (FileInputStream.java:-2).
does-not-exist.txt (No such file or directory)
```

One straightforward fix would be to tweak the `cider-clojure-1.10--location` regexp to accept the `-` char
https://github.com/clojure-emacs/cider/blob/a74dff972c2f941cb562ac05251ebbf649eed4df/cider-eval.el#L571

But this seems to be an especially brittle way of doing things - are we sure that the regex fully describes all valid error messages? I had no idea that interop calls could result in negative line numbers, maybe it's due to the absence of .java sources or specific to the JDK that I'm running, and I'm certain there are more edge cases that aren't covered by this regex.

[...]

PR #3688, opened 2024-05-29
Handle negative line numbers in error messages

yuhan0:
(partially fixes) #3687
I took the opportunity to refactor some of the rx forms and add documentation, let me know if these should be separated out or squashed.

-----------------

Thanks!

manual][1] extremely useful.*


yuhan0:
Yup, the existing `cider-clojure-runtime-error-regexp` tests cover both those cases.

vemv (approved):

PR #3689, opened 2024-05-30
Show friendly error message if symbol is not found on ClojureDocs

katomuso:
When we use `cider-clojuredocs`, it shows the error message `Wrong type argument: char-or-string-p, nil` if a symbol we are looking for is not found. It will be nice if it will show something more meaningful instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 
 ### Bugs fixed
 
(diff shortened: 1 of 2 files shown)


katomuso:
> P.S. I'll be cutting a new release of CIDER soon, so this would be a good time to also add some documentation about `cider-cheatsheet` with a few screenshots, etc.

I was planning to add customizable variables for the cheatsheet tomorrow, and then update the documentation the day after tomorrow. Will this be in time for the new release?

bbatsov:
Take your time, there's no rush. It will be late for this release, at it happened just now, but there's no problem to include those changes in the next one.

PR #3690, opened 2024-05-31
Add customizable variables for cheatsheet

katomuso:
Add multiple customizable variables to control different aspects of `cider-cheatsheet` and `cider-cheatsheet-select` behavior.

### DIFF
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -46,6 +46,17 @@
   :type 'boolean
   :package-version '(cider . "1.15.0"))
 
+(defcustom cider-cheatsheet-default-action-function #'cider-doc-lookup
+  "Function to use on a var when it is selected.
+
+By default, documentation for a var is displayed using `cider-doc-lookup`,
+but it can also be set to `cider-clojuredocs-lookup` to show documentation
+from ClojureDocs or any other function accepting a var as an argument."
+  :type '(choice (const cider-doc-lookup)
+                 (const cider-clojuredocs-lookup)
+                 function)
+  :package-version '(cider . "1.15.0"))
+
 (defconst cider-cheatsheet-hierarchy
   '(("Documentation"
      ("REPL"
@@ -582,7 +593,7 @@ With a prefix argument FLAT, represent each candidate as a full path to var."
              (paths (mapcar (lambda (sections) (string-join sections " > ")) hierarchy))
              (path (completing-read "Select path: " paths))
              (var (car (last (split-string path " > ")))))
-        (cider-doc-lookup var))
+        (funcall cider-cheatsheet-default-action-function var))
     (let ((hierarchy cider-cheatsheet-hierarchy))
       (while (stringp (caar hierarchy))
         (let* ((sections (mapcar #'car hierarchy))
@@ -590,7 +601,7 @@ With a prefix argument FLAT, represent each candidate as a full path to var."
           (setq hierarchy (map-elt hierarchy section))))
       (let* ((vars (seq-mapcat #'cider-cheatsheet--expand-vars hierarchy))
              (var (completing-read "Select var: " vars)))
-        (cider-doc-lookup var)))))
+        (funcall cider-cheatsheet-default-action-function var)))))
 
 (cl-defun cider-cheatsheet--insert-hierarchy (hierarchy &optional (level 0))
   "Insert HIERARCHY with visual indentation for LEVEL."
@@ -604,7 +615,8 @@ With a prefix argument FLAT, represent each candidate as a full path to var."
         (insert-text-button var
                             'var var
                             'action (lambda (btn)
-                                      (cider-doc-lookup (button-get btn 'var)))
+                                      (funcall cider-cheatsheet-default-action-function
+                                               (button-get btn 'var)))
                             'help-echo (format "Show documentation for %s" var))
         (insert "\n")))))
 
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -41,6 +41,11 @@
 
 (defconst cider-cheatsheet-buffer "*cider-cheatsheet*")
 
+(defcustom cider-cheatsheet-auto-select-buffer t
+  "Whether to auto-select the cheatsheet popup buffer."
+  :type 'boolean
+  :package-version '(cider . "1.15.0"))
+
 (defconst cider-cheatsheet-hierarchy
   '(("Documentation"
      ("REPL"
@@ -613,7 +618,8 @@ With a prefix argument FLAT, represent each candidate as a full path to var."
 (defun cider-cheatsheet ()
   "Display cheatsheet in a popup buffer."
   (interactive)
-  (with-current-buffer (cider-popup-buffer cider-cheatsheet-buffer)
+  (with-current-buffer (cider-popup-buffer cider-cheatsheet-buffer
+                                           cider-cheatsheet-auto-select-buffer)
     (read-only-mode -1)
     (insert (cider-cheatsheet--buffer-contents))
     (read-only-mode 1)
--- a/cider-cheatsheet.el
+++ b/cider-cheatsheet.el
@@ -34,6 +34,8 @@
 (require 'seq)
 (require 'subr-x)
 
+(defconst cider-cheatsheet-buffer "*cider-cheatsheet*")
+
 (defconst cider-cheatsheet-hierarchy
   '(("Documentation"
      ("REPL"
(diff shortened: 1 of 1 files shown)


katomuso:
@bbatsov Can I be added to the `cider-cheatsheet.el` package authors?

cider-cheatsheet.el
```diff
@@ -34,6 +34,49 @@
...
+(defgroup cider-cheatsheet nil
+  "Clojure cheatsheet in CIDER."
+  :prefix "cider-cheatsheet-"
+  :group 'cider)
+
+(defconst cider-cheatsheet-buffer "*cider-cheatsheet*")
+
+(defcustom cider-cheatsheet-auto-select-buffer t
+  "Whether to auto-select the cheatsheet popup buffer."
+  :type 'boolean)
```
bbatsov, line 46: You should also add `:package-version` properties to all `defcustoms` (saying `1.15.0` as the version).

cider-cheatsheet.el
```diff
@@ -34,6 +34,49 @@
...
+
+(defcustom cider-cheatsheet-doc-function #'cider-doc-lookup
+  "Function used to display documentation on var.
+
+Accepts var as an argument when var is selected in the cheatsheet."
+  :type '(choice (const :tag "Doc" cider-doc-lookup)
+                 (const :tag "ClojureDocs" cider-clojuredocs-lookup)
+                 function))
+
+(defcustom cider-cheatsheet-select-path nil
```
bbatsov, line 56: Probably it'd better for this to be an option taking some symbol (the name of the type of behavior to use) instead of a boolean - it would make the code more flexible if we decide to add different options down the road.

cider-cheatsheet.el
```diff
@@ -34,6 +34,49 @@
...
+  :prefix "cider-cheatsheet-"
+  :group 'cider)
+
+(defconst cider-cheatsheet-buffer "*cider-cheatsheet*")
+
+(defcustom cider-cheatsheet-auto-select-buffer t
+  "Whether to auto-select the cheatsheet popup buffer."
+  :type 'boolean)
+
+(defcustom cider-cheatsheet-doc-function #'cider-doc-lookup
```
bbatsov, line 48: Maybe use a more generic name like `cider-cheatsheet-(default)-action-function`? And obviously in the buffer view we can also have more actions bound to different keybindings (the default could be the one triggered by RET)

cider-cheatsheet.el
```diff
@@ -34,6 +34,49 @@
...
+
+(defcustom cider-cheatsheet-select-path-separator " > "
+  "Separator to use between elements of a full path to a var.
+
+Affects `cider-cheatsheet-select' when `cider-cheatsheet-select-path'
+is non-nil.  Can't contain characters present in elements of a path,
+such as '/' and '&'."
+  :type 'string)
+
+(defcustom cider-cheatsheet-indentation-function
```
bbatsov, line 71: I'm not quite certain that many people will want to modify this. Do you have any particular use-case in mind?

bbatsov:
> @bbatsov Can I be added to the cider-cheatsheet.el package authors?

Sure.

cider-cheatsheet.el
katomuso: Done.

cider-cheatsheet.el
katomuso: I think that it's a good idea to create a variable like `cider-cheatsheet-select-type` or `cider-cheatsheet-select-display`, which accepts symbols like `multi` (default) and `path`. However, one small consequence of this approach is that we won't be able to use `cider-cheatsheet-select` with a prefix argument anymore as it doesn’t make sense to negate the value of this variable as it did with `cider-cheatsheet-select-path`, which accepted a boolean. Will this be okay?

katomuso: Actually, we can. We just need to check the exact prefix argument value and use the value of `cider-cheatsheet-select-type` accordingly, as it is done in `cider-jack-in-universal`, right?

cider-cheatsheet.el
katomuso: I thought that, since it is related to UI, some users might want to modify how it looks. However, I agree that most users won't do this, and it is actually not that safe to modify that variable, as it can break `cider-cheatsheet-select`. So, I think it would be a good idea to move it to a constant.

cider-cheatsheet.el
katomuso: I'm not exactly sure what you mean. Maybe you can point me to some place where such logic is already implemented, so I can understand it better? I think some examples of values for such a variable will also be helpful.

katomuso:
> > @bbatsov Can I be added to the cider-cheatsheet.el package authors?
> 
> Sure.

Thank you! I'll update the package commentary in a separate PR after this one is finished.

cider-cheatsheet.el
katomuso: Oh, I assumed you meant `cider-cheatsheet-select-path-separator`. If it is `cider-cheatsheet-indentation-function`, then some people prefer 2 spaces, others 4 spaces, tabs, or other indication of indentation. This use-case is what I had in mind when adding this as a customizable variable.

cider-cheatsheet.el
katomuso: If I understand correctly, you're suggesting renaming it to a more generic name because it's not guaranteed that by selecting a var, the user wants to show documentation. Did I understand that correctly?

bbatsov: Yeah, exactly. You can check the namespace navigator or apropos where you can trigger different actions in the buffer displaying the results (e.g. you can show the docs or go to the definition for some var), and there's some default action triggered by `RET`. I'm not sure if that makes much sense here, but it's another case where we are dealing with var names, so we might aim for some consistency.

cider-cheatsheet.el
bbatsov: That's one option, another would be to just drop the prefix argument and rely only on the defcustom.

cider-cheatsheet.el
bbatsov: Seems like premature optimization to me. I usually wait for the users to request some config options if I have doubts that there would be demand for them.

cider-cheatsheet.el
katomuso: I think it will be better to drop the prefix argument as most users will probably always use one way or the other.

cider-cheatsheet.el
katomuso: I agree, and it actually looks more generic, which is a good thing.

cider-cheatsheet.el
katomuso: I think that's a good approach, so for now, they will be removed.

katomuso:
I've made all the changes as discussed in the review:

- Renamed `cider-cheatsheet-doc-function` to `cider-cheatsheet-default-action-function` and tweaked docstring
- Renamed `cider-cheatsheet-select-path` to `cider-cheatsheet-select-type`, tweaked docstring and removed prefix argument for `cider-cheatsheet-select`
- Removed `cider-cheatsheet-select-path-separator` and `cider-cheatsheet-indentation-function`

cider-cheatsheet.el
```diff
@@ -57,6 +57,17 @@ from ClojureDocs or any other function accepting a var as an argument."
...
+(defcustom cider-cheatsheet-select-type 'multi
+  "Controls how cheatsheet is displayed when using `cider-cheatsheet-select`.
+
+If `multi`, then provide a multi-step selection process where you need to go
+section by section until you find a var.
+
+If `path`, then represent each completion candidate as a full path to a var."
+  :type '(choice (const multi)
+                 (const path))
+  :package-version '(cider . "0.15.0"))
```
bbatsov, line 69: 1.15.0 :-)

cider-cheatsheet.el
```diff
@@ -34,6 +34,40 @@
...
+
+By default, documentation for a var is displayed using `cider-doc-lookup`,
+but it can also be set to `cider-clojuredocs-lookup` to show documentation
+from ClojureDocs or any other function accepting a var as an argument."
+  :type '(choice (const cider-doc-lookup)
+                 (const cider-clojuredocs-lookup)
+                 function)
+  :package-version '(cider . "0.15.0"))
+
+(defcustom cider-cheatsheet-select-type 'multi
```
bbatsov, line 60: Let me think a bit about the names `multi` and `path`, because I'm not sure those are optimal here.

cider-cheatsheet.el
katomuso: My bad, fixed!

cider-cheatsheet.el
katomuso: Sure. I think so as well, but it was tough to come up with something better.

katomuso: Some alternatives I came up with are 'multi-step and 'full-path, or 'nested and 'flat.

katomuso: Well, come to think of it, maybe just having a prefix argument for `cider-cheatsheet-select` will be enough for now.

bbatsov: Fair enough. :-)

bbatsov:
Thanks!

PR #3692, opened 2024-05-31
[inspector] Add function and binding for toggling view mode

alexander-yakushev:
Following https://github.com/clojure-emacs/cider-nrepl/pull/879, this PR adds support for view mode toggling. Should not be merged before cider-nrepl release and subsequent version bump (pending).


alexander-yakushev:
Ready for review.

vemv (approved):
Nice!

doc/modules/ROOT/pages/debugging/inspector.adoc
```diff
@@ -76,6 +76,10 @@ You'll have access to additional keybindings in the inspector buffer
 | `cider-inspector-set-max-atom-length`
 | Set a new maximum length above which nested atoms (non-collections) are truncated
 
+| kbd:[v]
+| `cider-inspector-toggle-view-mode`
+| Switch the rendering of the current value between `:normal` and `:object` view mode.
```
vemv, line 81: If possible expand this section to increase discoverability

PR #3693, opened 2024-05-31
Support adding -Djdk.attach.allowAttachSelf to jack-in params to enable nrepl JVMTI agent

alexander-yakushev:
Also, add `cider-enable-nrepl-jvmti-agent` customizable to control this behaviour.

For now, the required Java property is injected for Leiningen and Clojure CLI jack ins. It probably should also do so for gradle and maven?

Additionally, I need advice on where to document this best, and if this can be tested somehow.

--- 

### DIFF
--- a/cider.el
+++ b/cider.el
@@ -173,7 +173,6 @@ then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
   :safe #'stringp
   :package-version '(cider . "1.1"))
 
-
 (defcustom cider-clojure-cli-global-aliases
   nil
   "Global aliases to include when jacking in with the clojure CLI.
@@ -185,7 +184,6 @@ then concatenated into the \"-M[your-aliases]:cider/nrepl\" form."
   :type 'string
   :safe #'stringp
   :package-version '(cider . "1.14"))
-
 
 (defcustom cider-shadow-cljs-command
   "npx shadow-cljs"
@@ -376,6 +374,14 @@ The repl dependendcies are most likely to be nREPL middlewares."
   :safe #'booleanp
   :version '(cider . "0.11.0"))
 
+(defcustom cider-enable-nrepl-jvmti-agent nil
+  "When t, add `-Djdk.attach.allowAttachSelf' to the command line arguments.
+This allows nREPL JVMTI agent to be loaded.  It is needed for evaluation
+interruption to properly work on Java 21 and above."
+  :type 'boolean
+  :safe #'booleanp
+  :version '(cider . "1.15.0"))
+
 (defcustom cider-offer-to-open-cljs-app-in-browser t
   "When nil, do not offer to open ClojureScript apps in a browser on connect."
   :type 'boolean
@@ -536,7 +542,7 @@ Throws an error if PROJECT-TYPE is unknown."
   "List of dependencies where elements are lists of artifact name and version.")
 (put 'cider-jack-in-dependencies 'risky-local-variable t)
 
-(defcustom cider-injected-nrepl-version "1.1.2"
+(defcustom cider-injected-nrepl-version "1.2.0-beta2"
   "The version of nREPL injected on jack-in.
 We inject the newest known version of nREPL just in case
 your version of Boot or Leiningen is bundling an older one."
@@ -762,6 +768,8 @@ group:artifact:version notation and MIDDLEWARES are
 prepared as arguments to Clojurephant's ClojureNRepl task."
   (concat global-opts
           (unless (seq-empty-p global-opts) " ")
+          (when cider-enable-nrepl-jvmti-agent
+            "-Pjdk.attach.allowAttachSelf ")
           (cider--gradle-jack-in-property (append (cider--jack-in-required-dependencies) dependencies))
           " "
           params
@@ -813,7 +821,9 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
                       (seq-map (lambda (middleware)
                                  (concat "update-in :middleware conj "
                                          middleware))
-                               lein-middlewares))
+                               lein-middlewares)
+                      (when cider-enable-nrepl-jvmti-agent
+                        `(,(concat "update-in :jvm-opts conj -Djdk.attach.allowAttachSelf"))))
               " -- ")
    " -- "
    (if (not cider-enrich-classpath)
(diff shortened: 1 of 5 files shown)


cider.el
```diff
@@ -813,7 +818,9 @@ removed, LEIN-PLUGINS, LEIN-MIDDLEWARES and finally PARAMS."
                       (seq-map (lambda (middleware)
                                  (concat "update-in :middleware conj "
                                          middleware))
-                               lein-middlewares))
+                               lein-middlewares)
+                      (when cider-enable-nrepl-jvmti-agent
+                        `(,(concat "update-in :jvm-opts conj " (shell-quote-argument "-Djdk.attach.allowAttachSelf")))))
```
vemv, line 823: Lein `update-in` tends to play poorly with everything else (plugins, profiles, etc) IME - I wish it worked as one would imagine, but quite commonly it doesn't.

We already use enrich-classpath for adding misc JVM flags. In that sense, that project slightly grew in scope from "replace the classpath with some entries attached at the tail" to also "add whatever flags are needed for cider-nrepl to work".

I'd suggest adding the flag there, and have fewer code paths here.

(It's also consistent with the current state of things - enrich-classpath is not a third-party thing but our only approach to have misc key features working)

bbatsov:
> For now, the required Java property is injected for Leiningen and Clojure CLI jack ins. It probably should also do so for gradle and maven?

Yeah, that would be ideal. This might also be a good time to purge the Boot stuff. 

> Additionally, I need advice on where to document this best, and if this can be tested somehow.

The documentation can go here I guess - https://docs.cider.mx/cider/usage/code_evaluation.html#configuration

I think there were a few tests checking that the params for Lein and friends were properly generated, so you can add something there. 

I'm not sure where's the best place to add this exactly, but your proposal seems reasonable.

cider.el
bbatsov: I don't think that's a good idea, as enrich-classpath is optional functionality and the interrupt functionality is kind of central to the eval workflows.  From my perspective it'd be best to keep them separate, although I agree we should be careful where/how exactly we set the `:jvm-opts`.

vemv: Well, it's optional but in the end it brings feature that every JVM Clojure dev wants (namely those that have lived in Orchard for many years but had no way of being used).

We're not too far away from being in a position of enabling it by default.

In specific terms, Enrich adds the following flags, smartly, only for further easing stuff in Orchard:

* `--add-opens=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED`
* `--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED`

As you can imagine, either asking users to add that by hand or adding it in cider are both less practical/maintainable.

vemv: With that said, we already use update-in in a lot of places, so I don't mind a lot.

bbatsov: I'll reiterate here that we should not make assumptions that people are using CIDER with `cider-nrepl` (even when it comes to `cider-jack-in`. That's why I like to keep whatever is at the nREPL core level separate from CIDER's additional functionality.

alexander-yakushev:
Added tests, updated docs. Ready for review.

doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -161,6 +161,12 @@ NOTE: CIDER internally increases the timeout to 30 seconds for the first sync ev
 
 == Configuration
 
+=== Enable evaluation interrupts on Java 21 and newer
+
+Special variable `cider-enable-nrepl-jvmti-agent` has to be enabled for
+interrupts to work properly on Java 21+. See
```
bbatsov, line 167: In the other section you've written Java 20+, so I'm guessing one of the two is not correct. :-)

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -204,6 +204,14 @@ You can further customize the command line CIDER uses for `cider-jack-in` by
 modifying the some options. Those differ a bit between the various tools,
 so we'll examine them tool by tool.
 
+==== Enabling nREPL JVMTI agent
+
+Since version 1.2.0, nREPL ships together with a native JVMTI agent, so that the
+eval interrupts properly work on Java 20 and later. To enable the agent, the
+Java process should be launched with `-Djdk.attach.allowAttachSelf`. CIDER will
+do it automatically during jack-in if you change
```
bbatsov, line 212: "... if/when `cider-enable-nrepl-jvmti-agent` is set to `t`." would read better IMO.

doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -204,6 +204,14 @@ You can further customize the command line CIDER uses for `cider-jack-in` by
 modifying the some options. Those differ a bit between the various tools,
 so we'll examine them tool by tool.
 
+==== Enabling nREPL JVMTI agent
+
+Since version 1.2.0, nREPL ships together with a native JVMTI agent, so that the
+eval interrupts properly work on Java 20 and later. To enable the agent, the
```
bbatsov, line 210: 20 or 21?

doc/modules/ROOT/pages/usage/code_evaluation.adoc
```diff
@@ -161,6 +161,12 @@ NOTE: CIDER internally increases the timeout to 30 seconds for the first sync ev
 
 == Configuration
 
+=== Enable evaluation interrupts on Java 21 and newer
+
+Special variable `cider-enable-nrepl-jvmti-agent` has to be enabled for
```
bbatsov, line 166: "Special -> The configuration variable" might read better here.

doc/modules/ROOT/pages/usage/code_evaluation.adoc
alexander-yakushev: Yeah, I know I sowed confusion with this one. I have to pick one. Technically, it's JDK20+, but since JDK20 is not supported by CIDER, and in a year everyone will forget what 20 is, perhaps it makes sense to write 21 as the first LTS version where this is needed? What do you think?

bbatsov: Let's go with Java 21.

alexander-yakushev: Done

cider.el
```diff
@@ -903,9 +914,10 @@ your aliases contain any mains, the cider/nrepl one will be the one used."
          (deps (format "{:deps {%s} :aliases {:cider/nrepl {:main-opts [%s]}}}"
                        (string-join all-deps " ") main-opts))
          (deps-quoted (cider--shell-quote-argument deps command)))
-    (format "%s-Sdeps %s -M%s:cider/nrepl%s"
+    (format "%s%s-Sdeps %s -M%s:cider/nrepl%s"
             ;; TODO: global-options are deprecated and should be removed in CIDER 2.0
             (if global-options (format "%s " global-options) "")
+            (if cider-enable-nrepl-jvmti-agent "-J-Djdk.attach.allowAttachSelf " "")
```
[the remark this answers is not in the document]
alexander-yakushev: No difference at all. They will not conflict either if set in both places.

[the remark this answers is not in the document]
alexander-yakushev: Do you mean this could be put into the alias instead of a top-level property? Yeah, we could probably do that. Can be changed in the future easily.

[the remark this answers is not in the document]
yuhan0: Yeah, and since we're now going the route of injecting jvm-opts to clojure-cli, it might be worth revisiting the decision back in #3023 and see if the `-XX:-OmitStackTraceInFastThrow` flag could also be added.

PR #3696, opened 2024-06-02
[complete] Prioritize flex style if flex style is requested

alexander-yakushev:
This has been briefly discussed here https://github.com/clojure-emacs/cider/pull/3659#discussion_r1602166071 before making `basic` the prioritized one.

The current setup produces a bug when Compliment returns a list of candidates that matches `flex` style, but only one of the items on that list matches `basic` style. In that case, Emacs will eagerly complete that single candidate without presenting a list.

Example:

`Unsa` instantly completes to `UnsatisfiedLinkError` (because that's a class that is always imported to all namespaces).

Desired behavior:

image

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -297,7 +297,7 @@ Only affects the `cider' completion category.`"
     (unless found-styles
       (setq found-styles '(styles basic)))
     (unless (member 'flex found-styles)
-      (setq found-styles (append found-styles '(flex))))
+      (setq found-styles (apply #'list 'styles 'flex (cdr found-styles))))
     (add-to-list 'completion-category-overrides (apply #'list 'cider found-styles (when found-cycle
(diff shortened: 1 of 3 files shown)


bbatsov (approved):
Looks good to me. Thanks!

cider-completion.el
```diff
@@ -297,7 +297,7 @@ Only affects the `cider' completion category.`"
     (unless found-styles
       (setq found-styles '(styles basic)))
     (unless (member 'flex found-styles)
-      (setq found-styles (append found-styles '(flex))))
+      (setq found-styles (apply #'list 'styles 'flex (cdr found-styles))))
```
bbatsov, line 300: Maybe you can add here a comment similar to the description of the ticket, so people would understand better the need for this bit of code. Feel free do directly commit the comment if you agree with me it'd be useful.

PR #3698, opened 2024-06-02
Fix another error regex bug, limit check to JVM runtime

### LINKED ISSUE #3687: Inline error overlays not displaying

yuhan0:
Fixes #3687

Not sure if the last commit is technically a breaking change.. the inconsistent var naming was bugging me, and these were introduced recently so it's unlikely that other packages depend on them. 

Side note: `eldev lint` produces lots of linter errors unrelated to the changes here, it seems like the CI isn't flagging them so they've been accumulating from other PRs.

-----------------

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -594,7 +594,8 @@ It delegates the actual error content to the eval or op handler."
     ":" (group-n 3 (one-or-more (any "-" digit)))
     (optional
      ":" (group-n 4 (one-or-more (any "-" digit))))
-    " - "))
+    " - ")
+  "Regexp matching various non-error messages, e.g. reflection warnings.")
 
 ;; Please keep this in sync with `cider-clojure-compilation-error-regexp',
 ;; which is a subset of these regexes.
@@ -605,8 +606,6 @@ It delegates the actual error content to the eval or op handler."
    'nogroup)
   "A few example values that will match:
 \"Reflection warning, /tmp/foo/src/foo/core.clj:14:1 - \"
-\"CompilerException java.lang.RuntimeException: Unable to resolve symbol: \\
-lol in this context, compiling:(/foo/core.clj:10:1)\"
 \"Syntax error compiling at (src/workspace_service.clj:227:3).\"
 \"Unexpected error (ClassCastException) macroexpanding defmulti at (src/haystack/parser.cljc:21:1).\"")
 
@@ -618,8 +617,6 @@ lol in this context, compiling:(/foo/core.clj:10:1)\"
 but excluding warnings such as reflection warnings.
 
 A few example values that will match:
-\"CompilerException java.lang.RuntimeException: Unable to resolve symbol: \\
-lol in this context, compiling:(/foo/core.clj:10:1)\"
 \"Syntax error compiling at (src/workspace_service.clj:227:3).\"
 \"Unexpected error (ClassCastException) macroexpanding defmulti at (src/haystack/parser.cljc:21:1).\"")
 
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -925,8 +925,11 @@ depending on the PHASE."
               (and cider-show-error-buffer
                    (member phase (cider-clojure-compilation-error-phases))))
              ;; Only show overlays for things that do look like an exception (#3587):
-             (or (string-match-p cider-clojure-runtime-error-regexp err)
-                 (string-match-p cider-clojure-compilation-error-regexp err)))
+             ;; Note: only applicable to JVM Clojure error messages (#3687)
+             (if (cider-runtime-clojure-p)
+                 (or (string-match-p cider-clojure-runtime-error-regexp err)
+                     (string-match-p cider-clojure-compilation-error-regexp err))
+               t))
     ;; Display errors as temporary overlays
     (let ((cider-result-use-clojure-font-lock nil)
           (trimmed-err (funcall cider-inline-error-message-function err)))
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -564,7 +564,7 @@ It delegates the actual error content to the eval or op handler."
 ;; Reference:
 ;; https://github.com/clojure/clojure/blob/clojure-1.10.0/src/clj/clojure/main.clj#L251
 ;; See `cider-compilation-regexp' for interpretation of match groups.
-(defconst cider-clojure-1.10--location
+(defconst cider--clojure-1.10-location
   '(sequence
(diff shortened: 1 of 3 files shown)


cider-eval.el
```diff
@@ -925,7 +915,9 @@ depending on the PHASE."
               (and cider-show-error-buffer
                    (member phase (cider-clojure-compilation-error-phases))))
              ;; Only show overlays for things that do look like an exception (#3587):
-             (or (string-match-p cider-clojure-runtime-error-regexp err)
+             ;; Note: only applicable to JVM Clojure error messages (#3687)
+             (or (not (cider-runtime-clojure-p))
```
bbatsov, line 919: I think it'd be better to move the runtime check outside the `or`, as that would read better.

bbatsov:
> the inconsistent var naming was bugging me, and these were introduced recently so it's unlikely that other packages depend on them.

Yeah, that's fine. They were meant to be private, anyways, and it's extremely unlikely that someone decided to do something with them directly. The PR looks good at a glance, but I'll take a closer look tomorrow.

cider-eval.el
yuhan0: Something like this?
```lisp
(if (cider-runtime-clojure-p)
    (or (string-match-p cider-clojure-runtime-error-regexp err) 
        (string-match-p cider-clojure-compilation-error-regexp err))
  t)
```

bbatsov: Yep, or some `cond` if you plan to add more runtime-specific parsing.

yuhan0: I don't plan to myself, the`if` reads better for now.

yuhan0:
For reference, I based the regexes directly off the format strings here:
https://github.com/clojure/clojure/blob/clojure-1.10.0/src/clj/clojure/main.clj#L264

And made the judgement that errors thrown in `:print-eval-result`and  `:read-eval-result`  phases (not covered previously) would be considered "runtime errors" by Cider.

vemv (commented):
LGTM in general, however:

* I take it that you verified that the `Refactor error-matching regexes` is covered?
* `(cider-runtime-clojure-p)` seems to consider clj and cljs as one

yuhan0:
> I take it that you verified that the Refactor error-matching regexes is covered?

Yup, although I didn't add a test for the `:read-eval-result` phase, (not even sure how one would trigger the error)

> (cider-runtime-clojure-p) seems to consider clj and cljs as one

Aren't Clojuresicrpt repls a layer running inside a Clojure process? (throwing JVM errors upon macroexpansion etc.) 
I'm not a regular cljs user, not sure what's the appropriate thing to do there

vemv:
> Yup, although I didn't add a test for the :read-eval-result phase, (not even sure how one would trigger the error)

That would seem very much valuable -sounds like a question that you could get answered on #clojure 

> Aren't Clojuresicrpt repls a layer running inside a Clojure process? (throwing JVM errors upon macroexpansion etc.)

Easier and safer to check. I think it is queriable anyway whether the connection is a cljs one (in the same way that you can query the port out of a repl object).

If you need a modern project with cljs, I have https://github.com/reducecombine/icd.scroll at hand - it's my generic repro project with a very clear README.

yuhan0:
Just a heads up but I won't be able to devote much time over the next week on this, probably not to the extent of setting up and debugging an unfamiliar cljs env.

As far as I'm concerned this PR is a regression fix for 3616, I could separate out the more subjective commits if it helps with getting things merged.

bbatsov:
I'm OK with the code in the current state and in general and given that "perfect" is the enemy of "done", I'll just go ahead and merge it. Still, it'd be nice to follow-up on this in the future and refine the error message handling for the various runtime environments.

PR #3700, opened 2024-06-03
JVMTI followup

alexander-yakushev:
1. Per @yuhan0 suggestion, I've changed the way how allowAttachSelf property is injected for tools.deps. It looks marginally cleaner this way.
2. Updated the docs clarifying that allowAttachSelf can (and should) be specified manually when the REPL is not started through jack-in.

### DIFF
--- a/doc/modules/ROOT/pages/basics/up_and_running.adoc
+++ b/doc/modules/ROOT/pages/basics/up_and_running.adoc
@@ -206,11 +206,33 @@ so we'll examine them tool by tool.
 
 ==== Enabling nREPL JVMTI agent
 
-Since version 1.2.0, nREPL ships together with a native JVMTI agent, so that the
-eval interrupts properly work on Java 21 and later. To enable the agent, the
-Java process should be launched with `-Djdk.attach.allowAttachSelf`. CIDER will
-do it automatically during jack-in if `cider-enable-nrepl-jvmti-agent` is set to
-`t`.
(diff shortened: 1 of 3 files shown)


doc/modules/ROOT/pages/basics/up_and_running.adoc
```diff
@@ -206,11 +206,15 @@ so we'll examine them tool by tool.
...
-`t`.
+Since version 1.2.0, nREPL includes a native JVMTI agent which makes the eval
+interrupts properly work on Java 21 and later. To enable the agent, the Java
+process should be launched with `-Djdk.attach.allowAttachSelf`. CIDER will do it
+automatically during jack-in if `cider-enable-nrepl-jvmti-agent` is set to `t`.
+
+IMPORTANT: `cider-enable-nrepl-jvmti-agent` has no effect if you start a REPL
+process outside of Emacs and connect to it with `cider-connect`. In that
+scenario, you have to add `-Djdk.attach.allowAttachSelf` Java property manually
+through the means of the build tool.
```
bbatsov, line 217: Maybe you can also add here some snippets for `lein` and the Clojure CLI, so it'd be a tiny bit easier for the majority of the users to figure out what they have to do.

alexander-yakushev: Done.

bbatsov:
Thanks!

PR #3701, opened 2024-06-03
ClojureDocs formatting tweaks

katomuso:
I plan to rework how ClojureDocs buffer is displayed (improve some formatting, make it more modular and customizable) and want to start with some small improvements.

This particular change:
- Removes annoying whitespace that appears before every line (after the first one) of the docstring.
- Changes how buttons are displayed so that `*` and newline are no longer part of the button.

This is how it looked before:
cider-cheatsheet-formatting-tweaks-old

This is how it looks now:
cider-cheatsheet-formatting-tweaks-new

### DIFF
--- a/cider-clojuredocs.el
+++ b/cider-clojuredocs.el
@@ -116,9 +116,12 @@ opposite of what that option dictates."
     (let ((arglists (nrepl-dict-get dict "arglists")))
       (dolist (arglist arglists)
         (insert (format " [%s]\n" arglist)))
-      (insert "\n")
-      (insert (nrepl-dict-get dict "doc"))
-      (insert "\n"))
+      (when-let* ((doc (nrepl-dict-get dict "doc"))
+                  ;; As this is a literal docstring from the source code and
+                  ;; there are two spaces at the beginning of lines in docstrings,
+                  ;; we remove them to make it align nicely in ClojureDocs buffer.
+                  (doc (replace-regexp-in-string "\n  " "\n" doc)))
+        (insert "\n" doc "\n")))
     (insert "\n== See Also\n\n")
     (if-let ((see-alsos (nrepl-dict-get dict "see-alsos")))
         (dolist (see-also see-alsos)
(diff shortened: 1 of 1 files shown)


cider-clojuredocs.el
```diff
@@ -116,17 +116,23 @@ opposite of what that option dictates."
     (let ((arglists (nrepl-dict-get dict "arglists")))
       (dolist (arglist arglists)
         (insert (format " [%s]\n" arglist)))
-      (insert "\n")
-      (insert (nrepl-dict-get dict "doc"))
+      (when-let ((doc (nrepl-dict-get dict "doc")))
+        (insert "\n")
+        (let ((lines (mapcar (lambda (line)
+                               (replace-regexp-in-string "^\s+" "" line))
```
bbatsov, line 122: Why not use `string-trim-left`? 

One thing to consider here, though, is how much to strip exactly, as I'm guessing that some docstrings might have intentional indentation on some lines.

katomuso: You are right, `string-trim-left` will suffice in that case.

I'm not sure why we get these spaces in the docstring in the first place. Maybe it has something to do with ClojureDocs data.

But we can be sure that every line after the first one contains two additional spaces at the beginning, so maybe just removing exactly two spaces will be good enough.

katomuso: Well, they are here because this is a literal string from the source code. As there are generally two spaces at the beginning of lines in docstrings, I guess it will be okay.

bbatsov: Okay. I'm also not sure where those spaces are coming from, but I guess your solution is unlikely to break something in the formatting.

cider-clojuredocs.el
```diff
@@ -116,17 +116,18 @@ opposite of what that option dictates."
     (let ((arglists (nrepl-dict-get dict "arglists")))
       (dolist (arglist arglists)
         (insert (format " [%s]\n" arglist)))
-      (insert "\n")
-      (insert (nrepl-dict-get dict "doc"))
-      (insert "\n"))
+      (when-let ((doc (nrepl-dict-get dict "doc")))
+        (insert "\n" (replace-regexp-in-string "\n  " "\n" doc) "\n")))
```
bbatsov, line 120: I'd add a note here regarding the need to do this.

katomuso: I've added a note.

PR #3704, opened 2024-06-05
Refactor code used to insert sections into ClojureDocs buffer

katomuso:
I've moved the code used to insert sections into separate functions, as it is easier to modify sections that way. Maybe, in the future, I will also create a generic function to insert sections, so we can reduce boilerplate.

In places where we need to insert an empty line, I've used `newline` explicitly instead of the previously used additional `\n` in content insertions. This way, it will be more obvious where we are inserting actual content and where we are inserting just empty lines.


bbatsov:
Thanks!

PR #3706, opened 2024-06-07
Minor adjustments to docstrings

katomuso:
Tweak docstring of `cider-sync-request:clojuredocs-lookup` as it actually performs `clojuredocs-lookup` op and not `resource` op.

Tweak docstrings of `cider-var-info` and `cider-member-info` as it seems that what these functions actually return is an nREPL dict and not an alist. Also, I'm not sure what "with list cdrs" means.

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -548,7 +548,7 @@ Used only when the info nREPL middleware is not available."
     var-info))
 
 (defun cider-var-info (var &optional all)
-  "Return VAR's info as an alist with list cdrs.
+  "Return info for VAR as an nREPL dict.
 When multiple matching vars are returned you'll be prompted to select one,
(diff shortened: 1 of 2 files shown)


cider-client.el
```diff
@@ -559,7 +559,7 @@ unless ALL is truthy."
       (if all var-info (cider--var-choice var-info)))))
 
 (defun cider-member-info (class member)
-  "Return the CLASS MEMBER's info as an alist with list cdrs."
+  "Return info on MEMBER of CLASS as an nREPL dict."
```
bbatsov, line 562: on -> for

katomuso:
@bbatsov I've applied the suggested changes.

bbatsov:
Thanks!

PR #3708, opened 2024-06-08
Split `cider-docstring--dumb-trim` into two single-purpose functions

katomuso:
The function `cider-docstring--dumb-trim` performs both formatting and trimming. I propose splitting it into two separate functions: `cider-docstring--trim` and `cider-docstring--format` because these two functionalities are needed in different contexts. For example, we need to format the docstring when displaying it in the documentation buffer, and we need to trim the docstring when displaying it in the minibuffer or in a popup along with completions.

### DIFF
--- a/cider-docstring.el
+++ b/cider-docstring.el
@@ -27,6 +27,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'seq)
 (require 'shr)
 (require 'subr-x)
 
@@ -140,20 +141,14 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
         second-attempt
         first-attempt)))
 
-(defun cider-docstring--trim (s &optional n)
-  "Returns up to the first N lines of string S,
+(cl-defun cider-docstring--trim (string &optional (max-lines cider-docstring-max-lines))
+  "Returns up to the first MAX-LINES lines of string STRING,
 adding \"...\" if trimming was necessary.
 
-N defaults to `cider-docstring-max-lines'."
-  (when s
-    (let* ((n (or n cider-docstring-max-lines))
-           (lines (split-string s "\n"))
-           (lines-length (length lines))
-           (selected-lines (cl-subseq lines 0 (min n lines-length)))
-           (result (string-join selected-lines "\n")))
-      (if (> lines-length n)
-          (concat result "...")
-        result))))
+MAX-LINES defaults to `cider-docstring-max-lines'."
+  (let* ((lines (split-string string "\n"))
+         (string (string-join (seq-take lines max-lines) "\n")))
+    (concat string (when (> (length lines) max-lines) "..."))))
 
 (defun cider-docstring--format (s)
   "Performs formatting of S, cleaning up some common whitespace issues."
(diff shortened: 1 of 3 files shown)


katomuso:
To easily verify that it doesn't change the output, you can use the following snippet:
```elisp
(let* ((info (cider-var-info "clojure.core/reduce"))
       (doc (nrepl-dict-get info "doc")))
  (string= (cider-docstring--dumb-trim doc)
           (cider-docstring--trim (cider-docstring--format doc))))
```

cider-docstring.el
```diff
@@ -139,27 +141,21 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
...
-                                (replace-regexp-in-string "\\`[ ]+" "" f))
-                              selected-lines
-                              "\n")))
-      (if (> lines-length n)
-          (concat result "...")
-        result))))
+(cl-defun cider-docstring--trim (string &optional (max-lines cider-docstring-max-lines))
+  "Return MAX-LINES of STRING, adding \"...\" if trimming was necessary."
+  (let* ((lines (split-string string "\n"))
+         (string (string-join (seq-take lines max-lines) "\n")))
```
vemv, line 147: A nil value for the second argument of `seq-take` appears to be undocumented, so I'd prefer to have an explicit default for it

cider-docstring.el
```diff
@@ -139,27 +141,21 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
...
-           (selected-lines (cl-subseq lines 0 (min n lines-length)))
-           (result (mapconcat (lambda (f)
-                                ;; Remove spaces at the beginning of each line, as it is common in many clojure.core defns:
-                                (replace-regexp-in-string "\\`[ ]+" "" f))
-                              selected-lines
-                              "\n")))
-      (if (> lines-length n)
-          (concat result "...")
-        result))))
+(cl-defun cider-docstring--trim (string &optional (max-lines cider-docstring-max-lines))
```
vemv, line 144: cider-docstring-max-lines is unused

vemv (commented):
LGTM!

I appreciate starting over more cleanly.

cider-docstring.el
katomuso: The value of `cider-docstring-max-lines` is the default for `max-lines` if `max-lines` is not specified explicitly, so there is no need to use it in the body. So, if we call `cider-docstring--trim` without a second argument, `cider-docstring-max-lines` is used, right? So I'm not sure what you mean by saying that `cider-docstring-max-lines` is unused.

cider-docstring.el
katomuso: You mean if the `max-lines` argument is `nil`, right? Well, I don't really think there should be a default value because it should always be a number, but if anything, I think it should just return the input `string` if `max-lines` is `nil`.

katomuso: I mean, why assume that `nil` can be passed? In which cases can it be useful?

cider-docstring.el
vemv: Ah ok! I wasn't immediately familar with that cl-defun syntax.

vemv:
Thanks!

PR #3712, opened 2024-06-09
Change the way `cider-docstring--format` performs formatting

### LINKED ISSUE #3709: `cider-docstring--format` breaks some formatting
Currently, the `cider-docstring--format` function used for formatting docstrings breaks some formatting by moving every single paragraph (indicated by `.  `) to a new line and removing a variable number of spaces from the beginning of each line. Let me demonstrate how it looks.

Snippet to get original docstring:

```elisp
(thread-first
  (cider-var-info "clojure.core/reduce")
  (nrepl-dict-get "doc"))
```

Original docstring:

```
f should be a function of 2 arguments. If val is not supplied,
  returns the result of applying f to the first 2 items in coll, then
  applying f to that result and the 3rd item, etc. If coll contains no
  items, f must accept no arguments as well, and reduce returns the
  result of calling f with no arguments.  If coll has only 1 item, it
  is returned and f is not called.  If val is supplied, returns the
  result of applying f to val and the first item in coll, then
  applying f to that result and the 2nd item, etc. If coll contains no
  items, returns val and f is not called.
```

Snippet to get formatted docstring:

```elisp
(thread-first
  (cider-var-info "clojure.core/reduce")
  (nrepl-dict-get "doc")
  cider-docstring--format)
```

Formatted docstring:

```
f should be a function of 2 arguments. If val is not supplied,
returns the result of applying f to the first 2 items in coll, then
applying f to that result and the 3rd item, etc. If coll contains no
items, f must accept no arguments as well, and reduce returns the
result of calling f with no arguments.

If coll has only 1 item, it
is returned and f is not called.

If val is supplied, returns the
result of applying f to val and the first item in coll, then
applying f to that result and the 2nd item, etc. If coll contains no
items, returns val and f is not called.
```

It seems that nobody has noticed this strange formatting because `cider-docstring--format` is currently used only for docstrings in the minibuffer and in completions and not many people look at the docstrings there (though it will also be beneficial to use it for docstrings in documentation buffers).

As discussed in #3701, removing two spaces from the beginning while formatting the docstring will be good enough.

katomuso:
Fix #3709

### DIFF
--- a/cider-docstring.el
+++ b/cider-docstring.el
@@ -149,13 +149,10 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
 
 (defun cider-docstring--format (string)
   "Return a nicely formatted STRING to be displayed to the user."
-  (let* ((string (replace-regexp-in-string "\\.  " ".\n\n" string)) ;; improve the formatting of e.g. clojure.core/reduce
-         (string (mapconcat (lambda (line)
-                              ;; Remove spaces at the beginning of each line, as it is common in many clojure.core defns:
-                              (replace-regexp-in-string "\\`[ ]+" "" line))
-                            (split-string string "\n")
-                            "\n")))
-    string))
+  ;; As this is a literal docstring from the source code and
+  ;; there are two spaces at the beginning of lines in docstrings,
+  ;; we remove them to make it align nicely when it is displayed.
+  (replace-regexp-in-string "\n  " "\n" string))
 
 (provide 'cider-docstring)
 ;;; cider-docstring.el ends here
--- a/cider-docstring.el
+++ b/cider-docstring.el
@@ -148,10 +148,14 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
     (concat string (when (> (length lines) max-lines) "..."))))
 
 (defun cider-docstring--format (string)
-  "Return a nicely formatted STRING to be displayed to the user."
-  ;; As this is a literal docstring from the source code and
-  ;; there are two spaces at the beginning of lines in docstrings,
-  ;; we remove them to make it align nicely when it is displayed.
+  "Return a nicely formatted STRING to be displayed to the user.
+
+We need to format the docstring before displaying it to the user
+because it is obtained from the source code.  For example, this means
+that it usually has two spaces before each line used for indentation
+\(see https://guide.clojure.style/#docstring-indentation).  While displaying
+the docstring to the user, we usually want to control indentation and
+other aspects of the presentation, so we format it before displaying."
   (replace-regexp-in-string "\n  " "\n" string))
 
 (provide 'cider-docstring)
--- a/cider-clojuredocs.el
+++ b/cider-clojuredocs.el
@@ -25,15 +25,15 @@
 
 ;;; Code:
 
+(require 'subr-x)
+(require 'url-vars)
+
 (require 'cider-client)
 (require 'cider-common)
-(require 'subr-x)
+(require 'cider-docstring)
 (require 'cider-popup)
 (require 'cider-util)
-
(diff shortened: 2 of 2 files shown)


cider-docstring.el
```diff
@@ -149,13 +149,10 @@ Prioritize rendering as much as possible while staying within `cider-docstring-m
...
 (defun cider-docstring--format (string)
   "Return a nicely formatted STRING to be displayed to the user."
-  (let* ((string (replace-regexp-in-string "\\.  " ".\n\n" string)) ;; improve the formatting of e.g. clojure.core/reduce
-         (string (mapconcat (lambda (line)
-                              ;; Remove spaces at the beginning of each line, as it is common in many clojure.core defns:
-                              (replace-regexp-in-string "\\`[ ]+" "" line))
-                            (split-string string "\n")
-                            "\n")))
-    string))
+  ;; As this is a literal docstring from the source code and
```
bbatsov, line 152: You can probably mention in the docstring that this private function just trims leading whitespace or even remove it completely, as I now I don't see much need for it. 

This change is user-facing, so it also needs a changelog entry. Might be a good idea to add some unit tests verifying how docstrings get displayed.

katomuso: We already use this function in a couple of places, and I think we should use it to format the docstring in the Doc buffer as well (so we can control indentation specifically in code for formatting Doc buffer because currently without proper formatting of the docstring, we have accidental indentation), so I strongly believe we should leave this function be.

katomuso: Regarding the docstring of the function, it will be unclear why we trim two spaces if we omit the specific detail that this is because the string is a literal string from the source code.

If you want, I can add the whole comment to the docstring, though there can be more rules, and it may not be that clear for which formatting rule this comment corresponds to.

bbatsov: Fair enough, just keep it as is, then. 

I find the terminology "literal docstring from the source code" to be a bit confusing (perhaps something like "raw Clojure docstring with its original formatting" would be better), but overall I don't have a great alternative proposal either. My remark for the description of the Elisp function was prompted mostly because it might be obvious to people why we need to reformat the Clojure docstrings at all.

katomuso: I see, that's a valid remark. Indeed, we should add some reasons to the docstring as to why we do formatting at all.

katomuso: I will also think about some unit tests.

katomuso: I've updated the docstring and also removed the comment as it felt redundant. I've also found a section in the community style guide related to docstring indentation (https://guide.clojure.style/#docstring-indentation). It might be worthwhile to include it in the docstring of the function.

Regarding unit tests, they don't make much sense for this one function, but they do make sense for the whole `cider-docstring.el` package. However, it first needs to be refactored, as it is a 10-year-old package and hard to make sense of.

Also, I was thinking about adding a changelog entry after another PR, in which we start using `cider-docstring--format` in the Doc buffer. Or should it be two separate entries?

bbatsov: > I've updated the docstring and also removed the comment as it felt redundant. I've also found a section in the community style guide related to docstring indentation (https://guide.clojure.style/#docstring-indentation). It might be worthwhile to include it in the docstring of the function.

Good idea. 

> Regarding unit tests, they don't make much sense for this one function, but they do make sense for the whole cider-docstring.el package. However, it first needs to be refactored, as it is a 10-year-old package and hard to make sense of.

Fine by me.

> Also, I was thinking about adding a changelog entry after another PR, in which we start using cider-docstring--format in the Doc buffer. Or should it be two separate entries?

Sure, you can add this later.

katomuso: Okay, I've updated the docstring. Can we merge now?

PR #3713, opened 2024-06-11
Improve performance of nrepl-dict-get

yuhan0:
See #3711 

These changes do not affect the semantics of nrepl-dict-get (covered by existing tests), but byte compilation will raise a warning on any 3-argument usage, hopefully decreasing the risk of its future removal.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/nrepl-dict.el
+++ b/nrepl-dict.el
@@ -28,7 +28,8 @@
 ;;; Commentary:
 ;;
 ;; Provides functions to interact with and create `nrepl-dict's.  These are
-;; simply plists with an extra element at the head.
+;; simply plists with an extra element at the head, and using `equal' for
+;; comparison of string keys.
 
 ;;; Code:
 (require 'cl-lib)
@@ -199,7 +200,7 @@ If NO-JOIN is given, return the first non nil dict."
           (t `(,dict1 ,dict2)))))
 
 
-;;; Dbind
+;;; Destructuring-bind of string keys
 (defmacro nrepl-dbind-response (response keys &rest body)
   "Destructure an nREPL RESPONSE dict.
 Bind the value of the provided KEYS and execute BODY."
--- a/nrepl-dict.el
+++ b/nrepl-dict.el
@@ -45,12 +45,11 @@
     (maphash (lambda (k v) (nrepl-dict-put dict k v)) hash)
     dict))
 
-(defun nrepl-dict-p (object)
+(defsubst nrepl-dict-p (object)
   "Return t if OBJECT is an nREPL dict."
-  (and (listp object)
-       (eq (car object) 'dict)))
+  (eq (car-safe object) 'dict))
 
-(defun nrepl-dict-empty-p (dict)
+(defsubst nrepl-dict-empty-p (dict)
   "Return t if nREPL dict DICT is empty."
   (null (cdr dict)))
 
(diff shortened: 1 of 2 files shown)


nrepl-dict.el
```diff
@@ -66,9 +66,10 @@ If dict is nil, return nil.  If DEFAULT not provided, and KEY not in DICT,
 return nil.  If DICT is not an nREPL dict object, an error is thrown."
   (when dict
     (if (nrepl-dict-p dict)
-        (if (nrepl-dict-contains dict key)
-            (lax-plist-get (cdr dict) key)
-          default)
+        (or (lax-plist-get (cdr dict) key)
```
bbatsov, line 74: I'd probably add some comments here, so someone won't "optimize" away the optimizations in the future.

bbatsov:
Btw, how about removing such checks:

```
    (if (nrepl-dict-p dict)...
```

They obviously have some overhead and seem pointless in practice as we do expect a dict to be pass. I've never been found of so defensive style of programming. (although I do get you're getting nicer errors when you mess up something this way)

bbatsov:
Ah, I see you've inlined `nrepl-dict-p`, so I'm guessing it's not as big of a problem now.

yuhan0:
Yeah, I didn't specifically profile the effect of `nrepl-dict-p` but I'd guess that the overhead from it is negligible anyway.

One other thought I had was whether we could overload all of these functions to accept a hash-table as argument, that should be a non-breaking change and allow us to swap out the implementations of those huge dicts under the hood, similar to how Clojure treats PersistentArrayMap  / PersistentHashMap?

Something like

```clj
(defun nrepl-dict-get (dict key)
  (cond ((null dict) nil)
        ((nrepl-dict-p dict)
         <<current-implementation>>)
        ((hash-table-p dict)
         (gethash key dict))
        (t
         (error "Not an nREPL dict object: %s" dict))))
```

nrepl-dict.el
yuhan0: Added, and rebased to resolve the conflicts

bbatsov:
> One other thought I had was whether we could overload all of these functions to accept a hash-table as argument, that should be a non-breaking change and allow us to swap out the implementations of those huge dicts under the hood, similar to how Clojure treats PersistentArrayMap / PersistentHashMap?

I think that's a pretty reasonable idea, as I was thinking something quite similar over the weekend when we started to discuss this. It'd be an easy and fairly transparent change.

katomuso:
> One other thought I had was whether we could overload all of these functions to accept a hash-table as argument, that should be a non-breaking change and allow us to swap out the implementations of those huge dicts under the hood, similar to how Clojure treats PersistentArrayMap / PersistentHashMap?

Sounds really interesting. It will be nice to see some examples of where we can actually use hash tables instead of dicts.

bbatsov:
@katomuso Well, we can just pick some size after which dicts become hashes (e.g. 10 elements) and this would be done transparently by nrepl-dict constructors. If you check the related discussion you'll see that @yuhan0 found some pretty big dicts in some instances (that are frequently consulted).

katomuso:
Yeah, I've checked the discussion. I guess it is just hard to wrap my head around how exactly this should work. Will be looking for further updates on the topic!

PR #3714, opened 2024-06-11
Show progress when evaluating files using `cider-load-all-files`

katomuso:
Show progress in echo area while recursively evaluating files in directory using `cider-load-all-files`.

As I've noticed that messages from evaluations keep popping up, so it's very hard to see progress message, I've used `inhibit-message` around `cider-load-file`. One possible downside of this is that error messages are not displayed as well (though it will be really hard to notice some error message while messages pop up one after the other).

Also, I've tweaked the prompt of `cider-load-all-files` because it might not be clear how files are loaded as the word "beneath" is ambiguous (it may mean one level of depth which is not the case).

Closes #3615

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1839,7 +1839,7 @@ all ns aliases and var mappings from the namespace before reloading it."
 Useful when the running nREPL on remote host.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
 all ns aliases and var mappings from the namespaces being reloaded"
-  (interactive "DLoad files beneath directory: \nP")
+  (interactive "DRecursively load files in directory: \nP")
   (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
          (reporter (make-progress-reporter "Loading files" 0 (length files))))
     (seq-do-indexed (lambda (file idx)
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1840,8 +1840,14 @@ Useful when the running nREPL on remote host.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
 all ns aliases and var mappings from the namespaces being reloaded"
   (interactive "DLoad files beneath directory: \nP")
-  (mapcar (lambda (file) (cider-load-file file undef-all))
-          (directory-files-recursively directory "\\.clj[cs]?$")))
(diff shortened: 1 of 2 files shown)


bbatsov:
> As I've noticed that messages from evaluations keep popping up, so it's very hard to see progress message, I've used inhibit-message around cider-load-file. One possible downside of this is that error messages are not displayed as well (though it will be really hard to notice some error message while messages pop up one after the other).

Yeah, that's not a big problem - after all most people will get some error buffers anyways. 

Just add a changelog entry and we're good to go here.

cider-eval.el
```diff
@@ -1839,9 +1839,15 @@ all ns aliases and var mappings from the namespace before reloading it."
 Useful when the running nREPL on remote host.
 When UNDEF-ALL is non-nil or called with \\[universal-argument], removes
 all ns aliases and var mappings from the namespaces being reloaded"
-  (interactive "DLoad files beneath directory: \nP")
-  (mapcar (lambda (file) (cider-load-file file undef-all))
-          (directory-files-recursively directory "\\.clj[cs]?$")))
+  (interactive "DRecursively load files in directory: \nP")
+  (let* ((files (directory-files-recursively directory "\\.clj[cs]?$"))
+         (reporter (make-progress-reporter "Loading files" 0 (length files))))
```
bbatsov, line 1844: Perhaps we should mention the directory name in the message as well.

cider-eval.el
```diff
@@ -1839,9 +1839,15 @@ all ns aliases and var mappings from the namespace before reloading it."
...
+    (seq-do-indexed (lambda (file idx)
```
bbatsov, line 1845: I would have probably used a `dolist` here, but I'm old-school. :D

katomuso: Well, it's a chore to manually manipulate the counter, and I couldn't pass up such a sweet function!

cider-eval.el
katomuso: Good idea, will do.

katomuso: Actually, in every reporter message, there is already a name for the loaded file (the `file` argument of `(progress-reporter-update reporter (1+ idx) file)`), so I think there is no need to change anything.

katomuso:
I've updated the changelog.

bbatsov:
Thanks!

PR #3719, opened 2024-06-29
clean duplicate define-key at cider-sesman-browser-map

conao3:
This keybind first introduced in [the commit](https://github.com/clojure-emacs/cider/commit/6a9bc1cc05a4e393b31c5dfb79c70179fb3233f4).  There are two lines of key bindings to `C-c C-q`, probably due to a simple mistake.

-----------------

- [x]All tests are passing (`eldev test`)

Thanks!

manual][1] extremely useful.*


conao3:
Or, just remove the line

bbatsov:
I think it's better to just remove the duplication.

conao3:
Thanks, @bbatsov!
I just remove duplicate one.

bbatsov:
Thanks!

PR #3726, opened 2024-07-20
[Fix #3725] Use defvaralias for x-gtk-use-system-tooltips on emacs29

### LINKED ISSUE #3725: Defining a now-deprecated variable `x-gtk-use-system-tooltips`
Hi,

In Emacs 29, the `x-gtk-use-system-tooltips` was deprecated in favor of `use-system-tooltips`. See: https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.29#n881

In cider, `x-gtk-use-system-tooltips` is defined as a variable:

https://github.com/clojure-emacs/cider/blob/c698e1454cb416ad55c739961f9df514da9556f0/cider-mode.el#L1065

This breaks compatibility with the [PDF Tools](https://github.com/vedang/pdf-tools/) package. That is, if I load cider before loading PDF Tools, then it fails with the error:

```
File mode specification error: (error Don’t know how to make a buffer-local variable an alias: x-gtk-use-system-tooltips)
```

To me it seems that they have addressed the issue with the renaming of `use-system-tooltips` here:

https://github.com/vedang/pdf-tools/commit/f9ab15ee19f42f742a88b979a0fe6d097449645a

but that this conflicts with cider explicitly defining the variable.

My guess would be that cider could fix this in the same way that PDF Tools does it?

larstvei:
Hi,

This should fix the issue with defining the now deprecated variable`x-gtk-use-system-tooltips` as a variable, as discussed here: #3725.

### DIFF
--- a/cider-mode.el
+++ b/cider-mode.el
@@ -1062,7 +1062,12 @@ property."
 
 
 ;;; Minor-mode definition
-(defvar x-gtk-use-system-tooltips)
+
+(if (and (> emacs-major-version 28)
+         (not (boundp 'x-gtk-use-system-tooltips)))
+    ;; The x-gtk prefix has been dropped in Emacs 29
+    (defvaralias 'x-gtk-use-system-tooltips 'use-system-tooltips)
+  (defvar x-gtk-use-system-tooltips))
 
 ;;;###autoload
 (define-minor-mode cider-mode


bbatsov:
Can you push again here? Seems you ran into some issues with the CI.

larstvei:
Sure thing! I found a missing word in the comment to fix 👍

bbatsov:
Thanks!

PR #3738, opened 2024-09-03
Bump cider-nrepl to 0.50.2

alexander-yakushev:


alexander-yakushev:
Going ahead with this.

PR #3739, opened 2024-09-06
Correctly quote allowAttachSelf argument for Leiningen

alexander-yakushev:
Context: https://clojurians.slack.com/archives/C0617A8PQ/p1725626644003359


bbatsov:
That was fast! Looks good to me.

alexander-yakushev:
Big thanks to @jpmonettas for both the discovery and the suggested fix!

PR #3746, opened 2024-10-09
[completion] Fix custom completion style and make it default for fuzzy matching

alexander-yakushev:
Back to completion woes.

Some time ago we moved the official way to enable fuzzy Compliment-driven completion from a custom completion style to `flex`. This worked correctly until https://github.com/company-mode/company-mode/issues/1493 – cider completion now no longer suggests candidates for class shortnames like e.g. `Persistent`. Nothing has been broken in Company, vice versa, Company has became more like corfu and `complete-at-point`, so it was already broken in those clients.

I tried to fix it for `flex` but did not succeed. It looks like owning a custom completion style is still the proper way to go, so I'm resurrecting it with this PR and undeprecating the previous way to enable "fuzzy" completion. The current _partial_ compatibility with `flex` remains, so if folks prefer to use that, they still can.

I've also fixed our custom completion style so that it is compatible with Emacs machinery – something that users complained about before.

### DIFF
--- a/doc/modules/ROOT/pages/usage/code_completion.adoc
+++ b/doc/modules/ROOT/pages/usage/code_completion.adoc
@@ -41,9 +41,12 @@ CIDER defines a specialized completion category through the `cider-complete-at-p
 added to `completion-at-point-functions`, establishing a dedicated completion category named
 `cider`.
 
-The CIDER completion at point function supports most completion styles, including
-`partial-completion`, `orderless` and `flex` (read more below).
-
+The CIDER completion at point function supports most completion styles,
+including `partial-completion`, `orderless` and `flex`. It also supports a
+custom completion style that is confusingly named `cider` too. Activating it
+provides a richer set of completion candidates (see
+xref:usage/code_completion.adoc#fuzzy-candidate-matching[fuzzy candidate
+matching]).
 
 Sometimes the user may want to use a different completion style just for the CIDER
 complete at point function. That can be achieved by setting
@@ -57,8 +60,6 @@ complete at point function. The following snippet accomplishes that:
 
 This specifies that the `cider` completion category should employ the basic completion style by
 default.
-
-You can also enable the `flex` completion style by activating xref:usage/code_completion.adoc#fuzzy-candidate-matching[fuzzy candidate matching].
 
 == Auto-completion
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Seems we constantly manage to both fix something and break something on the completion side of things. Let me think about this for a bit.

alexander-yakushev:
Sure. But what I understood from @dgutov words is that a custom completion style is an OK solution for many packages, bot something to shun from.

bbatsov:
I'm mostly worried about this part:

> Please go ahead and ping me if you have further questions or problems adapting the style implementation. This is still a breaking change in company-mode's behavior, and I'd like to see all major completion providers onboard before tagging the next release.

I'm not sure if this will work fine with the current stable version of company, so I wanted to investigate. But if you're sure about this I guess we can proceed.

alexander-yakushev:
It should be fine. The changes in Company introduced problems for the current `flex`-based solution but didn't touch the old custom completion-style solution. Besides, it's opt-in anyway, so it wouldn't impact users that don't have it enabled in the first place.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -41,9 +41,12 @@ CIDER defines a specialized completion category through the `cider-complete-at-p
 added to `completion-at-point-functions`, establishing a dedicated completion category named
 `cider`.
 
-The CIDER completion at point function supports most completion styles, including
-`partial-completion`, `orderless` and `flex` (read more below).
-
+The CIDER completion at point function supports most completion styles,
+including `partial-completion`, `orderless` and `flex`. It also supports a
+custom completion style that is confusingly named `cider` too. Activating it
```
bbatsov, line 46: Perhaps we should name it `fuzzy-cider`, `cider-fuzzy`, `cider-flex` or something like this, so it's less confusing?

alexander-yakushev: I think there is value in retaining the current name for compatibility reasons. Besides, I always had a dare to make it a default someday, so I'm not sure if emphasizing `fuzzy` would be necessary. It's just the way of completion that CIDER provides and that's it. But in the end, I don't have a strong opinion about the name of the completion style.

dgutov:
@bbatsov 

> I'm not sure if this will work fine with the current stable version of company, so I wanted to investigate.

The new release came out a month ago, so this won't be a problem. I indeed hoped to get more backends onboard first, but that only had partial success. OT2H, the problem had been present with all (?) other UIs except company-mode, so it can't be too bad.

@alexander-yakushev

> a custom completion style is an OK solution for many packages, not something to shun from

I agree this is more complicated than we'd want to, but for now it's what's recommended for CAPF. Also see eglot's and lsp-mode's styles.

PR #3750, opened 2024-10-21
[completion] Use ignore instead of try-completion-function in CIDER completion style

### LINKED ISSUE #3749: Error after enabling `cider` completion style
## Expected behavior

`cider` completion style works without errors.

## Actual behavior

Error:
```
completion--in-region-1: Wrong type argument: listp
```
in the echo area and in the `*Messages*` buffer.

1. Start `emacs -q`
2. Evaluate:
```emacs-lisp
(package-initialize)
(require 'cider)
(cider-enable-cider-completion-style)
```
3. jack-in to a clojure project.
4. type something and try to use `complete-symbol` or `C-M-i`.

The feature was introduced in the recent PR - https://github.com/clojure-emacs/cider/pull/3746

### CIDER version information

```
;; CIDER 1.16.0 (Kherson), nREPL 1.3.0
;; Clojure 1.12.0, Java 23
```

Clojure CLI 1.12

### Emacs version

30

### Operating system

macOS

### JDK distribution

openjdk version "23" 2024-09-17
OpenJDK Runtime Environment Homebrew (build 23)
OpenJDK 64-Bit Server VM Homebrew (build 23, mixed mode, sharing)

### Stack trace

```
Debugger entered--Lisp error: (wrong-type-argument listp #("print" 0 1 (type "function" ns "clojure.core")))
  completion--do-completion(#<marker at 231 in app.clj> 235)
  completion--in-region-1(#<marker at 231 in app.clj> 235)
  #f(compiled-function (start end collection predicate) #<bytecode 0x1edeb26f2986a456>)(#<marker at 231 in app.clj> 235 #f(compiled-function (prefix pred action) #<bytecode -0x971aca080484915>) nil)
  apply(#f(compiled-function (start end collection predicate) #<bytecode 0x1edeb26f2986a456>) (#<marker at 231 in app.clj> 235 #f(compiled-function (prefix pred action) #<bytecode -0x971aca080484915>) nil))
  #f(compiled-function (funs global args) #<bytecode 0x1802ed4d7104dedc>)(nil nil (#<marker at 231 in app.clj> 235 #f(compiled-function (prefix pred action) #<bytecode -0x971aca080484915>) nil))
  completion--in-region(#<marker at 231 in app.clj> 235 #f(compiled-function (prefix pred action) #<bytecode -0x971aca080484915>) nil)
  completion-in-region(#<marker at 231 in app.clj> 235 #f(compiled-function (prefix pred action) #<bytecode -0x971aca080484915>) nil)
  completion-at-point()
  complete-symbol(nil)
  funcall-interactively(complete-symbol nil)
  command-execute(complete-symbol)
```

### Screenshot

image

alexander-yakushev:
Should fix #3749.

Turns out the API of a custom completion style's "try-completion function" doesn't match Emacs' own `try-completion` because why the hell should it, right? The API expected from that function is also not documented anywhere.

However, I've seen that `eglot` simply uses `ignore` instead of implementing that function and everything still works. I've verified locally that the completion is still operational with `ignore` there in both company-mode and `M-x complete-symbol`.

Asking @rrudakov to kindly verify the fix locally (you can replace the body of `cider-completion-try-completion` with just returning nil to verify).

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -256,14 +256,6 @@ in the buffer."
 
 ;; Fuzzy completion for company-mode
 
-(defun cider-completion-try-completion (string collection pred _)
-  "Return longest common substring of all completions of STRING in COLLECTION,
-also pass PRED to `try-completion'.
-
-This function is only needed to be a correct citizen in
-`completion-styles-alist'."
-  (try-completion string collection pred))
-
 (defun cider-company-unfiltered-candidates (string &rest _)
   "Return CIDER completion candidates for STRING as is, unfiltered."
   (cider-complete string))
@@ -276,7 +268,8 @@ This function is only needed to be a correct citizen in
 ;;  which introduced `cider-company-enable-fuzzy-completion')
 (add-to-list 'completion-styles-alist
              '(cider
-               cider-completion-try-completion
+               ;; Use `ignore' in place of "try-competion function".
+               ignore
                cider-company-unfiltered-candidates
                "CIDER backend-driven completion style."))
 


rrudakov:
Thanks @alexander-yakushev! This fixes the error.

One thing I noticed with the cider completion style, the matched characters are not highlighted in the candidates list.

With built-in `flex` style it looks like this:
image

With `cider` completion style:
image

Another issue I noticed is that "fuzzy" matching doesn't really work, if I type `Instnt` with `cider` completion style I'm not getting any completion suggestions, which is not the case with `flex` style.

alexander-yakushev:
Wait, did you really typed `Instnt`, triggered the completion, and got all of the candidates on the first screenshot?

rrudakov:
> Wait, did you really typed `Instnt`, triggered the completion, and got all of the candidates on the first screenshot?

Yes, it works with `flex` completion style.

rrudakov:
I'm not sure if it works with company though, but it definitely works with built-in completion and with corfu.

alexander-yakushev:
Are you sure you haven't triggered completion at `Inst` and then typed in the additional characters?

rrudakov:
> Are you sure you haven't triggered completion at `Inst` and then typed in the additional characters?

Ah, sorry, looks like this is it. I didn't realize it actually makes a difference.

rrudakov:
So I guess with `flex` styles the list is fetched once and then filtered on the client's side and with `cider` style it's fetched on every typed character, right?

alexander-yakushev:
> and with cider style it's fetched on every typed character, right?

I don't think this should happen. I think it still happens on the frontend but with stricter rules than `flex` matching.

alexander-yakushev:
And the whole ordeal of moving from `flex` happened because it doesn't work correctly with completing `Persistent`, for example.

alexander-yakushev:
Regarding the lack of highlighting matched characters – yeah, I noticed that as well. It still highlights correctly if the completion prefix is the actual prefix of the candidate:
image
But there is no highlight for non-prefix candidates like you've showed. I currently don't know if there is a way to fix this.

alexander-yakushev:
> Am I missing something?

Can you please try the following:

1. Enter `Persistent` into the editor.
2. Do `M-x completion-at-point`.

What does it do?

rrudakov:
> It still highlights correctly if the completion prefix is the actual prefix of the candidate

I'm pretty sure it's because `basic` completion style is used for that.

rrudakov:
> > Am I missing something?
> 
> Can you please try the following with `flex` style:
> 
> 1. Enter `Persistent` into the editor.
> 2. Do `M-x completion-at-point`.
> 
> What does it do?


https://github.com/user-attachments/assets/c6b5dc99-e14e-4962-b2f9-b94c33596c00

alexander-yakushev:
Interesting. On my machine, a different thing happens. Let me try to reproduce with `emacs -q`.

rrudakov:
I've just tried to reproduce with `emacs -q` and I think I was able to reproduce your issue. `(Persistent)` becomes `(.Persistent)` and then I cannot get any more completion candidates. I'm curious what do I have in my config to make it actually work.

rrudakov:
So, after evaluating `(setopt completion-category-overrides '((cider (styles . (flex)))))` completion starts working as expected.

Funny that setting `(setopt completion-styles '(flex))` doesn't fix that.

alexander-yakushev:
> I'm curious what do I have in my config to make it actually work.

That would be great to know indeed 😄.

rrudakov:
> So, after evaluating `(setopt completion-category-overrides '((cider (styles . (flex)))))` completion starts working as expected.
> 
> Funny that setting `(setopt completion-styles '(flex))` doesn't fix that.

Too soon, actually it's `(setopt completion-ignore-case t)`. @alexander-yakushev could you please confirm?

rrudakov:
Full snippet to make it work in `emacs -q`:
```emacs-lisp
(package-initialize)
(require 'cider)
(setopt completion-styles '(flex))
(setopt completion-ignore-case t)
```

alexander-yakushev:
Thanks, I could reproduce that indeed. Still, this looks like a side effect. The problem comes from the fact that completion-at-point machinery looks through all candidates returned by the backend and tries to grow a common substring from them. It happens so for `Persistent` and other similar queries, there is a common prefix `.` in all of them, that `.` is prepended, completion is triggered again, and Compliment returns nothing for such query. I can imagine that enabling `completion-ignore-case` makes it so there is some candidate that no longer matches `.Persistent`??? Now that I typed it out it sounds even more confusing.

rrudakov:
image
There are candidates in the list that has lower case `p` as a first match, so I think you're right.  I think `Persistent` prefix is sent to the backend and the list of candidates is fetched, but then because of this `ignore-case` settings, there is no more common prefix, for some candidates it's `.` for others it's `p`, so we see the unfiltered list.

alexander-yakushev:
I'll take the liberty of merging this ASAP since the original issue is quite nasty.

PR #3753, opened 2024-10-29
Add cider-log-show-frameworks command

r0man:
This command renders the available log frameworks in a buffer. It shows the log framework name, the website and javadocs urls and the levels for now.

This is mostly to address @vemv's suggestion to ask the user to run `(cider-sync-request:log-frameworks)`. I hooked the command up at the usual places and added it with a small troubleshooting section to the docs.

We could get even more sophisticated and show the current appenders and consumers but I need a user interface designer for this. Let's do this at some other point.

-----------------

### DIFF
--- a/cider-log.el
+++ b/cider-log.el
@@ -48,6 +48,20 @@ Example values: \"Logback\", \"Timbre\"."
   :safe #'stringp
   :type 'string)
 
+(defcustom cider-log-frameworks-buffer "*cider-log-frameworks*"
+  "The name of the log frameworks popup buffer."
+  :group 'cider
+  :package-version '(cider . "1.17")
+  :safe #'stringp
+  :type 'string)
+
+(defcustom cider-log-auto-select-frameworks-buffer t
+  "Whether to auto-select the log frameworks popup buffer."
+  :group 'cider
+  :package-version '(cider . "1.17")
+  :safe #'booleanp
+  :type 'boolean)
+
 (defcustom cider-log-appender-id "cider-log"
   "The name of the default log appender."
   :group 'cider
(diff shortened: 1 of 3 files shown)


vemv:
Thanks a lot!

LGTM, but I cannot do a particularly deep dive atm.

btw, one thing that you might want to hammock is the behavior of selecting the sole available framework if there's only one.

Maybe we can keep the behavior but also `(message "cider-log-mode: only one framework available, auto-selecting <framework>")`. That way we leave something in the 'logs' so that things are less surprising when they go wrong.

Cheers - V

cider-log.el
```diff
@@ -48,6 +48,20 @@ Example values: \"Logback\", \"Timbre\"."
   :safe #'stringp
   :type 'string)
 
+(defcustom cider-log-frameworks-buffer "*cider-log-frameworks*"
+  "The name of the log frameworks popup buffer."
+  :group 'cider
+  :package-version '(cider . "1.16.1")
```
bbatsov, line 54: That should be 1.17. (the next release)

cider-log.el
```diff
@@ -48,6 +48,20 @@ Example values: \"Logback\", \"Timbre\"."
...
+  :safe #'stringp
+  :type 'string)
+
+(defcustom cider-log-auto-select-frameworks-buffer t
+  "Whether to auto-select the log frameworks popup buffer."
+  :group 'cider
+  :package-version '(cider . "1.16.1")
```
bbatsov, line 61: Same here.

cider-log.el
```diff
@@ -1032,6 +1046,26 @@ the CIDER Inspector and the CIDER stacktrace mode.
 
 ;; Framework actions
 
+(transient-define-suffix cider-log-frameworks ()
+  "Show the log frameworks in a buffer."
```
bbatsov, line 1050: "Show the available log..." would be a bit better IMO.

cider-log.el
```diff
@@ -1032,6 +1046,26 @@ the CIDER Inspector and the CIDER stacktrace mode.
 
 ;; Framework actions
 
+(transient-define-suffix cider-log-frameworks ()
```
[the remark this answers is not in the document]
r0man: Yes, I changed it to `cider-log-show-frameworks`.

cider-log.el
r0man: Yes, I agree

bbatsov:
Thanks!

PR #3757, opened 2024-11-14
[inspector] Fix def-current-value selecting wrong REPL

alexander-yakushev:
Inspector's `cider-inspector-def-current-val` (hotkey `d`) behaves incorrectly if multiple REPLs are connected. It rebinds our carefully managed `cider-inspector--current-repl` with `(cider-current-repl)` which returns whatever in non-file and non-REPL buffers.

Perhaps, this is another reminder to make a generic "we know this buffer is associated with this REPL" variable and account  for it in `(cider-current-repl)`.


bbatsov (approved):

bbatsov:
> Perhaps, this is another reminder to make a generic "we know this buffer is associated with this REPL" variable and account for it in (cider-current-repl).

Agreed.

Issue #3763, opened 2024-12-18
Completion Popup causes `cider-docview-render` to error when symbol being completed does not have a docstring.


vedang:
## Expected behavior
Completion works without throwing a backtrace

## Actual behavior
```
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  replace-regexp-in-string("\n  " "\n" nil)
  cider-docstring--format(nil)
  (cider-docstring--trim (cider-docstring--format fetched-doc))
  ...
  cider-docview-render(#<buffer *cider-doc*> #("some-symbol" 0 1 (ns "tech.unravel.some.ns.core" type "function")) (dict "status" ("done") "arglists-str" "[some]" "column" 1 "file" "file:/path/to/core.clj" "id" "73" "line" 107 "name" "some-symbol" "ns" "tech.unravel.some.ns.core" "session" "d0de2276-b315-40fc-a5dc-e2c3a2720e01" "time-stamp" "2024-12-18 16:01:59.332041000") :compact)
  cider-create-doc-buffer(#(("some-symbol" 0 1 (ns "tech.unravel.some.ns.core" type "function")) :compact)
  cider-create-compact-doc-buffer(#("some-symbol" 0 1 (ns "tech.unravel.some.ns.core" type "function")))
  corfu-popupinfo--get-documentation(#("some-symbol" 0 1 (ns "tech.unravel.some.ns.core" type "function")))
  corfu-popupinfo--show(#("some-symbol" 0 1 (ns "tech.unravel.some.ns.core" type "function")))
```

## Steps to reproduce the problem
- Use `corfu` for completion
- Try to complete a symbol where the first completion does not have a docstring
- Wait a short while (2s by default) for corfu to try creating a info popup
- See the backtrace

### CIDER version information

```
;; CIDER 1.16.1 (Kherson), nREPL 1.3.0
;; Clojure 1.12.0-2, Java 23.0.1
```

### Lein / Clojure CLI version

Clojure CLI version 1.12.0.1488

### Emacs version

GNU Emacs 31.0.50 (build 1, aarch64-apple-darwin22.6.0, NS appkit-2299.77 Version 13.6.7 (Build 22G720)) of 2024-07-02

### Operating system

MacOS version 15.2

### JDK distribution

openjdk version "23.0.1" 2024-10-15

PR #3764, opened 2024-12-18
Fix: Do not call cider-docstring fns unless we have a doc

vedang:
Sometimes, there is no "doc" attribute in the nrepl-dict. In such cases, `fetched-doc` is nil, which causes `cider-docstring--*` functions to error.

We should only call these functions when we have a docstring to render.

Fixes: #3763

-----------------

- [-] All tests are passing (`eldev test`)
- [-] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes
- [-] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/cider-doc.el
+++ b/cider-doc.el
@@ -428,10 +428,11 @@ in a COMPACT format is specified, FOR-TOOLTIP if specified."
                                                             "doc-first-sentence-fragments" (nrepl-dict-get info "doc-first-sentence-fragments"))))
          (fetched-doc (nrepl-dict-get info "doc"))
          (doc     (or rendered-fragments
-                      (if compact
-                          (cider-docstring--trim
-                           (cider-docstring--format fetched-doc))
-                        fetched-doc)
+                      (when fetched-doc
+                        (if compact
+                            (cider-docstring--trim
+                             (cider-docstring--format fetched-doc))
+                          fetched-doc))
                       (unless compact
                         "Not documented.")))
          (url     (nrepl-dict-get info "url"))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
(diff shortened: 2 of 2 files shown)


vemv (approved):
lgtm, thanks!

cider-doc.el
```diff
@@ -428,10 +428,11 @@ in a COMPACT format is specified, FOR-TOOLTIP if specified."
...
-                      (if compact
-                          (cider-docstring--trim
-                           (cider-docstring--format fetched-doc))
-                        fetched-doc)
+                      (when fetched-doc
+                        (if compact
+                            (cider-docstring--trim
+                             (cider-docstring--format fetched-doc))
+                          fetched-doc))
                       (unless compact
```
bbatsov, line 436: Btw, this `unless` looks weird to me. Shouldn't this also check `fetched-doc`?

vedang: My understanding of the code is:

1. Return `rendered-fragments` if we already have them
2. Else return the `fetched-doc` (handle optional `compact` flag, depending on where the doc is being rendered)
3. Else return "Not documented" for when we are rendering in *Help* buffer, or return nothing if rendering an e.g. completion popup.

Only code-path 2 deals with fetched-doc. Others don't.

bbatsov: This part I get - I'm just confused how can a check about `compact` be related to the presence or absence of a docstring. I know it's not directly related to your PR - it's just something that looks wrong to me, so I thought I should mention it.

vedang: If the caller is asking for a compact docstring, and no docstring exists, return nil. If the caller is asking for an "expanded" docstring, and no docstring exists, return "not documented".

bbatsov:
Just add a changelog for your bigfix and we're good go.

PR #3767, opened 2025-01-01
Bump nrepl to 1.3.1 and cider-nrepl to 0.51.0

alexander-yakushev:
I think some documentation can be reworked too mentioning that certain jump-to and javadoc parsing will now work for certain classes without enrich. But we can do it later once other people confirm it works.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Changes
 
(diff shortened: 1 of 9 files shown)


bbatsov (approved):

bbatsov:
> I think some documentation can be reworked too mentioning that certain jump-to and javadoc parsing will now work for certain classes without enrich. But we can do it later once other people confirm it works.

Fine by me.

PR #3773, opened 2025-02-04
Add Emacs HEAD to CI matrix

purcell:
See #3606.


purcell:
Seems okay, but I'll refrain from merging this myself.

bbatsov:
I thought we had the HEAD in the CI, but we probably removed it while it was causing some issues in the past. Thanks!

PR #3774, opened 2025-02-14
[overlays] Fix overlay slowness when huge result is displayed

alexander-yakushev:
Since calling `string-width` on a big string is very slow, displaying a big overlay (that we truncate later anyway) induces an unnecessary delay.

Here, I solve this by checking first for a too big of a result (10000 characters) that we certainly definitely going to truncate.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 ### Bugs fixed
 
 - [#3763](https://github.com/clojure-emacs/cider/issues/3763): Fix `cider-docview-render` completion popup error when symbol being completed does not have a docstring.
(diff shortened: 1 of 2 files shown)


bbatsov:
> Here, I solve this by checking first for a too big of a result (10000 characters) that we certainly definitely going to truncate.

How did you pick this size? Overall - your approach seems reasonable, although probably I'd some comments explaining it in the code.

alexander-yakushev:
I indeed just spitballed it. Now I think we can base the approximation on the screen width.

alexander-yakushev:
Updated, no more magic values.

bbatsov:
Yeah, that's much nicer. Thanks!

bbatsov (approved):

PR #3777, opened 2025-02-26
Inspector cleanup

alexander-yakushev:
1. Remove code responsible for rendering Javadoc fragments (already dropped from cider-nrepl).
2. Remove compatibility code that was introduced together with the javadoc fragment code – use the new approach exclusively. I'm not sure what the compatibility code was for; perhaps, for cases when people relied on those functions in their Emacs config somehow. Anyway, a few years has passed already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,19 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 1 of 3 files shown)


bbatsov:
We might want to do the OS integration tests just on one version of Emacs, as those are taking forever. I came back to merge this PR only to see the CI is still in progress one hour later.

PR #3779, opened 2025-02-27
Fix cider-find-keyword for clojure-ts-mode

rrudakov:

- [ ] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes. There are warnings, but not because of my changes.

Thanks!

Currently `cider-find-keyword` doesn't work with `clojure-ts-mode`, because it relies on `clojure-keyword-face` text property (which is different for keywords for `clojure-ts-mode`). I've introduced a small helper which uses text property for `clojure-mode` and tree sitter based predicate for `clojure-ts-mode`.

### DIFF
--- a/cider-util.el
+++ b/cider-util.el
@@ -69,6 +69,10 @@ Setting this to nil removes the fontification restriction."
   "Return non-nil if current buffer is managed by a ClojureC major mode."
   (derived-mode-p 'clojurec-mode 'clojure-ts-clojurec-mode))
 
+(defun cider-clojure-ts-mode-p ()
+  "Return non-nil if current buffer is managed by a Clojure[TS] major mode."
+  (derived-mode-p 'clojure-ts-mode))
+
 (defun cider-util--clojure-buffers ()
   "Return a list of all existing `clojure-mode' buffers."
   (seq-filter
@@ -106,6 +110,18 @@ which nREPL uses for temporary evaluation file names."
 If BUFFER is provided act on that buffer instead."
   (with-current-buffer (or buffer (current-buffer))
     (or (cider-clojurec-major-mode-p))))
+
+(defun cider-keyword-at-point-p (&optional point)
+  "Return non-nil if POINT is in a Clojure keyword.
+
+Take into consideration current major mode."
+  (let ((pos (or point (point))))
+    (if (and (cider-clojure-ts-mode-p)
+             (fboundp 'clojure-ts--keyword-node-p)
+             (fboundp 'treesit-node-parent)
+             (fboundp 'treesit-node-at))
+        (clojure-ts--keyword-node-p (treesit-node-parent (treesit-node-at pos)))
+      (member 'clojure-keyword-face (text-properties-at pos)))))
 
 
 ;;; Thing at point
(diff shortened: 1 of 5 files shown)


rrudakov:
Looks like it's not that simple. CIDER currently doesn't have `clojure-ts-mode` as a dependency.

rrudakov:
Not sure how to fix the linter error. Any suggestions are welcome.

The `treesit-*` functions are only used if `treesit` feature is available, but I don't know how to tell linter about it.

rrudakov:
I've figured it out :)

cider-util.el
```diff
@@ -107,6 +117,17 @@ If BUFFER is provided act on that buffer instead."
   (with-current-buffer (or buffer (current-buffer))
     (or (cider-clojurec-major-mode-p))))
 
+(defun cider-keyword-at-p (point)
```
bbatsov, line 120: Perhaps `cider-keyword-at-point-p` instead?

cider-util.el
```diff
@@ -69,6 +73,12 @@ Setting this to nil removes the fontification restriction."
   "Return non-nil if current buffer is managed by a ClojureC major mode."
   (derived-mode-p 'clojurec-mode 'clojure-ts-clojurec-mode))
 
+(defun cider-clojure-ts-mode-p ()
+  "Return non-nil if current buffer is managed by a Clojure[TS] major mode."
+  (and (featurep 'treesit)
+       (featurep 'clojure-ts-mode)
+       (derived-mode-p 'clojure-ts-mode)))
```
bbatsov, line 80: Are the other two checks really needed?

cider-util.el
rrudakov: I was thinking about it, but here we pass `point` as an argument. Should I rename it?

cider-util.el
rrudakov: Probably not. I was trying to silent the linter. I've removed it now.

cider-util.el
bbatsov: Yeah, I think that'd be better. You can also consider making the `point` optional and just getting it automatically if not passed it, which is a common convention for this type of functions.

cider-util.el
```diff
@@ -41,6 +41,10 @@
 ;; clojure-mode and CIDER
 (require 'clojure-mode)
 
+;; clojure-ts-mode and treesit if available
```
bbatsov, line 44: I don't think those requires are needed, as if the major mode is enabled obviously they've been required anyways, plus you have `fboundp` for anything that can come from them.

cider-util.el
rrudakov: Done.

cider-util.el
rrudakov: Done.

bbatsov:
Let me think about this a bit, as I've got a small concern about leaking implementation details from the major modes in CIDER and I'm wondering what's the best way to abstract this away. Some work will definitely need to be done in this directly to fully embrace `clojure-ts-mode`. (unless it takes so much time to do it, that we'll merge the two modes first, as is scheduled for Emacs 32 :D )

bbatsov:
That's what I was thinking as well, otherwise I would have done it by now. Perhaps we can wrap all direct calls to `clojure-mode` in some `cider-foo-bar` functions in the same file in CIDER, so we can all such glue code is contained within a single file. No better ideas come to mind right now.

bbatsov:
I'll merge this as is for now and I'll think about the proper isolation of the major modes later.

PR #3793, opened 2025-03-20
Remove features that relied on printed exception parsing

alexander-yakushev:
This PR removes features that depended on the now removed `analyze-stacktrace` op which used Haystack parsing functionality:

- Manual `cider-stacktrace-analyze-string` and `cider-stacktrace-analyze-at-point` functions.
- Automatic stacktrace parsing in log viewer.

To replace this functionality, the latest cider-nrepl now provides much better ability to jump to functions and methods printed in the stacktrace (see https://github.com/clojure-emacs/orchard/pull/320). I went through examples/tests in Haystack repository, and the improved Orchard can jump to most of them directly from any buffer, including REPL, e.g.:
```
clojure.lang.AFn.applyToHelper
clojure.lang.Compiler$InvokeExpr.eval
nrepl.middleware.interruptible-eval/evaluator/fn/fn
clojure.main/repl/read-eval-print/fn
clojure.core$eval.invokeStatic
clojure.core$with_bindings_STAR_.doInvoke
clojure.main$repl$read_eval_print__9110$fn__9113.invoke
```

-----------------


cider-stacktrace.el
```diff
@@ -930,48 +930,6 @@ through the `cider-stacktrace-suppressed-errors' variable."
...
-     `("op" "analyze-stacktrace"
-       "stacktrace" ,stacktrace
-       ,@(cider--nrepl-print-request-plist fill-column))
-     (lambda (response)
-       (setq causes (nrepl-dbind-response response (class status)
-                      (cond (class (cons response causes))
-                            ((and (member "done" status) causes)
-                             (cider-stacktrace--analyze-render causes)))))))))
-
-(defun cider-stacktrace-analyze-at-point ()
```
bbatsov, line 960: I think it's good to replace these obsolete commands with some message explaining why there were removed. I don't think they had many users, being relatively new, but adding a message like "Use x instead" is never a bad idea IMO.

bbatsov (approved):
Looks good to me, sans my small remark that it might be a good idea to convert the remove interactive commands to just a message saying that users are supposed to use something else instead. (before we fully remove them a few releases down the road) You can also gut them and use `make-obsolete` https://www.gnu.org/software/emacs/manual/html_node/elisp/Obsolete-Functions.html

cider-stacktrace.el
alexander-yakushev: Good point! Will do.

PR #3797, opened 2025-03-27
[completion] Enable cider-completion-style by default

alexander-yakushev:
I'm feeling lucky about enabling our custom `cider-completion-style` by default. We had some troubles with it in the past – it worked not according to how completion spec expected it, but now these issues are fixed as far as I can tell. The other reason why it took long to fix it is that this was optional and I don't think many users enabled it, so we didn't receive enough data about the bug we had.

Anyway, I've been using this since forever and I'm confident now that this is a net positive feature to have on for everybody. I've modified `cider-enable-cider-completion-style` to accept `ARG = -1` to have a way for users to disable this if they don't want it (in the same way as they now enable it).

-----------------

### DIFF
--- a/cider-completion.el
+++ b/cider-completion.el
@@ -283,27 +283,31 @@ DEPRECATED: please use `cider-enable-cider-completion-style' instead."
   (interactive)
   (cider-enable-cider-completion-style))
 
-(defun cider-enable-cider-completion-style ()
-  "Enables `cider' completion style for CIDER in all buffers.
+(defun cider-enable-cider-completion-style (&optional arg)
+  "Enables or disables `cider' completion style for CIDER in all buffers.
 
 This style supports non-prefix completion candidates returned by the
-completion backend.  Only affects the `cider' completion category."
+completion backend.  Only affects the `cider' completion category.  If ARG
+is `1' or nil, enables the custom completion style; if `-1', disables it."
   (interactive)
-  (let* ((cider (assq 'cider completion-category-overrides))
-         (found-styles (assq 'styles cider))
-         (new-styles (if found-styles
-                         (cons 'styles (cons 'cider (cdr found-styles)))
-                       '(styles cider basic)))
-         (new-cider (if cider
-                        (cons 'cider
-                              (cons new-styles
-                                    (seq-remove (lambda (x) (equal 'styles (car x)))
-                                                (cdr cider))))
-                      (list 'cider new-styles)))
-         (new-overrides (cons new-cider
-                              (seq-remove (lambda (x) (equal 'cider (car x)))
-                                          completion-category-overrides))))
-    (setq completion-category-overrides new-overrides)))
+  (if (= arg -1)
+      (setq completion-category-overrides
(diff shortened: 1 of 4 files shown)


cider-completion.el
```diff
@@ -283,27 +283,35 @@ DEPRECATED: please use `cider-enable-cider-completion-style' instead."
...
+                                                  (cdr cider))))
+                        (list 'cider new-styles)))
+           (new-overrides (cons new-cider
+                                (seq-remove (lambda (x) (equal 'cider (car x)))
+                                            completion-category-overrides))))
+      (setq completion-category-overrides new-overrides))))
+
+;; Enable cider-completion-style by default. Users can opt out of this by
+;; running (cider-enable-cider-completion-style -1)
+(cider-enable-cider-completion-style 1)
```
bbatsov, line 314: I think we should do this in the minor-mode, as it's not very nice to run code at the top-level in general.

cider-completion.el
```diff
@@ -283,27 +283,35 @@ DEPRECATED: please use `cider-enable-cider-completion-style' instead."
...
 
-(defun cider-enable-cider-completion-style ()
-  "Enables `cider' completion style for CIDER in all buffers.
+(defun cider-enable-cider-completion-style (&optional arg)
+  "Enables or disables `cider' completion style for CIDER in all buffers.
 
 This style supports non-prefix completion candidates returned by the
-completion backend.  Only affects the `cider' completion category."
+completion backend.  Only affects the `cider' completion category.  If ARG
+is `1' or nil, enables the custom completion style; if `-1', disables it."
```
bbatsov, line 291: Might be more idiomatic for anything non-nil to enable it and for `nil` to disable it.

bbatsov:
No objections from me!

cider-completion.el
alexander-yakushev: The problem is that people currently have this function in their configs, and it now accepts zero arguments. To reuse the same function, we need to make ARG optional, and it means it arrives as nil when people want to enable.

alexander-yakushev: @bbatsov So is this OK to leave as is or is there a better solution?

bbatsov: Yeah, I guess it is.

doc/modules/ROOT/pages/usage/code_completion.adoc
```diff
@@ -139,31 +113,30 @@ without needing to hit an extra key, please customize:
...
+  to `java.util.function.BiFunction`.
+
+You can learn all completion scenarios and features
+https://github.com/alexander-yakushev/compliment/wiki/Examples[here].
 
-CIDER provides a function to enable the `cider` completion style for CIDER-specific
-completions. If you wish to enable that, you can add this to your config:
+If you only want to receive standard prefix-restricted completions (where the
+candidate must contain the prefix at the beginning verbatim), you can disable
+this feature by addint this to your config:
```
bbatsov, line 133: addint -> adding

PR #3799, opened 2025-03-31
ci: Use only one setup-emacs

jcs090218:
We only need `jcs090218/setup-emacs` since it supports all three platforms. See https://emacs-eldev.github.io/eldev/#continuous-integration.


bbatsov:
Thanks!

PR #3806, opened 2025-04-09
Add client-info to clone op

ericdallo:
As discussed on [this thread](https://clojurians.slack.com/archives/C17JYSA3H/p1744123473420459), we want to have client information about the nrepl client, we discussed about adding this extra info in `clone` op as optional information.

Related to https://github.com/nrepl/nrepl/pull/370

-----------------

Thanks!

manual][1] extremely useful.*


ericdallo:
@bbatsov I think the tests failing are not related to this PR

alexander-yakushev:
Looks like it is related to the PR:
image

ericdallo:
That's weird, the PR just add optional data, any clues @alexander-yakushev ?

alexander-yakushev:
Yeah, it is weird. This works correctly locally. @bbatsov any ideas?

bbatsov:
See https://github.com/clojure-emacs/cider/blob/master/test/nrepl-server-mock.el

alexander-yakushev:
Looks like the mock expects the clone message to be exactly this, without the extra fields: https://github.com/clojure-emacs/cider/blob/master/test/nrepl-server-mock.el#L46

ericdallo:
makes sense, just wondering how to get civer-version on that file

ericdallo:
I think I got it :)

PR #3810, opened 2025-04-13
Add command to print the current value of the inspector.

r0man:
This PR adds the interactive `cider-inspector-print-current-value` command that prints the current value of the inspector to the `cider-inspector-value-buffer` with the usual pretty printing machinery.

The NREPL part is here: https://github.com/clojure-emacs/cider-nrepl/pull/933

When pressing "P" in the inspector the currently inspected value is printed in a new popup buffer (on the left side in the picture).

cider-inspect-print-current-value

I see it as the inspectors brother of `cider-pprint-eval-last-sexp` which I use all the time.

This is the functionality I initially suggested, before working on the pretty mode. I think it is still useful. What I would like to have is a quick way to pretty print the current value of the inspector in the same familiar way my eye is used to, as the Clojure pretty printer (or what we are using in CIDER). 

The new pretty mode does print more pretty, but differently, it uses "inspector mode" printing.

That's why I think this could be a useful addition. 

Wdyt?

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+- [#3810](https://github.com/clojure-emacs/cider/pull/3810): Add the `cider-inspector-print-current-value` command to print the current value of the inspector.
 - [#3813](https://github.com/clojure-emacs/cider/pull/3813) Add support for pretty printing values in the inspector.
 - [#3802](https://github.com/clojure-emacs/cider/issues/3802): Inspector analytics.
 - [#3802](https://github.com/clojure-emacs/cider/issues/3802): Inspector table view-mode.
--- a/cider-inspector.el
+++ b/cider-inspector.el
@@ -30,6 +30,7 @@
 (require 'cl-lib)
 (require 'easymenu)
 (require 'seq)
+(require 'cider-client)
 (require 'cider-eval)
 
 ;; ===================================
@@ -146,6 +147,7 @@ Can be turned to nil once the user sees and acknowledges the feature."
     (define-key map [(shift tab)] #'cider-inspector-previous-inspectable-object)
     (define-key map "p" #'cider-inspector-previous-inspectable-object)
     (define-key map "P" #'cider-inspector-toggle-pretty-print)
+    (define-key map (kbd "C-c C-p") #'cider-inspector-print-current-value)
     (define-key map ":" #'cider-inspect-expr-from-inspector)
     (define-key map "f" #'forward-char)
     (define-key map "b" #'backward-char)
@@ -404,6 +406,18 @@ current-namespace."
     (cider-inspector--render-value result)
     (message "Defined current inspector value as #'%s/%s" ns var-name)))
 
+(defun cider-inspector-print-current-value ()
+  "Print the current value of the inspector."
+  (interactive)
+  (cider-ensure-connected)
+  (cider-ensure-op-supported "inspect-print-current-value")
+  (let ((buffer (cider-popup-buffer cider-result-buffer nil 'clojure-mode 'ancillary)))
+    (cider-nrepl-send-request
+     `("op" "inspect-print-current-value"
+       ,@(cider--nrepl-print-request-plist fill-column))
+     (cider-popup-eval-handler buffer)
+     (cider-current-repl))))
+
 (defun cider-inspector-tap-current-val ()
   "Sends the current Inspector current value to `tap>'."
   (interactive)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+- [#3810](https://github.com/clojure-emacs/cider/pull/3810) Add the `cider-inspector-print-current-value` command to
```
bbatsov, line 5: You forgot the `:` after the PR link. Also - this should be under "New features".

bbatsov:
I'm fine with this suggestion. @alexander-yakushev what do you think?

alexander-yakushev:
With pretty-printing mode now merged, do you still desire to have this feature, @r0man? If so, we'll need some other key binding.

cider-inspector.el
```diff
@@ -30,13 +30,15 @@
...
 (require 'seq)
+(require 'cider-client)
 (require 'cider-eval)
 
 ;; ===================================
 ;; Inspector Key Map and Derived Mode
 ;; ===================================
 
 (defconst cider-inspector-buffer "*cider-inspect*")
+(defconst cider-inspector-value-buffer "*cider-inspect-value*")
```
alexander-yakushev, line 41: Is it possible to reuse `*cider-result*` buffer for this instead of creating yet another buffer type?

r0man: I was actually using that before, but then thought a new buffer is better. But, yes, I can change it to use that one instead.

alexander-yakushev: It is best to reuse the existing facility that powers `*cider-result*` so that we don't need to add any new code for setting up the buffer, making it readonly, establishing `q` binding, etc.

r0man: I changed it to use the `*cider-result*` buffer.

cider-inspector.el
```diff
@@ -404,6 +406,26 @@ current-namespace."
...
+     ,@(cider--nrepl-print-request-plist fill-column))
+   (lambda (response)
+     (with-current-buffer (get-buffer-create cider-result-buffer)
+       (let ((inhibit-read-only t))
+         (nrepl-dbind-response response (status value)
+           (cond ((member "done" status)
+                  (goto-char (point-min)))
+                 ((member "value" response)
+                  (insert value)
+                  (newline)))))))
```
alexander-yakushev, line 426: Is it possible to use `cider-emit-into-popup-buffer` here instead of custom code? Maybe even reuse `cider-popup-eval-handler` as a whole somehow?

cider-inspector.el
```diff
@@ -146,6 +147,7 @@ Can be turned to nil once the user sees and acknowledges the feature."
     (define-key map [(shift tab)] #'cider-inspector-previous-inspectable-object)
     (define-key map "p" #'cider-inspector-previous-inspectable-object)
     (define-key map "P" #'cider-inspector-toggle-pretty-print)
+    (define-key map "x" #'cider-inspector-print-current-value)
```
r0man, line 150: Do you have a better suggestion for this?

r0man: I just saw you other comment. I change it to `C-p`.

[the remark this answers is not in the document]
r0man: Well, `C-p` is usually bound to `previous-line`. So maybe better not?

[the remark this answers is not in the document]
bbatsov: I'd go with something like `r` (I guess `render` can be a mnemonic for it) - in the absence of meaningful names I go for those that are easy to type.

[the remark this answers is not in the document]
alexander-yakushev: `M-p`?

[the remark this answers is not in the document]
alexander-yakushev: Yeah, that's a great suggestion.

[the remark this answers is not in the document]
r0man: I went with `C-c C-p` now. Are you fine with this @bbatsov ?

cider-inspector.el
r0man: Using `cider-popup-eval-handler` now.

cider-inspector.el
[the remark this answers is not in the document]
bbatsov: That's normally a minor mode keybinding (`C-c something`), so it's a bit weird IMO, but I don't have strong reservations for it.

alexander-yakushev (approved):
Looks very clean now!

alexander-yakushev:
Published cider-nrepl 0.55.3.

r0man:
Thank you @alexander-yakushev !

alexander-yakushev:
I updated to 0.55.3 in master.

r0man:
@alexander-yakushev Should update the cider-nrepl version in this PR?

alexander-yakushev:
> I updated to 0.55.3 in master.

r0man:
@alexander-yakushev Ok, should I merge it or wait for something?

alexander-yakushev:
Ah, sorry, thought there's something else left to modify in the PR.

r0man:
Alright, thanks for merging and your help on this!

PR #3812, opened 2025-04-16
Remove Boot support

alexander-yakushev:
Guess it is time to go.

I've kept the customizables and marked them obsolete. The functions I removed outright.

-----------------


bbatsov (approved):
No objections from me. Boot seems extremely dead these days.

Issue #3786, opened 2025-03-13
nrepl-bencode doesn't sort dictionaries by key


alexander-yakushev:
From https://clojurians.slack.com/archives/C17JYSA3H/p1741890145125339.

It appears that CIDER doesn't adhere to Bencode spec which requires dictionary keys to be sorted alphabetically. This hasn't been a problem so far because the bencode reader on nREPL side doesn't validate the order of keys. Still, it will be rigorous to produce correct values according to the selected format.

## Expected behavior

```
(nrepl-bencode '("b" 2 "a" 1))
=> "l1:ai1e1:bi2ee"
```

## Actual behavior

```
(nrepl-bencode '("b" 2 "a" 1))
=> "l1:bi2e1:ai1ee"
```

PR #3814, opened 2025-04-21
[Fix: #3786] Sort dictionaries by key in nrepl-bencode

iarenaza:
CIDER doesn't adhere to Bencode spec which requires dictionary keys to be sorted alphabetically. This hasn't been a problem so far because the bencode reader on nREPL side doesn't validate the order of keys. Still, it will be rigorous to produce correct values according to the selected format.

[Re: #3786]

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/nrepl-client.el
+++ b/nrepl-client.el
@@ -423,13 +423,30 @@ decoded message or nil if the strings were completely decoded."
       (erase-buffer)
       (cons string-q response-q))))
 
+(defun nrepl--bencode-dict (dict)
+  "Encode DICT with bencode.
+According to the Bittorrent protocol specification[1], when bencoding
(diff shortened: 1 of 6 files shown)


nrepl-client.el
```diff
@@ -423,13 +423,25 @@ decoded message or nil if the strings were completely decoded."
       (erase-buffer)
       (cons string-q response-q))))
 
+(defun nrepl--bencode-dict (dict)
+  "Encode DICT with bencode."
```
bbatsov, line 427: Probably we should mention in the docstring that the keys need to be sorted as per the spec, as this is probably not obvious to people not familiar with bencode - until recently even I didn't know this was in the spec. :D

PR #3818, opened 2025-05-07
save-mark-and-excursion that nasty call to check-parens

IrrenWirr:
nasty (check-parens)
it mess up marks selection in my idiosyncratic config of emacs

i.e when activating a mark, then moving around, and suddenly `cider-completion-get-info-context-at-point' get triggered by a timer, and mess-up my desired mark-region by another mark-region created from a mix of my starting point of activating mark and then end point of `check-parens'

So I put it in (save-mark-and-excursion) instead of only (save-excursion)

The only one that was messing with my usage was in "cider-completion-get-info-context-at-point" 

but who knows, maybe "cider-completion-get-context-at-point" also mess someone's emacs usage

So to be safe, did it to both, also because kinda hard to pinpoint the issue from a user perspective with an emacs loaded full of so many modes & customization, because its triggered by a timer...

-----------------

Thanks!

### DIFF
--- a/cider-completion-context.el
+++ b/cider-completion-context.el
@@ -51,7 +51,7 @@
 Note that this context is slightly different than that of
 `cider-completion-get-context-at-point': this one does not include
 the current symbol at point."
-  (when (save-excursion
+  (when (save-mark-and-excursion
           (condition-case _
               (progn
                 (up-list)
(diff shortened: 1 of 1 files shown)


alexander-yakushev:
Looks reasonable to me. Thanks!

PR #3819, opened 2025-05-07
Update interactive_programming.adoc

bigfoote:
fixed the link to "Additional Resources" from an anchor on this page to the actual, separate page for them. Seems like they must have been split off into their own page at some point, but this link wasn't updated accordingly.

-----------------

Thanks!

manual][1] extremely useful.*


doc/modules/ROOT/pages/usage/interactive_programming.adoc
```diff
@@ -58,4 +58,4 @@ understand the key concepts better. Here are a few ideas:
 NOTE: As CIDER is evolving rapidly, some of the information in those videos would likely be outdated by the time you get to watch them. Still,
 the core ideas of interactive programming are immutable, so whatever differences you come to observe and experience will likely be superficial.
 
-You can find more CIDER demos on the xref:additional_resources["Additional Resources"] page.
```
bbatsov, line 61: The actual problem is that the `.adoc` extension is missing for the `xref`.

PR #3823, opened 2025-05-17
Add command to pprint-eval-and-replace

### LINKED ISSUE #3822: Feature request: add command to evaluate sexp, pretty print result and replace sexp
**Problem statement**

At times, I have to use a lengthy datastructure to debug some of my code - for example, a 20KiB JSON response from a server, converted to a Clojure map, like this:

```
(def test-datastructure {:a :really :long "map" :with {:non-trivial "structure"} ... })
```

And I can iterate on that, but it becomes a pain to edit a small part of it: emacs is not great with long lines in general, and it's hard to tell which part of the structure exactly is being edited, if there's a lot of nesting happening.

**Possible solution**

It'd be great to have a command to evaluate the sexp and pretty-print the result, then replace the sexp with the result. It's kind of a combination of already existing `cider-eval-last-sexp-and-replace` and `cider-pprint-eval-last-sexp`.

**Alternatives**

There are multiple alternative solutions, all of which require invoking additional commands:
 - `cider-pprint-eval-last-sexp-to-repl` works, but requires bringing up REPL buffer and kill-yanking from it
 - `cider-pprint-eval-last-sexp` evals in a dedicated buffer, with the same problem
 - `cider-pprint-eval-last-sexp-to-comment` also works, but requires even more text manipulations.

I propose a dedicated command instead that could be used with single invocation.

**Additional context**

I'm not the only one with the problem -- please see https://clojurians.slack.com/archives/C099W16KZ/p1745507056198819 .

I'll be issuing a PR addressing this shortly.

trueneu:
Add command to eval sexp, pretty-print the result and replace the sexp with the result.

It may come in handy should the CIDER user need to define a large datastructure to debug their code, and change it slightly later.

The behaviour is the same as `cider-eval-print-last-sexp` with optional argument, but with additional replacing the original sexp.

NOTE: 
 - I haven't found tests related to printing, apologies if I missed them.
 - I feel there is refactoring potential here since I just copy-pasted the function essentially, but my elsip-fu is too weak to carry it out ATM.

Closes #3822

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New features
 
+- [#3822](https://github.com/clojure-emacs/cider/issues/3822): Add `cider-pprint-eval-last-sexp-and-replace` interactive command.
+
 ### Changes
 
 - [#3816](https://github.com/clojure-emacs/cider/issues/3816): **(Breaking)** Remove enrich-classpath support from cider-jack-in.
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -1036,6 +1036,22 @@ buffer."
                             nil
                             (cider--nrepl-pr-request-plist))))
 
+(defun cider-pprint-eval-last-sexp-and-replace ()
+  "Evaluate the expression preceding point and replace it with its
+pretty-printed result."
+  (interactive)
+  (let ((last-sexp (cider-last-sexp)))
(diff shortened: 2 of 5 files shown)


trueneu:
Scratch this, @rrudakov was kind enough to point out `cider-format-edn-last-sexp`.

PR #3824, opened 2025-05-29
Bump cider-nrepl to 0.56.0

alexander-yakushev:
Together with upgrading the version, I've also spliced some redundant functions insider `cider-inspector.el`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### Changes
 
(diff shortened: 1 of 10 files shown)


bbatsov (approved):

bbatsov:
For the version upgrades - for a long time I was thinking of adding a lein task or a Ruby script to do this (and updating the version when cutting a release), but I was too lazy to do it. And I guess I still am! 😅

PR #3830, opened 2025-07-18
Fix inspector not using the appropriate REPL connection

ak-coram:
I believe 9dc5107e8b48dbd5839df03ce533c4afa3682fb5 introduced an issue where the inspector is potentially not using the correct nrepl connection for evaluation. This is because `cider-nrepl-send-sync-request` defaults to `(cider-current-repl 'any 'ensure)` when given no connection, while calling `cider-current-repl` without arguments infers the correct type from the current buffer. This PR simply adds back the calls to `cider-current-repl` to the inspector.

-----------------

### DIFF
--- a/cider-client.el
+++ b/cider-client.el
@@ -170,9 +170,10 @@ nREPL connection."
 (defun cider-nrepl-op-supported-p (op &optional connection skip-ensure)
   "Check whether the CONNECTION supports the nREPL middleware OP.
 Skip check if repl is active if SKIP-ENSURE is non nil."
-  (nrepl-op-supported-p op (or connection (cider-current-repl nil (if skip-ensure
-                                                                      nil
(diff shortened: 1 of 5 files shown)


bbatsov:
Hmm, perhaps it's the default that should be adjusted, as I find the difference in the behavior a bit weird.

ak-coram:
@bbatsov So you want me to change this PR into changing the defaults of `cider-nrepl-send-sync-request` or leave it as is?

bbatsov:
I think changing the default would be better, as it's reasonable to expect it'd be using `cider-current-repl`.

ak-coram:
@bbatsov are we good?

bbatsov:
Just add a changelog entry for the bug-fix and we're good to go.

bbatsov (approved):

PR #3833, opened 2025-08-06
Fix error popup on pprint-eval-last-sexp

### LINKED ISSUE #3827: ddfc99b6 breaks cider--handle-stacktrace-response
## Expected behavior

Exceptions thrown from cider-eval pop up an error buffer when an exception is thrown.

## Actual behavior

An error buffer is not popped up, and the following gets printed to `*Messages*`:

```
error in process filter: Selecting deleted buffer
```

I tried to poke around the code and I think I narrowed it down to `cider--handle-stacktrace-response`.  When this function gets called, `source-buffer` is bound to `#<killed buffer>`.  I'm not sure how this buffer is getting killed but I don't recall killing it.

Reverting `ddfc99b6` seems to fix it for me.

evaluate: `(throw (ex-info "hello" {}))`

```
CIDER 1.19.0-snapshot (package: 1.19.0-snapshot)
GNU Emacs 30.1 (build 2, aarch64-apple-darwin24.3.0, NS appkit-2575.40 Version 15.3.1 (Build 24D70)) of 2025-02-26
Clojure 1.12.0
java version "24" 2025-03-18
Java(TM) SE Runtime Environment (build 24+36-3646)
MacOS 15.5 arm64
```

aiba:
Empirically, this fixes #3827 .  I have no idea why this works, but it seems to work.  Maybe this should not be merged and there is a better fix, but I'm sending this PR to reference from the issue.

### DIFF
--- a/cider-eval.el
+++ b/cider-eval.el
@@ -915,7 +915,7 @@ This is used by pretty-printing commands."
            (cider-popup-buffer-quit-function t)))
        ;; also call the default nrepl-err-handler, so that our custom behavior doesn't void the base behavior:
        (when nrepl-err-handler
-         (funcall nrepl-err-handler buffer)))
+         (funcall nrepl-err-handler (current-buffer))))
      ;; content type handler:
      nil
      ;; truncated handler:


alexander-yakushev:
Fixed by #3837. Thank you for the correct pointer to where the problem was!

PR #3837, opened 2025-08-27
[eval] Fix broken stacktrace response when C-c C-p throws an exception

alexander-yakushev:
This addresses #3827.

-----------------


aiba:
Working great, thank you!!!

PR #3839, opened 2025-09-02
Add jack-in support for ClojureCLR

rene-descartes2021:
Adds jack-in support for ClojureCLR.

Only changes are in the CIDER (elisp) side.
No changes to the cider-nrepl (Clojure) side are made.

Allows jacking into a ClojureCLR nREPL server **clojure/clr.tools.nrepl** [\[1\]](https://github.com/clojure/clr.tools.nrepl) with CIDER.

**clojure/clr.tools.nrepl** is at present a port of **babashka/babashka.nrepl** [\[2\]](https://github.com/babashka/babashka.nrepl) to ClojureCLR. Which might later change to a port of **nrepl/nrepl** [\[3\]](https://github.com/nrepl/nrepl) once that port to ClojureCLR is working.

So, at present jacking into ClojureCLR nREPL server this PR is set to is just like jacking into a Babashka nREPL server. Default middleware are in those repos not in cider-nrepl. W.r.t. those I see that middleware can be specified/added-to via the `:xform` parameter, but at present the jack in for Babashka doesn't support that so I didn't add it for the jack into ClojureCLR. I guess someone might want to do that with CIDER?

Once the **nrepl/nrepl** port to ClojureCLR is working (in the subdir of **clojure/clr.tools.nrepl**) then better integration with CIDER can be considered, including adapting **cider-nrepl** to ClojureCLR. That's my understanding. I guess that's the direction things are going.

Related: #3361

~~#### Problems with this PR:~~

~~For unknown reasons eventually the sync fails and things don't seem to work.~~

~~At that point reconnecting via CIDER commands doesn't seem to work. Hadn't diagnosed what to do about this, I've come from dotnet-land wanting to use Clojure: I've never used CIDER before so I don't know what to expect. Probably user-error I guess.~~

**EDIT:** Maybe spacing in the docs might be wrong. I eyeballed the formatting. I guess linting might complain if I got that wrong.

#### My use-case/motivation:
For ClojureCLR I think the trend is to use inf-clojure with the socket REPL server, not Cider with an nREPL server. For Calva the vscode plugin for Clojure, [looks like](https://blog.agical.se/en/posts/how-to-create-a-really-simple-clojureclr-dependency-tool/) they've looked into using **clojure/clr.tools.nrepl** (nREPL) for ClojureCLR but I don't use vscode.

I'm using Doom emacs and the clojure module uses CIDER not inf-clojure and I didn't want to deal with remapping hotkeys & other configuration. So I wanted to explore using **clojure/clr.tools.nrepl** as-is with CIDER.

I thought to submit this PR as a basis for others to continue in the effort using & improving CIDER with ClojureCLR.

-----------------

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/modules/ROOT/pages/caveats.adoc
+++ b/doc/modules/ROOT/pages/caveats.adoc
@@ -60,13 +60,11 @@ from your Emacs config.
 
 == ClojureCLR Support
 
-CIDER currently has very basic support for ClojureCLR (via Arcadia's nREPL server). The reasons for this are the following:
+CIDER currently has very basic support for ClojureCLR (via either a port of Babashka's nREPL server or Arcadia's nREPL server). The reasons for this are the following:
 
-* nREPL itself runs only on the JVM (because it leverages Java APIs
-internally). There's an
-https://github.com/clojure/clr.tools.nrepl[nREPL port for ClojureCLR], but
-it's not actively maintained and it doesn't behave like the Clojure nREPL.
-* `cider-nrepl` uses a lot of Java code internally itself.
+* The https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[nrepl/nrepl port to ClojureCLR] is not yet working
+* `cider-nrepl` uses a lot of Java code internally itself and would need to be adapted/ported like any
+  other clojure library adapted/ported to ClojureCLR.
 
 Those issues are not insurmountable, but are beyond the scope of our current roadmap.
 If someone would like to tackle them, we'd be happy to provide assistance.
--- a/doc/modules/ROOT/pages/caveats.adoc
+++ b/doc/modules/ROOT/pages/caveats.adoc
@@ -64,19 +64,6 @@ loaded. As a workaround remove
 
 from your Emacs config.
 
-== ClojureCLR Support
-
-CIDER currently has very basic support for ClojureCLR (via Arcadia's nREPL server). The reasons for this are the following:
-
-* nREPL itself runs only on the JVM (because it leverages Java APIs
-internally). There's an
-https://github.com/clojure/clr.tools.nrepl[nREPL port for ClojureCLR], but
-it's not actively maintained and it doesn't behave like the Clojure nREPL.
-* `cider-nrepl` uses a lot of Java code internally itself.
-
-Those issues are not insurmountable, but are beyond the scope of our current roadmap.
-If someone would like to tackle them, we'd be happy to provide assistance.
-
 == Injecting dependencies and Leiningen pedantic: abort mode
 
 Because injection currently creates an override of the nREPL dependency that
--- a/doc/modules/ROOT/pages/platforms/clojureclr.adoc
+++ b/doc/modules/ROOT/pages/platforms/clojureclr.adoc
@@ -2,27 +2,27 @@
 
 == Current Status
 
-ClojureCLR on CIDER is not great due to the lack of a fully-functional nREPL
-server for ClojureCLR. There are currently two options:
+You will get basic CIDER functionality with ClojureCLR. Three nREPL server options:
 
-- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: A direct (but incomplete) port of the reference Clojure nREPL server.
+- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: At present port of babashka's nREPL server (https://github.com/babashka/babashka.nrepl[babashka.nrepl]).
+- https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[port of nrepl/nrepl]: A non-working, work-in-progress port of nrepl/nrepl, which may
+  ultimately better integrate with CIDER once CIDER's middleware (cider-nrepl) is also adapted/ported.
 - https://github.com/arcadia-unity/Arcadia/blob/master/Editor/NRepl.cs[Arcadia's nREPL]: A basic, but working nREPL implementation in C#.
 
-If you need to use CIDER with ClojureCLR today Arcadia's nREPL is your only usable option. That being said - `clr.tools.nrepl` is a much
-more sophisticated project and ideally we should get it over to the finish line.
+An alternative to CIDER & a nREPL server is inf-clojure with ClojureCLR's stock socket REPL server.
 
 == Usage
 
 NOTE: Contributions welcome!
 
-As `cider-jack-in` doesn't support ClojureCLR projects out-of-the-box currently, you'll need to start an nREPL server externally and
-connect to it with `cider-connect`.
+`cider-jack-in-universal` will jack into a clr.tools.nrepl server as long as a `deps-clr.edn` file
+exists in the project directory, otherwise you may call `cider-jack-in-universal` with prefix
+argument 6, by either `M-6` or `C-u 6` followed by `M-x cider-jack-in-universal`.
 
 == Plans
 
 In an ideal world we'll achieve the following objectives:
 
-- out-of-the-box ClojureCLR support with `cider-jack-in`
 - feature parity between Clojure's nREPL implementation and `clr.tools.nrepl` (the project can use some help)
 - adapting `cider-nrepl` for ClojureCLR (some of its codebase is JVM-specific)
 
--- a/doc/modules/ROOT/pages/platforms/clojureclr.adoc
+++ b/doc/modules/ROOT/pages/platforms/clojureclr.adoc
@@ -8,10 +8,9 @@ You will get basic CIDER functionality with ClojureCLR. Two nREPL server options
 - https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[port of nrepl/nrepl]: A non-working, work-in-progress port of nrepl/nrepl, which may
   ultimately better integrate with CIDER once CIDER's middleware (cider-nrepl) is also adapted/ported.
 
-CIDER currently has basic support for ClojureCLR via a port of Babashka's nREPL
-  server. The reasons for the basic support is that: `cider-nrepl`, the CIDER
-  middleware, uses a lot of Java code internally itself and would need to be
-  adapted/ported like any other clojure library adapted/ported to ClojureCLR.
+CIDER's support is basic because `cider-nrepl`, the middleware behind most of
+CIDER's features, uses a lot of Java internally and would need to be ported to
+ClojureCLR like any other Clojure library.
 
 Alternatives to using CIDER with ClojureCLR include https://github.com/clojure-emacs/inf-clojure[inf-clojure] and https://github.com/nrepl/neat[neat].
 
--- a/doc/modules/ROOT/pages/platforms/clojureclr.adoc
+++ b/doc/modules/ROOT/pages/platforms/clojureclr.adoc
@@ -4,7 +4,7 @@
 
 You will get basic CIDER functionality with ClojureCLR. Two nREPL server options:
 
-- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: At present port of babashka's nREPL server (https://github.com/babashka/babashka.nrepl[babashka.nrepl]).
+- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: currently a port of https://github.com/babashka/babashka.nrepl[babashka.nrepl] to ClojureCLR.
 - https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[port of nrepl/nrepl]: A non-working, work-in-progress port of nrepl/nrepl, which may
   ultimately better integrate with CIDER once CIDER's middleware (cider-nrepl) is also adapted/ported.
 
--- a/doc/modules/ROOT/pages/platforms/clojureclr.adoc
+++ b/doc/modules/ROOT/pages/platforms/clojureclr.adoc
@@ -2,27 +2,31 @@
 
 == Current Status
 
-ClojureCLR on CIDER is not great due to the lack of a fully-functional nREPL
-server for ClojureCLR. There are currently two options:
+You will get basic CIDER functionality with ClojureCLR. Two nREPL server options:
 
-- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: A direct (but incomplete) port of the reference Clojure nREPL server.
-- https://github.com/arcadia-unity/Arcadia/blob/master/Editor/NRepl.cs[Arcadia's nREPL]: A basic, but working nREPL implementation in C#.
+- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: At present port of babashka's nREPL server (https://github.com/babashka/babashka.nrepl[babashka.nrepl]).
+- https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[port of nrepl/nrepl]: A non-working, work-in-progress port of nrepl/nrepl, which may
+  ultimately better integrate with CIDER once CIDER's middleware (cider-nrepl) is also adapted/ported.
 
-If you need to use CIDER with ClojureCLR today Arcadia's nREPL is your only usable option. That being said - `clr.tools.nrepl` is a much
-more sophisticated project and ideally we should get it over to the finish line.
+CIDER currently has basic support for ClojureCLR via a port of Babashka's nREPL
+  server. The reasons for the basic support is that: `cider-nrepl`, the CIDER
+  middleware, uses a lot of Java code internally itself and would need to be
+  adapted/ported like any other clojure library adapted/ported to ClojureCLR.
+
+Alternatives to using CIDER with ClojureCLR include https://github.com/clojure-emacs/inf-clojure[inf-clojure] and https://github.com/nrepl/neat[neat].
 
 == Usage
 
 NOTE: Contributions welcome!
 
-As `cider-jack-in` doesn't support ClojureCLR projects out-of-the-box currently, you'll need to start an nREPL server externally and
-connect to it with `cider-connect`.
+`cider-jack-in-universal` will jack into a clr.tools.nrepl server as long as a `deps-clr.edn` file
+exists in the project directory, otherwise you may call `cider-jack-in-universal` with prefix
+argument 6, by either `M-6` or `C-u 6` followed by `M-x cider-jack-in-universal`.
 
 == Plans
 
 In an ideal world we'll achieve the following objectives:
 
-- out-of-the-box ClojureCLR support with `cider-jack-in`
 - feature parity between Clojure's nREPL implementation and `clr.tools.nrepl` (the project can use some help)
 - adapting `cider-nrepl` for ClojureCLR (some of its codebase is JVM-specific)
 
--- a/lisp/cider-jack-in.el
+++ b/lisp/cider-jack-in.el
@@ -207,6 +207,30 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.14.0"))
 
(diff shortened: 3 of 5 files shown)


cider.el
```diff
@@ -298,6 +298,26 @@ By default we favor the project-specific shadow-cljs over the system-wide."
...
+  :package-version '(cider . "1.20.0"))
+
+(defcustom cider-clr-parameters
+  "-X clojure.tools.nrepl/start-server!"
+  "Params passed to ClojureCLR to start an nREPL server via `cider-jack-in'."
+  :type 'string
+  :safe #'stringp
+  :package-version '(cider . "1.20.0"))
+
+(defcustom cider-clr-nrepl-sha "a58009f03b489b51194834466a2ee7040dad5861"
```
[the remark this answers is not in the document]
rene-descartes2021: Yes it does that's what I tried at first but Clojure isn't designed that way for git dependencies. For maven just specifying `:mvn/version` works, but for a git dep if only `:git/tag` is specified it errors and still requires `:git/sha` anyway. So rather than have end user having to specify two things to version it I just elected for the full SHA.

Dunno why it was designed that way, at a glance I'd expect the git tag should be sufficient, but I guess maybe the lack of trust is from the possibility of a bad actor in a supply chain attack changing the tag in the repo? Always specifying the SHA at least in part should mitigate that:
https://clojure.org/reference/deps_edn#deps_git_tag

cider.el
```diff
@@ -298,6 +298,26 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.14.0"))
 
+(defcustom cider-clr-command
```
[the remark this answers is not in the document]
rene-descartes2021: > I wonder if we should used the more descriptive naming "clojure-clr-cli" (or the shorter "clr-cli") instead of just "clr" here,

So `cider-clojure-clr-cli-command` in full? Anywhere else? I wasn't really sure how to name things I just tried to follow any patterns I could make out.

> in case at some point we want to add a different way to jack in for Clojure CLR projects.

A different way being for Arcadia? That makes sense I think. I personally don't use Unity. Looking around in their issues/PRs I think they've made do just connecting with CIDER manually.

From what I see the commercial support for Arcadia collapsed and they suggested someone else fork it and maintain it. 

Looking at the 3 forks active after Arcadia's last commit, they didn't last more than a year, and all had some nREPL changes/fixes [[1]](https://github.com/longestpath/Arcadia/commits/longestpath-dev)[[2]](https://github.com/Folcon/Arcadia/commits/master/)[[3]](https://github.com/benjamin-asdf/Arcadia/commits/master/).

Well that's a tangent just exploring ideas.

Maybe instead the features within the Arcadia fork of ClojureCLR will get de-fragmented back in/around ClojureCLR by a motivated person(s), I think that would be the best path rather than forking Arcadia... in their videos they described their rationale for the fork and I think it makes sense now to de-fragment. But I wouldn't worry about it till a motivated person shows up and comment.

So I won't worry about supporting a different way to jack into ClojureCLR or a variant of it. Upcoming Clojure LLVM/C++ dialect Jank might better suit their Unity and performance use-case more anyway... or maybe not, I'm not sure.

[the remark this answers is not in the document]
bbatsov: > From what I see the commercial support for Arcadia collapsed and they suggested someone else fork it and maintain it.

Ah, I wasn't aware of this. I did remember that for a while that was the only way to run nREPL on .NET, but I guess things have changed since then.

But I was mostly think how with Clojure there were many different build tools over the years - I think the `dotnet` CLI itself is a fairly "new" development in the .NET community. I learned about it only when I started to play with F# last year. :D

.github/workflows/test.yml
```diff
@@ -27,17 +27,21 @@ jobs:
         # emacs_version: ['27.2', '28.2', '29.3', '30.1', 'snapshot']
         emacs_version: ['27.2', '28.2', '29.3', '30.1']
         java_version: ['21']
+        dotnet_version: ['8.0.413', '9.0.304']
```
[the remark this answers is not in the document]
rene-descartes2021: Oh ok just one .net version is fine with me. I'll cut out the non-LTS one if that's a thing.

I thought maybe to grab the version from where I got that workflow pattern, maybe keep it in sync, but couldn't find it again. It was one of dmiller's or the E-A-Griffin author's repos on GitHub. I found [this one](https://github.com/jank-lang/clojure-test-suite/pull/101) but that one I guess relies on a globals.json in a pre-existing dotnet install or a default? Sorry memory lost, writing notes for myself.

rene-descartes2021:
Ok I see the jack-in cleanup in #3883 #3884 #3885 docs, changelog. I see the new [cider-register-jack-in-tool](https://github.com/clojure-emacs/cider/blob/master/doc/modules/ROOT/pages/basics/up_and_running.adoc#universal-jack-in) usage. Very nice! 👍 

~~Well, I think I'll just add example of cider-register-jack-in-tool for ClojureCLR in the relevant spot in the docs. Still tinkering on that.~~

Reworking my changes to cider.el now.

~~Do you still want the ClojureCLR integration test in this PR? I did update it to dotnet 10 but didn't push it yet. **EDIT:** I imagine it'd be reworked to include a call to cider-register-jack-in-tool and then use that.~~

rene-descartes2021:
`make test` and `make integration-test` not working locally for me now, e.g. `eldev -dtT -p test --test-type integration`. Says stuff like:

```
Activating dependency package ‘clojure-ts-mode’...
Debugger entered--Lisp error: (error "Loading file /data/data/com.termux/files/home/dev/clr/cider/.eldev/30.2/local/packages/cider-1.22.0snapshot/cider-apropos.elc failed to provide feature ‘cider-apropos’")
#<subr require>(cider-apropos)
```

But just `eldev test` and `eldev test --test-type integration` run as expected but I still get the integration test quirks mentioned earlier. The "Project:" prompt and the cljr runs as expected but the test thinks it fails.

Maybe I should just drop the integration test from this PR? Since:
- fixing it is probably beyond me at this time
- you earlier mentioned adversity to making the test-matrix slower

Also I might have the wrong stuff in the cider-register-jack-in-tool call in cider.el. I put nbb in there wasn't really sure what to put. So... Well that's all I have for today.

bbatsov:
Thanks for sticking with this! Here's where I'd land:

**Drop the integration test.** Given the timeouts you can't easily reproduce and my general reluctance to grow the test matrix, the unit tests on the elisp logic are enough. We can revisit if/when the nrepl/nrepl ClojureCLR port stabilizes — at that point integration coverage starts to be worth the CI cost.

**`cider-register-jack-in-tool` shape.** Mirroring how babashka is registered (cider.el:849-853), ClojureCLR should look roughly like:

```elisp
(cider-register-jack-in-tool 'clojure-clr-cli
                             :command-var 'cider-clojure-clr-cli-command
                             :params-var 'cider-clojure-clr-cli-parameters
                             :project-files '("deps-clr.edn")
                             :universal-prefix-arg 6)
```

That replaces the static defcustoms with `clojure-clr-cli` as the registry key (leaves room for an Arcadia or other ClojureCLR variant later — addresses the naming question I had earlier). I think `nbb` snuck into your draft because you used the nbb registration as a template — that should just be `'clojure-clr-cli`.

**The SHA pinning** — when you rework, please drop in a brief comment explaining that Clojure git deps require `:git/sha` even when `:git/tag` is set ([deps_edn ref](https://clojure.org/reference/deps_edn#deps_git_tag)). Just so future readers don't wonder.

**`make test` locally** — that error usually means a stale `.eldev/` build dir; try `eldev clean` and rebuild. Unrelated to this PR though, don't let it block you.

No rush, ping me when the rework is up.

rene-descartes2021:
@bbatsov Rework is done.

doc/modules/ROOT/pages/caveats.adoc
```diff
@@ -60,13 +60,11 @@ from your Emacs config.
 
 == ClojureCLR Support
 
-CIDER currently has very basic support for ClojureCLR (via Arcadia's nREPL server). The reasons for this are the following:
+CIDER currently has very basic support for ClojureCLR (via either a port of Babashka's nREPL server or Arcadia's nREPL server). The reasons for this are the following:
```
bbatsov, line 63: I think we can remove mentions of Arcadia, as the project has been dead for a few years now. Less is more.

doc/modules/ROOT/pages/caveats.adoc
```diff
@@ -60,13 +60,11 @@ from your Emacs config.
...
 
-* nREPL itself runs only on the JVM (because it leverages Java APIs
-internally). There's an
-https://github.com/clojure/clr.tools.nrepl[nREPL port for ClojureCLR], but
-it's not actively maintained and it doesn't behave like the Clojure nREPL.
-* `cider-nrepl` uses a lot of Java code internally itself.
+* The https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[nrepl/nrepl port to ClojureCLR] is not yet working
```
bbatsov, line 65: Both those items will be quite easy to tackle today with the help of AI agents (as the outcome is fairly well specified, so it'd be mostly a matter of time and tokens), but I'm not sure if anyone has interest in filling in the blanks there. 

Perhaps we should remove this server as well, so we won't confuse users?

I'm also thinking we should probably suggest `neat` as a good alternative of CIDER for ClojureCLR, as it's simpler and makes no assumptions about the runtime.

bbatsov:
Overall things look good to me. I suggested a couple of small doc improvements to simplify things for the users and it'd be nice if you can rebase on top of the current `master`

doc/modules/ROOT/pages/caveats.adoc
rene-descartes2021: Ok I did that. Forgot to mention earlier David Miller, main maintainer of ClojureCLR said in talks that he did integrate Arcadia stuff into mainline ClojureCLR. Just the Arcadia folks wanted to do stuff faster and experimentally under different philosophy in the pursuit of their goals. So no problem removing references.

doc/modules/ROOT/pages/caveats.adoc
rene-descartes2021: > but I'm not sure if anyone has interest in filling in the blanks there

EDIT: *shrugs*

> Perhaps we should remove this server as well, so we won't confuse users?

Oh, it's fine to leave it I think.

ClojureCLR stuff is quite sparse and this CIDER web page comes out high on the Google search results when searching for ClojureCLR so I think it's fine and useful writing broad perspective for person searching for answers and the lay of the land.

> I'm also thinking we should probably suggest neat as a good alternative of CIDER for ClojureCLR

I put the neat and inf-clojure suggestions as alternatives.

I took out the ClojureCLR section out of caveats.adoc and merged it into platforms/clojureclr.adoc. Jack-in works now.

> as it's simpler and makes no assumptions about the runtime.

I figured something more generic about non-Java runtimes and neat would make sense there in that caveats document, your thinking on that. Well not part of this PR my thinking.

Sorry took me a while to... Context switch back into this and read new things.

lisp/cider-jack-in.el
```diff
@@ -894,6 +948,13 @@ Throws an error if PROJECT-TYPE is unknown."
                              :project-files '("basilisp.edn")
                              :dispatch-prefix-arg 5)
 
+(cider-register-jack-in-tool 'clojure-clr-cli
+                             :command-var 'cider-clojure-clr-cli-command
+                             :params-var 'cider-clojure-clr-cli-parameters
+                             :project-files '("deps-clr.edn")
+                             :inject-fn #'cider--clojure-clr-cli-inject-deps
+                             :universal-prefix-arg 6)
```
rene-descartes2021: Oh I see this universal-prefix-arg param changed, gotta fix that.

bbatsov (commented):
Thanks for sticking with this, it's close. A few small things inline, plus some tests to add.

lisp/cider-jack-in.el
```diff
@@ -225,6 +225,30 @@ By default we favor the project-specific shadow-cljs over the system-wide."
   :safe #'stringp
   :package-version '(cider . "1.14.0"))
 
+(defcustom cider-clojure-clr-cli-command
+  "cljr"
+  "The command used to execute ClojureCLR."
+  :type 'string
+  :group 'cider
+  :safe #'stringp
+  :package-version '(cider . "2.0.1"))
```
bbatsov, line 234: These should be `2.1.0`, 2.0.1 is already out. Same for the two below.

lisp/cider-jack-in.el
```diff
@@ -668,6 +692,30 @@ Does so by concatenating PARAMS and DEPENDENCIES."
...
 
+(defun cider-clojure-clr-cli-jack-in-dependencies (params dependencies &optional command)
+  "Create ClojureCLR clr.core.cli jack-in dependencies.
+Does so by concatenating DEPENDENCIES, and PARAMS into a
+suitable `cljr` invocation and quoting, also accounting for COMMAND if
+provided."
+  (let* ((all-deps (thread-last dependencies
+                                (cider--dedupe-deps)
+                                (seq-map (lambda (dep)
+                                           (if (listp (cadr dep))
```
bbatsov, line 703: The only caller passes a `(lib sha)` pair, so the `listp` branch is dead code carried over from the clojure-cli version. Something like this does the same job and also drops the doubled space before the params:

```elisp
(let* ((deps (format "{:deps {%s}}"
                     (mapconcat (lambda (dep)
                                  (format "%s {:git/sha \"%s\"}" (car dep) (cadr dep)))
                                (cider--dedupe-deps dependencies)
                                " ")))
       (deps-quoted (cider--shell-quote-argument deps command)))
  (if params
      (format "-Sdeps %s %s" deps-quoted params)
    (format "-Sdeps %s" deps-quoted)))
```

lisp/cider-jack-in.el
```diff
@@ -727,6 +775,12 @@ COMMAND is the resolved jack-in command, used to handle PowerShell quoting."
    (cider-add-clojure-dependencies-maybe cider-jack-in-dependencies)
    (cider-jack-in-normalized-nrepl-middlewares)))
 
+(defun cider--clojure-clr-cli-inject-deps (params _project-type _command)
+  "Inject CIDER deps into PARAMS for a ClojureCLR CLI project."
+  (cider-clojure-clr-cli-jack-in-dependencies
+   params
+   `(("io.github.clojure/clr.tools.nrepl" ,cider-clojure-clr-cli-nrepl-sha))))
```
bbatsov: `command` should be passed through so the PowerShell quoting kicks in, same as `cider--clojure-cli-inject-deps`. .NET users are more likely than most to be on Windows.

```suggestion
(defun cider--clojure-clr-cli-inject-deps (params _project-type command)
  "Inject CIDER deps into PARAMS for a ClojureCLR CLI project.
COMMAND is the resolved jack-in command, used to handle PowerShell quoting."
  (cider-clojure-clr-cli-jack-in-dependencies
   params
   `(("io.github.clojure/clr.tools.nrepl" ,cider-clojure-clr-cli-nrepl-sha))
   command))
```

lisp/cider-jack-in.el
```diff
@@ -894,6 +948,13 @@ Throws an error if PROJECT-TYPE is unknown."
                              :project-files '("basilisp.edn")
                              :dispatch-prefix-arg 5)
 
+(cider-register-jack-in-tool 'clojure-clr-cli
```
bbatsov, line 953: `deps-clr.edn` also needs to go into `cider-build-tool-files` in cider-util.el, otherwise `cider-project-dir` won't find the root of a CLR-only project. And `clojure-clr-cli` should be added to the `:type` choices of `cider-jack-in-default` and `cider-preferred-build-tool` so it shows up in customize.

lisp/cider-jack-in.el
```diff
@@ -894,6 +948,13 @@ Throws an error if PROJECT-TYPE is unknown."
...
+                             :command-var 'cider-clojure-clr-cli-command
+                             :params-var 'cider-clojure-clr-cli-parameters
+                             :project-files '("deps-clr.edn")
+                             :inject-fn #'cider--clojure-clr-cli-inject-deps
+                             :dispatch-prefix-arg 6)
```
bbatsov, line 958: There are no tests yet. The cheap ones that would cover this, all in test/cider-jack-in-tests.el:

- add `clojure-clr-cli` to the tool list in the `cider-jack-in-tools registry` spec
- add `deps-clr.edn` to the faked files in the `cider--identify-buildtools-present` spec and expect `clojure-clr-cli` in the result
- a spec for the deps string next to the clojure-cli one, e.g.

```elisp
(describe "cider-clojure-clr-cli-jack-in-dependencies"
  (it "produces a -Sdeps invocation with the clr.tools.nrepl git dep"
    (let ((result (cider-clojure-clr-cli-jack-in-dependencies
                   nil '(("io.github.clojure/clr.tools.nrepl" "abc123")))))
      (expect result :to-match "\\`-Sdeps ")
      (expect result :to-match "io.github.clojure/clr.tools.nrepl")
      (expect result :to-match ":git/sha")
      (expect result :to-match "abc123")))
  (it "appends params after the deps"
    (expect (cider-clojure-clr-cli-jack-in-dependencies
             "-X clojure.tools.nrepl/start-server!"
             '(("io.github.clojure/clr.tools.nrepl" "abc123")))
            :to-match "-X clojure.tools.nrepl/start-server!\\'")))
```

- one for `cider-inject-jack-in-dependencies` with `'clojure-clr-cli`, checking that `cider-clojure-clr-cli-nrepl-sha` ends up in the output

doc/modules/ROOT/pages/platforms/clojureclr.adoc
```diff
@@ -2,27 +2,31 @@
 
 == Current Status
 
-ClojureCLR on CIDER is not great due to the lack of a fully-functional nREPL
-server for ClojureCLR. There are currently two options:
+You will get basic CIDER functionality with ClojureCLR. Two nREPL server options:
 
-- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: A direct (but incomplete) port of the reference Clojure nREPL server.
-- https://github.com/arcadia-unity/Arcadia/blob/master/Editor/NRepl.cs[Arcadia's nREPL]: A basic, but working nREPL implementation in C#.
+- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: At present port of babashka's nREPL server (https://github.com/babashka/babashka.nrepl[babashka.nrepl]).
```
bbatsov:
```suggestion
- https://github.com/clojure/clr.tools.nrepl[clr.tools.nrepl]: currently a port of https://github.com/babashka/babashka.nrepl[babashka.nrepl] to ClojureCLR.
```

doc/modules/ROOT/pages/platforms/clojureclr.adoc
```diff
@@ -2,27 +2,31 @@
...
+- https://github.com/clojure/clr.tools.nrepl/tree/master/partial-nrepl-nrepl-port[port of nrepl/nrepl]: A non-working, work-in-progress port of nrepl/nrepl, which may
+  ultimately better integrate with CIDER once CIDER's middleware (cider-nrepl) is also adapted/ported.
 
-If you need to use CIDER with ClojureCLR today Arcadia's nREPL is your only usable option. That being said - `clr.tools.nrepl` is a much
-more sophisticated project and ideally we should get it over to the finish line.
+CIDER currently has basic support for ClojureCLR via a port of Babashka's nREPL
+  server. The reasons for the basic support is that: `cider-nrepl`, the CIDER
+  middleware, uses a lot of Java code internally itself and would need to be
+  adapted/ported like any other clojure library adapted/ported to ClojureCLR.
```
bbatsov: The continuation lines are indented, which is a literal-block tell in AsciiDoc (it renders fine here, but it's easy to break). Tightened the wording while at it:

```suggestion
CIDER's support is basic because `cider-nrepl`, the middleware behind most of
CIDER's features, uses a lot of Java internally and would need to be ported to
ClojureCLR like any other Clojure library.
```

PR #3854, opened 2025-12-31
Make item-lines buttons in cider-ns-browser

tomdl89:
Especially for evil's RET, and clicking with the mouse

-----------------

- [ ] All code passes the linter (`eldev lint`) which is based on [`elisp-lint`](https://github.com/gonewest818/elisp-lint) and includes -- it doesn't, but not because of this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Fine by me. Thanks!

PR #3856, opened 2026-01-15
fix: Make cider close robust to directory removal

hugoduncan:
## Summary

When using git worktrees, it is very easy to remove the worktree before a REPL running in that directory is closed. Currently in this case, CIDER complains the cider-history file cannot be written and errors.

This PR modifies `cider-repl--history-write` so that if the file is not writable due to the parent directory no longer existing, it emits a warning via `message` and skips writing the file (returns early without error).

## Changes

- Modified `cider-repl--history-write` to check if parent directory exists when file is not writable
- If parent directory missing: emit warning and return early (no error)
- If parent exists but file not writable: keep existing error behavior
- Added buttercup tests for history write edge cases

## Commits

- `7bb8f848` fix: Handle missing parent directory in history write
- `8ed71452` test: Add tests for history write edge cases

---

- [x] All code passes the linter (`eldev lint`)
- [ ] You've updated the user manual (if adding/changing user-visible functionality)

### DIFF
--- a/test/cider-repl-tests.el
+++ b/test/cider-repl-tests.el
@@ -329,10 +329,14 @@ PROPERTY should be a symbol of either 'text, 'ansi-context or
              (cider-repl-history-size 100))
         (unwind-protect
             (progn
-              ;; Create a read-only file
               (write-region "" nil history-file)
               (set-file-modes history-file #o444)
+              ;; Root bypasses file permissions, so mock file-writable-p
+              (when (zerop (user-uid))
+                (spy-on 'file-writable-p :and-call-fake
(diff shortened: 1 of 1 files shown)


bbatsov:
One test seems to be failing:

```
cider-repl--history-write when file is not writable but parent directory exists raises an error
FAILED: Expected `(cider-repl--history-write history-file)' to throw a child signal of `error', but instead it returned `("(+ 1 2)")'
```

bbatsov:
Thanks!

I've fixed separate the outdated macOS images that were causing unrelated failures.

PR #3882, opened 2026-04-20
Correct load-path in hacking docs after move of *.el to lisp/

metayan:
Update [hacking docs][1] with correct `load-path`, now that the `.el` files have been moved to the `lisp` subdirectory.
Only change to documentation, so no Changelog entry.


bbatsov:
Thanks!

PR #3899, opened 2026-05-03
WIP: prepl backend (architectural exercise, not for merge)

bbatsov:
This branch prototypes a Clojure prepl (`clojure.core.server/io-prepl`) backend for CIDER. It is **not** intended for merge; it's a thinking-out-loud branch left as a reference artifact.

**1. Useful as a separation exercise.** Adding a second client forced the question "what's actually nREPL-specific in CIDER vs. what's general?". The eval-handler keyword API (#3892, already merged) was step one; building a second consumer surfaced a few more leaky-abstraction warts. `nrepl--mark-id-completed` quietly assumed the request tables existed on every connection buffer. `cider-jack-in-tools` was entirely nREPL-shaped (params, middleware injection). The eval-* user commands called nREPL request functions directly rather than going through an abstraction. Even with prepl shelved, those are real findings.

**2. Prototyping confirmed prepl isn't a good fit.** Two reasons:

- *Dispatch complexity.* Every eval-shaped command grew a backend branch. Source-buffer integration needed `cider-current-backend`. The jack-in registry needed a `:backend` slot. None of this is hard individually; all of it is overhead future features would have to keep paying.
- *Capability mismatch.* The features people pick CIDER over `clj` for (debugger, structured stacktrace, inspector, profiler, test runner UI) are the same ones prepl can't host. A prepl-on-CIDER user would get a degraded experience compared to nREPL-on-CIDER while paying the same dispatch tax in the codebase.

**3. Things worth bringing forward independently:**

- `nrepl--mark-id-completed` tolerating nil request tables. A small defensive fix, correct even without a second backend.
- `dev/design/prepl-support.md` documents "we considered prepl, here's what it'd cost". Worth keeping as a reference for the next time the question comes up.
- The free-port helper (`make-network-process :server t :host "127.0.0.1" :service t`) is generic enough to live in `cider-util.el` if a second consumer ever shows up.

Leaving the branch archived. If demand ever materializes (a serious Basilisp prepl push, say), the work is on file.


bbatsov:
This research lead to the creation of https://batsov.com/articles/2026/05/12/port-a-minimalist-prepl-client-for-emacs/

It also convinced me that for CIDER it's best to remain fully focused on nREPL for reasons I've outlined in https://batsov.com/articles/2026/05/20/nrepl-forever/

PR #3900, opened 2026-05-08
Fix to Emacs hang when viewing ClojureDocs for "/"

metayan:
Emacs hangs terribly with 100% CPU usage and enormous memory consumption if it is forced to highlight "" in a buffer.

Requesting ClojureDocs for / triggers the issue.
This PR prevents it, by correctly removing the namespace - keeping only the symbol name.

-----------------

- [x] All code in the relevant file passes the linter (`eldev lint`)

-----------------

With cursor after a lone /
Don't do `C-c` `C-d` `C-c`
because it will send Emacs on a memory-eating spree and never return.
Occasionally it's possible to interrupt with `C-g`, but usually one needs to kill the Emacs process.

### CIDER version information
```
;; CIDER 1.22.0-snapshot, nREPL 1.7.0
;; Clojure 1.12.4, Java 25.0.1
```
`cider` master branch
and any version before that, probably since at least 2022

### Emacs version
30.2
28.2

### Operating system
macOS and Linux

### DIFF
--- a/test/cider-clojuredocs-tests.el
+++ b/test/cider-clojuredocs-tests.el
@@ -48,3 +48,11 @@
     (expect (cider-clojuredocs-url "even?" "clojure.core") :to-equal "https://clojuredocs.org/clojure.core/even_q")
     (expect (cider-clojuredocs-url nil "clojure.core") :to-be nil)
     (expect (cider-clojuredocs-url "even?" nil) :to-be nil)))
+
+(describe "cider-clojuredocs-lookup-remove-namespace"
+  (it "removes namespace from symbol to highlight only the name"
+    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core//") :to-equal "/")
+    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core/+") :to-equal "+")
+    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core/subs") :to-equal "subs")
+    (expect (replace-regexp-in-string "^.+?/" "" "clojure.string/trim") :to-equal "trim")
+    ))
--- a/test/cider-clojuredocs-tests.el
+++ b/test/cider-clojuredocs-tests.el
@@ -49,10 +49,14 @@
     (expect (cider-clojuredocs-url nil "clojure.core") :to-be nil)
     (expect (cider-clojuredocs-url "even?" nil) :to-be nil)))
 
-(describe "cider-clojuredocs-lookup-remove-namespace"
-  (it "removes namespace from symbol to highlight only the name"
-    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core//") :to-equal "/")
-    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core/+") :to-equal "+")
-    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core/subs") :to-equal "subs")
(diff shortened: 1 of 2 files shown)


test/cider-clojuredocs-tests.el
```diff
@@ -48,3 +48,11 @@
     (expect (cider-clojuredocs-url "even?" "clojure.core") :to-equal "https://clojuredocs.org/clojure.core/even_q")
     (expect (cider-clojuredocs-url nil "clojure.core") :to-be nil)
     (expect (cider-clojuredocs-url "even?" nil) :to-be nil)))
+
+(describe "cider-clojuredocs-lookup-remove-namespace"
+  (it "removes namespace from symbol to highlight only the name"
+    (expect (replace-regexp-in-string "^.+?/" "" "clojure.core//") :to-equal "/")
```
bbatsov: You're not really testing `cider-clojuredocs-lookup-remove-namespace`, but only the built-in `replace-regexp-in-string` with your test.

bbatsov:
Nice catch on the bug! The fix works. Two thoughts while we're here:

The regex could be tighter and line up with patterns already in the codebase. `cider-namespace-qualified-p` (cider-util.el) and `cider-resolve--prefix-regexp` (cider-resolve.el) both use `[^/]+/` for matching namespace prefixes, and the latter is even used the same way (with `replace-regexp-in-string ... ""`):

```elisp
(replace-regexp-in-string "\\`[^/]+/" "" sym)
```

Same behavior on `clojure.core//` → `/`, but more explicit about intent ("non-slash chars + slash") and `\`` is more precise than `^` (string start vs. line start).

The other thing: the new tests exercise the regex inline, not the production code — if someone tweaks the regex inside `cider-clojuredocs-lookup`, the tests pass regardless. Worth extracting a tiny helper and pointing the tests at that, mirroring how `cider-clojuredocs-replace-special` is factored:

```elisp
(defun cider-clojuredocs--strip-ns (sym)
  "Strip the namespace prefix from SYM, leaving just the symbol name."
  (replace-regexp-in-string "\\`[^/]+/" "" sym))
```

And then:

```elisp
(describe "cider-clojuredocs--strip-ns"
  (it "removes the namespace from qualified symbols"
    (expect (cider-clojuredocs--strip-ns "clojure.core/subs") :to-equal "subs")
    (expect (cider-clojuredocs--strip-ns "clojure.string/trim") :to-equal "trim"))
  (it "preserves a slash symbol name (the bug being fixed)"
    (expect (cider-clojuredocs--strip-ns "clojure.core//") :to-equal "/"))
  (it "preserves operator-like names"
    (expect (cider-clojuredocs--strip-ns "clojure.core/+") :to-equal "+"))
  (it "leaves unqualified symbols unchanged"
    (expect (cider-clojuredocs--strip-ns "subs") :to-equal "subs")))
```

test/cider-clojuredocs-tests.el
metayan: Yes, the intention with the test is to show that the regex behaves as intended. Didn't want to break out the `replace-regexp-in-string` into a new function.

Edit: Saw your comment below now.

metayan:
Clever with the "is" comments in the tests. Thanks. Learning...
I'll modify the PR accordingly.

metayan:
PR modified - with an extra test for unqualified "/" added.
Do you have a way of squashing the three commits into one, or should I do that somehow?

bbatsov:
Yeah, I can squash when merging. Thanks!

PR #3918, opened 2026-05-14
cider--completing-read-port: Default to 7888

Wilfred:
Whilst 7888 isn't the official port (see https://github.com/nrepl/nREPL/issues/6 discussion), it's frequently used in practice. Set that as the default value in the minibuffer if we couldn't infer a port.

Most nREPL implementations do create a `.nrepl-port` file but it's nice to have 7888 suggested to save a few keystrokes.


bbatsov:
I'm OK with the proposal. Can you rebase the PR on top of `master`?

Wilfred:
Done :)

PR #3920, opened 2026-05-16
Optimize cider-get-ns-name for clojure-ts-mode buffers

kommen:
Using `clojure-ts-find-ns` is a bit more efficient as it leverages tree-sitter.

Plus, there seem to be some egde cases where `clojure-find-ns` within a `clojure-ts-mode` buffer makes Emacs hang, as `clojure-find-ns`'s `(while t (up-list nil t t))` loop never terminates because `up-list` delegates to `treesit-forward-sexp`, which not always seems to signal an error.

Unfortunately I wasn't able to reproduce the error case. But as the change is a performance improvement as well, I hope this is acceptable.

### DIFF
--- a/lisp/cider-util.el
+++ b/lisp/cider-util.el
@@ -142,10 +142,14 @@ instead."
         (cider--text-or-limits bounds (point) end)))))
 
 (defun cider-get-ns-name ()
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #3930, opened 2026-06-02
[inspector] Render a help block with keybindings

alexander-yakushev:
Has been yearning for a long time to do this. The help message looks like this:

image

- Helps both the new users with basic commands and adds to discoverability of new features for experienced users.
- Hideable with `?` (or by customizing `cider-inspector-display-help`).
- Allows us to remove hints rendering from Orchard (which are also incorrect if the user redefines the bindings).

Also updated doc page for the inspector. Removed keybindings table to not having to maintain it in one more place (let it live in Emacs now) and directed the configuration to `customize-group` instead of re-listing them all again.

-----------------

### DIFF
--- a/lisp/cider-inspector.el
+++ b/lisp/cider-inspector.el
@@ -112,11 +112,16 @@ by clicking or navigating to them by other means."
   :type 'boolean
   :package-version '(cider . "0.27.0"))
 
-(defcustom cider-inspector-display-analytics-hint t
-  "When true, display hint about analytics feature for eligible objects.
-Can be turned to nil once the user sees and acknowledges the feature."
+(defcustom cider-inspector-display-analytics-hint nil
+  "Obsolete; no longer used."
   :type 'boolean
   :package-version '(cider . "1.18.0"))
+(make-obsolete 'cider-inspector-display-analytics-hint nil "1.22")
+
+(defcustom cider-inspector-display-help t
+  "When true, render help at the bottom of the inspector page."
+  :type 'boolean
+  :package-version '(cider . "1.22.0"))
 
 (defvar cider-inspector-uninteresting-regexp
   (concat "nil"                      ; nils are not interesting
@@ -164,6 +169,7 @@ Can be turned to nil once the user sees and acknowledges the feature."
     (define-key map "P" #'cider-inspector-toggle-pretty-print)
     (define-key map "S" #'cider-inspector-toggle-sort-maps)
     (define-key map "D" #'cider-inspector-toggle-only-diff)
+    (define-key map "?" #'cider-inspector-toggle-display-help)
     (define-key map (kbd "C-c C-p") #'cider-inspector-print-current-value)
     (define-key map ":" #'cider-inspect-expr-from-inspector)
     (define-key map "f" #'forward-char)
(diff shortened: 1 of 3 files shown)


lisp/cider-inspector.el
```diff
@@ -112,11 +112,16 @@ by clicking or navigating to them by other means."
...
-(defcustom cider-inspector-display-analytics-hint t
-  "When true, display hint about analytics feature for eligible objects.
-Can be turned to nil once the user sees and acknowledges the feature."
+(defcustom cider-inspector-display-analytics-hint nil
+  "Obsolete; no longer used."
   :type 'boolean
   :package-version '(cider . "1.18.0"))
+(make-obsolete 'cider-inspector-display-analytics-hint nil "1.22")
+
+(defcustom cider-inspector-display-help t
```
bbatsov, line 121: I'm wondering if we need a defcustom for this. Doesn't it make more sense to just always display something like "Press ? for help" or whatever? 

Btw, you all local keybindings are also available when doing `C-h m`, although I guess this might be more confusing for people who are not very used to Emacs.

lisp/cider-inspector.el
```diff
@@ -735,6 +744,60 @@ that value.
...
+          (max-len 0))
+      (goto-char (point-max))
+      (insert "\n\n")
+      (insert (propertize "--- Keybindings:\n" 'font-lock-face 'font-lock-comment-face))
+      ;; Pre-compute key strings and calculate maximum length
+      (dolist (cmd-pair commands)
+        (let* ((cmd (car cmd-pair))
+               (desc (cdr cmd-pair))
+               (keys (mapcar (lambda (k) (propertize (key-description (where-is-internal k cider-inspector-mode-map t))
+                                                'font-lock-face 'help-key-binding))
```
bbatsov: This looks weirdly aligned to me.

lisp/cider-inspector.el
```diff
@@ -735,6 +744,60 @@ that value.
           (t
            (error "No clickable part here")))))
 
+(defun cider-inspector--render-help ()
+  "Render a keybinding help block at the bottom of the buffer."
+  (save-excursion
+    (let ((commands '((cider-inspector-toggle-display-help . "Show/hide this message")
```
bbatsov: Perhaps this should be some var/const outside the function as it inflates its body quite a bit.

bbatsov:
Looks good to me overall, although I left a few small comments here and there. I'm not sure we should remove the keybindings table as it have the value of being searchable online. In general all the keybindings are viewable from within Emacs, but I've still replicated many of them in the docs just because I know some people are likely to discover something in the manual than organically.

That being said - I don't think strongly about the removal; just sharing my take on why sometimes redundancy is not a bad thing.

lisp/cider-inspector.el
alexander-yakushev: Yeah, damn `lambda => λ` substitution.

lisp/cider-inspector.el
alexander-yakushev: `C-h m` is fine but you have to know about it, the resuts there are not sorted and grouped, and you don't know what is important and what is not. I like the approach taken here because I grew to love it very much in Magit's interactive rebase. 

image

Having the cheatsheet right there, inline, is incredibly helpful. Magit can affort having it there all the time because rebases very rarely span many many lines. With inspector, the rendered data can be of different number of lines, and having the help all the time adds visual noise to experienced developers. But for new users, it is a great discovery mechanism. So, the custom allows to have both – it is enabled by default, so everybody gets to see the help and know it's there, they get to disable it for the current session with `?` (so they learn the binding), and they can also disable it by default forever by customizing.

alexander-yakushev:
OK, I don't mind keeping the old doc. Just figured it wouldn't be necessary if you always get to see the commands the moment you open the inspector. I'll bring it back.

Responded to other comments and addressed them.

bbatsov (approved):

PR #3932, opened 2026-06-05
Set `default-directory` in `cider-test-report-mode` buffers to `(cider-project-dir)`

Zzull:
Sets `default-directory` in `cider-test-report-mode` buffers to the value of `(cider-project-dir)` to let other Emacs commands like `magit-status` or `project-find-file` be aware of the current project.

Sesman seems to still work, meaning that pressing RET with POINT at "expected:" or "actual:" jumps correctly to the corresponding test file.

-----------------

Thanks!

manual][1] extremely useful.*


bbatsov:
Thanks for digging into this @Zzull. Looking at it, the test-report buffer isn't really special - it's just the most visible case of CIDER popup buffers not having a consistent notion of which session/project they belong to. So I went with a central fix in #3962: every popup gets pinned to its originating REPL and adopts that session's project dir, which covers `magit`/`project.el` awareness here and drops the old `*Messages*` hack along the way. Going to close this in favor of that, but the nudge was much appreciated.

PR #3934, opened 2026-06-12
Fix multiple evaluation in nrepl-dbind-response

yuhan0:
`nrepl-dbind-response` spliced the response expression directly into each binding, causing it to be evaluated once per key. When the expression has side effects (e.g. `cider-nrepl-send-sync-request`), this fires the operation multiple times.

Discovered via `cider-inspector-tap-at-point` tapping the value twice. (Note: this was the only callsite which passes a side-effectful expression directly, as a result of refactor 9dc5107e8b48dbd5839df03ce533c4afa3682fb5; all other 50+ call sites pass plain variables or pure computations)

-----------------

### DIFF
--- a/lisp/nrepl-dict.el
+++ b/lisp/nrepl-dict.el
@@ -230,10 +230,12 @@ If NO-JOIN is given, return the first non nil dict."
   "Destructure an nREPL RESPONSE dict.
 Bind the value of the provided KEYS and execute BODY."
   (declare (debug (form (&rest symbolp) body)))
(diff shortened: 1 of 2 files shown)


yuhan0:
Side note: while looking at the macroexpansion I also wondered if there was any point coalescing the repeated calls to `nrepl-dict-get`  into a single pass (currently O(n*k) in the number of keys and bindings), but a quick probe showed that the numbers involved were small enough to be negligible, so probably not worth the complexity cost

bbatsov:
Thanks!

PR #3957, opened 2026-06-16
Fix Emacs 31+ byte-compiler warnings

p4v4n:
- Replace obsolete `when-let`/`if-let` macros
- Replace obsolete `any` keyword with `non-newline` in nREPL address rx
- Add missing requires/forward-decl where the byte-compiler complained
- Bump Eldev version to 1.11.2

No functional changes.


bbatsov:
Thanks!

PR #3970, opened 2026-06-18
Inline macro stepping (cider-macrostep)

bbatsov:
Draft / Phase 1 of an in-house, inline macrostep experience (#1850). Up for hands-on play-testing - not done yet.

**What works now:**
- `cider-macrostep-expand` expands the macro form at point one step **inline in the source buffer**, highlighted, and enters a transient read-only `cider-macrostep-mode`.
- In the mode: `e`/`=`/`RET` expand (including macros inside a previous expansion), `c`/`u`/`DEL` collapse the innermost expansion, `q` collapse everything and exit.
- Each expansion is an overlay that remembers the exact text it replaced, so collapsing restores the original verbatim and nested expansions peel back in the right order. Edits go through `with-silent-modifications` + `atomic-change-group`, so the buffer's modified flag and undo aren't disturbed.
- No dependency on the external `macrostep` package, and **no new middleware** - macro detection uses the existing `info` op and expansion uses `cider/macroexpand`.

**Deliberately not here yet (follow-ups):**
- Highlighting the expandable macro sub-forms and `n`/`p` navigation between them - this is the opt-in cider-nrepl "which heads are macros" query (Phase 2).
- Per-level gensym coloring, compiler-macro/inline handling (Phase 3).
- Rewiring/replacing the existing separate-buffer commands and giving the inline flow a default keybinding - deferred until the inline flow proves out in real use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [#3963](https://github.com/clojure-emacs/cider/pull/3963): Attach `cider-scratch` buffers to a specific session (one scratchpad per session), with a configurable eval destination (cljc-style by default).
 - [#3966](https://github.com/clojure-emacs/cider/pull/3966): Add `cider-set-eval-destination` and `cider-cycle-eval-destination` (`C-c C-M-d`) to override, per buffer, which REPL type (clj, cljs or multi) evaluations dispatch to.
 - [#3969](https://github.com/clojure-emacs/cider/pull/3969): Show ClojureDocs examples inline in the `*cider-doc*` buffer, toggled with `e` (`cider-docview-clojuredocs-examples`) or shown automatically when `cider-doc-show-clojuredocs-examples` is enabled.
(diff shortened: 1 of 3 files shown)


yuhan0:
Any reason we aren't piggybacking on the existing macrostep package? I believe it has language-agnostic extension points for this purpose: 
https://github.com/emacsorphanage/macrostep/blob/d0928626b4711dcf9f8f90439d23701118724199/macrostep.el#L161-L175

There's already an existing impl https://github.com/nbfalcon/macrostep-geiser which is supposed to support CIDER too (disclaimer: haven't actually tried it out)

bbatsov:
Yeah, I'm well aware of both, but I have an aversion to 3rd party deps these days and given this is pretty simple to do I decided it was easier for both me and the end users to just go with our own implementation. This also allows us to deviate from the "classic" macrostep functionality easily.

PR #3974, opened 2026-06-18
Hint to load the namespace on unresolved-var failures

bbatsov:
Where #3973 *shows* the namespace load state, this makes the *failure modes* graceful: when you act on a var whose namespace isn't loaded, you get told what to do instead of a cryptic dead end.

[#3972](https://github.com/clojure-emacs/cider/pull/3972) taught the macroexpansion commands to say "its namespace may not be loaded; try `C-c C-k`" instead of silently doing nothing. This generalizes that:

- Extracts a shared `cider-ensure-var-resolved` gate (next to `cider-ensure-connected` / `cider-ensure-op-supported`).
- De-duplicates it out of both macro helpers (`cider-macrostep` and `cider-macroexpansion`) - pure refactor, same behavior.
- Applies it to `cider-find-var` and `cider-doc`, which previously failed with the unhelpful "Symbol not resolved".

The message points at `cider-load-buffer` via `substitute-command-keys`, so it shows whatever key it's actually bound to.

Scoped to the "ensure loaded" half. The "in sync / stale" detection lives in #3973; a later iteration could have this gate consult that state too, but keeping them independent makes each easier to review. The gate can easily be pushed into more call sites (eldoc, inspect, etc.) too.


bbatsov:
Going to close this in favor of #3979, which takes the same approach further: it tells a genuinely unloaded namespace apart from a typo, a missing require, or an out-of-sync var, works without cider-nrepl, and extends the hints to ClojureDocs, xref and browse-ns as well.

PR #4018, opened 2026-06-25
Improvements to cider-eval-to-comment

yuhan0:
Some improvements to `cider-eval-defun-to-comment ` commands:

**Prevent value and stderr comments from being interleaved**

#3148 introduced printing of stderr output (e.g. reflection warnings, compile errors) into the comments, however these were naively concatenated with the value strings as they streamed in, producing out-of-order or interleaved text in some cases.
Stderr is now accumulated separately and appended after the value with a newline.

**Replace stale eval comments on subsequent re-evals**
Easier to explain with the following gif:

https://github.com/user-attachments/assets/90908331-cdf7-474d-9e39-7c7178bbf59e

Importantly, it only matches against the exact `cider-comment-prefix`, and deletion only extends as far as the continued prefix.

### Open question
Do we want to introduce a separate toggle defcustom for this behavior? I personally don't see a good reason a user would prefer the current behaviour of accumulating stale results. It's also easy enough to 'bypass' by adding even a single newline between the form and the eval comment, which breaks the match detection. 

That was my original plan - 2771d8b3999e6c3aa0eb7e599eadb4a007b562f7 adds this defcustom (defaulting to t) and the following commit 7df655e43bfa061d3fa45feda60d6964c9922d0b removes it entirely making the detect-and-replace behavior always active. 

Will either drop the latter commit or squash away the back-and-forth changes depending on the decision.

**Also fixes**
Minor refactor bringing the point marker coercion (from #2607) into the handler instead of requiring callsites to pass a marker type. No user-facing change.

(disclosure: minor LLM assistance used in writing the test boilerplate, which I reviewed in full. All other changes certified organic and tested in person)

No changelog and user manual edits yet, pending the above question

-----------------

### DIFF
--- a/lisp/cider-eval.el
+++ b/lisp/cider-eval.el
@@ -511,28 +511,63 @@ result."
                     (insert (concat comment-prefix res (or comment-postfix "") "\n"))))
                 (cider--maybe-set-eval-register res)))))
 
+(defun cider-maybe-delete-multiline-comment (comment-prefix continued-prefix comment-postfix)
+  "Delete the region after the point in the form of a multiline comment.
+The region must begin with COMMENT-PREFIX, followed by multiple lines
+beginning with CONTINUED-PREFIX at the same indentation,
+and optionally end with COMMENT-POSTFIX."
+  (let ((pre-rgx (rx (group-n 1 (* (any " \t"))) ;; may be indented
+                     (literal comment-prefix))))
+    (when (looking-at pre-rgx)
+      (let* ((cont-rgx (rx (literal (match-string 1)) ;; match indentation rigidly
+                           (literal continued-prefix)))
+             (post-rgx (rx (literal (match-string 1))
+                           ;; trim the leading newline, NOTE string-trim messes with match data
+                           (literal (if (string-prefix-p "\n" comment-postfix)
+                                        (substring comment-postfix 1)
+                                      comment-postfix))))
+             (start (point))
+             (end (save-excursion
+                    (goto-char (match-end 0))
+                    (if (string-prefix-p ";" comment-prefix)
+                        ;; line comment - skip past any similarly indented comments
+                        (progn
+                          (forward-line 1)
+                          (while (and (not (eobp))
+                                      (looking-at-p cont-rgx))
+                            (forward-line 1)))
+                      ;; otherwise it's probably some sort of discarded form like #_
+                      (clojure-forward-logical-sexp 1))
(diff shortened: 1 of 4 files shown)


yuhan0:
(sorry, something seems to have broken when i fastforwarded the last 2-3 days worth of commits - currently investigating, will mark when it's ready for review)

yuhan0:
Fixed the bug, which was due to me splitting out some WIP / personal changes from this PR - it should be ready for review now.
The spell checker failure looks like an existing false positive, unrelated to anything here:
```
  ./test/cider-log-tests.el:100: hel ==> help, hell, healuvx codespell --ignore-words=codespell.txt
```

bbatsov:
@yuhan0 The changes look good to me. 

I'd only propose to clean up the commit log and add a changelog entry.

yuhan0:
I found another inconsistency while reviewing it for correctness - the existing `cider-eval-defun-to-comment` appends a newline to the output
https://github.com/clojure-emacs/cider/blob/8bda52abb4ae7a31eadf3056830f1364a7208137/lisp/cider-eval.el#L510

whereas the pprint variants do not, resulting in the following repro:

```clojure
(+ 2 3)
(str "foo"
     "bar")
```

With point on the end of line 1, and `M-x cider-pprint-eval-defun-to-comment`:

**Expected:**
```
(+ 2 3)
;; => 5
(str "foo"
     "bar")
```

**Actual:**
```
(+ 2 3)
;; => 5(str "foo"
     "bar")
```


Added a separate commit 1839b58db2bbdbb50e072d2d7a23e23f27339864 to address this - always adding a newline regardless of the `insert-before` arg.

lisp/cider-eval.el
```diff
@@ -510,47 +511,90 @@ result."
...
           (beg (point))
           (col (current-indentation)))
       ;; only the first line gets the normal comment-prefix
       (insert (concat comment-prefix (pop lines)))
       (dolist (elem lines)
         (insert (concat "\n" continued-prefix elem)))
-      (indent-rigidly beg (point) col)
-      (unless (string= comment-postfix "")
-        (insert comment-postfix)))))
+      (insert comment-postfix)
```
yuhan0, line 562: We could also insert the newline explicitly here i.e. `(insert comment-postfix "\n")` or afterwards in the handler, but that broke a bunch of existing tests so I went with the minimal change.

bbatsov:
Thanks!

PR #4112, opened 2026-07-15
Relax the overzealous macroexpansion guard

### LINKED ISSUE #4111: Overzealous `cider-ensure-macro` checks
## Expected behavior

`cider-macroexpand-1` (and friends) should expand `let`, `fn`, `loop`, `letfn`, since these are macros in `clojure.core` that handle destructuring of their bindings and delegate to `*`-suffixed special forms. 

## Actual behavior

Macroexpansion now refuses to expand `let` etc , throwing the user-error:

```
cider-ensure-macro: 'let' is a special form; there's nothing to macroexpand
```

This is a recent regression due to https://github.com/clojure-emacs/cider/commit/da0e416fc9ed71a0b09f4220f5987584a53ee314 - before the refactor they were handled correctly.

The updated check relies on `:special-form` metadata on the public var, via `cider-var-info`:

https://github.com/clojure-emacs/cider/blob/4b1b983da04405daa0fcb99fc8b567b217d64c41/lisp/cider-client.el#L784

Which is tricky to rely on due to Clojure's implementation of special forms:

```clojure
(:special-form (meta #'clojure.core/let)) ;=> true
(special-symbol? 'clojure.core/let) ;=> false
(special-symbol? 'let*) ;=> true
```

The same issue affects `fn`, `loop`, and `letfn`.

This also broke the use of `cider-macroexpand` as a general reader syntax normalization tool - e.g. expanding `::auto/resolved` keywords, namespace-prefixed maps, or `#()` reader syntax - which used to work as a side effect of macroexpansion being a no-op on non-macro forms, and now throws `'form' is not a macro` instead.

1. Jack in to a clj REPL
2. Place point on `(let [x 1] x)` and `M-x cider-macroexpand-1`.
3. Observe the `'let' is a special form` error, instead of expansion to `(let* [x 1] x)`

### CIDER version information

TODO: paste the version string shown at REPL startup, e.g.
```
;; CIDER 2.0.0-snapshot (package: 2.0.0-snapshot), nREPL 1.3.1
;; Clojure 1.12.5, Java 21.0.11
```

NA

### Emacs version

31.0.50

### Operating system

macOS 15.7.7

### JDK distribution

NA

bbatsov:
The 2.0 macroexpansion diagnostics (da0e416f) were stricter than intended in three ways:

- `let`, `fn`, `loop` and `letfn` were rejected as special forms, although they're real macros that merely carry `:special-form` documentation metadata (they delegate to the `*`-suffixed true specials). The macro check now comes before the special-form check.
- Non-symbol forms (auto-resolved keywords, `#()` and other reader syntax, literals) errored with "'form' is not a macro". They now pass through again - expansion is a no-op server-side, but the reader still normalizes them usefully.
- `cider-macroexpand-all` gated on the top-level operator even though a fully-recursive expansion can reach macros in nested sub-forms, e.g. `(inc (when x 1))`. The `-all` commands no longer consult the guard (matching what `cider-macrostep-expand-all` already did).

The useful parts of the diagnostics are unchanged: an unresolved operator still points at loading the namespace, a true special form (`if`, `do`, ...) still explains there's nothing to expand, and a plain function still gets "not a macro" on single-step expansion.

Fixes #4111

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+- [#4111](https://github.com/clojure-emacs/cider/issues/4111): Fix the macroexpansion commands refusing to expand `let`, `fn`, `loop` and `letfn` (macros that double as special forms), restore the no-op expansion of non-macro forms (useful for normalizing reader syntax like `::auto/kw` and `#(...)`), and stop gating `cider-macroexpand-all` on the top-level operator (a fully-recursive expansion can reach macros in nested sub-forms).
+
 ## 2.0.0 (2026-07-15)
 
 ### New features
--- a/lisp/cider-client.el
+++ b/lisp/cider-client.el
@@ -772,19 +772,24 @@ last one being the out-of-sync case where the namespace is loaded but stale."
   "Signal a helpful `user-error' unless OPERATOR names a resolvable macro.
 This turns the silent \"nothing happens\" case (e.g. the namespace hasn't
 been evaluated yet) into an actionable message, distinguishing an unresolved
-symbol from a special form or an ordinary (non-macro) var."
-  (cond
-   ((not (cider--symbol-operator-p operator))
-    (user-error "`%s' is not a macro" (or operator "form")))
-   (t
+symbol from a special form or an ordinary (non-macro) var.
+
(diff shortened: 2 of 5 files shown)


yuhan0:
Thanks! I pulled the changes, can confirm it fixes the issue :+1:

PR #4113, opened 2026-07-17
fix: apply cider-print-quota to load-file requests

### LINKED ISSUE #3052: "cider-eval-buffer" cannot be configured to *not* send evaluation result to echo area
## Expected behavior
Cider should allow to configure where to send the result of the last expression to, if a buffer/file gets evaluated
## Actual behavior
Cider always sends the evaluation result of the last expression in a filer/buffer on "cider-load-file" to the echo area of Emacs.
In case this expression is large, emacs will crash/being unresponsive immediately.
1.  Create a empty clj file
2. Add a last expression into the file which produces a complex, deep, map
3. evaluate the buffer with "cider-eval-buffer"
4. Cider will send the result of 2) to the echo area and crash

### CIDER version information
latest

mmustafasenoglu:
## Problem

When using `cider-eval-buffer` / `cider-load-buffer`, the evaluation result of the last form is displayed in the echo area via `cider--display-interactive-eval-result`. If this result is large (e.g., a deep map or huge string), Emacs freezes or crashes because the full result is sent to the echo area.

While `cider-print-quota` (default 1 MB) truncates results for `eval` requests, the `load-file` op used by `cider-load-buffer` did **not** include the print middleware parameters, so `cider-print-quota` was not enforced.

## Fix

Add `cider--nrepl-print-request-plist` to the `load-file` request in `cider-load-file-request`, so that `cider-print-quota`, `cider-print-fn`, and other print settings are honored — matching the behavior of `eval` requests.

**Before:**
```elisp
(cider-nrepl-send-request `("op" "load-file"
                            "file" ,file-contents
                            "file-path" ,file-path
                            "file-name" ,file-name)
                          ...)
```

**After:**
```elisp
(cider-nrepl-send-request `("op" "load-file"
                            "file" ,file-contents
                            "file-path" ,file-path
                            "file-name" ,file-name
                            ,@(cider--nrepl-print-request-plist))
                          ...)
```

## Why this works

`cider--nrepl-print-request-plist` already provides all the standard print middleware parameters:
- `nrepl.middleware.print/quota` → `cider-print-quota` (default 1 MB)
- `nrepl.middleware.print/print` → `cider-print-fn`
- `nrepl.middleware.print/stream?` → streaming flag
- `nrepl.middleware.print/buffer-size` → `cider-print-buffer-size`
- `nrepl.middleware.print/options` → other print options

These parameters were already being sent for `eval` requests but were missing from the `load-file` request path.

## Related issue

Fixes #3052

### DIFF
--- a/lisp/cider-client.el
+++ b/lisp/cider-client.el
@@ -806,7 +806,8 @@ positional shim retained for backward compatibility."
   (cider-nrepl-send-request `("op" "load-file"
                               "file" ,file-contents
                               "file-path" ,file-path
-                              "file-name" ,file-name)
+                              "file-name" ,file-name
+                              ,@(cider--nrepl-print-request-plist))
                             (or callback
                                 (cider-load-file-handler (current-buffer)))
                             connection))
@@ -1279,3 +1280,4 @@ end-of-input, and cancelling (quit) interrupts the evaluation."
 (provide 'cider-client)
 
 ;;; cider-client.el ends here
+
--- a/lisp/cider-client.el
+++ b/lisp/cider-client.el
@@ -807,7 +807,7 @@ positional shim retained for backward compatibility."
                               "file" ,file-contents
                               "file-path" ,file-path
                               "file-name" ,file-name
-                              ,@(cider--nrepl-print-request-plist))
+                              ,@(cider--nrepl-pr-request-plist))
                             (or callback
                                 (cider-load-file-handler (current-buffer)))
                             connection))
@@ -1280,4 +1280,3 @@ end-of-input, and cancelling (quit) interrupts the evaluation."
 (provide 'cider-client)
 
 ;;; cider-client.el ends here
-


lisp/cider-client.el
```diff
@@ -806,7 +806,8 @@ positional shim retained for backward compatibility."
   (cider-nrepl-send-request `("op" "load-file"
                               "file" ,file-contents
                               "file-path" ,file-path
-                              "file-name" ,file-name)
+                              "file-name" ,file-name
+                              ,@(cider--nrepl-print-request-plist))
```
bbatsov: I'd reach for `cider--nrepl-pr-request-plist` here instead. The eval commands that echo their result to the minibuffer (`cider-eval-last-sexp`, `cider-eval-region`) use the pr variant, and it carries the quota too, so it fixes the crash while keeping the echoed result single-line and consistent with eval. `cider--nrepl-print-request-plist` would switch load-file's result to pretty-printed and streamed, which is a behavior change I don't think we want on this path.

lisp/cider-client.el
```diff
@@ -1279,3 +1280,4 @@ end-of-input, and cancelling (quit) interrupts the evaluation."
 (provide 'cider-client)
 
 ;;; cider-client.el ends here
+
```
bbatsov: Stray trailing newline, drop this.

bbatsov (commented):
Good catch, this is a real gap. load-file sends no print params, so the last form's value comes back unbounded (I checked against a live server: a modest map came back at 82KB, and it can be much bigger), while eval bounds it via the quota. I confirmed load-file does honor the quota when we send it, so the approach is sound.

One change I'd want before merging (inline), and I'd keep #3052 open rather than close it. That issue is really about making the result destination configurable (send it to the REPL, or nowhere), which this doesn't add. Bounding the size fixes the crash, which is worth doing on its own, but it's a different thing from what the reporter asked for, so maybe reword to "Addresses #3052". A test asserting the load-file request carries the quota would be good too.

mmustafasenoglu:
Done. Thanks for the review!

- `cider--nrepl-print-request-plist` → `cider--nrepl-pr-request-plist` (pr variant keeps echoed result single-line and consistent with eval commands)
- Removed the stray trailing newline

bbatsov:
Thanks, the code looks good now. One last thing before I merge: could you change "Fixes #3052" to "Addresses #3052" in the description? That issue is really about making the result destination configurable (send it to the REPL, or nowhere), which this doesn't add, so I'd rather it not auto-close on merge. After that this is good to go.

PR #4157, opened 2026-08-07
Make cider-doctor show actual - untruncated - Clojure version

metayan:
`cider-doctor` reported
  ✓  Clojure 1.13.0
instead of
  ✓  Clojure 1.13.0-alpha6

`cider-minimum-clojure-version` is still compared to the truncated version string.

-----------------

- [✓] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
- [✓] You've added tests (if possible) to cover your change(s)
- [✓] All tests are passing (`eldev test`)


bbatsov:
Good catch. Thanks!

